PR #478, opened 2026-07-27
fix: hardcoded gts actions id-s moved to gts plugin

GeraBart:
## Summary by CodeRabbit

- **New Features**
  - Added hierarchy-aware lifecycle action type resolution for loading, mounting, and unmounting extensions.
  - Improved mount strategy selection/validation, including derived action-type handling and stricter mixed-strategy rejection.
  - Added stronger domain governance checks to require declared actions have matching registered handlers (and disallow extras).
- **Bug Fixes**
  - Improved decompression failure errors to include the original cause.
- **Chores**
  - Updated tooling and dependency versions.
- **Tests**
  - Expanded coverage for hierarchy-aware action dispatch and mount strategy validation.


packages/gts-plugin/src/plugin.ts
```diff
@@ -26,6 +26,11 @@ import {
   type JsonEntity,
 } from '@globaltypesystem/gts-ts';
 import type { TypeSystemPlugin, ValidationResult } from '@gears-frontx/mfes';
+import {
+  FRONTX_ACTION_LOAD_EXT,
+  FRONTX_ACTION_MOUNT_EXT,
+  FRONTX_ACTION_UNMOUNT_EXT,
+} from '@gears-frontx/mfes';
```
itechmeat: The three constants still live in `packages/mfes/src/constants.ts` and this file imports them back out of `@gears-frontx/mfes`, so the GTS literals are still inside the runtime and MFES-1 still fails - what moved is an indirection, not the ownership. Two new comments now assert the opposite (`DefaultMfeRegistry.ts:442` "the runtime never spells a concrete type-format literal", and lines 315-317 here), which will read as settled fact to the next person. Would it work to define the literals in gts-plugin and drop `constants.ts`, moving `INFRASTRUCTURE_LIFECYCLE_ACTIONS` behind the same port? If the intent is to land that in a follow-up, could the comments say so instead of claiming it is already done?

packages/mfes/src/runtime/DefaultMfeRegistry.ts
```diff
@@ -428,71 +431,107 @@ export class DefaultMfeRegistry extends MfeRegistry {
...
-    if (requireMount && !declaredActions.includes(FRONTX_ACTION_MOUNT_EXT)) {
-      // @cpt-begin:cpt-frontx-algo-extension-domain-governance-strategy-cardinality:p1:inst-sc-missing-required
+    // Non-string entries are treated as non-matching (F-009 hardening) rather than
+    // letting typeSystem.isTypeOf throw on a malformed declaredActions entry.
+    const hasMountExtOrDerivative = declaredActions.some(
+      (id) => typeof id === 'string' && this.typeSystem.isTypeOf(id, mountExtActionId)
+    );
+    const hasUnmountExtOrDerivative = declaredActions.some(
+      (id) => typeof id === 'string' && this.typeSystem.isTypeOf(id, unmountExtActionId)
+    );
```
itechmeat, line 455: After this switch from `includes` to `isTypeOf`, a domain that declares only a derived `mount_ext` passes registration — but handler resolution stays exact (`actions-chains-mediator.ts:393` is a plain `Map.get(actionTypeId)`), and `INFRASTRUCTURE_LIFECYCLE_ACTIONS` stays an exact-match Set feeding both the contract rule-3 exemption (`validation/contract.ts:98`) and the mediator's `infrastructureActionTypes` (`actions-chains-mediator.ts:223`). So the derived action registers but a dispatched base `mount_ext` finds no handler and the derived id is not treated as infrastructure. The five new tests only assert `registerDomain` does not throw, so nothing covers that. Should is-a reach handler resolution and the infrastructure set as well, or should this gate stay exact-match until it does?

packages/gts-plugin/src/plugin.ts
GeraBart: thanks, good catch. i missed that

itechmeat (approved):

Issue #476, opened 2026-07-27
[ARCH] Ecosystem layer enforcement is orphaned and failing after the framework/template split


tscbmstubp:
## Description

`npm run arch:sdk` and `npm run arch:layers` both fail on develop. Neither is gated by CI or pre-commit, so the failures are invisible.

```
$ npm run arch:sdk
❌ SDK @gears-frontx/mfes: Zero @gears-frontx deps: Found @gears-frontx/gts-plugin
❌ Framework: Only SDK dependencies: Invalid deps: @gears-frontx/auth,
   @gears-frontx/frontx-template-standard, @gears-frontx/gts-plugin
   Passed: 20  Failed: 2

$ npm run arch:layers
❌ SDK config: Has sdk-no-frontx-imports rule: Rule missing
   Passed: 32  Failed: 1
```

Verified against develop @ 14651af6.

## Root cause

`internal/depcruise-config/layer-constants.cjs` still encodes the pre-split L1/L2/L3 model:

```js
const SDK_PACKAGES = Object.freeze(['state', 'api', 'i18n', 'mfes']);
```

`state` and `i18n` moved to `template-standard/packages/` in #456 and are no longer ecosystem packages. `internal/depcruise-config/` also still ships `framework.cjs`, `react.cjs`, and `screenset.cjs` — for two packages that emigrated and one concept that no longer exists.

The `mfes → gts-plugin` failure is a false positive. That edge is the type-substrate port mandated by `architecture/DESIGN.md` (the `GTS -- "type-substrate port" --> MFES` edge in the pillar diagram, line 159) and specified in `architecture/features/type-substrate-port/`. The rule is stale, not the code.

## Why it went unnoticed

`.github/workflows/main.yml:72` and `.pre-commit-config.yaml:33` both run only `arch:check`, which passes (CI is green on develop head). `arch:sdk` and `arch:layers` are unreferenced anywhere in CI or hooks.

## Suggested fix

Rewrite `layer-constants.cjs` for the pillar model (Tooling → Core Framework, with the type-substrate port as an allowed edge), drop `framework.cjs`/`react.cjs`/`screenset.cjs` from the ecosystem's copy, then either wire both scripts into CI or delete them if `arch:check` has genuinely superseded them. Leaving them present-but-failing is the worst of the three options.

GeraBart:
I think the concept of layers have become obsolete after the recent changes. I'd suggest removing npm run arch:layers

[...]

Issue #470, opened 2026-07-23
Split template-standard into template-shell + template-mfe


GeraBart:
## Description
`template-standard` (`frontx-template.json`) currently claims exclusive ownership of one undivided tree: `src-app/app/*` (layout, routing, actions, api, events, themes — shell concerns) plus `src-app/mfe_packages/*` (demo-mfe, `_blank-mfe`, widgets-fixture-a/b — individual MFE package concerns), alongside shared root config (`package.json` workspaces, `vite.config.ts`, `scripts/dev-all.ts`, `build-mfes.ts`).

This is explicitly the reference scenario the architecture already anticipates: a 2026-07-16 multi-template-composition exploration names "shell + mfes" as its own canonical example and flags a shell+2-mfe spike as recommended prototype scope. The assembly mechanism itself (`uniformApply`, ownership-boundary staging, pre-flight conflict check) is already implemented in `packages/cli/src/scaffold/` — this is template content restructuring + validation, not new CLI mechanism work.

Work:
1. Carve `mfe_packages/*` plus its dev/build scripts into a standalone `template-mfe`, with its own `frontx-template.json` manifest and ownership boundaries.
2. Leave `template-shell` with `src-app/app`, root workspace config, and `dev-all.ts` orchestration.
3. Re-declare `exclusiveSubtrees`/`sharedFiles` on both templates per ADR-0031, resolving how shared root files (`package.json`, `tsconfig*.json`, `vite.config.ts`) are owned/merged across the two — currently claimed exclusively by one template.
4. Exercise the pre-flight conflict check (ADR-0032) when both templates are assembled into one repo.
5. Test three seeding scenarios: shell alone, mfe alone, and shell+mfe combined (monorepo) — including `dev:all` + browser click-through per the project's done-gate standard.

## Estimate
4–5 engineer-days

## Risks / open questions
- Depends on ADR-0020/0031/0032 being fully `accepted` — several composition ADRs are currently still `proposed`; splitting before acceptance risks rework.
- Shared root files are currently claimed exclusively by one template; splitting must define ownership vs. `sharedFiles` merge rules — this is the untested part of the mechanism.
- Whether `template-mfe` seeded alone (no shell) is a supported scenario or only "add-template into existing shell" is supported needs a decision.
- No second template exists today (`template-standard` is the only `template-*` directory) — this is the first real exercise of the multi-template code path outside unit tests (`assembler.test.ts`, `conflict.test.ts`); expect to surface gaps.

[...]

Issue #469, opened 2026-07-23
Add //<subdir> addressing to template source-spec; retire temporary npm template-publish workflow


GeraBart:
## Description
Folder/repo-based template discovery already exists: the CLI resolves templates via `host:owner/repo@ref` and fetches the whole GitHub repo as a tarball (`packages/cli/src/spec-parser/parse.ts`, `adapters/github-fetch.ts`, ADR-0016 `template-acquisition-and-location`), materializing it into `~/.frontx/inventory`. There is no npm-registry lookup anywhere in this resolution path.

The only npm-publish step in the repo, `.github/workflows/publish-packages.yml` (publishing `@gears-frontx/frontx-template-standard`), is explicitly marked `PART 2 (TEMPORARY)` in its own comments and is not consumed by the CLI's resolver at all.

CF Studio's `cfs kit install <owner/repo[@ref]>` uses a functionally identical whole-repo-fetch mechanism, plus one capability FrontX lacks: explicit subdirectory addressing via `//<subdir>` (`cfs kit install git/<url>[//<subdir>][@<kit>]`), letting one kit/template be a subfolder of a larger repo rather than requiring one repo per template.

Work:
1. Extend the source-spec grammar to `host:owner/repo//path@ref`, parsed in `spec-parser`.
2. Update the tarball-unpack step to filter/extract only the matched subtree when a subdir is specified.
3. Add tests covering subdir-addressed resolution.
4. Amend ADR-0016 to document the new addressing form.
5. Retire the temporary npm-publish step in `publish-packages.yml` once `template-standard` (or its successors) no longer needs it.

## Estimate
2–3 engineer-days

## Risks / open questions
- Original task premise ("migrate from npm packages to folder discovery") was factually incorrect — folder discovery already exists; scope here is narrower (subdir addressing) than initially assumed.
- CF Studio itself has no automated kit-discovery/store yet (listed only as planned/future) — nothing further to replicate beyond the `//<subdir>` syntax.
- GitHub API rate limits/auth for private repos are unaffected either way (already handled via optional bearer token in `github-fetch.ts`).
- Decide whether to delete `publish-packages.yml`'s template-publish part now or wait for `template-standard`'s extraction into its own repo — unrelated dependency, not blocking this issue.

PR #481, opened 2026-07-28
feat: subtree-addressed templates and manifest-declared identity (#469, #477)

### LINKED ISSUE #477: Stop deriving template identity from the repository name
## Description

Template identity across the whole CLI is derived from the repository name, not from the manifest. `packages/cli/src/resolver/resolve.ts:10` sets `const name = ref.repo`, and that value then becomes the inventory key (`FsInventoryIndex.record` writes `entries[entry.name]`), the on-disk content path (`resolveInstalledContentPath(root, name)`), the `templateRef` accepted by `seed` and `add`, and the `templateIdentity` written into `.frontx/provenance.json`.

The manifest's own `name` field is required by validation (`packages/cli/src/manifest/validate-contract.ts`) but never read: grepping `manifest\.name` across `packages/cli/src` outside tests returns zero hits. A template can therefore declare `"name": "@gears-frontx/frontx-template-standard"` and be tracked everywhere as `frontx-template-standard`, which is exactly what a seeded project's provenance records today.

Two consequences follow.

**A name collision merges two templates instead of rejecting them.** `TemplateInventory.install` writes unconditionally, with no existence check and no warning, and `FsContentStore.write` does `mkdirSync(recursive)` followed by a write-through of each file without clearing the directory first - only `replace`, used by `update-local`, calls `rmSync`. Installing `github:acme/ui-template@v1` and then `github:contoso/ui-template@v1` therefore does not replace one with the other: the second template's files are written into the first one's directory, matching paths overwritten and non-matching ones left in place. The index entry is replaced wholesale, so `frontx list` reports only the second template while the content path holds a mixture of both. Nothing in the output signals this.

**It blocks the two-templates-per-repository case.** Once #469 lands `//<subdir>` addressing, `github:owner/repo//template-shell@ref` and `github:owner/repo//template-mfe@ref` - which is exactly what #470 produces if both templates stay in this repository - resolve to the same `name`. Subdir addressing without an identity change ships broken for its primary use case.

**Relationship to #469.** This is not an independent change. Identity derivation lives in the same call path #469 modifies (`spec-parser` → `resolveToInventory`), shares its tests and its ADR-0016 amendment, and #469 is what makes the collision reachable in the primary use case. Treat this as an acceptance criterion of #469 rather than a separate pull request; it is filed separately only so the defect stays visible if #469 is rescheduled.

Verified against develop @ 14651af6.

Work:

1. Decide the identity source: the manifest's `name`, the full source-spec including the subdir, or a composite. Whichever is chosen, the required-but-unused `name` field in `frontx-template.json` either becomes load-bearing or is dropped from the contract.
2. Apply it in `resolveToInventory` and thread it through the inventory index, the installed content path, and provenance's `templateIdentity`.
3. Reject a colliding install explicitly instead of merging into an existing content path.
4. Decide what happens to inventories and `.frontx/provenance.json` files already written under the old identity.
5. Amend ADR-0016 and ADR-0019 if the identity contract changes.

## Estimate

1–2 engineer-days

## Risks / open questions

- Changing `templateIdentity` invalidates existing provenance records. `occupiedBoundariesFromProvenance` looks each record up in the inventory and skips silently when the lookup misses, so a stale identity drops that template's boundaries from the pre-flight conflict check instead of failing loudly.
- Identity taken from the manifest is only knowable after the fetch and after parsing the bundle envelope, while `resolveToInventory` currently computes the name before fetching. The resolver gains a parse step.
- A scoped manifest name such as `@gears-frontx/frontx-template-standard` contains `/` and `@`, which makes it an awkward directory name under the inventory root and an awkward argument to `seed`, `add` and `update-local`. If the manifest name wins, a normalisation rule is part of the contract.

itechmeat:
# Subtree-addressed templates, and identity from the manifest

Closes #477. Implements #469 except its work item 5, deferred with reasons below, so #469 stays open.

Base `develop` @ `37b09ee3`. 33 files, +1836 / -152, of which 6 are new.

One change, not two: both issues modify the same call path (`spec-parser` -> `resolveToInventory` -> `TemplateInventory`), share their tests, and #469 is what makes #477's defect reachable in the case #469 exists to serve. #477 asks to be treated as an acceptance criterion of #469 rather than a separate pull request.

## What changes for a developer

| Before | After |
|---|---|
| One template per repository. A family of templates meant a family of repositories. | `frontx install github:acme/templates//shell@v1.0.0` installs the template living in `shell/`. Only that subtree is materialized, re-rooted so the template does not know where it lived. |
| A template was tracked under the repository name. The `name` its manifest declares was required by validation and read by nothing. | A template is tracked under its manifest-declared identity: the inventory key, the installed content path, the argument to `seed` / `add` / `update-local`, and `templateIdentity` in provenance. |
| Two templates whose repositories shared a name merged into one content path - matching files overwritten, the rest left behind, `frontx list` showing one template and the disk holding a mixture. No signal. | The second install is refused, naming both sources. Nothing is written. |
| `github:acme/a/b@v1` parsed with `repo = "a/b"`, then 404ed or created a nested directory. `:owner/repo@ref` and `owner/repo@` parsed with an empty host or selector. | All three are parse errors before any fetch. |
| Re-installing at a newer ref wrote through the existing content path, so a file the new version dropped survived on disk. | A re-install at the same address clears the path first. |
| `frontx validate` accepted an identity such as `../outside`, which install then refused. | Path usability is part of the manifest contract, so validation and install agree. |

## Why the reference gained a segment rather than a second input

Recorded in **ADR-0033** (new). Weighed against a separate subtree input and one-repository-per-template. The segment wins because a reference is stored and re-resolved, not consumed once - provenance keeps it and `frontx upgrade` re-resolves through it - so a second value that can disagree with the first would have to be threaded through every one of those surfaces.

ADR-0033 **supersedes ADR-0017** rather than amending it. Both issues ask for an amendment, but the cf-studio kit forbids editing an accepted ADR's decision (*"ALWAYS create a NEW ADR with a SUPERSEDES reference"*). ADR-0033 therefore restates the complete reference shape, including the mandatory `host:` prefix and `@ref` selector, so no part of the grammar stays owned by a superseded record.

**Identity is not a new decision.** ADR-0018 already decides that the manifest declares identity; ADR-0027 assigns the field schema to the owning FEATURE. `const name = ref.repo` was simply not conforming, which is why this half needs no ADR of its own. In-tree evidence that the two disagreed: `template-standard/frontx-template.json` declares `"name": "@gears-frontx/frontx-template-standard"` and claims `.frontx/ai/@gears-frontx/frontx-template-standard` as its own subtree, while the runtime tracked it as `frontx-template-standard`.

## Resolution

The old path was four steps: parse, `identity := repo`, fetch, write the whole repository.

```mermaid
flowchart TB
  A1["parse host:owner/repo[//subtree]@ref"] --> A2["build fetch address<br/>(no subtree) and the<br/>re-resolvable source-spec<br/>(with subtree)"]
  A2 --> A3["fetch repository tarball"]
  A3 --> A4{"subtree named?"}
  A4 -- yes --> A5["narrow to subtree,<br/>re-root every path"]
  A4 -- no --> A6["keep as acquired"]
  A5 --> A7["read manifest,<br/>identity := manifest.name"]
  A6 --> A7
  A7 --> A8{"identity occupied by<br/>a different address?"}
  A8 -- yes --> A9["refuse, write nothing"]
  A8 -- no --> A10["write under<br/>declared identity"]
```

Two properties of that order are load-bearing, each with a test:

- **The fetch address excludes the subtree; the recorded source-spec includes it.** Acquisition stays whole-repository, so the transport is unchanged. But `provenance.sourceSpec` is what `frontx upgrade` re-parses, so dropping the segment there would silently re-resolve the repository root - a wrong diff rather than an error.
- **Re-rooting is not cosmetic.** `readManifestFromContent` and `FsContentStore` look the manifest up by its exact unprefixed filename, and the store materializes every bundle key verbatim. Filtering without stripping the prefix would nest the template a level deeper and hide its manifest.

## The escape re-rooting creates

`shell/../sibling/evil.txt` is inside the repository and becomes `../sibling/evil.txt` once `shell/` is stripped. The segment the parser validated is well-formed; narrowing manufactures the escape, so the guard cannot live in the parser. It refuses the whole narrowing (`escaping-path`) rather than dropping the key, so the refusal is visible instead of a partial template being installed. `FsContentStore.assertWithinRoot` would also catch it, but by throwing past the result union, and the in-memory store every unit test uses has no such guard.

## Two guards, and the migration that keeps them from locking anyone out

Identity is now the key to a content path, so two operations that could not collide now can.

```mermaid
flowchart TB
  I["install"] --> IC{"identity already<br/>in the index?"}
  IC -- no --> IW["write"]
  IC -- yes --> IA{"same template ADDRESS?<br/>(reference minus @ref)"}
  IA -- yes --> IW2["write: a version bump<br/>is not a collision"]
  IA -- no --> IR["refuse, naming both sources"]

  U["update-local"] --> UL{"entry exists?"}
  UL -- no --> UN["refuse: not installed"]
  UL -- yes --> UI{"fetched manifest declares<br/>the entry's identity?"}
  UI -- yes --> UW["replace content"]
  UI -- no --> UR["refuse: that spec resolves<br/>to a different template"]
```

The install guard compares the **address**, not the whole source-spec: comparing the spec would refuse an ordinary `@v1.0.0` -> `@v2.0.0` bump while leaving the real hazard - the same address re-installed over itself - unguarded.

The update guard exists because `update-local` takes a new source-spec for an existing entry, and a spec can now name a subtree: pointing an entry at a sibling subtree would substitute one template for another under the entry's identity, and `replace` clears the content path first, so the occupant would be gone.

A third refusal follows from identity being a path: `@acme/tools` and `@acme/tools/extra` are two index keys but not two directories, so a bounded update of the outer one would clear the inner one from disk while leaving it listed.

### Provenance written before this change

Work item 4 of #477, and its named risk. `occupiedBoundariesFromProvenance` used to **skip** a record it could not resolve, silently dropping that template's boundaries from the pre-flight conflict check - so `frontx add` would pass a claim over ground an applied template owns and overwrite it, the one outcome that check exists to prevent.

Failing loudly is necessary but not sufficient: the recorded identity is a repository name, and reinstalling from the recorded source-spec now lands under the manifest identity, so an identity-only lookup would fail every pre-existing repository with a recovery instruction that can never succeed. The record's **source-spec** is what survived the change, so an identity miss falls back to matching by source address:

```mermaid
flowchart TB
  P["provenance record"] --> L1{"identity in inventory?"}
  L1 -- yes --> OK["boundaries established"]
  L1 -- no --> L2{"exactly one installed template<br/>with the same source address?"}
  L2 -- yes --> OK
  L2 -- "several" --> AMB["abort: occupied-source-ambiguous"]
  L2 -- none --> NI["abort: occupied-not-installed,<br/>naming frontx install &lt;sourceSpec&gt;"]
  OK --> M{"manifest satisfies<br/>the contract?"}
  M -- no --> UM["abort: occupied-manifest-unreadable"]
  M -- yes --> C["submit to the conflict check"]
```

The migration is silent where it can be and loud only where the template genuinely is not installed - and there the message names a recovery that works. The occupant is still reported under the identity the **record** holds, because that is what this repository's region markers and provenance already carry.

An identity hit is trusted only when the entry came from the address the record names. An inventory written before the collision guard existed can hold that key pointing at an unrelated template, and checking its boundaries in place of the real occupant's would let a claim over the real owner's ground pass.

## Artifacts

| Artifact | Change |
|---|---|
| `ADR/0033-source-spec-subdirectory-addressing.md` | New. The complete reference shape, every delimiter rule, and the consequences - including that the registry namespace no longer supplies identity uniqueness for free, and that N sibling templates from one repository acquire that repository N times. |
| `ADR/0017-source-spec-syntax.md` | `status: superseded`, `superseded_by`, status history. Body untouched. |
| `DESIGN.md` | Seven references repointed; the new ADR added to the Pillar 2 inventory; 0017 marked superseded. |
| `features/template-resolution/FEATURE.md` | Parse and resolve algorithms rewritten, a sixth state transition, a new DoD for manifest identity and collision refusal, six acceptance criteria. |
| `features/composed-provenance/FEATURE.md` | A provenance record must retain the subtree segment. |
| `features/cli-scaffolding/FEATURE.md` | The add flow gains the occupied-boundaries resolution step, its abort, and the matching error scenario. |

## Deliberately not done

**#469 work item 5 - retiring the temporary npm template-publish block.** The issue calls the decision open and non-blocking, and three things argue for later:

1. The workflow's own comment conditions removal on the standard template moving to its own repository. That has not happened and is not part of #470.
2. `PART 2` publishes **seven** packages: the template plus all six `template-standard/packages/*`, five of which carry a `latest` dist-tag. The block conflates "temporary template publish" with "publication of the app layer that #456 relocated".
3. A seeded project does not need those six - the template's `exclusiveSubtrees` includes `packages`, so they travel as source - but they have consumers outside this repository.

Recipe for whoever removes it: delete lines 163-340, set the `summary` job's `needs` to `[detect-ecosystem, publish-ecosystem]`, drop its second `report` call. No `@cpt` markers live in that file and `scripts/ecosystem-version-policy.mjs` only touches `PART 1`.

**Identity parity on the upgrade path.** `update-local` now refuses a substituted identity; `frontx upgrade` re-resolves from a stored source-spec without the same check, so a repository reorganisation between two versions can have it compute a change set for one template against another's content. ADR-0033 records this as a consequence of the new shape.

Outside this PR for two reasons: it belongs to `upgrade-changeset`, a fifth FEATURE whose scope is the change set rather than identity; and it interacts with the legacy-provenance case above, where a record legitimately carries a different identity than the template declares, so a strict comparison would re-create exactly the lockout this PR removes. The fix has to reuse address-first resolution rather than add a second rule beside it. Blast radius is a wrong diff rather than lost content, and the change set is review-gated before anything is written.

**`.ai/targets/CLI.md`.** Seven of its twelve sections describe the pre-#456 CLI - `presets/`, `templates/`, `copy-templates.ts`, `screenset create|copy`, none of which exist. This PR adds a STATUS banner naming them and pointing at the real source of truth; deleting them is a separate change, since `.ai/GUIDELINES.md`'s routing table is stale for six other packages too.

## Import edges and markers

No edge between packages is added or removed: everything below is inside `@gears-frontx/cli`, and nothing new points at `adapters/` from core.

| New edge | Why it is allowed |
|---|---|
| `resolver/` -> `manifest/` | The resolver reads the acquired manifest to learn the declared identity. Both are core and pure; `manifest/` imports nothing from `resolver/`. |
| `resolver/`, `manifest/`, `adapters/` -> `bundle/` | `bundle/` is a new core module owning the envelope shape, so readers stop carrying their own copy and producers import the marker instead of restating it. Core does not import the adapters. |
| `spec-parser/`, `resolver/`, `manifest/`, `inventory/` -> `paths/` | `paths/` is a new core module owning the relative-path predicate a subtree segment and a declared identity both answer to. Leaf module, imports nothing. |
| `scaffold/` -> `spec-parser/` | Resolving a provenance record by source address needs the reference parsed and its selector dropped, which is the parser's function rather than a second copy. |
| `inventory/` -> `paths/`, `inventory/` -> `spec-parser/` (widened) | The collision guard compares addresses and refuses nesting identities. |

Both new leaf modules, `bundle/envelope.ts` and `paths/relative-path.ts`, are internal: neither is re-exported from the package barrel.

Sixteen markers are added, each against an instruction this PR also adds to a FEATURE, plus the DoD `cpt-frontx-dod-template-resolution-manifest-identity`. No marker was moved off an instruction and no instruction lost its marker. `inst-resolve-name` keeps its id and changes meaning, since the step it names is still "establish the template's identity". `cfs validate`: 54 artifacts, 0 errors, 0 warnings, coverage 117/117.

## Public surface and behavioural breaks

- `occupiedBoundariesFromProvenance` (exported from the barrel) returns a result union instead of an array, and takes the installed set as a third argument.
- `addTemplate` takes a `listInstalledFn` seam.
- `AddTemplateResult.reason` gains `occupied-not-installed`, `occupied-manifest-unreadable`, `occupied-source-ambiguous`. All three map to the user-error exit code, whereas the sibling `manifest-unreadable` maps to the internal one - a pre-existing inconsistency left alone.
- `validateManifestContract` rejects a `name` unusable as a repository-relative path, so a template that previously passed `frontx validate` and failed at install now fails validation.
- `StructuredRef` gains `subtree?: string`, **absent** rather than `undefined` when the reference carries no segment, so an older reference parses into the identical four-part shape.
- `github:owner/repo/extra@ref` and a malformed subtree segment are now parse errors. Both previously parsed and failed later or silently.
- The identity a template is tracked under changes for every template whose manifest name differs from its repository name. Nothing migrates the inventory index itself, so a stale entry keyed by a repository name stays listable until it is reinstalled.

## Automated review round

| Finding | Action |
|---|---|
| `local-fetch.ts` builds bundle keys with `path.join`, so on Windows a subtree-addressed install of a local template fails as `empty-subtree` and a backslash key reads as an escaping path. | Fixed. A bundle key is POSIX-separated on every host, because the envelope is a wire format rather than a host path; the rule now lives with the envelope, which owns the shape. `FsContentStore`'s read-back walker had the same latent defect - it made the round trip it documents lossy on Windows - and is fixed with it. |
| `update-local`'s argument error said `<name>` while the usage text says `<identity>`. | Fixed. |
| The `frontx add` example in `QUICK_START.md` used a repository-style name. | Fixed - it now uses the identity form the same document establishes two sections earlier. |
| The re-install regression asserted final content through the in-memory store, whose `write` and `replace` are both one `Map.set`, so it passed whichever the code chose. | Fixed with a store double carrying the real adapter's write/replace asymmetry, checked by the red-to-green flip. |
| Deduplicate provenance-derived occupied boundaries by resolved installed entry, since two records for one template are compared against each other and read as a self-conflict. | Declined. |

The declined finding describes a real mechanism: `checkAssemblyConflicts` compares all pairs of the combined claim set, so two occupied claims over one exclusive subtree do abort the apply. The state is not reachable. A second record for one template would have to be written by an apply, and the conflict check refuses that apply on exactly that ground; `frontx upgrade` rewrites one record in place and carries its `templateIdentity` over. What remains is a hand-edited provenance file, and such a file is genuinely wrong, because a template owns its boundaries once. Deduplicating would absorb that silently - the inverse of why this PR made an unresolvable record fail loudly instead of skipping it.

Also not acted on: the docstring-coverage threshold, 68% against a required 80%. The repository's own style rules have comments carry constraints rather than restate signatures, and the threshold is not a CI gate here.

## Verification

```
npm run test:unit    489 tests (CLI 212, was 168)
npm run type-check   pass
npm run lint         pass
npm run arch:check   pass
npm run arch:deps    pass
npm audit --audit-level=high --omit=dev   pass
cfs validate         0 errors, 0 warnings
cfs spec-coverage    thresholds met
```

The branch is rebased onto `develop` after #478, which the audit gate needs: the twelve high advisories failing on the previous `develop` all sit in the eslint / typescript-eslint chain #478 bumps. This PR touches no dependency file.

`arch:sdk` and `arch:layers` fail on `develop` and are wired into neither CI nor the pre-commit hook (#476). They fail identically here.

Both guard tests were checked for vacuity by breaking the guard and confirming the test goes red, then restoring it. The first version of the collision test compared the result against the same fixture string it fed the second installer, and the first version of the re-install test could not see which store operation was chosen; both passed with the behaviour removed.

## Suggested reading order

1. `ADR/0033` - the decision and its consequences.
2. `spec-parser/parse.ts` - the grammar and the six rejections.
3. `resolver/narrow-subtree.ts`, `resolver/resolve.ts` - narrowing, re-rooting, identity.
4. `inventory/TemplateInventory.ts` - the two guards.
5. `scaffold/materialize.ts`, `commands/add-template.ts` - the loud failure and the address migration.
6. `__tests__/source-spec-subtree.test.ts`, `__tests__/template-identity.test.ts` - fixtures deliberately divergent, so a manifest name never coincides with its repository segment.

## Questions for review

1. Is the address fallback for legacy provenance the right migration, or should a `frontx` command own it explicitly?
2. Should the upgrade-path identity check land as the immediate follow-up, or wait until legacy provenance is known to be gone?
3. `github:owner/repo/extra@ref` now fails where it previously parsed. It never worked end to end, but it is a behavioural break - acceptable pre-1.0?

## Summary by CodeRabbit

- **New Features**
  - Templates can now be installed from repository subdirectories using `host:owner/repo//subtree@ref`.
  - Template tracking now uses the identity declared in the template manifest (not the repo reference).
  - Added stricter validation for subtree segments, manifest identity usability, and identity collisions.
- **Bug Fixes**
  - Existing-project “add template” flows now better detect unresolved/ambiguous installed templates and avoid unsafe file writes.
  - `update-local` now refuses updates when the resolved template identity doesn’t match the installed entry.
- **Documentation**
  - Updated QUICK_START, README, and CLI help to match the new source-spec and identity semantics.
  - Redirected “Getting Help” guidance to architecture docs and removed outdated AI guideline references.


architecture/ADR/0017-source-spec-syntax.md
```diff
@@ -1,6 +1,9 @@
 ---
-status: accepted
+status: superseded
```
GeraBart: I would suggest to avoid superseding ADRs. ADR has an outcome-agnostic name. So instead of superseding the ADR, it's better to change the outcome of the decision making. We don't want any history, because it bloats documents. SDLC documents should precisely coved the current project state, and have zero historical footprints.

.ai/targets/CLI.md
GeraBart, line 1: I think this file as well as the whole ".ai" folder can be removed. This is uncleaned legacy. AI tooling now belongs to the frontx CF Studio kit

GeraBart:
Reviewed the code and the artifact chain. The change is solid — the parser's delimiter rules hold under adversarial reading, narrowing refuses rather than partially installs, and the test set covers every rejection ADR-0033's Confirmation section asks for. One blocking item and three smaller ones.

## Blocking — the nesting refusal is implemented but unspecified

`TemplateInventory.install` refuses an identity that nests inside an installed one (`pathsNest`, `paths/relative-path.ts:38`), with its own error message, and `template-identity.test.ts:238` asserts it. No artifact describes that behavior:

- `features/template-resolution/FEATURE.md` step 8 (`inst-resolve-collision-check`) covers only *"the local inventory already tracks that identity for a different template address"*.
- `cpt-frontx-dod-template-resolution-manifest-identity` covers the same single case.
- Neither the Install flow's Error Scenarios list nor §6 Acceptance Criteria mentions nesting.

Because the nesting branch is marked `inst-resolve-collision-fail`, `cfs validate` is green — traceability cannot see a marker that covers more behavior than its instruction describes, which is exactly why semantic conformance has to be checked separately. Please fix the artifact, not the code: add the instruction under `inst-resolve-collision-check`, the error scenario, the DoD sentence, and one acceptance criterion. The implementation is right; the spec is what is missing.

## Recommended

**1. CodeRabbit's dedupe concern is resolved against a different mechanism than the one it raised.** `checkAssemblyConflicts` (`scaffold/conflict.ts:62-63`) pairs *every* claim, occupied-vs-occupied included. `occupiedBoundariesFromProvenance` still pushes one claim per record keyed on `record.templateIdentity`, so a repository whose `provenance.json` already holds two records for one template — a pre-manifest-identity record plus a manifest-identity one — now has both resolve successfully (one by identity, one by source address) and collide with each other. Every `frontx add` in that repository aborts naming one template under two identities, with no in-CLI recovery. Fresh flows can no longer create the pair, so this is legacy data only, but the fix that closed the thread changed resolution rather than adding the dedupe. Dedupe the occupied claims by resolved installed entry before submitting them.

**2. The malformed-subtree parse error under-reports its own rule.** The message names only *"no empty, `.` or `..` segments"*, while `isSafeRelativePath` also rejects surrounding whitespace, `\`, `:` and control characters. `github:acme/t//a:b@v1` is refused with a message that does not say why. Two things: enumerate the full rule in the message (the manifest validator's identity violation already does), and record the `:` and `\` exclusion in ADR-0033's consequences next to the `@` one that is already there — a subtree path containing `:` is unaddressable for the same reason and deserves the same visibility.

**3. Two doc tables missed the `<identity>` rename.** `README.md:166` and `QUICK_START.md:110` still read `frontx update-local <name> <spec>` while `usageText()` and the argument error now say `<identity>`. Same nit as the `cli.ts` one, two occurrences behind.

Separately, I left two comments on ADR-0017 and `.ai/targets/CLI.md` about the ADR supersession policy and the legacy `.ai/` folder — both are about direction rather than this diff's correctness, so I have kept them out of this list.

architecture/ADR/0017-source-spec-syntax.md
itechmeat: Agreed, done in `5b5e6335`: ADR-0033 is folded back into ADR-0017 — the name is outcome-agnostic, the outcome is edited in place, the previous shape stays as a considered option, and the supersede/status-history metadata is gone.

A side benefit sold me on it: ADR-0016/0019/0028 cite this record's ID as the shape owner (0028 says the anchor must not be orphaned), so the supersede had left three accepted ADRs pointing at a dead record.

.ai/targets/CLI.md
itechmeat: Done in `00acba4e`: `.ai/` is removed together with everything that only existed to read it — the prompt-tests workflow (it also triggered on `.claude/**`), the pointer commands in `.claude`/`.cursor`/`.windsurf`, the copilot-commands mirror, template-standard's legacy command sets (their add-MFE role already lives in the template's own `.frontx/ai` bundle), and the promptfoo toolchain. Live tooling is repointed to the kit and `architecture/`.

One note for the record: `.cf-studio/config/core.toml` configures only the sdlc kit today, so the removed guidelines have no in-repo successor until the FrontX studio kit lands.

itechmeat:
@GeraBart 

> Reviewed the code and the artifact chain.

All four items are in:

- **Blocking** — fixed in the spec, code untouched: the nesting branch is now stated in `inst-resolve-collision-check`, the Install flow's error scenarios, the DoD, and a new acceptance criterion (`e3d93d5c`).
- **Dedupe** — implemented as you framed it: one claim per resolved installed entry. You were right that I had closed the thread against resolution, not the dedupe. Tests cover both sides — no false self-conflict, and the surviving claim still refuses a real intruder (`e3d93d5c`).
- **Parse error** names the whole rule now, with a test for `:`; the `:`/`\` exclusion is recorded in the ADR consequences next to the `@` one.
- Both `update-local <identity>` table rows updated.

Your two side comments are addressed too — replied in their threads (`5b5e6335`, `00acba4e`).

GeraBart (approved):

Issue #483, opened 2026-07-28
[ARCH] Make template validation gates runnable and keep CI/tool paths current


tscbmstubp:
## Description

One coherent failure class: checks and configuration exist in this repo but either never run, or point to paths that the template split (#470/#487/#489) already moved or removed. Each item below is a validation gate that looks live but isn't currently doing its job.

This used to include a `template-standard/`-specific item, but that tree no longer exists post-split — the CI-coverage gap for templates now lives entirely under item 1, scoped to `template-mfe` (and coordinated with the workflow shape PR #491 is introducing for `template-*/` in general). This issue is about durable CI/config coverage after the split, not about template ownership semantics.

### 1. CI validation for `template-mfe`

`.github/workflows/main.yml` still has an "Install MFE dependencies" step that iterates `src/mfe_packages/*/` — a path that hasn't existed at the repo root since the split. Packages now live under `template-mfe/src-app/mfe_packages/*/`. The step currently matches nothing, so it silently no-ops and MFE dependency installation never runs in CI.

- Integrate the correct validation into the template-validation workflow rather than patching the dead step in place. Coordinate with #491 (adds a `template-validate` job that globs `template-*/` and runs each template's own `npm ci` + build + `arch:deps` + `lint` + `test:unit`) and #501 (makes `template-mfe`'s packages actually resolve in the monorepo).
- Account for the intentionally minimal `template-mfe/package.json` (no `arch:deps`/`lint`/`test:unit` scripts by design, per #501): commands invoked against it should use `--if-present` where appropriate so the job doesn't turn red on a harness that's deliberately script-less. This is called out explicitly in #501's review notes as an open conflict between the two PRs — whichever lands second needs to reconcile it.

### 2. Stale Knip configuration

`knip.json` still names pre-split paths:

```json
"entry": ["src/app/main.tsx", "src/screensets/*/*[Ss]creenset.tsx"],
"project": ["src/**/*.{ts,tsx}"]
```

`src/` no longer exists at the repo root. Update entry/project patterns to current paths, and make the check fail or report clearly if a configured source root matches nothing — right now a misconfigured, all-empty glob passes silently instead of surfacing as a problem.

### 3. Architecture guard typing

[...]

Issue #482, opened 2026-07-28
[DOCS] .ai/GUIDELINES.md ROUTING points at packages that no longer exist after the framework/template split


tscbmstubp:
## Description

The root `CLAUDE.md` makes `.ai/GUIDELINES.md` the single source of truth for FrontX development, and its AI WORKFLOW section *requires* reading the routed target file before changing any code:

> - Route: select target file from Routing section.
> - Read: MUST read target file before changing code or docs.
> - FORBIDDEN: Making changes based on assumed rules without reading target file.

The ROUTING table still describes the pre-split repository. Of its 14 routed paths, 2 survive:

| Routed path | Status |
|---|---|
| `packages/api` | exists |
| `packages/cli` | exists |
| `packages/react`, `packages/studio` | **empty `node_modules` husks** — no tracked files |
| `packages/state`, `packages/screensets`, `packages/i18n`, `packages/auth`, `packages/framework` | gone (moved to `template-standard/packages/` in #456) |
| `presets/standalone`, `presets/monorepo` | gone |
| `src/mfe_packages`, `src/screensets`, `src/themes` | gone (now `template-standard/src-app/`) |

Meanwhile these real trees are unrouted: `packages/mfes`, `packages/gts-plugin`, `packages/cyber-pilot-kit-frontx`, `internal/eslint-config`, `internal/depcruise-config`, `template-standard/`.

The layer headings are also the retired model — `### SDK Layer (L1)`, `### Framework Layer (L2)`, `### React Layer (L3)` — rather than the pillar model of `architecture/DESIGN.md` §1.3 that #476 rewrote enforcement around. So an agent touching `packages/mfes` or `packages/gts-plugin` gets no route at all, and one reading the table for context gets a layering that no longer exists.

## Impact

Higher than ordinary doc rot, because the workflow is mandatory rather than advisory: the table is the first thing an agent consults and it cannot satisfy "MUST read target file" for most of the current tree. This is the documentation counterpart of #476 — the same post-split drift, in the file that instructs the tooling.

## Notes

Closed issue #356 covered `.ai/` staleness after the *MFE* migration (removed APIs in `.ai/targets/*`, missing rules, event terminology) — a different cause and different content; it did not touch ROUTING. Worth noting that #356's item 3 also looks unaddressed: `GUIDELINES.md` still opens REPO INVARIANTS with "Event-driven architecture only (see EVENTS.md)".

[...]

Issue #479, opened 2026-07-28
[FEATURE] Migrate to pnpm


hello1101n:
both gears-frontx and template-standard should be migrated

tscbmstubp:
Data point for this migration, found while working #476.

The committed `package-lock.json` was written by a newer npm than the one pinned expectations assume. On npm 11.7.0, any `npm install` silently rewrites it — `npm install --package-lock-only` produced **153 deletions**, stripping the `libc` metadata arrays that newer npm emits for optional platform binaries:

```diff
       "arm64"
     ],
     "dev": true,
-    "libc": [
-      "glibc"
-    ],
     "license": "SEE LICENSE IN LICENSE.md",
```

So whoever runs `npm install` on an older npm quietly downgrades the lockfile for everyone else, and the next person on a newer npm re-adds it — lockfile churn that looks like a real diff and carries no signal. I worked around it for #476 by hand-editing the two entries I needed and verifying with `npm ci --dry-run --offline` rather than letting npm rewrite the file.

Two things this suggests for the migration:

1. Whatever the outcome, pin the tool. `"packageManager"` in `package.json` (plus `engines`) makes the resolver version explicit instead of implicit — worth doing as a stopgap even before pnpm lands, since it makes this class of churn impossible.
2. Root `prepare` is `prek install`, which **fails outright** (exit 2) when hooks cannot be installed — for example when `core.hooksPath` is set, which `prek` refuses to override:
   ```
   error: Cowardly refusing to install hooks with `core.hooksPath` set.
   ```
   Because it runs from `prepare`, this fails `npm install` itself rather than just skipping hook setup. If the migration touches lifecycle scripts anyway, making hook installation non-fatal would remove an onboarding trap.

tscbmstubp:
Correction to my comment above: `packageManager` is already pinned — `"packageManager": "npm@10.9.4"` in root `package.json` (added in 317a7516), alongside `engines.npm: ">=10.0.0"`. I missed it before rebasing.

That doesn't change the observation, it sharpens it: the pin is declarative only unless Corepack is enabled, so it isn't being enforced. The machine I hit this on runs npm 11.7.0 against a pin of 10.9.4 and a lockfile written by something newer still — three different versions, no warning from any of them. So the gap is enforcement rather than declaration, which is worth deciding deliberately as part of the pnpm move (pnpm respects `packageManager` via Corepack the same way).

The `prepare: prek install` point stands unchanged.

PR #484, opened 2026-07-28
fix(arch): rewrite ecosystem layer enforcement for the pillar model, and make it actually enforce

### LINKED ISSUE #476: [ARCH] Ecosystem layer enforcement is orphaned and failing after the framework/template split

tscbmstubp:
Closes #476.

## What #476 asked for

> Rewrite `layer-constants.cjs` for the pillar model (Tooling → Core Framework, with the type-substrate port as an allowed edge), drop `framework.cjs`/`react.cjs`/`screenset.cjs` from the ecosystem's copy, then either wire both scripts into CI or delete them if `arch:check` has genuinely superseded them.

One squashed commit does that — and closes the vacuous-rule pattern it kept uncovering: guards that were green because they could not match anything, which is the same failure mode as an unrun guard.

## Rewrite for the pillar model

- `layer-constants.cjs` re-expressed as the DESIGN §1.3 pillar model: Core Framework (`api`, `gts-plugin`, `mfes`) and Tooling (`cli`, `cyber-pilot-kit-frontx`), replacing `SDK_PACKAGES` and the L1/L2/L3 chain.
- `sdk.cjs` → `core.cjs`; `framework.cjs`, `react.cjs`, `screenset.cjs` deleted. Those governed packages that emigrated to `template-standard/` in #456; the template enforces its own layering in its self-owned config.
- `mfes → gts-plugin` is no longer a false positive — the type-substrate port is an allowed edge, and `REQUIRED_ECOSYSTEM_EDGES` now asserts it *positively*, so silently dropping it fails rather than quietly passing.
- `sdk-layer-tests.ts` → `pillar-layer-tests.ts`. A missing manifest now fails instead of being skipped; the old skip is how #476 stayed hidden.
- `arch:sdk` → `arch:pillars`, `arch:deps:sdk` → `arch:deps:core`.

## Make the rules actually fire

The rewritten rules were green, but several could not match anything. The Core Framework rules keyed on `to: 'node_modules/@gears-frontx/'`, which is the one shape a workspace import almost never has — `preserveSymlinks` is false, so `@gears-frontx/mfes` resolves *through* the symlink to `packages/mfes/dist/index.cjs`. A deliberate forbidden import returned `✔ no dependency violations found`.

- Rules now match all three shapes an import can resolve to: symlink-resolved in-tree, under `node_modules/`, and the bare specifier an unresolvable template-side package falls back to (which is what keeps the template boundary enforced in the import graph, not just in manifests).
- `core-no-react` matched only `node_modules/react(/|$)`, so `react-dom` and bare-specifier React imports bypassed it entirely; both React packages now use the same three-shape patterns.
- The template's cross-package layer chain (`sdk-no-framework-import`, `framework-no-react`, `react-no-sdk`) could not fire in *either* tree state: installed, `exclude.path: 'packages/.*/dist'` deleted the resolved target node from the graph; uninstalled, the bare specifier never matched the source-path-only `to` patterns. `dist` moves from `exclude` to `doNotFollow` (visible as an un-traversed leaf rather than deleted) and `pkgTargets()` covers both shapes. `exclude` now path-excludes nothing — anything listed there is unusable as a rule's `to`, which is what silently disabled enforcement in the first place.
- `tsPreCompilationDeps: true` — without it the port's own `import type` edge is invisible. `no-circular` is overridden to skip type-only legs: the two cycles this surfaces in `mfes`/`api` are erased before emit and are not runtime cycles.
- **Verified by deliberate violation**, since green means nothing until it has gone red: L3→L1, L1→L2, L2→L3, sibling runtime import, sibling type-only import, template-side import, port provider overreaching, `react` and `react-dom` imports — each caught; the legitimate `gts-plugin → mfes` port and L3→L2 edges stay legal.

Manifest-side gaps the guard's own docstring claimed to cover:

- `internal/*` is a root workspace like `packages/*`, so its packages are now in scope (name → directory map, since a name no longer implies a location).
- Allowlists split `dependencies`/`peerDependencies` from `devDependencies`, so a runtime edge can't pass on a dev-group allowance while a legitimate shared lint-config devDep isn't reported as a layer violation. `packages/api` now declares the two config packages its own configs import (it relied on root hoisting).
- `cyber-pilot-kit-frontx → cli` removed from the allowlist — the ADR specifies command-surface-only with no package edge, so permitting it contradicted the rule the root config enforces.

Wiring, per "wire them in or delete them":

- `arch:deps:core`, `arch:pillars`, `arch:layers` all run in CI and pre-commit. `arch:deps:core` needs no build (a missing `dist` falls back to the bare specifier, which the rules match too); `arch:layers` builds `@gears-frontx/eslint-config` itself, because it reads that package's gitignored `dist/` and the pre-commit hook failed on a fresh clone without it.
- Both guards derive the repo root from their own location rather than `process.cwd()`, so they report the same thing from a subdirectory.

Also restores the React-*library* rule lost when the template's per-package configs were deleted — the same-named `framework-no-react` there guards the workspace package, not the library. `node_modules` had to come out of that config's `exclude` (which drops modules outright, unlike `doNotFollow`) or the restored rule would have been vacuous in the same way.

Guard count: 6 assertions → 15.

Template cleanup: dropped `arch:deps:sdk|framework|react` from the template's `package.json` and deleted the four orphaned per-package `.dependency-cruiser.cjs` files.

> **Correction — this paragraph originally claimed "No enforcement lost", and that was wrong.** Both halves of the justification were. Corrected in the review round below; left visible rather than quietly rewritten, since the whole point of this PR is that a guard's own claims about itself are worth checking.
>
> - The ownership-boundary argument holds for the four **per-package config files** only. The three npm scripts pointed at the template's *own* `internal/depcruise-config` copy — created by #470 precisely so it would not reach outside.
> - Enforcement **was** lost, twice: error-severity `no-circular` over the template packages (the surviving `no-circular-screenset-deps` is `warn` and scoped to `src-app/screensets`), and `sdk-no-gears-frontx-imports`, whose absence made `state -> @gears-frontx/auth` legal.

## The `types.ts` eslint-disable saga (why an earlier CI run failed)

`packages/api/src/types.ts:39` keeps its `@typescript-eslint/no-restricted-types` disable on `JsonCompatible = JsonValue | object`. An earlier revision of this branch *removed* it because ESLint reported the directive unused — but that report came from stale local `node_modules` holding ESLint 9, which only ever applies the root flat config. The repo pins ESLint 10.8.0, which resolves nested configs: for files under `packages/api`, CI applies `packages/api/eslint.config.js` → `sdkConfig` → `baseConfig`, where the rule bans `object`. So the directive looked unused locally but is load-bearing in CI. The `object` member itself is intentional — `JsonCompatible` must accept interface types without index signatures, which no lint-compliant alternative (`Record<string, unknown>` etc.) allows.

Takeaway for reviewers: linting this repo with pre-`npm install` node_modules gives misleading results, including false "unused directive" reports.

## Verification

`npm run lint` (under ESLint 10.8.0), `test:unit`, `arch:check` 12/12, `arch:pillars`, `arch:layers`, `arch:deps:core`, `template-standard`'s `arch:deps`, and `npm ci --dry-run` all pass. Both scripts type-check clean under `tsc --strict` (they're outside the root tsconfig — see #483).

## Out of scope, filed separately

- #482 — `.ai/GUIDELINES.md` ROUTING points at packages that no longer exist (12 of 14 paths dead). Same post-split drift, in the file that instructs the tooling; needs the guidelines correction policy.
- #483 — `template-standard/`'s own checks and `scripts/` type-checking are exercised by nothing. Same "exists but never runs" shape, in corners #476 doesn't reach.
- A note on #479 about lockfile churn across npm versions, found while updating the lockfile here.

## Summary by CodeRabbit

* **New Features**
  * Added additional CI architecture validation for core dependencies, pillar layering, and guard configurations.
  * Added local pre-commit hooks to run the same architecture checks on every commit.
  * Improved dependency analysis to include type-only imports and handle bare-specifier imports reliably.
* **Documentation**
  * Updated package architecture descriptions and boundaries to reflect the Core Framework and pillar model.
* **Chores**
  * Reworked architecture verification scripts and removed the older SDK-layer and related template config overrides.

---

## Amendment: "pillar" retired before merge

The architecture discussion in #495 retired the term **"pillar"**, and reserved "layer" for the `libs` / `templates` / `projects orchestration` taxonomy. This PR introduced ~24 new "pillar" mentions — among them a filename, an npm script, a pre-commit hook id, and a package description — so those names are changed to taxonomy-neutral ones **before** merge rather than being renamed a second time by the follow-up artifact work. (These files were already renamed once this quarter, from the retired SDK→framework→react vocabulary; neutral names end that cycle.)

Renamed:

| Before | After |
|---|---|
| `scripts/pillar-layer-tests.ts` | `scripts/package-edge-tests.ts` |
| `npm run arch:pillars` | `npm run arch:edges` |
| `arch-pillars` pre-commit hook | `arch-edges` |
| `internal/depcruise-config` description "…pillar layering" | "…boundary enforcement" |

23 files, no behaviour change. Beyond the referencing sites (`package.json`, `.pre-commit-config.yaml`, `main.yml`, `.cf-studio/config/artifacts.toml`, `layer-constants.cjs`, `core.cjs`, `index.cjs`, `verify-guard-configs.ts`, `.dependency-cruiser.cjs`, `packages/api/CLAUDE.md`, `llms.txt`, `template-shell/package.json`), also:

- `packages/cli/src/{index,commands/install,discovery/types}.ts`, its install-hook test, and `cyber-pilot-kit-frontx/src/extensions/discover-and-activate.ts` — "cross-pillar edge (F16 ← F10)" → "cross-package edge". Under the merged model both ends of that edge sit in the same layer, so the annotation no longer described a boundary.
- `eslint.config.js`, `template-shell/eslint.config.js`, `.dependency-cruiser.cjs` — "non-Pillar-1 packages" → "template-side packages", and the three section banners now name what they enforce (`CORE SUBSTRATE` / `CLI` / `AI TOOLING BOUNDARY ENFORCEMENT`).
- `.cf-studio/config/AGENTS.md` — "per-pillar package constraints" → "boundary constraints".
- `scripts/ecosystem-version-policy.mjs:34` — the citation quoted DESIGN's literal "three co-equal pillars"; it now cites §1.3 instead of quoting text that is about to change.

**Deliberately unchanged:**

- **No guard logic.** All 15 package-edge assertions, the three-shape import matching, the manifest edge allowlists, and the positive required-edge assertions are unchanged in behaviour. The membership sets are the same sets; only labels moved. `arch:edges`, `arch:guards` and `arch:deps:core` pass.
- **Nothing named `core`** — `core.cjs`, `arch:deps:core`, `CORE_FRAMEWORK_PACKAGES`. Only "pillar" was retired, and "core" is a plausible permanent name for the substrate partition; renaming it now would be a third rename.
- **`architecture/**`** — the PRD/DESIGN/DECOMPOSITION/ADR sweep is a separate unit tracked in #495. `ADR-0028` and `ADR-0029` cite "pillar parity" as a decision driver and are `accepted`, hence immutable for decision and rationale, so that sweep needs the new ADR to record how they carry forward.
- **`README.md`, `QUICK_START.md`, and the kit's `ecosystem-boundaries.md`** still say "pillar" (20 occurrences). Renaming them here would mean writing "three co-equal *parts*" — which keeps the tri-partite co-equality claim and merely swaps the noun, while the new model merges CLI and AI Tooling into one layer and #495 proposes removing the co-equality criterion outright. These need the ADR's vocabulary, not a synonym, so they belong with the artifact sweep.
- **The guards' own "layer" prose.** `layer-constants.cjs` and the depcruise configs still describe a "Core Framework layer"; that predates this PR. Re-expressing the guards in the new layer model is follow-up work, and the word for the partition *inside* `libs` — which is what `CORE_FRAMEWORK_PACKAGES` actually is — is still an open item on #495. This PR only stops introducing new model names, it does not pre-empt that decision.

The description above still uses "pillar" throughout and is left as written on purpose: it records why these guards were silently vacuous and how each rule was verified by deliberate violation. The vocabulary changed; the reasoning didn't.

---

## Amendment: review round (`656e4ebf`)

Four findings from @itechmeat, each verified against the tree before being fixed. Two were approval gates.

**The root config had the same defect as everything else this PR fixes.** The CI comment claimed `arch:check` covers the ecosystem import graph; it did not. `exclude` deleted `packages/*/dist` — the exact path a workspace import resolves to through `exports`/`main` — so a deliberate `@gears-frontx/api` import in `packages/mfes/src` returned `✔ no dependency violations found` in both tree states. `node_modules` was excluded too, which also disabled half of `frontx-mfes-4-type-format-dep`. Now: `dist` and `node_modules` under `doNotFollow`, a local `pkgTargets()` covering all three resolution shapes, and `tsPreCompilationDeps` with the `viaOnly` cycle exemption. Caught in both tree states, as runtime and as type-only.

**Nothing reconciled the layer lists against the workspaces.** The lists are hand-maintained while `workspaces` is a glob, so a new package was exempt from every guard by default — and `packages/telemetry` had already landed that way via #496. `arch:edges` now fails on any workspace no layer list knows, and on any listed package without an explicit allowlist entry. The rebase proved the point immediately by failing on `@gears-frontx/ui-kit`, which arrived on develop in the meantime.

Classification, both provisional and both flagged for #495:

- `telemetry` → Core Framework, deliberately the *strictest* layer: it forbids React and every cross-package edge, which is what the package already is. If #495 moves it somewhere looser, nothing was under-enforced meanwhile.
- `ui-kit` → a new `INTERIM_UNCLASSIFIED_PACKAGES` list rather than a layer. It is a React component library, so Core Framework would be a false assertion and Tooling a different one. Members are still allowlisted and still fail the coverage check if dropped; they just do not get a layer's rule set applied on a guess. The list should be empty.

**Two template rules had no survivor** — restored, see the correction above.

**Five dead config files deleted.** `template-shell/internal/depcruise-config/` had zero consumers and docstrings still naming the deleted scripts. `template-shell/internal/` no longer exists.

Also, while rebasing: develop's two new `frontx-ui-kit-interim-*` rules and its `ui-kit` addition to the telemetry rule were source-path-only, carrying the same vacuity this PR exists to remove. They use `pkgTargets()` now — the isolation rule genuinely fires where it previously did not.

Guard count: **21 assertions** (6 before this PR, 15 at review time).

Known and deliberately not fixed here: un-excluding `node_modules` feeds more edges to the nine `STUB` rules, which are `from: {}, to: {}` and warn on everything, taking `arch:deps` output from ~5,000 warning lines to ~8,900. Zero errors. That is the same vacuity in mirror image — a rule matching everything carries no more signal than one matching nothing — and it deserves its own change.


internal/depcruise-config/core.cjs
```diff
@@ -0,0 +1,98 @@
...
+        'CORE VIOLATION: The type-system provider may only import the runtime it implements the type-substrate port for. No other @gears-frontx import is permitted.',
+    },
+    {
+      name: 'core-no-react',
+      severity: 'error',
+      from: { path: CORE_FRAMEWORK_SRC_PATTERN },
+      to: { path: '(^|/)node_modules/react(/|$)' },
+      comment:
+        'CORE VIOLATION: Core Framework packages cannot import React. The substrate must stay UI-framework-agnostic (cpt-frontx-principle-agnostic-core).',
+    },
```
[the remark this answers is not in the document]
tscbmstubp: Valid — fixed in 822f1796, taking the suggested pattern.

The inconsistency was mine: I applied the three-shape reasoning to `headless-packages-no-react-library` in the template config and then failed to carry it back to `core-no-react`, in the same commit that documented the three shapes. Verified by deliberate violation from `packages/api/src` — `react-dom` now trips `core-no-react`, and plain `react` still does.

template-standard/.dependency-cruiser.cjs
```diff
@@ -92,9 +107,20 @@ module.exports = {
...
+    // `node_modules` is deliberately NOT excluded: `exclude` drops modules from
+    // the graph outright, so excluding it would leave
+    // `headless-packages-no-react-library` with no `to` node to ever match.
+    // `doNotFollow` above already keeps third-party modules as un-traversed
+    // leaves, which is all the size saving that mattered here.
     exclude: {
       dynamic: true,
-      path: 'packages/.*/dist|node_modules',
+      path: 'packages/.*/dist',
     },
```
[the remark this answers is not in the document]
tscbmstubp: Valid, and it turned out to be worse than described — fixed in 822f1796.

You were right that the resolved path is `packages/<pkg>/dist/...` because `exports`/`main` point there, and that `exclude` deletes the node the rules need. What that means concretely is that **three** rules could never fire, not just the React one: `sdk-no-framework-import`, `framework-no-react`, and `react-no-sdk` — the entire L1 -> L2 -> L3 chain this config exists to enforce.

And it failed in *both* tree states, for two different reasons:

- **Installed:** resolves to `packages/state/dist/index.js`, which `exclude.path: "packages/.*/dist"` then removes from the graph. Your finding.
- **Uninstalled** — the normal state for this tree, since it is consumed as a template rather than installed: the import is unresolvable, so `resolved` is the bare specifier `@gears-frontx/state`, which the source-path-only `to` patterns (`^packages/(state|i18n)/`) never matched either.

Took your `doNotFollow` suggestion for `dist`, and added a `pkgTargets()` helper so each rule covers both shapes. `exclude` now path-excludes nothing at all, since anything listed there is unusable as a `to` — that is precisely what had silently disabled both React and cross-package enforcement.

Verified by deliberate violation, each caught: L3 -> L1, L1 -> L2, L2 -> L3, and a `react-dom` import. The legitimate L3 -> L2 edge stays clean, and the baseline is clean at 614 modules / 1539 dependencies.

Worth noting these rules were vacuous before this PR too — nothing in CI runs this config, which is filed as #483.

internal/depcruise-config/core.cjs
```diff
@@ -0,0 +1,102 @@
...
+// so counting them here would report two pre-existing non-problems in `mfes` and
+// `api` and pressure someone into reshaping working code. The rule keeps its
+// name — `verify-guard-configs.ts` asserts base inheritance by that name — and
+// keeps full strength for cycles that survive to runtime.
+const TYPE_ONLY_DEPENDENCY_TYPES = ['type-only', 'type-import'];
+const baseNoCircular = base.forbidden.find((rule) => rule.name === 'no-circular');
+const runtimeOnlyNoCircular = {
+  ...baseNoCircular,
+  to: { ...baseNoCircular.to, dependencyTypesNot: TYPE_ONLY_DEPENDENCY_TYPES },
+};
```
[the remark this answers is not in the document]
tscbmstubp: Fixed in de7c875d: the exemption now lives in `viaOnly.dependencyTypesNot` (dependency-cruiser is pinned at 17.3.10, so the v16+ `viaOnly` sub-restriction is available). Validated both directions: a deliberate runtime cycle in `packages/api/src` is reported, a deliberate type-only cycle is not, and the clean baseline stays clean. `verify-guard-configs.ts` still passes (the rule keeps its `no-circular` name).

.github/workflows/main.yml
```diff
@@ -71,6 +71,21 @@ jobs:
     - name: Run architecture validation
       run: npm run arch:check
 
+    # arch:check covers the ecosystem import graph via the root config
```
itechmeat, line 82: This comment says arch:check covers the ecosystem import graph via the root config, but the root config was not part of this fix and still has the shape this PR fixes elsewhere: its exclude deletes packages/*/dist nodes from the graph, and its to: patterns match source paths only. I checked by adding a deliberate import of @gears-frontx/api into packages/mfes/src: arch:deps and the arch:check sub-checks stay green in both tree states, with and without dist. So the sentence asserts coverage that does not exist. Could we either apply the same three-shape fix to the root config here, or reword the comment and link a follow-up issue so the gap stays visible?

internal/depcruise-config/layer-constants.cjs
```diff
@@ -1,40 +1,206 @@
...
+// though the edge only exists at lint/build time. Directory names under
+// `internal/`.
+const INTERNAL_TOOLING_PACKAGES = Object.freeze(['eslint-config', 'depcruise-config']);
+
+// Every ecosystem package, in layer order.
+const ECOSYSTEM_PACKAGES = Object.freeze([
+  ...CORE_FRAMEWORK_PACKAGES,
+  ...TOOLING_PACKAGES,
+  ...INTERNAL_TOOLING_PACKAGES,
+]);
```
itechmeat, line 92: After the develop merge, packages/telemetry exists as a workspace package but none of the three new guards knows it: it is absent from every list in layer-constants.cjs, arch:edges runs 15 assertions with no mention of it, and arch:deps:core does not cruise its source. The lists are hand-maintained while the root workspaces field is a glob, and there is no reciprocal check, so any new package is silently unchecked - the same drift mode #476 describes, already live on this branch. Could the script scan the workspace globs and fail on a package that is missing from the layer lists? Where telemetry itself belongs is a separate question (DESIGN does not mention it either, that is #496's debt), but the guard should fail loudly on an unknown package rather than stay silent.

template-shell/.dependency-cruiser.cjs
```diff
@@ -65,21 +87,36 @@ module.exports = {
       name: 'sdk-no-framework-import',
       severity: 'error',
       from: { path: '^packages/(state|i18n)/' },
-      to: { path: '^packages/(framework|react)/' },
+      to: { path: pkgTargets('framework', 'react') },
```
itechmeat, line 137: [Not a blocker] The body says no enforcement was lost, but two rules did not survive: the deleted scripts carried a general no-circular at error severity over the template packages (the surviving no-circular-screenset-deps is warn and scoped to src-app/screensets only), and sdk-no-gears-frontx-imports forbade any @gears-frontx import from state/i18n, while the surviving sdk-no-framework-import covers only framework and react - so state importing @gears-frontx/api is now legal. Also, the three deleted scripts pointed at template-shell's own internal/depcruise-config copy (created by #470 precisely to avoid reaching outside), not across the ownership boundary - that reasoning applies to the four per-package config files only. Nothing ran these scripts before (#483), so nothing breaks today, but could we restore the two rules in the template root config, or state the loss explicitly?

template-shell/package.json
```diff
@@ -56,9 +56,6 @@
     "lint:framework": "eslint packages/framework/src --max-warnings 0",
     "lint:react": "eslint packages/react/src --max-warnings 0",
     "lint": "eslint . --max-warnings 0",
-    "arch:deps:sdk": "npx dependency-cruiser packages/state/src packages/i18n/src --config internal/depcruise-config/sdk.cjs --output-type err-long",
```
itechmeat, line 59: Deleting the three arch:deps:* scripts leaves template-shell/internal/depcruise-config/ (five .cjs files) with zero consumers - they still ship with the template and their docstrings still name the deleted scripts. Present-but-dead guard config is the exact shape this PR removes elsewhere; could they be deleted in the same change (or restored together with the rules from the previous point)?

itechmeat (changes requested):
Solid and much-needed work, and the direction is exactly right - the guards now demonstrably fire, and the earlier review threads were handled thoroughly. Most of my notes are minor polish, but two of them I consider approval gates (posted inline); both look fixable within this PR. Happy to approve once those are addressed.

.github/workflows/main.yml
tscbmstubp: Valid, and the fix is in `656e4ebf` rather than a reword — the sentence was worth making true.

Reproduced your test exactly: a deliberate `@gears-frontx/api` import in `packages/mfes/src` returned `✔ no dependency violations found` under the old root config, in both tree states. Two independent reasons, the same pair as the template config in #484's earlier thread:

- **Installed:** resolves through `exports`/`main` to `packages/api/dist/index.cjs`, which `exclude.path: 'packages/.*/dist'` then deletes from the graph.
- **Uninstalled:** unresolvable, so `resolved` is the bare specifier, which the source-path-only `to` never matched.

`node_modules` was in `exclude` too, which had a second casualty I had not noticed: `frontx-mfes-4-type-format-dep` keys half its target on `node_modules/@globaltypesystem/`, so that half could not fire either.

Changed:

- `dist` and `node_modules` move to `doNotFollow`. `exclude` now holds only fixture/generated trees that are not any rule's `to`.
- A local `pkgTargets()` covers all three resolution shapes, applied to the five `frontx-single-intra-ecosystem-edge-*` rules and to MFES-4.
- `tsPreCompilationDeps: true`, with the type-only cycle exemption in `viaOnly` — mirroring what `core.cjs` already does — since without it an `import type` crossing was invisible here.

Verified by deliberate violation, caught in both tree states and as both a runtime and a type-only import: `→ packages/api/dist/index.cjs` with `dist` present, `→ @gears-frontx/api` without.

One consequence worth flagging rather than burying: un-excluding `node_modules` puts more edges in the graph, and the nine `STUB` rules are `from: {}, to: {}`, so they warn on every single one. `arch:deps` output goes from ~5,000 warning lines to ~8,900. Still 0 errors. That is the same vacuity in mirror image — a rule matching everything carries no more signal than one matching nothing — but scoping the stubs is a separate change and I did not want to fold it in here.

internal/depcruise-config/layer-constants.cjs
tscbmstubp: Valid, and the reciprocal check is in `656e4ebf`. You were right that this is the interesting half — the specific missing package is a symptom, the missing check is the bug.

`arch:edges` now resolves the root `workspaces` globs and fails on any package no layer list knows. It also fails on a classified package with no explicit `ALLOWED_ECOSYSTEM_EDGES` entry, so `?? {}` cannot quietly stand in for a reviewed decision. An unsupported glob shape throws rather than being skipped — a glob the resolver silently ignored would hide exactly what it exists to find.

The rebase onto develop made the case for it better than I could have: `@gears-frontx/ui-kit` had landed in the meantime, and the check failed on it immediately.

On placement, where I made a judgment call you should feel free to reject:

- **`telemetry` → Core Framework, provisionally.** Into the *strictest* layer on purpose: it forbids React and forbids every cross-package edge, which is what the package already is (one third-party dep, zero `@gears-frontx` edges) and what the root config's `telemetry-standalone` rule already says. If #495 moves it somewhere looser, nothing was under-enforced meanwhile. Marked provisional in the comment with the ticket.
- **`ui-kit` → a new `INTERIM_UNCLASSIFIED_PACKAGES` list, not a layer.** It is a React component library, so Core Framework would be a false assertion and Tooling would be a different false one. That list is not an escape hatch: members are still allowlisted, still have to be added deliberately, and still fail the coverage check if dropped. They just do not get a layer's rule set applied on a guess. The list should be empty, and a non-empty entry is a debt with a ticket on it.

If you would rather `telemetry` also sit in the pending list until #495 rules, say so — that is a one-line change and I have no strong attachment to the classification, only to it not being silently absent.

While rebasing I also noticed develop's two new `frontx-ui-kit-interim-*` rules and its `ui-kit` addition to the telemetry rule were source-path-only, i.e. carrying the same defect this PR fixes. They now use `pkgTargets()` as well. Verified: a `@gears-frontx/ui-kit` import from `packages/telemetry/src` trips `frontx-ui-kit-interim-not-imported-by-ecosystem`, which it did not before.

`arch:edges` is 21 assertions, up from 15.

template-shell/.dependency-cruiser.cjs
tscbmstubp: You are right on all three points, and thank you for checking the claim rather than the summary — "no enforcement lost" was wrong as written.

Both rules are restored in `656e4ebf` rather than documented as lost:

- **`no-circular` at error severity**, over the template packages. Your reading is exact: the surviving `no-circular-screenset-deps` is `warn` and scoped to `src-app/screensets`, so it was not a substitute. Restored with the same `viaOnly.dependencyTypesNot` exemption `core.cjs` uses, since `tsPreCompilationDeps` is on in this config and type-level cycles are erased before emit.
- **`sdk-no-gears-frontx-imports`**, and your example was the live hole: `sdk-no-framework-import` names only `framework` and `react`, so `state → @gears-frontx/auth` was legal. The restored rule covers all six sibling packages with an own-package exemption. I kept `sdk-no-framework-import` alongside it deliberately — it states the layer chain and gives the better message when the chain is what broke — with a comment recording that the overlap is intentional.

And you are right about the ownership-boundary reasoning: the three deleted scripts pointed at template-shell's *own* `internal/depcruise-config` copy, created by #470 precisely to avoid reaching outside. That argument only ever applied to the four per-package config files, and I applied it to both. The PR description overstated it.

Verified by deliberate violation, each caught: a runtime cycle in `packages/state/src`, and `state → @gears-frontx/auth`. A type-only cycle is correctly not reported. Baseline stays clean.

template-shell/package.json
tscbmstubp: Agreed, and deleted in `656e4ebf` — all five files.

Confirmed they had zero consumers first: `template-shell/.dependency-cruiser.cjs` is standalone and requires none of them, nothing else in the tree references the directory, and no npm script survives that loaded them. Their docstrings still named `arch:deps:sdk|framework|react`, which is how they read as live config.

Taken together with the previous point, this is the "restored together with the rules" branch of your either/or: the two rules that mattered are back in the template's own config, and the dead copies are gone. `template-shell/internal/` no longer exists.

template-shell/.dependency-cruiser.cjs
tscbmstubp: Follow-up: the PR description itself is now corrected too, not just the code. The "No enforcement lost" paragraph carries an inline correction naming both wrong claims — the two rules that were actually lost, and the ownership-boundary argument applying only to the four per-package config files. Left visible rather than quietly rewritten, since a guard's own claims about itself being worth checking is the whole point of this PR.

internal/depcruise-config/layer-constants.cjs
tscbmstubp: Follow-up on the placement question, and on an inconsistency in my own answer above.

Offering you the telemetry/pending-list swap left two mechanisms for one situation with nothing saying why. `676b0917` writes the distinction down in `layer-constants.cjs` rather than resolving it by making the two uniform:

- **Provisional member of a real layer** (telemetry, in Core Framework). Use when a layer's rules are *all true* of the package today. It gets that layer's full enforcement — no React, no cross-package edge, and inclusion in the source set `arch:deps:core` cruises. #495 changes the label, not whether it is guarded.
- **`INTERIM_UNCLASSIFIED_PACKAGES`** (ui-kit). Use when *no* layer's rules are true, so joining one would assert something false.

The test is "are this layer's rules true of the package", not "is this layer a plausible eventual home". Moving telemetry to the interim list for symmetry would trade real enforcement for tidiness, which seems the wrong direction while the taxonomy is unsettled — but the offer stands if you disagree, and now the trade-off is explicit rather than buried in a review thread.

Separately, I owe you a correction on my claim above that the `frontx-ui-kit-interim-*` rules "did not fire before". That was true but I had asserted it before testing it. Now run against `origin/develop`'s actual config with a planted `@gears-frontx/ui-kit` import in `packages/telemetry/src`: `frontx-ui-kit-interim-not-imported-by-ecosystem` produces nothing in either tree state, and fires with the `pkgTargets()` version. One nuance my original wording glossed — with `dist` absent, develop still caught that particular import via `frontx-telemetry-1-no-template-content`, a different rule. The named interim rule was vacuous; the violation was not completely invisible in that one state.

Comment-only change, no behaviour difference: arch:edges 21/21, arch:deps:core and arch:guards unchanged.

itechmeat (approved):

GeraBart (approved):

PR #490, opened 2026-07-29
Split template-standard into template-shell + template-mfe

gs-layer:
Closes #470

## What

`template-standard` is split into two independently applicable templates, exercising the multi-template assembly path (ADR-0030/0031/0032) on real content for the first time outside unit tests.

- `e9f7eb84` **feat: split** — 736 files (707 pure renames): `template-standard/` -> `template-shell/` (identity + npm package renamed to `@gears-frontx/frontx-template-shell` across all three name axes), `src-app/mfe_packages/*` carved into `template-mfe/` (add-only), manifests per ADR-0031, AI bundles split per template namespace, all monorepo references updated (CI, root configs, kit guard, tests, `artifacts.toml`).
- `bae04b57` **fix: align** — template solution packages aligned with gts-plugin-owned action ids (upstream drift from `c679f1f3`; restores DTS builds and 306/306 + 135/135 framework/react tests).
- `4f40e749` **fix: conflict report** — assembly refusal now names each contested ground and its contestants (ADR-0032 Confirmation); adds 12 split e2e/pinning fixtures.

## Key decisions

- **No preset in this PR** — deferred: `referencedTemplates[].ref` carries no version/source (irreproducible, vs ADR-0017 drivers), upgrade targets `records[0]` (#488), fail-late resolution. Old identity `frontx-template-standard` now fails fast instead of silently changing behavior.
- **`template-mfe` is add-only** (`frontx add` into an applied shell). Standalone/exclusive ownership of root configs is mutually exclusive with add-ability; seeding it into an empty dir is characterized (works, produces a non-buildable tree, documented precondition in `src-app/mfe_packages/README.md`).
- **`sharedFiles: []` on both** — every shell->MFE link is glob/scan (workspaces glob, directory scanners); region-union is unusable for `add` until #487.
- Trailing-slash convention for directory boundaries (string-prefix matching audit: without it `"src"` claims all 333 MFE files).

## Validation

- Manifests VALIDATED; declaration coverage: shell 376/377 + manifest, mfe 338/340 + allow-list, 0 uncovered.
- Fixtures (CLI suite 183/183): union-fidelity seed+add parity, add-flow byte-invariance (374/374 shell files), dup-claim REFUSED naming ground/contestants, R3 characterizing (nested/spelling blind spot), coverage scan, known-defect pinning for #487/#488.
[...]

Issue #489, opened 2026-07-29
normalize MFE file: dependencies and undeclared mf-gts coupling in template-mfe


gs-layer:
## Description

MFE packages in `template-mfe` carry dangling `file:` dependencies and one undeclared coupling. `npm ls` reports `ELSPROBLEMS` and the resolution graph works partly by accident.

> **Revised 2026-07-30.** The original text gave two causes for the undeclared coupling; investigation disproved both, and they pointed the work in the wrong direction. Corrections are marked below. Half of this issue has since landed — see Status.

1. **Dangling `file:` deps**: `_blank-mfe`, `widgets-fixture-a`, `widgets-fixture-b` declare `@gears-frontx/{api,gts-plugin,mfes}` as `file:../../../packages/{api,gts-plugin,mfes}` — directories that do not exist. The same applies to `{auth,framework,i18n,react,state}`, which resolve to `template-shell/packages/*` in an assembled project but to nothing inside this monorepo, since `template-mfe/packages/` does not exist.

2. **Undeclared coupling**: all four MFEs — `_blank-mfe`, `demo-mfe`, `widgets-fixture-a`, `widgets-fixture-b` — import `@gears-frontx/frontx-template-shell/build/mf-gts` in their `vite.config.ts`; only `_blank-mfe` declares the dependency (in `devDependencies`, as `file:../../..`).

## Corrections to the original text

**The stated resolution mechanism was wrong.** The original attributed it to "hoisting caused by `_blank-mfe`'s declaration plus the root `overrides`". Neither holds:

- the monorepo's 14 root `overrides` contain **no** `@gears-frontx/*` entry;
- the actual provider is `template-shell/packages/framework`, which declares the shell both in `devDependencies` (`file:../..`) and as an optional `peerDependencies: "*"`. Combined with `overrides: {"@gears-frontx/frontx-template-shell": "file:."}` in the shell's root manifest — which reaches an assembled project byte-for-byte, `package.json` being the shell's exclusive subtree — that is what produces the symlink.

**"Removing `_blank-mfe` from the template would silently break the other three" is false.** Verified on a reproduced seeded project: removing the declaration from all four MFEs left the symlink alive and `mf-gts` resolving; removing `packages/framework`'s devDep as well (leaving only the optional peer) also left it working. The real fragility is elsewhere — the whole coupling rests on a `peerDependencies` entry inside the *other* template, which nothing in `template-mfe` can reference or protect.

[...]

Issue #485, opened 2026-07-29
[BUG] Seeded project cannot install: ecosystem dependencies are pinned by monorepo-local `file:` paths


itechmeat:
## Description

A project produced by `frontx seed` declares its FrontX ecosystem dependencies as `file:` paths that only resolve inside the gears-frontx monorepo. The first `npm install` in a freshly seeded project fails on missing paths, so the seed flow ends one step short of a working project.

The template's `file:` specifiers fall into two classes, and only one of them is broken:

**Escapes the template subtree - broken in a seeded project (5 specifiers, 2 files):**

- `template-standard/package.json:71-73`
  ```
  "@gears-frontx/api": "file:../packages/api",
  "@gears-frontx/gts-plugin": "file:../packages/gts-plugin",
  "@gears-frontx/mfes": "file:../packages/mfes",
  ```
- `template-standard/packages/framework/package.json:45-46`
  ```
  "@gears-frontx/mfes": "file:../../../packages/mfes",
  "@gears-frontx/gts-plugin": "file:../../../packages/gts-plugin",
  ```

All five resolve to the monorepo root's `packages/*`, which exists only where `template-standard/` sits next to `packages/`. A seeded project has no such sibling: the manifest's `ownershipBoundaries.exclusiveSubtrees` ships the template's own subtree and nothing above it.

**Stays inside the template subtree - fine as is:** the `file:../../../packages/*` references in `template-standard/src-app/mfe_packages/*/package.json` resolve to `template-standard/packages/*` (the template's own subpackages, shipped as sources via the `"packages"` boundary), and the `"file:."` self-reference in the root `package.json`. These travel with the project and keep resolving.

`template-standard/README.md` marks the pinning as a known alpha limitation ("This template pins the FrontX ecosystem packages by local path for pre-release development; published releases resolve them from the npm registry"), but no issue tracks it.

**Publishing is not the blocker.** All three packages are already in the npm registry at exactly the version the monorepo holds locally: `@gears-frontx/api`, `@gears-frontx/mfes` and `@gears-frontx/gts-plugin` all have `0.3.0-alpha.0` under the `alpha` dist-tag, matching `packages/*/package.json`. Swapping the five specifiers to `0.3.0-alpha.0` would work today with no extra publishing step.

[...]

PR #491, opened 2026-07-29
feat: run orphaned checks — template-standard CI validation and scripts/ type-checking (#483)

### LINKED ISSUE #483: [ARCH] Make template validation gates runnable and keep CI/tool paths current

tscbmstubp:
Closes #483.

## Part 1 — `template-standard/` checks now run in CI

New `template-validate` job in `main.yml`:
- Builds the ecosystem packages the template `file:`-links (`mfes`, `gts-plugin`, `api`) first, since npm installs those as symlinks and the template packages build against their `dist/`.
- Globs `template-*/` and runs each template's own `npm ci` + build + `arch:deps` + `lint` + `test:unit` — no hardcoded template names, so the job survives the planned template-shell/template-mfe split (#470) unchanged.

Running the checks locally immediately surfaced two breakages that had rotted unenforced, proving the issue's point:
1. **Template build was broken on develop**: c679f1f3 moved `FRONTX_ACTION_LOAD/MOUNT/UNMOUNT_EXT` from `@gears-frontx/mfes` to `@gears-frontx/gts-plugin`, but 8 files in the template's framework package still imported them from `mfes`. Imports re-pointed (gts-plugin was already a declared peer dep; downstream consumers go through framework's re-export chain, so nothing else changed).
2. **15 lint errors** from the compiler-powered `react-hooks` v6 rules (`refs`-during-render, `set-state-in-effect`) across the react package hooks and demo screens. All fixed with behavior-preserving restructures (render-time state adjustment, lazy `useState` initializers, deriving instead of storing) — no `eslint-disable` for these. The only disable added is on TanStack Table's `useReactTable` (`incompatible-library` is informational: the React Compiler skips that component). Also added `dist-lib/` to the template's eslint ignores.

## Part 2 — `scripts/` is now type-checked

- New `tsconfig.scripts.json`: Node lib, `types: ["node"]`, `checkJs` for the JSDoc-typed `.mjs` tooling; `scripts/manual-auth-tests/` excluded (run-by-hand demos importing template-owned packages not resolvable from the ecosystem root).
- Wired in as `type-check:scripts` inside `type-check:all`, so the existing CI step covers it with no workflow change.
- First-run backlog was ~48 errors — all fixed with type-only changes: proper typedefs (`PackageJson`, `Project`), narrowing production signatures to the structural subsets they consume (`SpawnedChildLike`, `DirEntryLike`, `FileSystemLike`) instead of casting test fakes, and JSDoc return-type drift corrections in `ecosystem-version-policy.mjs`. No runtime behavior changes.

Also widened the `check-json` pre-commit exclude from `tsconfig.json` to `tsconfig*.json` (tsconfigs are JSONC).

## Verification

- Root: `lint`, `type-check:all` (now incl. scripts), `arch:check`, `test:unit` — all exit 0; script tests 183/183.
- Template: `arch:deps`, `lint`, `build:packages`, `test:unit` — all exit 0.

## Out of scope (known leftovers)

- Template `arch:deps:sdk/framework/react` still reference `../internal/` — tracked as a sub-task on #470.
- One pre-existing type error in the template's `ExtensionDomainSlot.test.tsx` (mock missing new `TypeSystemPlugin` members) — only surfaces via the template's own type-check, which CI doesn't run yet.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved template app behavior when switching bridge instances, including theme, language, subscriptions, and text direction updates.
  * Ensured right-to-left languages correctly update text direction from initial and subsequent settings.

* **Tests**
  * Added coverage for bridge replacement and synchronized UI settings.
  * Expanded automated validation for templates, including builds, architecture checks, linting, type-checking, and unit tests.
  * Added dedicated validation and testing for repository scripts.
  * Improved test fixture error reporting and type coverage.


.github/workflows/main.yml
```diff
@@ -71,6 +71,59 @@ jobs:
...
+  # enforced only when someone runs them by hand — and they rot (#483). The
+  # glob loop deliberately avoids hardcoding template names so the job
+  # survives the planned template-shell/template-mfe split (#470).
+  template-validate:
+    name: Template Validate
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v5
```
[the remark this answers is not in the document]
tscbmstubp: Fixed in 27b79cd9 — added `permissions: contents: read` on the job and `persist-credentials: false` on checkout.

template-standard/src-app/mfe_packages/_blank-mfe/src/screens/home/HomeScreen.tsx
```diff
@@ -45,8 +45,18 @@ interface HomeScreenProps {
...
+  // this avoids an extra render and the set-state-in-effect anti-pattern. The effect
+  // below only subscribes for subsequent property changes.
+  const [theme, setTheme] = useState<string>(() => {
+    const initial = bridge.getProperty(FRONTX_SHARED_PROPERTY_THEME);
+    return initial && typeof initial.value === 'string' ? initial.value : 'default';
+  });
+  const [language, setLanguage] = useState<string>(() => {
+    const initial = bridge.getProperty(FRONTX_SHARED_PROPERTY_LANGUAGE);
+    return initial && typeof initial.value === 'string' ? initial.value : 'en';
+  });
```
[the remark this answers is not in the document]
tscbmstubp: Fixed in 27b79cd9 — theme/language are re-read during render when the `bridge` identity changes ("adjusting state during render" with a tracked `prevBridge`), in all four listed screens plus ProfileScreen.tsx, which had the same pattern.

template-standard/src-app/mfe_packages/demo-mfe/src/screens/profile/ProfileScreen.tsx
```diff
@@ -53,8 +53,18 @@ const languageModules = import.meta.glob('./i18n/*.json') as Record<
...
+  // this avoids an extra render and the set-state-in-effect anti-pattern. The effect
+  // below only subscribes for subsequent property changes.
+  const [theme, setTheme] = useState<string>(() => {
+    const initial = bridge.getProperty(FRONTX_SHARED_PROPERTY_THEME);
+    return initial && typeof initial.value === 'string' ? initial.value : 'default';
+  });
+  const [language, setLanguage] = useState<string>(() => {
+    const initial = bridge.getProperty(FRONTX_SHARED_PROPERTY_LANGUAGE);
+    return initial && typeof initial.value === 'string' ? initial.value : 'en';
+  });
```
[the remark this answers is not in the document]
tscbmstubp: Fixed in 27b79cd9 — the `dir` assignment moved into a `useEffect` keyed by `[language]`, covering the initial language; the duplicated logic in the subscription callback was removed. Applied uniformly to all five screens. (Note: this gap predates this PR — the develop version also only set `dir` inside the subscription callback.)

.github/workflows/main.yml
```diff
@@ -71,6 +71,59 @@ jobs:
...
+      uses: actions/setup-node@v5
+      with:
+        node-version: '25.x'
+        cache: 'npm'
+        cache-dependency-path: |
+          package-lock.json
+          template-*/package-lock.json
+
+    - name: Enable Corepack
+      run: corepack enable
```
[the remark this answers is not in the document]
tscbmstubp: Not changed, intentionally: this finding doesn't hold on GitHub-hosted runners. The Template Validate run for this PR executed `corepack enable` on Node 25.x and succeeded (run 30508953905, job "Template Validate", 2026-07-30T02:37:21Z — the step completed and `npm ci` proceeded normally), because the ubuntu-latest runner image provides corepack independently of the Node 25 tarball. The pre-existing `validate` matrix job has been running the identical step on 25.x as well. Removing or special-casing the step in only this job would diverge from the rest of the workflow; if Node-25 corepack provisioning ever becomes a real problem it should be fixed workflow-wide.

tscbmstubp:
Note for reviewers: [review 4814781500](https://github.com/constructorfabric/gears-frontx/pull/491#pullrequestreview-4814781500) was generated against `66e49fa8`, before the review-fix commit `27b79cd9` landed (CodeRabbit's re-review of that commit was rate-limited). Both of its findings — initial RTL direction for the lazily-read language, and theme/language resync on `bridge` replacement — are fixed in `27b79cd9` across all five MFE screens (via a `dir` effect keyed on `[language]` and a render-time `prevBridge` adjustment).

gs-layer:
### Code review

**Merge order: this PR must not be merged before #490.**

#490 ("Split template-standard into template-shell + template-mfe") relocates the entire `template-standard/` tree — 367 files to `template-shell/`, 341 to `template-mfe/`. 23 of the 38 files in this PR live under `template-standard/**`, and every one of them has a new home in #490:

- `template-standard/packages/**` -> `template-shell/packages/**` (framework import re-routing, `useApiStream`, `useCanAccess`, `useScreenTranslations`, `ExtensionDomainSlot`, `queryClient`)
- `template-standard/src-app/mfe_packages/**` -> `template-mfe/src-app/mfe_packages/**` (all 8 demo/blank MFE screen fixes, `data-table`, `date-picker`)

Merging this first would leave #490 to reconcile 23 modified files against a 733-file rename, which is exactly the case where rename detection is least reliable. Merging #490 first makes this PR the one that gets rebased, and it is the far smaller side of the two.

Direct file overlap between the two PRs is otherwise small and mechanical: `.github/workflows/main.yml` (#490 only edits one comment line — `template-standard/` -> `template-shell/` — and adds no job of its own), plus `scripts/check-test-dependency-versions.mjs`, `scripts/run-monorepo-unit-tests.test.mjs` and `scripts/test-runner/discovery.mjs` (a handful of lines each).

Worth noting the `template-*/` glob in the new job is the right call here — the comment already anticipates the split via issue #470, and after #490 lands the loop picks up both `template-shell/` and `template-mfe/` with no workflow change:

https://github.com/constructorfabric/gears-frontx/blob/27b79cd9629ef6731fc495bdec2a4fdc838b9ed4/.github/workflows/main.yml#L77-L81

---

Found 1 issue (minor):

1. The `eslint-disable` justification overstates what the rule permits. It says the warning "only means this component is skipped by compilation, not that it misbehaves", but `react-hooks/incompatible-library` also covers passing values returned by the unanalyzable hook into other memoized components — and `table` is passed to `<DataTablePagination table={table} />` further down the same file. Suggest narrowing the wording rather than the disable itself.

https://github.com/constructorfabric/gears-frontx/blob/27b79cd9629ef6731fc495bdec2a4fdc838b9ed4/template-standard/src-app/mfe_packages/demo-mfe/src/components/ui/data-table/data-table.tsx#L59-L62

Otherwise clean. Checked: the `FRONTX_ACTION_*` re-routing covers exactly the 8 files that imported those constants directly from `@gears-frontx/mfes` (downstream consumers resolve through the framework re-export); the render-time state adjustments in `useApiStream`, `useCanAccess` and `date-picker` compare primitive fields and converge in one pass; the RTL `dir` effect fires on first commit because `containerRef` is attached in both the skeleton and the loaded branch; the `scripts/` changes are type-only; no `@cpt-` marker was orphaned or unpaired.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

template-mfe/src-app/mfe_packages/demo-mfe/src/screens/uikit/UIKitElementsScreen.tsx
```diff
@@ -73,30 +80,31 @@ const languageModules = import.meta.glob('./i18n/*.json') as Record<
...
+  );
+  // The lazy initializers above run only on mount; if the host swaps the bridge
+  // instance, re-read its current properties during render ("adjusting state
+  // during render") — the subscription effect only delivers future changes.
+  const [prevBridge, setPrevBridge] = useState(bridge);
+  if (prevBridge !== bridge) {
+    setPrevBridge(bridge);
+    setTheme(readBridgeProperty(bridge, FRONTX_SHARED_PROPERTY_THEME, 'default'));
+    setLanguage(readBridgeProperty(bridge, FRONTX_SHARED_PROPERTY_LANGUAGE, 'en'));
+  }
```
itechmeat, line 101: The `RTL_LANGUAGES` const, the `readBridgeProperty` helper, the `prevBridge` re-read block and the `dir` effect are pasted verbatim into five screens, and only HomeScreen got a test for the new bridge-swap behavior. Four of the five copies live in demo-mfe, which already has `src/shared/` - would a shared hook there (e.g. `useBridgeProperty` + `useHostDirection`) cover them, with the test moving to the hook?

.github/workflows/main.yml
```diff
@@ -71,6 +71,64 @@ jobs:
...
+          (
+            cd "$template"
+            npm ci
+            npm run build:package --if-present
+            npm run build:packages --if-present
+            npm run arch:deps
+            npm run lint
+            npm run test:unit
+          )
+        done
```
itechmeat, line 261: The job runs build, arch:deps, lint and test:unit, but not the template's own `type-check` - by this PR's own logic that script stays orphaned, and the PR body itself names a type error in `ExtensionDomainSlot.test.tsx` that only the template's type-check would catch. Is that a deliberate scope cut, and if so could a follow-up issue carry it (same question for `build:mfes` / `generate:mfe-manifests`)?

template-mfe/src-app/mfe_packages/demo-mfe/src/screens/uikit/UIKitElementsScreen.tsx
```diff
@@ -123,6 +125,16 @@ export const UIKitElementsScreen: React.FC<UIKitElementsScreenProps> = ({ bridge
...
 
+  // Keep the Shadow DOM host's text direction in sync with the active language.
+  // An effect keyed by `language` (rather than logic inside the subscription
+  // callback) also covers the initial language, which never fires a callback.
+  useEffect(() => {
+    const rootNode = containerRef.current?.getRootNode();
+    if (rootNode && 'host' in rootNode) {
+      (rootNode.host as HTMLElement).dir = RTL_LANGUAGES.includes(language) ? 'rtl' : 'ltr';
+    }
+  }, [language]);
```
itechmeat, line 138: The comment at line 106 says "language + direction come from useScreenTranslations", but the new effect below now owns direction, and the screen keeps its own language subscription too. Could the comment be updated to match - or is direction actually meant to move into that shared hook?

itechmeat:
@tscbmstubp Merge-order heads-up: this PR points the template's FRONTX_ACTION_* imports at @gears-frontx/gts-plugin, but the published gts-plugin@0.3.0-alpha.0 predates that move, and #492 pins the template to the registry - so whichever merges second goes red. Suggestion: merge #492 first, publish the next ecosystem alpha, then rebase this PR bumping the pins to it.

gs-layer:
Found while working on #489 — the new `template-validate` job has a guard that will behave differently than intended once the split lands, and the difference is about to become a red CI rather than a silent gap.

### The guard skips half the split

```bash
for template in template-*/; do
  if [ ! -f "$template/package.json" ]; then
    continue
  fi
```

`template-mfe/` is an add-only overlay: it has **no root `package.json`** by design (its `frontx-template.json` declares only `src-app/mfe_packages/` and `.frontx/ai/@gears-frontx/frontx-template-mfe/`, and `package.json` is `template-shell`'s exclusive subtree, so the mfe template cannot claim one). So as written, the loop silently skips it and validates `template-shell` only.

The job's own comment says the glob "deliberately avoids hardcoding template names so the job survives the planned template-shell/template-mfe split (#470)". The glob does survive — the guard one line below is what defeats it. Since this PR exists to fix "checks exist, nothing runs" (#483), it seemed worth flagging that the same pattern reappears on the other half.

`template-shell` is genuinely covered: it has every script the job calls (`arch:deps`, `lint`, `test:unit`, `build:package`, `build:packages`).

### And it is about to flip to a hard failure

Work on #489 adds a monorepo-only build harness at `template-mfe/package.json` (commit `b63faffe` on `feature/489-template-mfe-harness`, not yet a PR). Without it the MFE packages resolve nothing inside the monorepo — `npm install` exits 0 while `require.resolve('@gears-frontx/frontx-template-shell/build/mf-gts')` throws `MODULE_NOT_FOUND`. The harness is deliberately undeclared in the manifest, so it never ships into a seeded project; it exists purely so this repo can build and type-check those packages.

Side effect: **the guard stops skipping `template-mfe`**, which is what you want — but the job then calls `arch:deps`, `lint` and `test:unit` on it **without `--if-present`** (only the two build steps carry that flag), and the harness deliberately carries no scripts. That is a failing job the first time both changes meet.

Two ways out, and it seems better to pick one before the merge rather than discover it in CI:

1. add `--if-present` to the three unflagged calls, matching the build steps; or
2. decide the harness should carry those three scripts, in which case #489 adds them.

Preference is (1) — it keeps the harness minimal and version-free, which is the property that stops it drifting out of step with `packages/*/package.json`. Happy to go with (2) if you would rather the job stay strict.

### Minor, same job

The ecosystem-build step is justified by:

> Templates link the ecosystem packages (mfes, gts-plugin, api) via file: paths into this repo's packages/ … npm installs those as symlinks, so the ecosystem dist/ must exist before the template's own packages can build against them.

#492 replaces those `file:` paths with exact registry pins, after which nothing is symlinked and the step becomes a slow no-op with a comment describing something that no longer happens. Nothing breaks — just worth a follow-up touch depending on merge order.

Also unrelated to this PR but adjacent: the pre-existing per-MFE loop at `.github/workflows/main.yml:41` iterates `src/mfe_packages/*/`, a path that exists nowhere in this repo (the real one is `src-app/mfe_packages/*`). It has been running on nothing — the same class this PR is fixing, if you want to sweep it here.

template-mfe/src-app/mfe_packages/demo-mfe/src/screens/uikit/UIKitElementsScreen.tsx
tscbmstubp: Agreed on the shape — tracked as #502, deliberately out of this PR.

One constraint worth recording: only four of the five copies can move. `no-cross-mfe-imports` is error-severity in `template-shell/.dependency-cruiser.cjs`, so `_blank-mfe` cannot import from `demo-mfe`, and `HomeScreen` — the one copy that has the bridge-swap test — is the one that has to keep its own. That matches existing convention rather than fighting it: `useScreenTranslations.ts` and its test already live verbatim in both packages' `src/shared/`, and demo-mfe's `shared/README.md` scopes that directory to reuse between that package's own screen entries.

So #502 is `useBridgeProperty` + `useHostDirection` in `demo-mfe/src/shared/` covering the four demo-mfe screens, with the test moving onto the hook; `_blank-mfe` keeps its copy, which seems right for a blank starter people copy from anyway. The rule's `^src-app/mfe_packages/shared/` escape hatch would cover all five, but no such package exists and creating one for a single hook pair is a bigger call than this warrants — noted in the issue, not assumed.

Your other point on this PR is fixed in c36941f9: the comment above the theme/language subscription no longer claims direction comes from `useScreenTranslations`, since the effect below owns it.

Also fixed there, from your review of the workflow: the template's own `type-check` now runs in both template paths. You were right that leaving it out contradicted the premise of the PR — and it caught nothing only because it passes, not because it was covered. `build:mfes` / `generate:mfe-manifests` are still not run; they sit behind the full `npm run build`, which I left out on cost grounds rather than principle, so say the word if you would rather the job run it.

.github/workflows/main.yml
tscbmstubp: Not a deliberate scope cut — just wrong on my part, and both are now in rather than deferred.

`type-check` landed in c36941f9, and the full `npm run build` in 7b3d5bc1, so `build:mfes` and `generate:mfe-manifests` are covered too. Both template paths now run `build` ahead of the other checks so they see the generated manifests.

I had left the build out on the assumption a full vite build was too expensive for CI. Measuring said otherwise: 1:12 for the standalone template, 2:03 for the composed one, and both were already green — so there was nothing to weigh. Standalone degrades cleanly (`build-mfes.ts` and `generate-mfe-manifests.ts` each report zero packages and skip); the composed run builds all four MFEs plus the host bundle.

Worth noting your premise held up under the wider sweep. Two more of the same kind turned up and are fixed here:

- `scripts/` had 183 vitest tests across 4 files that nothing ran — `test:unit` fans out to workspaces and `scripts/` is not one, so no project collected them, including the tests this PR itself adds to three of those files. Wired up via `vitest.scripts.config.ts`.
- the `Install MFE dependencies` step looped over `src/mfe_packages/*/`, a path that has not existed since the MFEs moved under the templates. Deleted.

template-mfe/src-app/mfe_packages/demo-mfe/src/screens/uikit/UIKitElementsScreen.tsx
tscbmstubp: Comment updated — c36941f9. It now reads:

```
// Subscribe to theme and language; direction is derived from `language` by
// the effect below, and useScreenTranslations only consumes the language.
```

which matches what the code actually does: the screen owns both subscriptions, the effect below owns direction, and `useScreenTranslations` only consumes the language rather than producing it.

On your second branch — yes, direction is meant to move, but into a hook that does not exist yet rather than into `useScreenTranslations`. That is #502 (`useBridgeProperty` + `useHostDirection` in `demo-mfe/src/shared/`), split out of this PR. Folding direction into `useScreenTranslations` specifically seems worth avoiding: it would give a translations hook a DOM side effect on the Shadow DOM host, and the two have different lifetimes — direction has to be set on first commit before any language change fires a callback, which is exactly why the effect is keyed on `language` rather than living in the subscription.

Worth knowing for #502: only four of the five copies can move. `no-cross-mfe-imports` is error-severity, so `_blank-mfe` cannot import from `demo-mfe`, and `HomeScreen` — the one copy that has the bridge-swap test — keeps its own. Same shape as `useScreenTranslations`, which is already duplicated verbatim across both packages' `src/shared/`.

.github/workflows/main.yml
```diff
@@ -71,6 +64,194 @@ jobs:
...
+            if (!s || !s.length) {
+              console.error("no exclusiveSubtrees declared in " + process.argv[1]);
+              process.exit(1);
+            }
+            console.log(s.join("\n"));
+          ' "$overlay/frontx-template.json")"
+
+          while read -r subtree; do
+            [ -n "$subtree" ] || continue
+            [ -d "$overlay/$subtree" ] || continue
```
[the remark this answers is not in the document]
tscbmstubp: Fixed in `29e8bc37`. Agreed, and it is the same class of bug this job exists to catch — a declared subtree that is not on disk means the manifest and the tree disagree, and skipping it leaves that overlay unvalidated while `mfe_count` still passes on another overlay's packages.

Both of `template-mfe`'s declared subtrees are present today, so this was latent rather than firing. Verified the guard by deleting `.frontx/ai/@gears-frontx/frontx-template-mfe/` from a copy of the overlay: the step now exits non-zero with `::error::… declares exclusiveSubtree … which does not exist`, where before it composed the remaining subtree and passed.

.github/workflows/main.yml
```diff
@@ -71,6 +64,194 @@ jobs:
...
+        if [ "$mfe_count" -eq 0 ]; then
+          echo "::error::composition produced no MFE packages under $shell/src-app/mfe_packages"
+          exit 1
+        fi
+        echo "==> Composed $mfe_count MFE package(s) into $shell"
+
+        # Each MFE ships a standalone package-lock.json whose file: targets
+        # point into template-mfe/packages/*, which does not exist. npm ignores
+        # nested locks inside a workspace, but drop them so nothing reads them.
+        rm -f "$shell"/src-app/mfe_packages/*/package-lock.json
```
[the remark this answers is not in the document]
tscbmstubp: Fixed in `29e8bc37`, and this one was real — thanks for catching it.

`cp -R src dest` copies *into* `dest` when `dest` already exists, so the packages would land at `src-app/mfe_packages/mfe_packages/*` and the `-mindepth 2 -maxdepth 2` count would miss them. As you noted, the failure would surface as a misleading "composition produced no MFE packages" rather than pointing at the nesting.

Now `mkdir -p "$shell/${subtree%/}"` + `cp -R "$overlay/$subtree/." "$shell/${subtree%/}/"`, which merges regardless of whether the destination pre-exists.

Verified against three cases:

| case | before | after |
| --- | --- | --- |
| destination absent (today's state) | 4 MFEs | 4 MFEs |
| destination pre-exists | **0** MFEs, `mfe_packages/mfe_packages/` created | 4 MFEs, no nesting, pre-existing file preserved |
| declared subtree missing | silently skipped | exits 1 (separate fix, `discussion_r3688495160`) |

`template-shell` has no `src-app/mfe_packages` today, which is why CI is green — this was latent. Worth noting it would not have stayed latent: `template-shell/.dependency-cruiser.cjs` already has an `^src-app/mfe_packages/shared/` escape hatch with no corresponding directory, so the first commit to add that shared folder would have tripped it.

tscbmstubp:
@gs-layer thanks — you were right, and it is fixed in `29e8bc37`.

I went with manifest-based classification rather than `--if-present`, because the guard's job is to decide *which kind* of template each directory is, and `package.json` presence is only a proxy for that. The templates already declare it: `template-shell` claims `package.json` in `ownershipBoundaries`, `template-mfe` claims only `src-app/mfe_packages/` and its `.frontx/ai/**` subtree. So the job reads the manifests and splits into self-contained vs overlay, mirroring the assembler's ownership semantics (`packages/cli/src/scaffold/assembler.ts`) instead of inferring from paths.

That matters for the #489 collision specifically. Once `template-mfe` gains a `package.json` for the `file:` normalization, a `-f package.json` guard — or `--if-present` — would classify it as self-contained and try to `npm ci && npm run build` an add-only overlay with no root manifest semantics, which is the red CI you predicted. I verified this against a stub `template-mfe/package.json`: the old guard classifies it self-contained, the manifest classifier still says overlay. `--if-present` would also have quietly weakened the standalone path, where a missing script *should* fail.

Overlays are now validated by composing them onto the shell the way `frontx seed` + `frontx add` would, then running the full check set against the composed tree — 4 MFEs overlaid, 125 files that eslint had never seen, `arch:deps` from 335 to 470 modules cruised.

Your `src/mfe_packages/*/` observation was also correct and is actioned: that loop in the `validate` job pointed at a path that does not exist, so it was a second orphaned check hiding in the same file. Removed, with a comment explaining why nothing replaced it.

On the ecosystem-build step's `file:`-symlink rationale being obsoleted by #492 — agreed in principle, but that step is doing real work independent of the pins: the composed tree's type-check and build resolve `@gears-frontx/*` through the workspace, so the packages have to be built first. I have reworded the comment to say that rather than leaning on the `file:` explanation. Happy to revisit once #492 lands if you think the step itself can go.

tscbmstubp:
@itechmeat this one has dissolved — the PR no longer touches `FRONTX_ACTION_*` or the pins at all, so there is no longer an ordering constraint against #492.

What happened: `develop` re-pointed those imports to `@gears-frontx/gts-plugin` independently, via `typeSystem.resolveMountExtActionId()` / `resolveUnmountExtActionId()`. After rebasing onto `develop`, those files dropped out of this PR's diff entirely. Current state:

```
$ git diff --name-only origin/develop..HEAD | wc -l
33
$ git diff origin/develop..HEAD | grep "FRONTX_ACTION_\|gts-plugin.*[0-9]\.[0-9]"
(no source or pin changes — only workflow comments)
```

So #491 and #492 can merge in either order. Worth flagging that the overlap left a real trap during the merge, not just a stale diff: `develop`'s version and the branch's version merged *cleanly* in `TestContainerProvider.ts` and produced a broken file — develop had dropped the constants from the `@gears-frontx/mfes` import while the branch added a new `gts-plugin` import for them, and the merge kept both, giving `TS6192: All imports in import declaration are unused` and a failed framework dts build. Clean textual merge, red CI. Only running the checks caught it.

Also for the record: #490 merged this morning, so the merge-order concern @gs-layer raised about the 733-file relocation is settled — this PR's 23 files under `template-standard/**` are now at their `template-shell/**` and `template-mfe/**` homes, and the rebase took them there rather than replaying the old paths.

package.json
```diff
@@ -52,8 +53,9 @@
     "prek:run": "prek run --all-files",
     "prepare": "prek install",
     "test": "npm run test:unit",
-    "test:unit": "node scripts/run-monorepo-unit-tests.mjs --run",
-    "test:unit:watch": "node scripts/run-monorepo-unit-tests.mjs --watch"
+    "test:unit": "node scripts/run-monorepo-unit-tests.mjs --run && npm run test:unit:scripts",
```
itechmeat: This PR turns `test:unit` into an `&&` chain, and the existing CI step still calls `npm run test:unit -- --coverage`. npm appends post-`--` args to the end of the expanded script, so `--coverage` now lands on the trailing `npm run test:unit:scripts` and is swallowed as an unknown npm config - I reproduced this on both npm 10.9.4 and 11.6.2, and `npm_config_coverage` does not enable vitest coverage. That would mean the 70/70/70/60 thresholds in `template-shell/vitest.shared.ts` stop being enforced everywhere, with only an `npm warn` line inside a green step named "Run unit tests with coverage gating" - which is the exact failure class this PR exists to close. Could the chaining move into a separate script (e.g. `test:unit:all`) so the flag keeps reaching the runner, or `--coverage` move into the config instead? Separately, the comment above that step points at `vitest.shared.ts`, which only exists under `template-shell/`.

template-shell/packages/react/src/hooks/useCanAccess.ts
```diff
@@ -75,26 +85,35 @@ export function useCanAccess<TRecord extends AccessRecord = AccessRecord>(
...
-  if (storedKey !== stableKey) {
-    setStoredKey(stableKey);
+  if (runState.app !== app || runState.auth !== auth || runState.stableKey !== stableKey) {
+    setRunState({ app, auth, stableKey });
     setStableQuery(query as AccessQuery);
+    // @cpt-begin:cpt-frontx-flow-auth-plugin-rbac-guard:p1:inst-pending-effect
+    // The re-pessimize half of the pending instruction lives here (render
+    // time) rather than in the effect below — see the comment there.
     setResult({ allow: false, isResolving: true });
+    // @cpt-end:cpt-frontx-flow-auth-plugin-rbac-guard:p1:inst-pending-effect
```
itechmeat, line 103: This adds a second `inst-pending-effect` marker region at :98-102, while the same instruction id already marks :107-119 - so one instruction now labels two disjoint pieces of code. More importantly, the no-auth behavior that region used to perform now lives at :144 as `return auth ? result : DENIED_NO_AUTH;`, outside every marker region (the file's ends are at 104/102/119/134/141). The same shape in `useScreenTranslations.ts:79-86`: the `inst-begin-load` region kept its markers but lost the `isLoading` state it was named for. Neither id appears in any document under `architecture/`, and artifacts.toml excludes these trees from validation, so nothing can catch the drift - would it be better to keep the pending behavior inside its region and drop the duplicate pair, rather than leave a marker chain that cannot be checked against anything?

.github/workflows/main.yml
```diff
@@ -71,6 +64,205 @@ jobs:
...
+        set -euo pipefail
+        node -e '
+          const fs = require("fs");
+          const selfContained = [], overlays = [];
+          for (const dir of fs.readdirSync(".").filter((d) => d.startsWith("template-")).sort()) {
+            if (!fs.existsSync(dir + "/frontx-template.json")) continue;
+            const b = JSON.parse(fs.readFileSync(dir + "/frontx-template.json", "utf8")).ownershipBoundaries || {};
+            const claims = [
+              ...(b.exclusiveSubtrees || []),
+              ...(b.sharedFiles || []).map((f) => (typeof f === "string" ? f : f.path)),
```
itechmeat: [Not a blocker] Three things in the classify/compose steps. The `typeof f === "string"` branch at :142 looks unreachable - `manifest/validate-contract.ts:191-213` rejects any non-object sharedFiles entry, so it guards a shape we forbid. At :144, because `claims` flattens exclusiveSubtrees and sharedFiles together, a template declaring `package.json` as a `region-union` shared file (a legal strategy) owns only a region of someone else's file yet classifies as self-contained. And the copy loop reads only `exclusiveSubtrees`, silently ignoring sharedFiles, while `scaffold/compose-shared-files.ts` exists for exactly that and `assembler.ts:15-19` treats both as ownership - which makes the comment at :183-185 about mirroring the assembler's semantics only half true. All three are latent today (both manifests ship `"sharedFiles": []`) - would an explicit refusal when an overlay declares a non-empty sharedFiles be cheaper than half-implementing the contract?

package.json
```diff
@@ -52,8 +53,9 @@
...
+    "test:unit:watch": "node scripts/run-monorepo-unit-tests.mjs --watch",
```
itechmeat, line 67: [Not a blocker] `test:unit` now chains `test:unit:scripts`, but `test:unit:watch` still runs only the monorepo runner - so the 183 tests this PR connects stay unrun in watch mode, which is the same gap one level down. The `&&` also means a failure in the workspace fan-out short-circuits before the scripts project runs, so a red run carries no signal about `scripts/` at all.

template-shell/packages/react/src/hooks/useApiStream.ts
```diff
@@ -70,18 +74,45 @@ export function useApiStream<TEvent>(
   /** When true, connect's promise resolution must tear down the new id instead of adopting it. */
   const disconnectRequestedRef = useRef(false);
   // Latest descriptor for connect/disconnect without tying effect or callbacks to object identity.
+  // Written in a deps-less effect (runs after every render, during commit) rather than
+  // during render itself — refs must not be mutated while rendering.
   const descriptorRef = useRef(descriptor);
-  descriptorRef.current = descriptor;
+  useEffect(() => {
+    descriptorRef.current = descriptor;
+  });
```
itechmeat: [Not a blocker] Here, `.../src/mfe/components/ExtensionDomainSlot.tsx:92-101` and `.../src/queryClient.tsx:440-447`.

The `ref.current = latest` write moves into a deps-less effect in three places, with the same rationale sentence pasted in each, and there is no `useLatest`-style helper anywhere in template-shell or packages - #502 covers the five MFE screens, so these three need their own pointer. Worth noting the obvious remedy mostly does not apply: React 19.2.4 does export `useEffectEvent`, but it wraps a function, so it fits `ExtensionDomainSlot`'s callbacks and not the two value-holding refs, and `@types/react` is pinned at 19.0.8 where the export does not exist yet. One shared helper for the value case, or a follow-up issue?

.github/workflows/main.yml
```diff
@@ -71,6 +64,205 @@ jobs:
...
+          while read -r subtree; do
+            [ -n "$subtree" ] || continue
+            # A declared subtree that isn't there means the manifest and the
+            # tree disagree. Skipping it would leave that overlay unvalidated
+            # while the aggregate count below still passes on another one's
+            # packages — the same vacuous pass this job exists to catch.
+            if [ ! -d "$overlay/$subtree" ]; then
+              echo "::error::$overlay declares exclusiveSubtree '$subtree', which does not exist"
+              exit 1
+            fi
```
itechmeat, line 320: [Not a blocker] The comment at :216-219 says the directory check stops one overlay riding on another's packages, but :220 only proves the directory exists, and the count at :237-238 is taken once after the loop - so an overlay whose subtree exists but holds no package.json still contributes zero and passes (`template-mfe/src-app/mfe_packages/README.md` is exactly that shape). Unreachable today with a single overlay, so either a per-overlay count inside the loop or a weaker comment would do. Small related thing: :237 hardcodes `$shell/src-app/mfe_packages`, which is the path-guessing the :183-185 comment says the copy loop reads the manifest to avoid.

template-shell/packages/react/src/hooks/useApiStream.ts
```diff
@@ -60,7 +60,11 @@ export function useApiStream<TEvent>(
 
   const [data, setData] = useState<TEvent | undefined>(undefined);
   const [events, setEvents] = useState<TEvent[]>([]);
-  const [status, setStatus] = useState<StreamStatus>('idle');
+  // Lazy initializer mirrors what the "connecting" render-adjustment below would set on
+  // a later (descriptorKey, mode, enabled) change — the adjustment's before/after
+  // comparison can't itself detect "this is the very first render", since its own
+  // `prevConnectKey` state is seeded from the same initial connectKey.
+  const [status, setStatus] = useState<StreamStatus>(() => (enabled ? 'connecting' : 'idle'));
```
itechmeat, line 120: [Not a blocker] `status` now initialises to `'connecting'` when enabled and the return is derived as `enabled ? status : 'idle'`, and since `enabled` defaults to true and nothing ever writes `'idle'` into state again, that value is unobservable for the default configuration - the first committed frame changed from `'idle'` to `'connecting'`. That is a public contract change on a hook carrying `@cpt-dod`/`@cpt-flow` markers. I checked before asking: `use-api-stream` appears in no file under `architecture/` and in none of the 18 FEATURE specs, so there is nothing to contradict - but should the status lifecycle be recorded somewhere before it becomes folklore?

.github/workflows/main.yml
```diff
@@ -30,20 +30,13 @@ jobs:
...
-            (cd "$mfe" && npm ci)
-          fi
-        done
+    # NOTE: an "Install MFE dependencies" step used to sit here, looping over
+    # `src/mfe_packages/*/`. That path has not existed in this repo since the
+    # MFEs moved under the templates — the loop matched nothing and the root
+    # `type-check:mfe` script it existed to feed is gone too, so it had been
+    # running on empty for a while. Removed rather than repointed: the MFE
+    # packages are template content, and the template-validate job below is
+    # what installs and checks them now (#483).
```
itechmeat: [Not a blocker] Two comments read as change-history rather than as descriptions of the code: the seven-line tombstone at `main.yml:33-39` for the removed "Install MFE dependencies" step, and `useScreenTranslations.ts:191-195` starting "`isLoading` was previously a stored flag". Both describe something that is not there any more, which git already carries. The invariant inside the second one is worth keeping though - "on error loadedLanguage is never advanced, so the mismatch alone gates isLoaded false" - just in the present tense.

scripts/test-runner/discovery.mjs
```diff
@@ -14,6 +14,15 @@ import { readdir, readFile } from 'node:fs/promises';
 import path from 'node:path';
 import { defaultRepoRoot } from './common.mjs';
 
+/**
+ * Structural subset of `node:fs`'s `Dirent` that this module actually reads.
+ * Narrowing away from `typeof readdir`'s real (heavily overloaded) return
+ * type lets unit tests inject a plain-object fake directory listing instead
+ * of constructing real `Dirent` instances.
+ *
+ * @typedef {{ name: string; isDirectory: () => boolean }} DirEntryLike
```
itechmeat: [Not a blocker] `DirEntryLike` is declared in both here and `check-test-dependency-versions.mjs:68`, and the shapes already differ - the second adds `isSymbolicLink`. No compile conflict since both are module-scoped, but same name / different contents tends to drift further. Worth one shared typedef, or distinct names?

itechmeat (changes requested):

package.json
tscbmstubp: Fixed as part of conflict resolution, not by keeping this chain — the merge with `develop` replaced this branch's own `test:unit:scripts` chaining mechanism with `develop`'s: `scripts/` is now registered as a `repo-scripts` project inside `run-monorepo-unit-tests.mjs` itself, so `test:unit` is back to a single command:

```
"test:unit": "node scripts/run-monorepo-unit-tests.mjs --run"
```

`npm run test:unit -- --coverage` now passes `--coverage` straight to that one invocation, so it can't be swallowed by a trailing chained script. Same reason the `test:unit:watch`-gap and `&&`-short-circuit comments below no longer apply — the chaining they're about is gone. Current head: d133bd78.

package.json
tscbmstubp: Resolved by the merge, same fix as the coverage-flag reply above: `repoScriptsProject` is now registered unconditionally in `discovery.mjs`'s `loadProjects()`, and both `test:unit` (`--run`) and `test:unit:watch` (`--watch`) call `loadProjects()` through the same runner. Watch mode now fans out over `scripts/` too — no separate chained script to short-circuit past. Current head: d133bd78.

.github/workflows/main.yml
tscbmstubp: Checked this against the current head and the original PR commit: `npm run type-check` is present in both halves — line 187 of the self-contained step and again in the composition step after the shell rebuild — and both were already there in the original commit (`cf3fce44`), unchanged by the conflict-resolution merge. The adjacent comment on `npm run build` (lines 182-185) is what covers `build:mfes`/`generate:mfe-manifests` before `type-check` runs. Looks like this one may have been filed against a partial read rather than an actual gap — let me know if I'm missing the step you meant.

tscbmstubp:
Re the merge-order heads-up: this is closed out now. #492 already merged into `develop` on 2026-07-31, before I touched this branch, and while resolving conflicts here I also bumped `template-shell`'s pins to `0.3.0-alpha.1` (the alpha that carries the `FRONTX_ACTION_*` move into `gts-plugin`). Template Validate is green against that on the current head (`d133bd78`), covering both the self-contained and shell+overlay composition paths.

template-mfe/src-app/mfe_packages/demo-mfe/src/screens/uikit/UIKitElementsScreen.tsx
tscbmstubp: Confirmed as filed — this file wasn't touched by the conflict-resolution merge, so the duplication across the five screens (only HomeScreen tested) stands as-is. That's a design call for @tscbmstubp, not something the merge changed.

template-mfe/src-app/mfe_packages/demo-mfe/src/screens/uikit/UIKitElementsScreen.tsx
tscbmstubp: Also unaffected by the merge — same text was already there in the original PR commit (cf3fce44). Leaving the wording call to @tscbmstubp.

template-shell/packages/react/src/hooks/useCanAccess.ts
tscbmstubp: Fixed in 831dd8ab, went with the fix you suggested:

- `useCanAccess.ts`: dropped the duplicate `inst-pending-effect` region at the render-time re-pessimize block (it's already inside `inst-stable-key`) and moved the marker to wrap the effect's guard clause *and* the final `return auth ? result : DENIED_NO_AUTH` — that return is the no-auth half of the same instruction, so it's now inside the region instead of outside every one.
- `useScreenTranslations.ts`: dropped `inst-begin-load` from the shared marker stack over `loadedLanguage`/`error` — it named a flag (isLoading) that's gone, and nothing in that state pair signals "load began" anymore. The other four ids sharing that declaration still apply and are untouched.

Verified: build, type-check, lint, and the react package's 137 unit tests all pass. Pushed to the branch, CI is running now.

.github/workflows/main.yml
tscbmstubp: Fixed in 5ea827c0, went with the refusal you suggested rather than half-implementing the contract:

- Dropped the `typeof f === "string"` branch — confirmed it's dead: `validate-contract.ts`'s sharedFiles loop (:191-213) rejects anything that isn't a well-formed object with `path`/`mergeStrategy`/`ownedRegions`, so a valid manifest can never hit the string case.
- `claims` now reads only `exclusiveSubtrees` — sharedFiles no longer feeds classification, so a region-union share of package.json can't get misread as full ownership.
- Added an explicit refusal right where `claims` used to flatten both in: if any manifest declares a non-empty `sharedFiles`, the job now fails with a message pointing at both steps that would need extending (Classify + composition), instead of silently misclassifying or silently skipping the merge.

Verified: ran the extracted classify script against the real manifests (unchanged output: template-shell self-contained, template-mfe overlay), and against a manifest with a synthetic non-empty sharedFiles entry to confirm the refusal actually fires. Pushed, CI running now.

template-shell/packages/react/src/hooks/useApiStream.ts
tscbmstubp: Deferred to a follow-up: #515.

Tried the shared `useLatest` helper first, and it exposed a real wrinkle worth recording rather than working around silently: `eslint-plugin-react-hooks`'s `exhaustive-deps` only recognizes a ref as omittable-from-deps when it's a literal `useRef(...)` in scope, not when it comes back from a custom hook. Adding the ref to deps clears the "missing dependency" error, but not the separate "ref value will likely have changed by the time this cleanup function runs" one — which fires wherever the ref is read inside a returned cleanup closure (`latestRef` here, `onDetachedRef` in ExtensionDomainSlot.tsx). That's exactly where `descriptorRef`/`onAttachedRef` differ from `latestRef`/`onDetachedRef`: the first two aren't read in a cleanup closure and convert cleanly, the second two are and don't. Zero existing `eslint-disable` for this rule anywhere in the codebase, so accepting one here would be a first — write-up and options are in #515 for whoever picks it up.

.github/workflows/main.yml
tscbmstubp: Fixed in 83f5dbee, went with the per-overlay count:

- Moved the package.json count inside the per-overlay loop, right after that overlay's own subtrees are copied — no longer one aggregate taken after every overlay has landed.
- Counted from `$shell/${subtree%/}` (the subtree path the manifest itself named), not the hardcoded `$shell/src-app/mfe_packages` — fixes the path-guessing you flagged too, same spot the composition step's own comment says the copy loop avoids by reading the manifest.
- An overlay whose subtree exists but holds no package.json now fails immediately, attributed to that overlay specifically.

Verified against a scratch copy of the real templates (unchanged count: template-mfe contributes 4), then reproduced your exact scenario with two synthetic overlays — a working one first, a docs-only one second. Confirmed the old aggregate would have passed (4 > 0) while the new per-overlay check fails specifically on the second one. Pushed, CI running now.

.github/workflows/main.yml
tscbmstubp: Fixed in 808f330b (DirEntryLike) and 0d5b893c (both change-history comments) — same push.

- Shared DirEntryLike moved to scripts/fs-types.mjs, imported by both check-test-dependency-versions.mjs and test-runner/discovery.mjs. discovery.mjs only ever called isDirectory()/name on these, so the wider shape (with isSymbolicLink) is a safe superset — confirmed by the scripts type-check, which caught two test fakes in run-monorepo-unit-tests.test.mjs missing it and needing the update.
- Trimmed the main.yml tombstone to the one present-tense fact worth keeping (MFE packages are checked in template-validate, not here).
- Reworded the useScreenTranslations.ts comment to state the invariant directly: on error, loadedLanguage never advances to currentLanguage, so the mismatch alone gates isLoaded false.

Verified: full root type-check:all, lint, and the scripts test suite (201 tests) all green, plus a targeted type-check on @gears-frontx/react. Pushed, CI running now.

scripts/test-runner/discovery.mjs
tscbmstubp: Fixed in 808f330b — shared it. Moved the (wider, isSymbolicLink-including) shape into scripts/fs-types.mjs and had both files import it via `@typedef {import('...').DirEntryLike}`. discovery.mjs only ever read isDirectory()/name off these entries, so widening it is a safe superset — the scripts type-check confirmed that, and caught two test fakes in run-monorepo-unit-tests.test.mjs that needed isSymbolicLink added to keep satisfying the type.

template-mfe/src-app/mfe_packages/demo-mfe/src/screens/uikit/UIKitElementsScreen.tsx
tscbmstubp: Correction on one part of this: the comment fix you noted as done in c36941f9 did not actually make it into the PR's current base. I checked cf3fce44 (the commit this PR started from after whatever squash/rebase happened) byte-for-byte, and the comment at UIKitElementsScreen.tsx:106-107 still had the pre-fix wording verbatim — so it got lost somewhere between c36941f9 and cf3fce44, not carried forward. Redone now in 7ec1384a (present on the branch as of this comment), reworded to remove the ambiguous 'derived from X by Y, and Z' parse rather than restoring the old wording verbatim. Worth knowing in case anything else from that commit didn't survive the squash either.

The type-check fix you mentioned in the same note did survive, for what it's worth — I found it already present in cf3fce44 when I checked main.yml separately.

template-shell/packages/react/src/hooks/useApiStream.ts
tscbmstubp: Deferred to a follow-up: #516.

Confirmed the lifecycle change as described — `status` now starts at `'connecting'` (was `'idle'`) for the default `enabled: true` config, and nothing writes `'idle'` back once a connection starts, so it's genuinely unobservable now. Agreed this should be written down given the `@cpt-dod`/`@cpt-flow` markers point at instructions that don't exist under `architecture/` — but which FEATURE it belongs under (or whether the markers were premature and should come back off) is an architecture-authoring call, not something I should improvise as a code fix. Options and context are in #516.

template-shell/packages/react/src/hooks/useCanAccess.ts
```diff
@@ -120,7 +138,9 @@ export function useCanAccess<TRecord extends AccessRecord = AccessRecord>(
       controller.abort();
     };
     // @cpt-end:cpt-frontx-flow-auth-plugin-rbac-guard:p1:inst-abort
-  }, [app, stableQuery]);
+  }, [app, auth, stableQuery]);
 
-  return result;
+  // @cpt-begin:cpt-frontx-flow-auth-plugin-rbac-guard:p1:inst-pending-effect
+  return auth ? result : DENIED_NO_AUTH;
+  // @cpt-end:cpt-frontx-flow-auth-plugin-rbac-guard:p1:inst-pending-effect
```
itechmeat: The `831dd8ab` fix moved this rather than resolving it: `inst-pending-effect` now opens twice - the pre-existing region (:106-118, unchanged lines; :89-100 at the merge-base) and the new pair this PR adds around the derived return (:143-145) - so one instruction id marks two disjoint regions. And the region and the code have come apart: :106-118 contains no pending-setting code (an early return plus the `alive`/`controller` declarations), while the actual `setResult({ allow: false, isResolving: true })` sits at the unchanged line :101 inside `inst-stable-key`, under the added comment at :98-100 that points at the other marker. Could the pending write and the derived return live under one region, or the flow gain a second instruction - rather than one id spanning two places with its implementation in a third?

.github/workflows/main.yml
```diff
@@ -79,6 +67,235 @@ jobs:
...
+    # This step is still required even though the pins no longer resolve here.
+    # The templates do not declare their whole toolchain: template-shell's
+    # eslint.config.js imports `typescript-eslint`, which is absent from its own
+    # devDependencies and resolves upward into the repo root's node_modules. Drop
+    # the root install and `npm run lint` below fails to load its config.
+    #
+    # A seeded project standing on its own therefore cannot run that lint at all;
+    # what covers it here is the monorepo's tree, not the template's. Tracked as
+    # part of the templates' declared-toolchain gap, not something this job can
+    # fix by installing differently.
```
itechmeat, line 195: The rewritten comment is honest that the pins resolve published tarballs and "NOT this repo's packages/" - which also means nothing in CI now goes red when packages/ and the templates drift, and CONTRIBUTING.md:119 documents the silence as expected. That is the reverse direction of what this job validates, and it used to be covered incidentally by the file: links. Is that an accepted gap, and if so could an issue carry it - e.g. a periodic job that relinks via dev:template:link and runs the template type-check against the working tree?

.github/workflows/main.yml
```diff
@@ -79,6 +67,235 @@ jobs:
...
+    - name: Install and build ecosystem packages
+      run: |
+        npm ci
+        npm run build:packages
```
itechmeat: The comment at :118-122 justifies keeping `npm ci` (the typescript-eslint hoist), but not `npm run build:packages` at :131 - with the pins on registry tarballs, no later step in this job reads the built packages/ output (checked every reference to packages/ in the file). Is the build still needed, or is it a leftover from the file:-links era?

.github/workflows/main.yml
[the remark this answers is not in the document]
itechmeat: The comment at :261-263 says the count is "not a hardcoded src-app/mfe_packages path", and :281 hardcodes exactly that literal for the nested-lock cleanup. The cleanup has a fair excuse (the manifest doesn't say which subtrees hold npm packages), but the juxtaposition reads as self-contradiction - could the rm derive from the copied subtrees, or its comment state why this one is allowed to hardcode?

CONTRIBUTING.md
```diff
@@ -114,11 +114,11 @@ npm run dev:template:link    # point the template's node_modules at the packages
...
-
-Once that follow-up lands, skipping the link becomes silent instead of loud: the template will build and pass its checks against the published alpha, and local edits to `packages/*` simply never appear. That is the steady state to watch for - the type errors above are the one period in which forgetting to relink announces itself.
```
itechmeat: Fix 0d5b893c removed both round-2 tombstones - and the same update writes four new history comments: CONTRIBUTING.md:121 and link-template-ecosystem.mjs:69-75 tell the same #485 alpha.0-to-alpha.1 story twice, useScreenTranslations.ts:78-82 explains which marker used to be here, fs-types.mjs:8-10 explains who declared the type before. None of them describe the system as it is; git carries the past. Keep the present-tense halves (the "missing link is silent" warning is genuinely useful) and drop the "used to" framing?

.github/workflows/main.yml
```diff
@@ -79,6 +67,235 @@ jobs:
...
+    name: Template Validate
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v5
+      with:
+        persist-credentials: false
```
itechmeat, line 161: Still only `template-validate` carries `permissions: contents: read` and `persist-credentials: false` - the whole file has exactly those two occurrences, no top-level block, and the five sibling jobs run on the default token with persisted credentials. A top-level `permissions: contents: read` plus the checkout flag on the five siblings would make this a property of the workflow rather than of one job.

itechmeat (approved):
Approved with comments

package.json
tscbmstubp: Fixed in an earlier round and carried through the rebase: `test:unit` is no longer an `&&` chain — it delegates to `scripts/run-monorepo-unit-tests.mjs`, which forwards post-`--` args (so CI's `npm run test:unit -- --coverage` reaches every project's vitest, including the new `repo-scripts` project) and aggregates exit codes instead of short-circuiting. The vitest.shared.ts thresholds are enforced again.

template-shell/packages/react/src/hooks/useCanAccess.ts
tscbmstubp: Fixed in c316d970: `inst-pending-effect` now marks exactly one region — the render-time `setResult({ allow: false, isResolving: true })` re-pessimize, which is the code the instruction names. The effect's early-return block and the derived `auth ? result : DENIED_NO_AUTH` return are unmarked (neither id appears under `architecture/`, and `cfs validate` stays green at 118/118). The `useScreenTranslations.ts` `inst-begin-load` comment is rewritten present-tense in the same commit.

.github/workflows/main.yml
tscbmstubp: All three resolved by the classify rewrite from the earlier round (kept through the rebase): classification now refuses any non-empty `ownershipBoundaries.sharedFiles` with a loud error instead of half-implementing the contract, which removes the unreachable `typeof f === "string"` branch, the region-union misclassification, and the copy loop's silent sharedFiles skip in one move — the day a manifest declares one, the job fails with instructions to extend both steps.

package.json
tscbmstubp: Covered by the runner-script rework: `test:unit:watch` goes through the same `run-monorepo-unit-tests.mjs` (defaulting to the `repo-scripts` project in watch mode, `--project=` to pick another), and the non-watch path aggregates per-project exit codes instead of `&&`-short-circuiting, so a red workspace run still reports `scripts/` results.

template-shell/packages/react/src/hooks/useApiStream.ts
tscbmstubp: Agreed these three need their own pointer — tracked in #515 (shared `useLatest` helper for the deps-less ref-write effects; blocked on eslint-plugin-react-hooks not recognizing refs returned from custom hooks, `useEffectEvent` only fits the callback-shaped one in `ExtensionDomainSlot`). #502 stays scoped to the five MFE screens.

.github/workflows/main.yml
tscbmstubp: Both parts addressed: the count is now taken per overlay inside the loop, from the subtrees the overlay's own manifest declares (an existing-but-empty subtree fails on its own, not hidden behind another overlay's packages), and in c316d970 the composed lockfile cleanup also reads those same subtree paths instead of hardcoding `$shell/src-app/mfe_packages` — the last path-guessing in the step.

template-shell/packages/react/src/hooks/useApiStream.ts
tscbmstubp: Deferred to #516: documenting `useApiStream`'s status lifecycle (including the `'idle'` → `'connecting'` first-frame change for the default config) against `architecture/`, so the contract question gets settled in the spec rather than in a comment here.

.github/workflows/main.yml
tscbmstubp: Both rewritten in the sweep that landed in c316d970: the main.yml tombstone is down to a two-line present-tense pointer at the step's old location ("MFE packages are template content — the template-validate job below installs and checks them"), and the `useScreenTranslations.ts` comment keeps only the present-tense invariant (loadedLanguage never advances on error, so the mismatch alone gates `isLoaded` false).

scripts/test-runner/discovery.mjs
tscbmstubp: Fixed: one shared `DirEntryLike` typedef in `scripts/fs-types.mjs` (the superset shape, with `isSymbolicLink`), imported by both `check-test-dependency-versions.mjs` and `test-runner/discovery.mjs`.

template-shell/packages/react/src/hooks/useCanAccess.ts
tscbmstubp: Fixed in c316d970 — the region follows the code this time rather than moving alongside it: `inst-pending-effect` opens once, around the render-time `setResult({ allow: false, isResolving: true })` under the `runState` comparison (the one place pending is set), and the two disjoint regions (the effect's early-return block and the derived return) lose their markers. `cfs validate` passes.

.github/workflows/main.yml
tscbmstubp: Accepted gap for this PR — this job's contract stays "what a seeded project installs" — and yes, an issue now carries the reverse direction: #518 proposes a periodic/`packages/**`-triggered job that relinks via `dev:template:link` and runs the template type-check against the working tree.

.github/workflows/main.yml
tscbmstubp: Leftover from the file:-links era — removed in c316d970. The step is now `npm ci` alone, renamed to "Install ecosystem toolchain", with the comment stating why the install suffices (typescript-eslint hoist) and why no build is needed (nothing in the job reads workspace `dist/` with the pins on registry tarballs).

CONTRIBUTING.md
tscbmstubp: All four rewritten present-tense in c316d970: CONTRIBUTING.md keeps the "forgetting to relink is silent" warning and drops the alpha.0 history paragraph; `link-template-ecosystem.mjs` tells the missing-link-is-silent story once, without the "used to warn the other way" framing; `useScreenTranslations.ts` states what the state pair signals now; `fs-types.mjs` says the typedef is shared so consumers cannot drift, not who declared it before.

.github/workflows/main.yml
tscbmstubp: Fixed in c316d970 exactly as suggested: one top-level `permissions: contents: read` block (the job-level one on template-validate is removed as redundant), and `persist-credentials: false` on all six checkouts — validate, template-validate, the three studio jobs, and dependency-graph.

tscbmstubp:
The Template Validate failure was an OOM in the composed-tree `arch:deps`, not a rule violation: #476's rewritten `template-shell/.dependency-cruiser.cjs` bounds traversal with `doNotFollow: '^node_modules'`, which is anchored to the path root — and the composed tree nests a `node_modules` under each overlaid `src-app/mfe_packages/*` workspace, so dependency-cruiser traversed the entire installed dependency universe and blew the default heap. First time the two changes met: develop has no composed run, and this branch pre-rebase had the old config.

Fixed in 8ee2eae0 by matching `(^|/)node_modules/` at any depth. Reproduced the composition locally: standalone 370 modules, composed 532, both green. The root `.dependency-cruiser.cjs` and `internal/depcruise-config/base.cjs` carry the same anchored pattern — latent there since the ecosystem tree stays under the heap today, but worth the same one-line hardening in a follow-up.

gs-layer:
Merge-order note from review: this PR should merge only after #558, then be rebased/rechecked. The current merge-ref `Template Validate` job fails in the composed `template-shell + template-mfe` tree while type-checking `_blank-mfe` (`src/components/ui/button.tsx` React type errors). That failure is in the same template-mfe React 19.2/type-alignment surface that #558 addresses, so merging #491 first would land the new validation job red on the current composition.

gs-layer:
Merge-order note from review: this PR should also merge only after #521, then be rebased/rechecked. #521 is the current code-altitude documentation for useApiStream status lifecycle, while #491 changes the same hook observable status behavior: initial enabled render becomes connecting, re-enable clears stale data before paint, and disabled status is derived as idle. If #491 lands before #521, #521 can reintroduce documentation for the pre-#491 lifecycle or create a stale conflict resolution path. Merge #521 first, then have #491 carry the runtime rewrite plus matching docs/tests on top.

GeraBart (approved):

PR #492, opened 2026-07-29
fix: exact registry pins for the template's ecosystem dependencies (#485)

itechmeat:
# Seeded projects install from the registry

Refs #485. #485 stays open until the follow-up pin bump to `alpha.1` proves the seeded build and smoke checks end-to-end.

Base `develop` @ `37b09ee3`. 2 commits, 14 files, +1810 / -13254; without the three lockfiles the code diff is 11 files, +99 / -25.

The first commit is the shipped-content fix and is complete on its own; the second adds the explicit in-monorepo dev loop that replaces what the `file:` paths used to provide implicitly. Nothing under `packages/cli/` changes - the fix lives entirely in the template's own content plus one root-level script, which is what lets it land independently of #481.

## What changes for a developer

| Before | After |
|---|---|
| `frontx seed` produced a project whose first `npm install` failed: five specifiers (`file:../packages/*` and `file:../../../packages/*`) resolve only where `template-shell/` sits next to the monorepo's `packages/` | The seeded project's install resolves end-to-end at this head: `@gears-frontx/{api,mfes,gts-plugin}` arrive from the npm registry at exact `0.3.0-alpha.0`, verified outside the monorepo with `npm install` and strict `npm ci`. Build and type-check currently fail against the pins alone on the `FRONTX_ACTION_*`/`DomainContext.typeSystem` exports moved to `gts-plugin` - a known issue documented in CONTRIBUTING - and verified to heal under `dev:template:link`. The original full build / type-check / 601-unit-test / browser-smoke green run was measured pre-#490-rebase, when those exports' published home was still `mfes` |
| A second, independent breaker: `template-shell/package-lock.json` recorded `"resolved": "../packages/api"` for all three, and its root version had drifted from `package.json` - a seeded `npm ci` failed on the lockfile alone, even with a fixed manifest | Lockfile regenerated from scratch with npm 11.12.1 (the version CI runs): zero `../` resolutions, versions consistent, integrity present |
| Nine more specifiers in the fixture/blank MFEs pointed at `template-shell/packages/{api,mfes,gts-plugin}` - directories that exist nowhere. Installs produced dangling symlinks that Node happened to fall through onto the hoisted copy | Same exact pins. Zero dangling symlinks under any `@gears-frontx` scope dir, checked recursively including nested MFE `node_modules` |
| Two nested MFE `package-lock.json` files shipped into every seeded project through the `src-app` ownership boundary | Deleted. npm never reads a workspace member's lockfile (`npm ci` inside an MFE warns `ignoring workspace config` and operates on the root tree), and both files recorded verbatim the dangling `file:` paths this fix removes. A fresh `npm ci` after deletion produces the identical tree - direct proof they were inert |
| `tsconfig.app.json` mapped the three ecosystem packages to `../packages/*/src` - the monorepo type-checked against source while a seeded project silently type-checked against published `.d.ts` | Mappings removed; both environments type-check against the same artifact |
| Editing `packages/mfes` was implicitly visible to the template via `file:` symlinks - and invisibly absent in a seeded project | Explicit opt-in: `npm run dev:template:link`. Documented in CONTRIBUTING together with the failure mode it prevents (silently developing against the published alpha) |

This PR also bumps `mfes` and `gts-plugin` to `0.3.0-alpha.1`, published on merge by PART 1 of `publish-packages.yml`; the template's own pins move to `alpha.1` in an immediate follow-up PR, since a lockfile cannot pin a tarball that does not exist yet.

## Why registry pins, and not the alternatives

Two facts collapse the option space.

**A rewrite step cannot be built where it would have to run.** Templates are acquired as a git tree (ADR-0016/0017, `github-fetch.ts`), so an npm-publish-time rewrite never touches what `frontx install` fetches. That leaves a seed-time rewrite inside the CLI - and materialization is a verbatim copy with no per-file transform seam, guarded by constraint CLI-1 (*the CLI knows nothing about template content*), backed by the grep-based check in `scripts/test-architecture.ts` that fails if `packages/cli/src` mentions any template package name - a deterrent rather than full enforcement: it would not catch a generic seed-time version rewrite that names no template package (a dedicated pin-drift guard for that lands separately as #498), so the constraint itself is the primary guard.

**Root `workspaces` does not address the bug at all.** A seeded project has no monorepo above it - workspaces only changes the in-monorepo loop, and at a real cost: npm does not recurse into nested `workspaces`, the root's 19 `overrides` would govern the template's tree in-monorepo while the template's own override set governs it when seeded (two different resolutions of the same `package.json` - the exact bug class this issue removes), and the shipped `template-shell/package-lock.json` would stop being exercised by the monorepo while remaining the artifact seeded projects install from.

So the shipped manifests carry registry-resolvable pins under every option; the only real question was the dev loop, answered by the link script below.

```mermaid
flowchart TB
  subgraph before["Before: file: paths"]
    B1["template-shell/package.json<br/>@gears-frontx/api: file:../packages/api"] --> B2{"where does<br/>../packages/api land?"}
    B2 -- "in the monorepo" --> B3["packages/api<br/>local sources, works"]
    B2 -- "in a seeded project" --> B4["nothing there<br/>npm install fails"]
  end
  subgraph after["After: exact registry pins"]
    A1["package.json<br/>@gears-frontx/api: 0.3.0-alpha.0"] --> A2["registry tarball,<br/>same bits in both worlds"]
    A3["dev:template:link<br/>opt-in, root-only, not shipped"] -.->|"overlays symlinks<br/>into node_modules"| A2
  end
  style B4 stroke-dasharray: 5 5
```

**Exact pins, not `^` and not a dist-tag.** `latest` is poisoned: `@gears-frontx/api` has `latest = 0.1.0` with the current version only under `alpha`. A dist-tag specifier is not reproducible in a source tree, and `^0.3.0-alpha.0` would silently absorb `0.3.0` final and every later alpha - unacceptable while the template is the ecosystem's reference consumer. Each ecosystem release now costs one deliberate template edit; at alpha cadence that is the right trade. This does not conflict with the no-exact-pin rule in `ecosystem-version-policy.mjs` - that rule governs one peer edge (`mfes` -> `gts-plugin`), not a consuming project's direct dependencies.

## What deliberately does not change

The intra-subtree `file:` specifiers (`src-app/mfe_packages/*` -> `template-shell/packages/*`, and the `file:.` self-reference) stay: they resolve to directories that travel with the seeded project. All 14 surviving `file:` specifiers were resolved programmatically - every one lands inside the template subtree.

One dangling declaration was removed rather than pinned: `_blank-mfe` declared the template self-reference against a path that never existed. Its three siblings consume the self-reference through the root `file:.` override without declaring it; `_blank-mfe` now matches them. (Pinning it as `"*"` instead crashes npm outright - Arborist's `canDedupe` hits `Invalid Version` once exact pins exist at the root - which is reproducible from a clean tree.)

## The dev loop

`file:` on a directory installs a symlink, so template development used to pick up local `packages/*` edits automatically. That implicitness is what let this bug ship: nothing distinguished "works because linked" from "works when seeded".

```mermaid
flowchart LR
  E["edit packages/*"] --> B["npm run build:packages"] --> L["npm run dev:template:link"] --> D["cd template-shell<br/>npm run dev"]
```

The script (`scripts/link-template-ecosystem.mjs`) symlinks exactly the three ecosystem packages into the template's `node_modules` and touches nothing else. The obvious alternative - `npm install --no-save --no-package-lock ../packages/*` - passes the "shipped files untouched" test but rebuilds npm's ideal tree (72 unrelated packages pruned) and replaces the `file:.` self-link with a packed snapshot, breaking the template's own rebuild-on-change loop. Deterministic symlinks avoid npm's tree reconciliation entirely; verified to leave `package.json` and `package-lock.json` byte-identical and the tree at the same entry count.

Running the loop against develop also surfaced a real latent break, filed separately: the template still imports `FRONTX_ACTION_*` from `@gears-frontx/mfes`, whose local sources no longer export them (moved to `gts-plugin` in `cd5ac585`). The pinned published `0.3.0-alpha.0` predates the move, so this PR is green - but the first `mfes` publish plus a routine pin bump will break the template unless the imports move with it.

## Verification

In-monorepo (npm 11.12.1 throughout): `npm ci` from scratch in `template-shell` (3 runs), template build / type-check / test:unit, root lint / type-check:all / test:unit / `arch:check` 12/12, pre-commit hooks on both commits. Template `npm run lint` has 32 pre-existing findings, proven identical on develop content (CI does not run it either way).

Seeded-project e2e, from a directory outside the repo: materialization was **simulated** by copying the git-tracked files under the manifest's `ownershipBoundaries.exclusiveSubtrees` - the real `frontx install github:...//template-shell@ref` path needs #481's subtree addressing and is worth re-running once both PRs land. In the seeded copy: `npm install` and strict `npm ci` (629 packages, exit 0) are green, and on-disk resolution was asserted (the three ecosystem packages are real registry directories at `0.3.0-alpha.0`; the six applied packages and the template self-reference are symlinks, zero dangling). The build + type-check + 601 unit tests + browser pass (demo MFE mounts via Module Federation, both widget fixtures render concurrently, no console errors) were measured pre-#490-rebase, when `FRONTX_ACTION_*`/`DomainContext.typeSystem` still lived in `mfes`; post-rebase, the pinned `0.3.0-alpha.0` predates their move to `gts-plugin`, so build and type-check fail against the pins alone until `dev:template:link` puts the local `dist/` in place (documented in CONTRIBUTING) - re-running the full green sequence against the pins is pending the follow-up bump to `alpha.1`.

## Follow-ups this PR makes visible, kept out of scope

- The pin surface is now 14 sites across 5 manifests plus the script's package list; nothing guards it against drifting from `packages/*/package.json` versions. A cheap CI assertion would make a missed bump loud instead of silent.
- The `FRONTX_ACTION_*` import drift above - fix is publish-coupled.
- `main.yml`'s per-MFE loop iterates `src/mfe_packages/*/`, a path that stopped existing at the template split - it has been silently running on nothing (same "checks exist, nothing runs" pattern as #483).
- PART 1 of `publish-packages.yml` (ecosystem publish) is now a hard prerequisite of the template's pins; a comment at its header records that it must survive the eventual PART 2 removal (#469 item 5).

## Summary by CodeRabbit

- **New Features**
  - Added a command to link locally built ecosystem packages into the shell template for development.
  - Added validation, clear errors, rollback support, and platform-specific linking, including Windows junctions.
  - Added dedicated unit-test support for repository tooling, including watch mode.

- **Documentation**
  - Documented the template development workflow, package linking, restoration steps, testing, and platform considerations.

- **Chores**
  - Updated publication workflows to ensure ecosystem packages are detected and published before the shell template.
  - Updated package versions for the latest ecosystem release.


gs-layer (changes requested):
## Review 

Everything below was checked against `refs/tmp/develop...refs/tmp/pr492`, the split branch (#490), and the live registry. Verdict at the bottom, plus a separate block on what has to be redone when this is rebased onto the split.

### The shipped fix is correct

The lockfile diff is exactly what it claims to be, with nothing extra riding along. Restricted to `@gears-frontx` entries:

```
~ node_modules/@gears-frontx/{api,mfes,gts-plugin}
    develop: resolved '../packages/api'            link=True   integrity=None
    PR:      resolved 'registry.npmjs.org/...tgz'  0.3.0-alpha.0  integrity=sha512-...
− 10 entries  src-app/mfe_packages/*/node_modules/@gears-frontx/*
              (all resolved 'packages/api' → template-standard/packages/api, which does not exist ⇒ dangling)
+ 0 entries
```

Zero `../` resolutions remain, root lock version matches `package.json`, integrity present on all three. All three versions really are on the registry, and `@gears-frontx/api` really does have `latest = 0.1.0` / `alpha = 0.3.0-alpha.0` — the poisoned-dist-tag argument holds.

---

## Findings

### 1. MAJOR — `dev:template:link` is documented as the working loop but fails on first run

`CONTRIBUTING.md:111`
```bash
npm run build:packages       # publish-shaped dist/ for the ecosystem packages
npm run dev:template:link    # point the template's node_modules at packages/* sources
cd template-standard && npm run dev
```

`template-standard/packages/framework/src/index.ts:100`
```ts
// MFE Action Constants (re-exported from @gears-frontx/mfes for convenience)
export { FRONTX_ACTION_LOAD_EXT, FRONTX_ACTION_MOUNT_EXT, FRONTX_ACTION_UNMOUNT_EXT } from '@gears-frontx/mfes';
```

`packages/mfes/src/` does not export those — they live in `packages/gts-plugin/src/constants.ts:14`. I unpacked both published tarballs: `mfes@0.3.0-alpha.0/dist/index.d.ts` has 8 `FRONTX_ACTION` occurrences, `gts-plugin@0.3.0-alpha.0/dist/index.d.ts` has 0. So the template stops type-checking the moment you link, and moving the import to `gts-plugin` would break the pinned path instead. The fix genuinely is publish-coupled — that assessment is right.

The defect is not in the code, it is that CONTRIBUTING presents the loop as working. The only feature of the second commit is broken at merge time, and that is recorded only in the PR description, which is not where a developer will look.

**Fix:** a paragraph next to the block — known break, `@gears-frontx/mfes` does not export `FRONTX_ACTION_*` locally since `cd5ac585`, the loop fails at type-check until the next publish, link to the issue. Having the script print the same warning would help too.

### 2. MAJOR — bumping the ecosystem version fails silently, and it breaks the very invariant this PR restores

The claim about `ecosystem-version-policy.mjs` is **correct**: `readEcosystemArtifacts` and `readDeclaredEdgeRange` read only `packages/*/package.json`, and `checkEdgeCompatibility` forbids an exact pin on exactly one edge — `mfes → gts-plugin` in `peerDependencies`. The template is never read. No conflict.

But there is also no guard. What happens on a `packages/*` bump with the template left alone:

- `publish-packages.yml` publishes the new version;
- the template keeps installing the old one (still on the registry) ⇒ `npm ci` green, CI green — **silent staleness**, not a break;
- `dev:template:link` then overlays *newer* local sources on top of an *older* pin ⇒ in-monorepo development runs against one artifact while a seeded project gets another. That is the same "two different resolutions of the same `package.json`" this PR uses to reject root workspaces.

Surface: 14 sites across 5 manifests, plus the hardcoded `linkedPackages` at `scripts/link-template-ecosystem.mjs:36`.

**Fix:** a `scripts/check-template-ecosystem-pins.mjs` modelled on the existing `scripts/check-test-dependency-versions.mjs` (already wired into `npm run lint` via `lint:deps`, so the pattern is there): walk the template's manifests, compare each `@gears-frontx/{api,mfes,gts-plugin}` against `packages/<n>/package.json.version`, fail on mismatch. And derive `linkedPackages` from `template-standard/package.json` instead of a literal, so the script stops being a 15th drift site. This is listed as a follow-up; given finding 1, it is worth landing here rather than deferring.

### 3. MAJOR — the link script's guard checks the wrong thing

`scripts/link-template-ecosystem.mjs:48`
```js
if (!fs.existsSync(path.join(source, 'package.json'))) {
  console.error(`Cannot link: packages/${name} is missing.`);
  process.exit(1);
}
```
…and 12 lines later, *after* everything is already linked:
```js
console.log('Run `npm run build:packages` first if the local dist/ is stale.');
```

`packages/<name>/package.json` always exists in a checkout, so the guard is vacuous. Meanwhile all three packages resolve exclusively through `dist/`:

```
api:        main ./dist/index.cjs   types ./dist/index.d.ts   exports {'.': dist/...}
mfes:       same
gts-plugin: same
```

After `npm run clean:artifacts`, or in a fresh clone, `dist/` is absent — the script silently links broken directories, exits 0, and the failure surfaces later as `Cannot find module '@gears-frontx/api'` inside the template build, far from the cause. The build hint is printed unconditionally and after the fact, so it reads as noise rather than a guard.

**Fix:** check the entry point (`dist/index.js`, or resolve `exports['.'].import`) and `exit 1` with the `npm run build:packages` command. Keeping the `package.json` check is fine, but it does not substitute for this one.

Idempotency is fine: I verified `fs.rmSync(symlink-to-dir, { recursive: true, force: true })` on Node 24.14 removes only the link and leaves `packages/*` contents intact. A second run converges to the same state.

### 4. MINOR — the script is untestable and breaks the `scripts/` convention

```
scripts/check-test-dependency-versions.mjs + .test.mjs
scripts/ecosystem-version-policy.mjs       + .test.mjs
scripts/run-monorepo-unit-tests.mjs        + .test.mjs
scripts/link-template-ecosystem.mjs        ← no test
```

The three existing `.mjs` scripts follow "exported core logic + `isEntryPoint` guard + `.test.mjs`" (see the `ecosystem-version-policy.mjs` docblock: *"Core logic is exported for unit tests"*). The new one is top-level imperative with side effects at import, exports nothing, and cannot be tested at all.

**Fix:** extract `export function linkEcosystemPackages({ rootDir, fs })`, add the `isEntryPoint` guard and `scripts/link-template-ecosystem.test.mjs`.

### 5. MINOR — no inverse operation, and `npm install` silently undoes the link

`CONTRIBUTING.md:119`: *"To go back to the pinned registry versions, run `npm ci` inside `template-standard`"* — 629 packages to undo three symlinks. Worse, the **unintentional** revert is not documented: any `npm install` inside the template (e.g. while adding a dependency) reifies the tree from the lockfile and restores the registry tarballs, silently. That is exactly the implicitness this PR sets out to remove.

**Fix:** an `--unlink` mode (or a separate `dev:template:unlink`), plus a line in CONTRIBUTING that any `npm install` in the template drops the links.

### 6. MINOR — Windows

`scripts/link-template-ecosystem.mjs:55`
```js
fs.symlinkSync(path.relative(scopeDir, source), target, 'dir');
```
`type: 'dir'` on Windows throws `EPERM` without Developer Mode or admin — and it happens *after* the `rmSync` on the line above, so the directory is already gone with no replacement, leaving the tree worse than before and recoverable only by `npm ci`. The standard workaround is `'junction'`, but junctions require an **absolute** target, so both the type and the path have to change (which invalidates the *"Relative link so the tree stays valid if the checkout moves"* comment). CONTRIBUTING makes no platform claims, so this is more "decide deliberately" than "bug": either `process.platform === 'win32' → junction + absolute`, or an explicit line saying the dev loop is macOS/Linux only.

### 7. MINOR — gap in the new gating when `detect-ecosystem` fails

`.github/workflows/publish-packages.yml:239`
```yaml
needs: [detect-template, publish-ecosystem]
if: >-
  !cancelled() && needs.detect-template.outputs.has_changes == 'true' &&
  (needs.publish-ecosystem.result == 'success' || needs.publish-ecosystem.result == 'skipped')
```
`publish-ecosystem` is `needs: detect-ecosystem` + `if: has_changes == 'true'`. If **`detect-ecosystem` fails**, GitHub marks `publish-ecosystem` as `skipped`, the condition passes, and `publish-template` runs in a state where nobody knows whether ecosystem versions changed. The declared hard prerequisite does not hold in exactly that case. It is self-correcting (the template's own `npm ci` fails loudly if the new pin is not published yet), hence minor rather than major.

**Fix:** add `needs.detect-ecosystem.result == 'success' &&` to the condition, and `detect-ecosystem` to `needs`.

### 8. NIT — "CI-enforced" is stronger in the description than the checks actually are

Both checks are real and inside `arch:check` — `scripts/test-architecture.ts:183` and `:191`. But neither would catch a seed-time rewrite:

- the grep at `:194` only matches `frontx-template-standard\|@gears-frontx/frontx-template`; a step rewriting `@gears-frontx/api` in someone else's `package.json` never trips it;
- the dep-cruiser rule `frontx-cli-1-no-bundled-template-content` (`.dependency-cruiser.cjs:150`) is about module import edges, not string literals.

So CI would block "the CLI knows the template's name" but not "the CLI rewrites ecosystem versions". The conclusion still stands — it rests on CLI-1 as a design constraint **and** on the verbatim copy below — but the "CI-enforced by two checks" wording should be softened, or the next reader will trust a guard that is not there.

Unrelated, worth its own issue: `.dependency-cruiser.cjs:93` still carries a dead STUB rule `cli-template-independence` with `severity: 'warn'`, `from: {}`, `to: {}`.

### 9. NIT — "the root's 14 `overrides`"

There are 19 in `package.json`. The argument is unaffected, but the description is the record of the decision, so the number is worth fixing.

---

## The rejected alternatives, checked against code

**Materialization is verbatim, there is no seam. Confirmed.** `template-standard/frontx-template.json` declares `sharedFiles: []`, and `package.json`, `package-lock.json` and `tsconfig.app.json` are all listed under `exclusiveSubtrees`. In `packages/cli/src/scaffold/compose-shared-files.ts:171` such a path takes this branch:

```js
if (entries.length === 1 && entries[0].mergeStrategy === 'exclusive') {
  await writeFileFn(`${targetDir}/${path}`, entries[0].content);
```

A byte-for-byte write with no transform point anywhere. The region-union machinery exists but only concatenates marker blocks, and only activates for `sharedFiles`, of which this template has none.

**CLI-1** — see finding 8: the constraint is real, the CI backing is weaker than stated, the conclusion holds.

**npm and nested workspaces — correct.** npm does not recurse into a nested package's `workspaces`. But the load-bearing part of the argument is not recursion, it is that a seeded project has no monorepo above it at all, which is true by construction — and confirmed by the fact that the template carries its own `workspaces: ["packages/*","src-app/mfe_packages/*"]` and its own `overrides: {"@gears-frontx/frontx-template-standard": "file:."}`, neither of which a root install would apply.

**Nested lockfile deletion — the claim is correct, reproducibility is not lost.** The template's root lock carries **22 entries** under `src-app/mfe_packages/*`, including each MFE's nested `node_modules` (`_blank-mfe/node_modules/@tanstack/react-query` and friends). The deleted locks recorded only dangling `resolved: 'packages/api'`. The `@gears-frontx` entry diff above is sufficient evidence: nothing added, 10 dangling links removed, 3 moved to the registry with integrity. No objection.

**`tsconfig.app.json` — the trade is right, but it removes the last signal, which reinforces findings 1–2.** Dropping the mappings hides nothing from CI, because CI does not type-check the template at all: root `type-check:all` = `type-check:host` + `packages/{mfes,gts-plugin,api}` + tests, no template; and in `publish-packages.yml` the template is only built (`build:package`, `build:packages`), never type-checked. On top of that, `.github/workflows/main.yml:41` iterates `for mfe in src/mfe_packages/*/` — there is no `src/` at the repo root (`git ls-tree` confirms), so that step runs zero iterations. "Both worlds type-check the same artifact" is an honest trade, but it moves all drift detection onto `dev:template:link`, which is itself broken (finding 1) and enforced by nothing (finding 2). Not a separate defect — an argument for not deferring 1 and 2.

**Verification — weaker at the acquisition edges, not weaker overall.** Given the verbatim-copy finding above, `materializeAssembly` physically cannot write anything but a byte-for-byte copy of the files under `exclusiveSubtrees` for *this* manifest, so copying by hand from `ownershipBoundaries` is equivalent to real materialization, and "simulation ≠ reality" does not apply to the pins. What genuinely remains unverified:

- the acquisition step — `github-fetch` with subtree addressing (#481), in particular whether the dotfile path `.frontx/ai/@gears-frontx/frontx-template-standard` lands in the fetched tree;
- provenance writing, and that `package-lock.json` actually arrives (it is in `exclusiveSubtrees`, but file enumeration happens upstream of `materialize`);
- `frontx add` of the MFE template on top of a shell project — post-split, out of scope here.

None of the three affects pin correctness. "Worth re-running once both PRs land" is the right call.

**PART 1 comment — satisfied.** It sits in the PART 1 banner (`:17–21`), it is unambiguous ("Removing the temporary PART 2 (#469 item 5) must leave PART 1 in place"), it is outside the block PART 2 removal would take with it, and it is echoed at `publish-template` (`:232`). No comment beyond finding 7 on the condition itself.

---

## Verdict: Request changes

The core of the PR is correct and verified in code: the pins resolve, the lockfile is clean, the dangling symlinks are gone, and both rejected alternatives survive scrutiny (with one over-stated claim). None of the findings breaks the shipped artifact.

Before merge I would only ask for two, both small: **finding 1** (a CONTRIBUTING paragraph about the known `FRONTX_ACTION_*` break — otherwise the second commit's only feature ships presented as working) and **finding 3** (check `dist/` instead of the vacuous `package.json` check). **Finding 2** should land either here or as a prioritised issue rather than a follow-up: without it the next `mfes` publish requires a coordinated three-place change that nothing forces.

**Separately, as a risk rather than a finding:** local `packages/mfes@0.3.0-alpha.0` and the published `0.3.0-alpha.0` have different public surfaces (verified from the tarball). That is a develop defect, not this PR's, but the PR cements the divergence into the shipped template. Publishing is version-gated, so the next `mfes` publish is impossible without a bump, and a bump without a synchronised pin edit plus the `framework` import move breaks the template. The sequence deserves to be written down in an issue.

---

## What has to be redone on the post-split structure

Checked against `refs/tmp/pr490`. All 7 conflicting files:

| File | State in #490 | How the #492 edit should land | Risk |
|---|---|---|---|
| `template-shell/packages/framework/package.json` | lines 45–46 still `file:../../../packages/{mfes,gts-plugin}`; **line 44 is a new #490 edit**: `frontx-template-shell: "file:../.."` (was `"*"` on develop) | take the #492 pins for `mfes`/`gts-plugin` **and keep** #490's `file:../..` | **High.** The textbook case where "take ours/theirs" drops one of the two edits. Resolve line by line. |
| `template-shell/tsconfig.app.json` | lines 36–37, 40–41, 44–45 — the same escaping `../packages/*/src` mappings | delete the same 6 lines | Low, 1:1 correspondence |
| `template-shell/README.md` | lines 124–127 — the same stale "Alpha note … by local path" | replace with the same text, `template-standard` → `template-shell` | Low |
| `template-shell/package.json` (not among the 7, but the edit must travel) | lines 69–71 — `file:../packages/{api,gts-plugin,mfes}` | three pins | Low |
| `.github/workflows/publish-packages.yml` | #490 touches 164–175 (banner), 195–197, 258, 305 (`template-standard`→`template-shell`); `publish-template` at 229–231 is still `needs: detect-template` | the auto-merge is correct — #492's edits (13–24, 232–244, 262–268) do not overlap | Low; after merging, **eyeball** that `needs: [detect-template, publish-ecosystem]` actually survived |
| `_blank-mfe/package-lock.json`, `demo-mfe/package-lock.json` (rename/delete) | moved to `template-mfe/src-app/mfe_packages/*/` | resolve as **delete** | **Medium, and it needs to be a conscious decision.** `template-mfe/` has neither a `package.json` nor a root lockfile (`git ls-tree` shows only `.frontx`, `README.md`, `frontx-template.json`, `src-app`). After the deletion, `template-mfe/` has no lockfile at all. That is formally right — the MFE template overlays a shell project whose lock governs the tree — but #490's `template-shell/package-lock.json` has **0** entries under `mfe_packages` (verified), so reproducibility of the assembled shell+mfe project is pinned by nothing that ships. That hole belongs to #489/#490, but applying #492 is what opens it. |
| `template-standard/package-lock.json` (rename/delete) | moved to `template-shell/package-lock.json`: **625** entries, **0** under `mfe_packages`, and still **3 escaping resolutions** `../packages/{api,gts-plugin,mfes}` | **regenerate with `npm install` in `template-shell`, do not merge the file** | **Highest risk in the rebase.** #492's lock has 758 entries, **including the 22 workspace-member entries under `src-app/mfe_packages/*`**, which no longer exist in `template-shell` (its `workspaces: ["packages/*","src-app/mfe_packages/*"]` now matches only the first glob). Taking #492's file as-is injects ~133 phantom entries. Correct order: apply the pins in `template-shell/package.json` and `packages/framework/package.json`, then `rm package-lock.json && npm install` in `template-shell`, then assert zero escaping resolutions. |

**What loses its meaning after the rebase:** the `_blank-mfe/package.json` hunk (dangling self-reference removal + three pins). #490 has already rewritten that file differently — `frontx-template-shell: "file:../../.."`, `mfes/gts-plugin/api: file:../../../packages/*` (lines 40–43) — and under `template-mfe/` those `../../../packages/*` point at `template-mfe/packages/*`, which does not exist. So the MFE half stays exactly where it was, which is #489 as agreed. Resolve that hunk in #490's favour and do not try to carry pins into `template-mfe/`.

**One thing that has to be added beyond the existing edits:** `dev:template:link` is hardwired to `template-standard/node_modules/@gears-frontx` (`scripts/link-template-ecosystem.mjs:33`), a path that disappears after the split. Both the script and the `dev:template:link` name need to move to `template-shell`. If finding 2 is addressed (deriving the list from the manifest), pull the template root into the same single constant so this becomes a one-line change.

itechmeat:
> ## Review — verified against code, not against the description

@gs-layer Rebased onto post-#490 develop per your rebase table: framework's file:../.. line preserved, lockfile regenerated from scratch (zero ../ resolutions), nested template-mfe lockfiles stay deleted; template-mfe pins left to #489 (its item 1 claims them).

### Findings:

1 - done, though inverted since your review: 1861e182 moved the imports to gts-plugin, whose published 0.3.0-alpha.0 predates the move, so the pinned build is now the broken path and dev:template:link the working one; CONTRIBUTING and the script say so, self-heals on the next ecosystem alpha.
2 - lands as #498.
3 - guard now checks each package's built entry before any write.
4 - exported-core convention, 12 tests.
5 - covered the silent link-reverts, skipped --unlink (can't restore tarball content anyway).
6 - win32 junction, posix symlink.
7 - detect-ecosystem == 'success' added. 8/9 - description updated.

gs-layer (changes requested):
## Review round 2

Re-checked against `refs/tmp/develop..refs/tmp/pr492` (develop `a0fd54aa`, head `4832dc04`), with an independent reproduction of the seeded build. Verdict at the bottom.

### Round 1 findings — status

| # | Round 1 finding | Status |
|---|---|---|
| 1 | `FRONTX_ACTION_*` break undocumented | **Closed.** Known-issue block in CONTRIBUTING, a warning printed by the script on success, and a test asserting the warning is printed |
| 2 | No pin-drift guard | **Deferred to #498**, a real open PR stacked on this branch. Accepted |
| 3 | Vacuous `package.json` guard instead of a `dist/` check | **Closed, and then some.** `builtEntryPointOf` reads `exports['.'].import → module → main`, and all three packages are validated *before the first write*, so a partial link state is no longer reachable |
| 4 | Script untestable | **Closed in form**, not in effect — see new finding 2 |
| 5 | No unlink, `npm install` silently reverts | **Closed.** Documented, including a reasoned explanation of why no `--unlink` can exist |
| 6 | Windows EPERM | **Closed.** `symlinkSpecFor` returns junction + absolute target on `win32`, covered by a test |
| 7 | `detect-ecosystem` failure slipped through the gate | **Closed.** `needs.detect-ecosystem.result == 'success'` added, with a comment explaining the skipped/failed ambiguity |
| 8 | "CI-enforced" overstated | **Closed.** Now "a deterrent rather than full enforcement" |
| 9 | "14 overrides" | **Closed** → 19 |

### The rebase onto the split landed correctly, including the risky part

`template-shell/package-lock.json` was regenerated rather than merged: **685** entries, **0** under `mfe_packages`, **0** escaping `../` resolutions, and the three ecosystem entries carry registry tarball URLs with integrity. Both edits survived in `template-shell/packages/framework/package.json` — the split's `"@gears-frontx/frontx-template-shell": "file:../.."` and this PR's pins. The `_blank-mfe/package.json` hunk was dropped, correctly leaving the MFE half to #489. No objections here.

---

## New findings

### 1. BLOCKER — the pinned template does not build, and the PR body still claims it does

Reproduced from a clean copy of `template-shell` at this head, outside the monorepo:

```
npm ci                    exit 0
npm run build:package     exit 0
npm run build:packages    exit 1    ← 7 TS errors
npm run type-check        exit 2    ← 15 TS errors across 5 files
```

```
packages/framework/src/index.ts(104,3): error TS2305: Module '"@gears-frontx/gts-plugin"' has no exported member 'FRONTX_ACTION_LOAD_EXT'.
packages/framework/src/plugins/microfrontends/actions.ts(18,3): error TS2305: …
packages/framework/src/plugins/microfrontends/base-domains.ts(33,3): error TS2305: …
packages/framework/src/plugins/microfrontends/index.ts(26,3): error TS2305: …
packages/framework/src/testing/TestContainerProvider.ts(81,34): error TS2339: Property 'typeSystem' does not exist on type 'DomainContext'.
```

Cause confirmed straight from the published tarballs:

- `gts-plugin@0.3.0-alpha.0` — `grep -c FRONTX_ACTION dist/index.d.ts` = **0**, while `template-shell/packages/framework/src/index.ts:107` now imports them from exactly there (after `1861e182` on develop);
- `mfes@0.3.0-alpha.0`, `dist/index.d.ts:930` — `interface DomainContext` has `mounter` and `lifecycleTrigger` but no `typeSystem`; local `packages/mfes/src/runtime/DomainContext.ts:47` adds it.

Two consequences.

**(a) The PR body's verification is now false.** Row 1 of the Before/After table — *"The seeded project installs and runs end-to-end … build, type-check, 601 unit tests"* — and the Verification section repeat a claim that the PR's own CONTRIBUTING known-issue block contradicts. That table describes a pre-rebase run, when the template still imported the constants from `mfes`, where the published tarball does have them.

**(b) `publish-template` is armed to fail.** In `.github/workflows/publish-packages.yml`, `working-directory: template-shell`:

```yaml
npm ci
npm run build:package
npm run build:packages   # ← this is the one that exits 1
```

It will not fire on this merge (no template package version is bumped), so it stays latent until the next bump.

**Suggested fix:** bump and publish `mfes` and `gts-plugin` — PART 1 is version-gated on `packages/*/package.json`, so this is fully in-tree — then pin to the published versions. That is precisely the ordering the new `needs: publish-ecosystem` edge in this PR encodes, and it turns the PR from "install fixed, build broken" into a complete fix for #485. At minimum, the Before/After table and the Verification section have to be corrected: they are the record of what was proven.

Separately: the figure "17 type errors" appears in three places (CONTRIBUTING, the `pinnedSurfaceDriftWarning` constant, a test comment). I measure 15 for `type-check` and 7 for `build:packages` — and `type-check` short-circuits at `type-check:app`, so `type-check:packages` and `type-check:mfe` never ran and the true total is ≥15 depending on what is counted. An exact number replicated in three places is brittle; "fails with type errors" would age better.

### 2. MAJOR — the new test file never runs in CI

`scripts/test-runner/discovery.mjs:21`
```js
export async function loadProjects(repoRoot = defaultRepoRoot) {
  const projects = [
    ...(await discoverWorkspaceProjects({ repoRoot })),
  ];
  return projects;
}
```

Executed live, it returns only the root workspaces:

```
workspace:api, workspace:cli, workspace:cyber-pilot-kit-frontx, workspace:gts-plugin, workspace:mfes
any project owning scripts/: false
```

The host project that used to own `scripts/` via `extraRootPaths` was removed by the split (`discovery.mjs:7-11`), even though `common.mjs:67-70` still documents that routing as live. So `npm run test:unit` does not pick up `scripts/link-template-ecosystem.test.mjs`.

The test itself is good — run from a standalone root it is **12/12 green**, covering all four refusal reasons, the write-nothing-on-late-failure guarantee, both platform branches, and the drift warning. It is only the wiring that is missing. Note also that `eslint.config.js:33` and `:112` both ignore `scripts/**`, so the new 274-line script is not linted either.

This hole predates the PR — `check-test-dependency-versions.test.mjs`, `ecosystem-version-policy.test.mjs`, `run-monorepo-unit-tests.test.mjs` and `test-runner/orchestration.test.mjs` are all dark for the same reason. But this is the PR that adds a test in response to review, so it should either register `scripts/` as a project (`{ kind: 'host', name: 'repo-scripts', rootPath: 'scripts' }`) or state plainly that the test is not enforced yet.

### 3. Context — why the PR's checks are green

Worth stating explicitly, because the green tick is easy to read as evidence against finding 1. Run [30602819260](https://github.com/constructorfabric/gears-frontx/actions/runs/30602819260), job `validate (24.14.x)`:

| Step | What it actually covers |
|---|---|
| `npm ci` | Root workspaces are `["packages/*", "internal/*"]`. `template-shell/` is not one, so **it is never installed** and its pins are never resolved in CI |
| `Install MFE dependencies` | Loops `src/mfe_packages/*/`. The log shows the command echo and an immediate `##[endgroup]` — **not one `==> npm ci in …` line**. There has been no root `src/` since the split |
| `npm run build:packages` | `config → mfes → gts-plugin → api → cli`. Not the template |
| `npm run type-check:all` | From the log: `type-check:host && type-check:packages && type-check:packages:test` → mfes, gts-plugin, api, api:test, cli:test. Not the template |
| `npm run lint` | `template-shell` is not installed, so `@gears-frontx/gts-plugin` is not even on disk to resolve against; and `scripts/**` is ignored outright |
| `npm run test:unit --coverage` | `Test Files 8 passed (8)` — the same 5 workspace packages. The template's own suite never runs; `discovery.mjs:7-11` says it "runs its own tests via its own package.json `test:unit` script", and nothing invokes that |

Nothing in `main.yml` enters `template-shell`. The only job that does is `publish-template`, which lives in a different workflow, triggers on `push` rather than `pull_request`, and is gated on a template version change. This is the same "checks exist, nothing runs" pattern flagged in #483: the split removed the host project from the runner and left the dead `src/mfe_packages/*/` loop behind, and no replacement for `template-shell` was added.

### 4. NIT — `template-mfe/` is left with no lockfile at all

After the two deletions the repository has exactly `package-lock.json` and `template-shell/package-lock.json`; `template-mfe/` has neither a `package.json` nor a lock. That is formally right — the MFE template overlays a shell project whose lock governs the tree — but `template-shell/package-lock.json` carries 0 entries under `mfe_packages`, so reproducibility of an assembled shell+mfe project is pinned by nothing that ships. That belongs to #489/#490, but applying this PR is what opens it.

---

## Verdict: REQUEST CHANGES

The engineering in this round is strong. The `dist/` guard came back better than requested — reading the entry point from `exports` and validating every package before the first write, so a half-linked tree is unreachable. The Windows branch, the refusal taxonomy and the tests are all correct, the workflow gate is fixed precisely, the description's overstated claims were corrected honestly, and the rebase onto the split — lockfile regeneration included — lost nothing.

Two asks, both small:

1. **Make the record match reality.** The Before/After table and the Verification section still assert a green seeded e2e that does not exist. And decide the merge order: publishing `mfes`/`gts-plugin` and pinning to those closes #485 completely instead of halfway, and keeps `publish-template` from breaking.
2. **Wire `scripts/` into the test runner**, or say explicitly that the new test is not enforced in CI yet.

itechmeat:
@gs-layer Ready for re-review. Since round 2:
1. mfes/gts-plugin bumped to 0.3.0-alpha.1 per your suggestion - they publish on merge, the template pins move to them in an immediate follow-up PR (a lockfile can't pin an unpublished tarball), and the body now says Refs #485 so the issue stays open until that follow-up proves the seeded build;
2. the record is corrected - the green e2e in the table is marked as pre-rebase, current pins fail on the moved exports (documented in CONTRIBUTING);
3. scripts/ is wired into the runner as repo-scripts - all five dark test files now run in test:unit;
4. beyond your asks: the linker now stages installed dirs aside and rolls back on failure instead of delete-then-link (with failure-injection tests), and the two stale vitest.config.mjs comments are fixed.

gs-layer (approved):
## Review round 3

Re-checked at head `1ac079b3` against develop `a0fd54aa`, including a live run of the documented dev loop on a real filesystem. Verdict at the bottom.

### Round 2 findings — status

**BLOCKER (the pinned template does not build; the PR body claimed it did; `publish-template` armed to fail) — resolved as far as one PR can.**

- **The record now matches reality.** `Closes #485` → `Refs #485`, with "#485 stays open until the follow-up pin bump to `alpha.1` proves the seeded build and smoke checks end-to-end". The Before/After row and the Verification section now separate install-green from build-red and mark the earlier green run explicitly as **pre-#490-rebase**.
- **A mechanism to clear it was added.** `packages/mfes` and `packages/gts-plugin` are bumped to `0.3.0-alpha.1`, which PART 1 publishes on merge; the template's own pins move in a second PR. Confirmed against the registry that `0.3.0-alpha.1` is not there yet, so leaving the pins on `alpha.0` is the only possible choice — "a lockfile cannot pin a tarball that does not exist yet" is correct.
- **This merge will not break CI.** `publish-template` is gated on `detect-template.has_changes`, and no template package version is bumped, so the job does not run.
- **The documented workaround is verified live, not asserted.** Full CONTRIBUTING loop on a real filesystem:

```
npm run build:packages (root)            exit 0
npm run dev:template:link                exit 0 → 3 links, correct relative targets
                                         leftover .frontx-link-backup: none
                                         framework/auth/state/i18n/react/studio untouched
template-shell: npm run build:package    exit 0
template-shell: npm run build:packages   exit 0    ← 0 TS errors
template-shell: npm run type-check       exit 0    ← 0 TS errors
```

Not one error on `FRONTX_ACTION_*` or `DomainContext.typeSystem`. "Running `dev:template:link` … and both go green" holds literally.

**MAJOR (the new test never runs in CI) — closed.**

```
host:repo-scripts root=scripts        ← new
workspace:api / cli / cyber-pilot-kit-frontx / gts-plugin / mfes

npm run test:unit  exit 0
  repo-scripts  5 files / 201 tests
  + 8/131, 27/230, 13/131, 6/53 across the packages
```

`npm run lint` exits 0. (My first run failed on an unbuilt `@gears-frontx/eslint-config` — my environment, not the PR.)

Removing `extraRootPaths` along with its branch in `routing.mjs` rather than patching around it is the right call: the mechanism stopped being needed, so it is gone instead of dangling.

### Claims I checked that turned out exact

| Claim | Measurement |
|---|---|
| eslint deferral: "69 of the 73 errors are `no-undef` …, the remaining 4 are real and pre-existing" | `eslint --no-ignore scripts/` → **73** exactly: 69 `no-undef`, 2 `no-useless-assignment`, 1 `preserve-caught-error`, 1 unused import. Precise to the number |
| Named `vitest.scripts.config.mjs` rather than `vitest.config.mjs` so `packages/gts-plugin` cannot inherit it | `packages/gts-plugin` is the **only** ecosystem package without a vitest config of its own; api/cli/mfes/cyber-pilot-kit-frontx all have one. The hazard is real |
| Glob anchored at `scripts/` because of `.claude/worktrees/*` | Sound — a worktree is a second full copy of the repo |
| The bump does not violate the version policy | `policy:version-check` exits 0: mfes declares `^0.3.0-alpha.0`, satisfied by `gts-plugin 0.3.0-alpha.1` |
| Root lockfile consistent | `packages/mfes` and `packages/gts-plugin` at `0.3.0-alpha.1` |
| Template lockfile | 685 entries, 0 under `mfe_packages`, 0 escaping resolutions, three registry pins with integrity, a single `mfes` copy |

---

## New in round 3

### 1. MINOR — the pin surface grew into `packages/gts-plugin`, and a divergence there yields two MFE runtimes

`packages/gts-plugin/package.json` now declares `dependencies: { "@gears-frontx/mfes": "0.3.0-alpha.1" }` — an exact pin that has to move in lockstep with the template's own `mfes` pin. They agree today: the lockfile holds one `node_modules/@gears-frontx/mfes` at `0.3.0-alpha.0`, and gts-plugin's recorded dependency is the same.

If they ever diverge — the template on `mfes@X` while the published `gts-plugin` depends on `mfes@Y` — npm hoists one and nests the other, and a seeded project ends up with **two copies of the MFE runtime substrate**. That is precisely the skew `ecosystem-version-policy.mjs` refuses on the peer edge:

> `"${declaredRange}" is an exact pin — an exact pin forces duplicate-runtime skew`

…but it can never see this one, because `readDeclaredEdgeRange` reads only `mfes.peerDependencies['@gears-frontx/gts-plugin']`.

**Ask:** make sure #498's drift guard covers `packages/gts-plugin`'s `mfes` dependency, not just the template manifests. Otherwise the guard will pass a tree that installs two runtimes.

### 2. NIT — the publish order leaves `gts-plugin@0.3.0-alpha.1` briefly uninstallable

`detect-ecosystem` builds its list with `for pkg_dir in packages/*/` — alphabetical, so `gts-plugin` before `mfes` — and `publish-ecosystem` publishes sequentially in that array order. So `gts-plugin@0.3.0-alpha.1` reaches the registry before the `mfes@0.3.0-alpha.1` it pins exactly. The window is one publish step inside a single job, and F1 explicitly disclaims dependency ordering, so this may be a deliberate non-issue — but PART 2 does `sort_by` for exactly this reason, and this PR is the first change that makes the edge live.

### 3. NIT — `restored` lists packages that had nothing to restore

`restoreInstalledTree`:
```js
fs.rmSync(linkPath, { recursive: true, force: true });
if (backupPath !== null) { fs.renameSync(backupPath, linkPath); }
restored.push(name);
```

Paired with the new "links a package the scope directory never had installed" path: if one package is absent from the scope and a later one fails, the report says *"Rolled @gears-frontx/api … back to the installed versions"* for a package that had no installed version and is simply absent again. The resulting state is correct and "nothing was left half-removed" is true, so this is wording rather than corruption — worth mentioning only because the last two commits exist to make exactly these words precise, and `LinkRollback`'s own JSDoc defines `restored` as "returned to the state `npm ci` left them in", which an absent package is not.

### 4. NIT — the glob matches only `.mjs` while CONTRIBUTING promises any new test file

`include: ['scripts/**/*.test.mjs']` versus CONTRIBUTING's *"A new test file under `scripts/` needs no wiring, but a new root-level directory of tests does."* A future `scripts/foo.test.ts` would silently not run — the same "collected by nothing" failure this config exists to fix. Either widen to `*.test.{mjs,ts}` or say `.mjs` in the doc. All five current files are `.mjs`, so nothing is dark today.

### 5. Unchanged — `template-mfe/` has no lockfile

The repository is left with `package-lock.json` and `template-shell/package-lock.json`. `template-shell/package-lock.json` carries no entries under `mfe_packages`, so reproducibility of an assembled shell+mfe project is pinned by nothing that ships. That belongs to #489/#490, but applying this PR is what opens it.

---

## Verdict: APPROVE WITH COMMENTS

The blocker is resolved as far as a single PR can take it, and what remains is recorded honestly rather than papered over. The engineering grew again in the right direction: staging by `rename` instead of `rm` makes a failed link cost a re-run rather than an `npm ci`; the rollback separates `restored` from `unrestored` instead of issuing a blanket "run `npm ci`"; and the two steps get their own `try` so the message names the syscall the filesystem actually refused. The test fake models *what an entry is* rather than only that a path exists, which is what lets a case prove the rollback put a directory back instead of leaving a hole. The eslint deferral is backed by a measurement, not a shrug.

None of the four remaining items blocks. The only substantive ask is finding 1, and it is really addressed to #498 rather than to this branch.

**Merge-order conditions** (not blockers, but worth stating so they do not get lost):

1. After merging, confirm PART 1 actually published `mfes@0.3.0-alpha.1` and `gts-plugin@0.3.0-alpha.1` before merging the follow-up that moves the pins.
2. The follow-up needs to bump `template-shell`'s own `version` too — otherwise `detect-template.outputs.has_changes` stays `false`, `publish-template` never runs, and the template's build against the new pins goes unverified in CI. That check is the one place CI enters `template-shell` at all.

GeraBart (approved):

PR #498, opened 2026-07-31
feat: template content self-containment validation, ecosystem pin-drift guard, and pins on the published alpha.1 (#493, #485)

### LINKED ISSUE #485: [BUG] Seeded project cannot install: ecosystem dependencies are pinned by monorepo-local `file:` paths

### LINKED ISSUE #493: [ARCH] Prepublish template validation does not check content self-containment
## Description

Prepublish template validation (`cpt-frontx-fr-cli-template-validate-prepublish`, FEATURE `template-manifest`) checks only the manifest itself: schema, required fields, ownership-boundary shape. Nothing checks the **content**: no rule forbids a file inside a declared exclusive subtree from referencing a filesystem path that resolves outside the template root.

This gap is exactly how the entire #485 class lived undetected. Three different carriers of the same escape, all passing validation:

- `file:../packages/*` specifiers in `package.json`
- `../packages/*/src` mappings in `tsconfig.app.json` `paths`
- `"resolved": "../packages/api"` entries in `package-lock.json`

```mermaid
flowchart LR
  v["frontx validate"] -->|"schema, fields,<br/>boundary shape"| m["frontx-template.json"]
  v -.->|"content not checked"| c["package.json / tsconfig /<br/>lockfile inside owned subtree"]
  c -->|"file:../.., paths,<br/>resolved: entries"| out["paths outside<br/>the template root"]
  style out stroke-dasharray: 5 5
```

#485 (PR #492) fixes the found instances, not the class. And the class has already reproduced: #489 records the same dangling `file:../../../packages/*` pattern in the newly created `template-mfe` (#470 / PR #490) - filed before any guard existed. Every new template inherits the trap until validation catches it.

Work:

1. New instruction in FEATURE `template-manifest` plus a DoD criterion: no file inside a declared exclusive subtree may reference a filesystem path that resolves outside the template root.
2. CI check implementing it. Generic and template-agnostic (inspects the validated candidate, knows no template names), so it stays safe for constraint CLI-1. Carriers to cover from day one: `file:` specifiers in `package.json`, `paths` in tsconfig files, `resolved` entries in lockfiles.
3. Pin-drift guard: CI comparison of `packages/*/package.json` versions against the template's registry pins. After PR #492 the pin surface is 14 sites across 5 manifests plus the package list in the dev link script; a manual bump that misses one site ships a mixed version set, and the dev loop actively masks it (it links local sources regardless of pins). The script's package list must derive from the same source as the pins, not duplicate it.

## Estimate

~1 engineer-day per check; no pending contract decisions - the pinning mechanism is already settled in #485 / PR #492.

## Risks / open questions

- Timing: should land before or together with the #470 split (PR #490), otherwise the two new templates start life unguarded - #489 shows this is not hypothetical.
- False positives: relative paths that stay inside the subtree are legitimate (the second class in #485) - the check must resolve paths, not pattern-match `../`.
- Lockfile `resolved` entries vary across npm versions - parse, don't grep.

Grounded in the #485 investigation (develop @ 37b09ee3) and its architect review; the 14-site pin count reflects PR #492.

itechmeat:
Closes #493. Closes #485.

## What this PR does

Two complementary safety nets for the template ecosystem, plus the pin bump that completes the #492 two-merge sequence:

1. **Content self-containment validation** (#493) - prepublish check that no file in a template's declared content-owning paths references a path resolving outside the template root. Carriers: `package.json` `file:` specifiers, tsconfig `paths`/`extends`/`references`/`files`/`include`/`exclude`/`typeRoots`/`outDir`, lockfile `resolved`/workspace keys (v1 and v2+). Segment-based resolution, fail-closed, wired into `validate:templates`.
2. **Ecosystem pin-drift guard** (#493, review ask from #492) - `scripts/template-pin-drift-check.mjs` fails CI when any pinned site (template manifests AND every `packages/*` manifest, e.g. `gts-plugin`'s exact `mfes` dependency) drifts from the ecosystem's actual versions. Fail-closed; `file:`/`link:`/`workspace:` specifiers are correctly not treated as version pins.
3. **Pins moved to the published `alpha.1`** (#485 follow-up promised in #492): `template-shell` now pins `mfes`/`gts-plugin` at `0.3.0-alpha.1` (`api` stays `0.3.0-alpha.0` - its latest published version), lockfile updated, template version bumped to `0.1.0-alpha.1` so `publish-template` runs on merge.

Also included per #492 round-3 review: registry publish order is now dependency-aware (`mfes` before `gts-plugin`), the linker rollback report distinguishes `restored` vs `cleared`, the scripts test glob covers `*.test.{mjs,ts}`, and CONTRIBUTING's known-issue block is replaced by the steady-state silent-staleness note.

## Proof for #485 item 3: seeded build outside the monorepo

Seed -> `npm ci` (npm 11.6.2) -> build -> type-check -> unit tests -> `vite preview` smoke, in a temp dir with no monorepo access; all `@gears-frontx/*` packages resolved from the real registry:

| Step | Result |
|---|---|
| `frontx seed @gears-frontx/frontx-template-shell ./my-app` | Apply complete |
| `npm ci` | 560 packages, exit 0 |
| `npm run build` | exit 0 |
| `npm run type-check` | exit 0 |
| `npm run test:unit` | 569 tests, exit 0 |
| `vite preview` + curl | HTTP 200, assets served |

Resolved from registry: `mfes@0.3.0-alpha.1`, `gts-plugin@0.3.0-alpha.1`, `api@0.3.0-alpha.0`.

Known pre-existing gap found by this run (not a regression, candidate for a separate issue): the seeded project cannot `npm run lint` - `eslint.config.js` imports `typescript-eslint`, which the template does not declare; it resolves only via monorepo root hoisting.

## Notes for reviewers

- The `template-shell` lockfile was updated surgically (13 lines, versions + registry integrity hashes only): npm 11's `--package-lock-only` write drops the `file:../..` self-link entry that it then demands on read. A clean `npm ci` from the committed lockfile passes. Removing the `file:../..` self-dependency would fix regeneration properly - follow-up candidate.
- `packages/framework` version is not bumped: only its devDependencies changed, consumers never install those.
- CodeRabbit's Docstring Coverage gate (65.79% < 80%) is a repo-level config threshold, unrelated to this diff.

## Summary by CodeRabbit

* **New Features**
  * Added template validation for references escaping declared boundaries.
  * Added dependency version pin-drift checks and automated template discovery and validation.
* **Bug Fixes**
  * Improved ecosystem linking rollback, stale-link warnings, and package publication ordering.
  * Added clearer failure reporting for invalid or unreadable template content.
* **Documentation**
  * Updated template development guidance for pin compatibility and linking behavior.
* **Tests**
  * Expanded coverage for validation, discovery, linking, filesystem handling, and version drift.


packages/cli/src/manifest/validate-content-self-containment.ts
```diff
@@ -0,0 +1,524 @@
...
+        violations.push({
+          field: fileRelPath,
+          message: `is a declared ${kind} carrier that could not be read (${describeError(error)}), so its path references cannot be checked`,
+        });
+        continue;
+      }
+
+      let parsed: unknown;
+      try {
+        parsed = JSON.parse(raw);
```
SysoevAndrey: TypeScript reads `tsconfig*.json` with a JSONC parser so comments and trailing commas are legal and `tsc` accepts them. But if it's parsed using `JSON.parse` an error will be thrown. So it would be a false positive and the message sends the developer to fix broken JSON that isn't broken

SysoevAndrey (approved):

packages/cli/src/manifest/validate-content-self-containment.ts
itechmeat: Good catch, thank you, fixed

gs-layer (changes requested):
Please address these before merging:

1. `packages/cli/src/adapters/fs-project-io.ts:95`  
The self-containment check currently passes when a declared ownership path is missing, is a broken symlink, or resolves outside the template. The adapter returns an empty file list, and the validator treats that as a successful inspection. This is fail-open behavior for a prepublish guard.  
Suggested fix: return an error for missing, broken, or out-of-root declared boundaries and make `frontx validate` fail instead of returning `[]`. Please add regression tests for a missing directory and a symlink outside the template root.

2. `packages/cli/src/manifest/parse-jsonc.ts:141`  
`parseJsonc()` does not handle a UTF-8 BOM. TypeScript accepts a BOM-prefixed `tsconfig.json`, but the validator rejects it even though the config is valid.  
Suggested fix: strip a leading BOM before processing comments and trailing commas, and add parser-level and `validateCommand` regression tests for BOM-prefixed tsconfig files.

itechmeat:
Two commits in this round fix inconsistencies that this PR's own new guards caught on their first CI runs. Both were introduced on develop by #476 and arrived here with the develop merge:

- template-shell/frontx-template.json still declared `internal/` as an exclusive ownership subtree after #476 deleted that directory as dead code. The fail-closed content check refused the missing boundary; the stale entry is dropped in d024d5b3.
- template-shell's `@gears-frontx/api` pin stayed at 0.3.0-alpha.0 after #476 bumped the package to 0.3.0-alpha.1. The pin-drift guard flagged it; the pin moves to the published alpha.1 in 309d77fc.

Both mismatches sit on develop today and are invisible there, because the guards that detect them only exist on this branch. Fixing them here rather than carrying them forward; nothing else in those files changed.

gs-layer (approved):

GeraBart (approved):

PR #500, opened 2026-07-31
fix: multi-template state integrity — provenance record set on upgrade, region carry-forward on add (#487, #488)

### LINKED ISSUE #487: add-template truncates region-union shared files to the incoming template's blocks
## Description

`frontx add` truncates a `region-union` shared file to only the incoming template's marker blocks. Blocks contributed by previously applied templates are silently destroyed.

Root cause (two cooperating defects in `packages/cli/src/scaffold/`):

1. `compose-shared-files.ts` assembles a region-union file **only from marker blocks present in the current assembly** — content outside markers is dropped, even with a single contributor.
2. `addTemplate` -> `materializeAssembly` passes only the **newly applied** templates' assembly; `composeSharedFiles` never reads the file already on disk in the target repository.

Consequence: `seed A` -> `add B` over a shared `region-union` file leaves only B's block. Verified on synthetic templates with **disjoint** region keys (the pre-flight conflict check correctly passes; the loss happens at materialization).

## Reproduction

Pinning test (marked as known defect): `packages/cli/src/__tests__/template-split.pinning.test.ts` -> "Fixture 7 (F6-pinning)". Single combined assembly preserves both blocks; `seed A -> add B` truncates.

## Impact

No template uses `sharedFiles` today (`template-shell`/`template-mfe` ship `sharedFiles: []` — decided in #470 partly **because** of this defect). This blocks any future real use of the shared-file tier declared in ADR-0031.

## Fix sketch

At `add` time, read the existing target file, extract marker blocks owned by templates **outside** the current assembly, and merge them with the incoming assembly's blocks before writing.

## References

- ADR-0031 (two-tier ownership boundary, declared merge), ADR-0032 (refuse-not-merge)
- Surfaced by the first real multi-template run in #470

### LINKED ISSUE #488: upgrade rewrites provenance.json as a single object, destroying sibling records in multi-template repos
## Description

`frontx upgrade` in a multi-template repository silently destroys the provenance records of every template except the first one, then reports success.

Chain (all confirmed by test):

1. `adapters/provenance-io.ts` bridges the record **set** onto the single-record seam as `records[0] ?? null` — upgrade always targets the first applied template.
2. `upgrade/apply.ts:127-131` writes `.frontx/provenance.json` back as **one object**, not an array — sibling records are gone.
3. The change set is empty when baseline and target resolve to the same inventory content, so the run looks like a clean no-op upgrade.
4. The next `frontx add` crashes: `for (const record of records)` over a non-array -> `TypeError: ... is not iterable` (`readProvenanceRecords` is an unchecked `JSON.parse` cast).

Violates ADR-0021 Confirmation (d) ("the other applied template and its provenance record are unaffected") and ADR-0019 (one record per applied template).

## Reproduction

Pinning tests (marked as known defect): `packages/cli/src/__tests__/template-split.pinning.test.ts` -> "Fixture 9 (B1-pinning)": `seed shell -> add mfe -> upgrade --yes` => provenance becomes a single object, the mfe record string is absent from the file; follow-on `addTemplate` rejects with "not iterable".

## Interim mitigation (shipped in #470)

Template docs state: "upgrade is not yet supported in a multi-template repository."

## Fix sketch

- `apply.ts`: read the full record set, update only the matching record, write the **array** back.
- Longer term: per-template upgrade target selection instead of `records[0]` (separate design decision).

## References

- ADR-0019, ADR-0021; surfaced and pinned in #470

gs-layer:
Two ways a multi-template repository silently loses state. Both were pinned as known defects in `template-split.pinning.test.ts` by #470; both pins are now correctness assertions.

Closes #488. Closes #487.

Base `develop` @ `a0fd54aa` (post-#490). Two commits, kept separate because they are asymmetric: #488 is a code-vs-spec divergence with no artifact change, #487 required amending the FEATURE first.

## Commit 1 — `frontx upgrade` destroyed sibling provenance records (#488)

`upgrade` in a shell+mfe repository rewrote `.frontx/provenance.json` as **one bare object**, deleting every applied template's record but the first, and reported success. The next `frontx add` then crashed with `TypeError: ... is not iterable`.

**The spec was already right.** `architecture/features/upgrade-changeset/FEATURE.md:133` reads "Update the selected applied template's provenance record to the newer version" — the code under that marker spread a single record and serialized it alone. So this commit touches no architecture artifact and adds no `@cpt-` markers; it makes the code do what `inst-app-update-prov` already says.

- `upgrade/apply.ts` parses the full record set from the content already snapshotted for rollback (never re-reads disk), replaces only the targeted record's `scaffoldedFromVersion`, and writes the **array** back.
- Validation of the set and lookup of the target record happen **before the first project write**. A malformed or inconsistent provenance file returns `{ok: false, message}` with the project byte-for-byte untouched, rather than throwing past the snapshot-restore window — `applyChangeSet` is declared as returning `ApplyResult`, and an escaping exception bypasses both `inst-app-restore-on-error` and `inst-app-return-failure`.
- `adapters/provenance-io.ts` rejects a non-array payload with an error naming the file and the shape found, instead of an unchecked cast surfacing as an unrelated `TypeError` in a distant caller.
- The single-record bridge still targets `records[0]` — per-template upgrade target selection is a separate design decision, deliberately out of scope — but now says on stderr which record it picked and which ones it could not.

Satisfies ADR-0019 (one record per applied template) and ADR-0021 Confirmation (d) ("the other applied template and its provenance record are unaffected").

**Left in place deliberately:** the template docs still say multi-template `upgrade` is unsupported. The destructive corruption is gone, but you still cannot *choose* which template to upgrade, so the statement remains true.

## Commit 2 — `frontx add` truncated region-union shared files (#487)

`add` composed a `region-union` file **only** from the incoming assembly's marker blocks, silently destroying blocks contributed by previously applied templates. Disjoint region keys, so the pre-flight conflict check correctly passed — the loss happened at materialization.

**Here the spec itself was incomplete**, unlike #488: `cpt-frontx-algo-cli-scaffolding-compose-shared-files` declared its input as the staged assembly plus a target path, and no step read the file already on disk. So this commit is spec-first.

Three on-disk cases, resolved as:

- **owner recorded in provenance** → carried forward **verbatim**. Re-deriving from the owner's installed content was rejected: `add` would silently up- or downgrade another template outside a reviewable change set.
- **owner in neither the assembly nor provenance** → **refused**, framed as evidence that the occupied set the pre-flight compared was incomplete. The instruction states outright that an on-disk block is **not** an ownership declaration — inferring a boundary from emitted output is the option ADR-0031 explicitly rejected (lines 43, 49, 90-92), and the wording must not revive it.
- **text outside markers** → out of scope, filed separately. That question is whether a region-union file is a concatenation of blocks or a host document with anchors, and the region-addressing schema belongs to `cpt-frontx-feature-template-manifest`, not to `cli-scaffolding`. Worth knowing: the current implementation already answers it de facto — composition is `orderedRegions.map(r => r.markerBlock).join('\n')`, so unmarked text is dropped even at seed, and no artifact records that as intentional.

**No ADR change.** The precedent is `inst-cs-if-span-overlap`: a materialization-time refusal condition that ADR-0032 never mentions, introduced at FEATURE level and accepted without touching the ADR. Refusing an unrecorded owner is the same move — a property of repository content that a comparison of declarations cannot observe.

**Also fixed here, and deliberately so:** `composeSharedFiles` used to write single-owner files inside the same loop that could later return `span-overlap`, so a refused assembly could already have written files — against ADR-0032's "a refused assembly writes zero files". Adding a new refusal reason on top of that would have made it worse. The function is now genuinely two-phase: compute every file and every refusal, then write.

## Two things for review

**The new parameters are optional**, defaulting to "nothing on disk, nothing applied" — which reproduces the pre-fix behaviour exactly. Every production path passes real values (`seedRepository`, `addTemplate`, `scaffoldComposedProject`, with `cli.ts` handing them `createFsReadProjectFileFn()`), verified by reading each call site. But `composeSharedFiles` and `materializeAssembly` are public re-exports of `packages/cli/src/index.ts`, so an external caller omitting the argument gets the old behaviour silently. The default exists because making it required would have broken 8 call sites in `template-split.e2e.test.ts`, which was being edited on the branch for #489. **Worth making required once both land.**

**The reverse marker scanner has a documented parsing limit.** It splits `identity:key` on the first whitespace boundary, then the first `:`. A region key containing whitespace, or a comment closer touching the token with no separating space, is outside what it can resolve. Root cause is upstream: the manifest contract places **no charset restriction on region keys**. Documented in the scanner's docstring rather than papered over.

## Verification

Main checkout, after rebasing onto post-#490 `develop` and re-running everything: `packages/cli` `test:unit` **235/235** (27 files), root `test:unit` green across all packages, `type-check` and `type-check:test` clean, `npx tsx scripts/test-architecture.ts` **12/12**, `cfs validate` 0 errors, `eslint --max-warnings 0` on every changed file.

A note for anyone verifying in a `git worktree`: the architecture check's baseline there is **11/12** — `API-1` fails regardless of any change, and a worktree has no `node_modules` of its own. Confirmed by stashing and re-running. Run it in the main checkout.

## Summary by CodeRabbit

* **New Features**
  * Preserves previously generated shared-file sections when adding templates or upgrading projects.
  * Supports projects containing provenance records for multiple templates.
  * Upgrade diagnostics identify the selected template and unavailable templates.

* **Bug Fixes**
  * Rejects malformed or conflicting file markers and provenance data before changes are made.
  * Preserves provenance records during upgrades and rollbacks.
  * Prevents partial writes when validation fails.
  * Provides clear, user-fixable refusal messages for unsafe materialization.


packages/cli/src/scaffold/compose-shared-files.ts
```diff
@@ -205,6 +325,58 @@ export async function composeSharedFiles(
...
+    // @cpt-begin:cpt-frontx-algo-cli-scaffolding-compose-shared-files:p1:inst-cs-carry-forward-recorded-blocks
+    // Every located block reaching here is, by the check above, either a
+    // contributor's own (stale) on-disk block — superseded by the fresh
+    // extraction from its installed content just below — or owned by a
+    // template recorded in provenance but NOT contributing to this
+    // assembly: exactly the ground `cpt-frontx-flow-cli-scaffolding-add-template`
+    // must leave untouched. Carried verbatim, never re-derived from
+    // installed content, so `add` cannot silently upgrade a template outside
+    // this assembly's own reviewable change-set.
+    const carriedBlocks = existingBlocks.filter((block) => !contributorIdentities.has(block.identity));
```
itechmeat, line 673: If the on-disk file carries the same identity:key twice, or nested markers, both copies land in carriedBlocks and both get written - the span-overlap check at compose-shared-files.ts:456 compares extracted regions only, and the Map at :432 silently drops the duplicate on the carried side. A collision between a carried and an extracted block is a hard refusal, so this looks asymmetric. Is that state genuinely unreachable, or should carried blocks get the same check?

packages/cli/src/adapters/provenance-io.ts
```diff
@@ -56,10 +88,30 @@ export async function readProvenanceRecords(repoRoot: string): Promise<Provenanc
...
   return async function readSingleProvenance(repoRoot: string): Promise<ProvenanceRecord | null> {
     const records = await readProvenanceRecords(repoRoot);
-    return records[0] ?? null;
+    const selected = records[0] ?? null;
+    if (selected !== null && records.length > 1) {
+      const others = records
+        .slice(1)
+        .map((record) => record.templateIdentity)
+        .join(', ');
+      process.stderr.write(
```
itechmeat: [minor, you can skip] This is now the only process.stderr.write in packages/cli/src outside cli.ts, and with `frontx upgrade --json` it interleaves an unstructured line into the machine-readable output. Could the "which record was picked" note travel back on the result and be printed by cli.ts like every other message - so the adapter stays decision-free and --json stays clean?

packages/cli/src/scaffold/compose-shared-files.ts
```diff
@@ -141,29 +161,127 @@ export function extractOwnedRegion(content: string, templateName: string, region
...
 export async function composeSharedFiles(
   assembly: StagedAssembly,
   targetDir: string,
   writeFileFn: WriteFileFn,
+  // Both default to "nothing already on disk, nothing already applied" —
+  // exactly the previous behavior — for the few call sites that genuinely
+  // have no target repository to reconcile against (e.g. a fresh in-memory
+  // fixture). Every real seed/add path (`seedRepository`, `addTemplate`,
+  // `scaffoldComposedProject`) always supplies its own real values.
+  readProjectFileFn: ReadProjectFileFn = async () => null,
```
itechmeat, line 518: [minor, you can skip] The default reproduces exactly the pre-fix behaviour, and composeSharedFiles / materializeAssembly are public re-exports - an external caller omitting the argument silently keeps the bug. The deferral until #489 lands sounds right, but could a TODO naming #489 sit at each default site, and could the Fixture-9 call at template-split.pinning.test.ts:358 pass the real reader - right now a test rewritten in this PR still runs the old seam?

packages/cli/src/scaffold/compose-shared-files.ts
```diff
@@ -141,29 +161,127 @@ export function extractOwnedRegion(content: string, templateName: string, region
...
  */
 export async function composeSharedFiles(
   assembly: StagedAssembly,
   targetDir: string,
   writeFileFn: WriteFileFn,
+  // Both default to "nothing already on disk, nothing already applied" —
+  // exactly the previous behavior — for the few call sites that genuinely
+  // have no target repository to reconcile against (e.g. a fresh in-memory
+  // fixture). Every real seed/add path (`seedRepository`, `addTemplate`,
+  // `scaffoldComposedProject`) always supplies its own real values.
```
itechmeat: The docstring says every real path including scaffoldComposedProject "always supplies its own real values", but scaffoldComposedProject keeps the same async () => null default and has no production caller - only composition.test.ts, which passes a null stub. Could the sentence drop it or say so? And "(see PR discussion)" won't be findable from the file later - if the manifest contract has a charset hole, an issue or the manifest FEATURE seems like the durable home for it.

packages/cli/src/scaffold/materialize.ts
```diff
@@ -166,8 +170,12 @@ export async function materializeAssembly(
   lookupFn: (name: string) => InventoryEntry | undefined,
   writeFileFn: WriteFileFn,
   provenanceWriteFn: ProvenanceWriteFn,
+  // Defaults to "nothing already on disk" — see composeSharedFiles for why
+  // that default is safe: it reproduces the pre-fix behavior exactly for a
+  // caller that has no target repository state to reconcile against.
+  readProjectFileFn: ReadProjectFileFn = async () => null,
 ): Promise<MaterializeResult> {
-  const composeResult = await composeSharedFiles(assembly, targetDir, writeFileFn);
+  const composeResult = await composeSharedFiles(assembly, targetDir, writeFileFn, readProjectFileFn, existingProvenance);
```
itechmeat, line 268: unrecorded-owner is a user-actionable refusal - its own message ends with "...and retry" - but it exits as EXIT_INTERNAL_ERROR: materialize.ts:178 flattens the union to {ok, message}, add-template relabels it provenance-failed, and cli.ts maps that to internal error. The union widening may be out of scope, but should this path exit as a user error? Right now the new reason field is also unreachable from any caller.

itechmeat (changes requested):

tscbmstubp (commented):
## Review: ⚠️ request changes

Both core fixes are correct, spec-aligned, and verified end-to-end against head `32e3e65b` in an isolated worktree:

- `cfs validate`: **PASS, 0 errors**, code coverage 119/119 — all seven new FEATURE instructions carry matching `@cpt-` markers
- `packages/cli` `test:unit`: **235/235** (27 files); `type-check` and `type-check:test` clean
- Arch check: 11/12 with only API-1 failing — **identical 11/12 at base `a0fd54aa` in the same worktree**, confirming the PR body's worktree-baseline note; not a PR regression
- Spec claims cross-checked against ADR-0019/0021/0031/0032 and both FEATUREs, including that commit 1 correctly needed no artifact change (`inst-app-update-prov` already says "the selected applied template's provenance record")

### On @itechmeat's five open threads — all valid, two substantive

| Thread | Assessment |
|---|---|
| Carried blocks bypass span-overlap/duplicate checks | **Valid, blocking.** Confirmed reachable: `locateAllMarkerBlocks` returns duplicate `identity:key` pairs and nested/overlapping spans; all pass the `carriedBlocks` filter and get composed, and the dedup `Map` at compose-shared-files.ts:432 hides duplicates from the new carried-key-collision invariant. Extracted regions are span-checked; carried blocks are trusted blindly — a hand-edited file is exactly the input class the `unrecorded-owner` refusal was built to distrust. |
| `unrecorded-owner` exits `EXIT_INTERNAL_ERROR` | **Valid, blocking.** Confirmed: `materializeAssembly` flattens to `{ok, message}` (materialize.ts:179), `addTemplate` relabels `provenance-failed` (add-template.ts:150), cli.ts:352–358 maps that to exit 2 — while the message itself ends "…and retry". The typed `reason: 'unrecorded-owner'` discriminant is unreachable from every caller. |
| Defaults keep old behaviour for external callers; Fixture-9 add runs the null seam | **Valid minor.** Confirmed: the `addTemplate` call at template-split.pinning.test.ts:358 omits `readProjectFileFn`. |
| Docstring overstates `scaffoldComposedProject` | **Valid minor.** Confirmed: no production caller — only composition.test.ts, which passes a null stub. |
| Adapter `stderr` write vs `--json` | **Valid minor** (reviewer-flagged as skippable). It's the only `process.stderr.write` in `src` outside cli.ts and also prints into test output. |

### Suggested resolution order

1. **Blocking:** symmetric duplicate/overlap integrity check for carried blocks; classify `unrecorded-owner` as a user error (exit 1) with the refusal reason threaded through `materializeAssembly`/`addTemplate`.
2. **Recommended:** tests for `carried-key-collision` and direct unit tests for `locateAllMarkerBlocks`; real fs reader at template-split.pinning.test.ts:358 + `TODO(#489)` at each default site; docstring fix + durable home for the region-key charset ambiguity.
3. **Follow-up:** adapter stderr note moved onto the result; `deps.writeProvenance` inside the restore-on-error boundary (inline comment).

Inline comments below cover findings not already raised in existing threads.

packages/cli/src/scaffold/compose-shared-files.ts
```diff
@@ -245,6 +417,42 @@ export async function composeSharedFiles(
...
+    );
+    const collidedCarried = extracted.find((region) =>
+      carriedByIdentityKey.has(`${region.templateName}\u0000${region.regionKey}`),
+    );
+    // @cpt-end:cpt-frontx-algo-cli-scaffolding-compose-shared-files:p1:inst-cs-if-carried-key-collision
+    if (collidedCarried) {
+      // @cpt-begin:cpt-frontx-algo-cli-scaffolding-compose-shared-files:p1:inst-cs-return-carried-key-invariant
+      return {
+        ok: false,
+        reason: 'carried-key-collision',
```
tscbmstubp, line 825: No test exercises this `carried-key-collision` refusal — the FEATURE marks `inst-cs-if-carried-key-collision`/`inst-cs-return-carried-key-invariant` `[x]`, and the markers keep `cfs validate` green, but nothing pins the behaviour. `locateAllMarkerBlocks` also has no direct unit tests (duplicate `identity:key` pairs, nested/overlapping spans, unterminated pairs, malformed markers) — precisely the input shapes the carried-block thread above turns on. Suggest adding both alongside the carried-block integrity fix, so the new trust boundary is pinned the same way the #487/#488 defects now are.

packages/cli/src/upgrade/apply.ts
```diff
@@ -124,11 +191,18 @@ export async function applyChangeSet(
...
+  // already validated above, before any file was written, so this step is a
+  // pure substitution into an already-known-good array — it cannot itself
+  // throw. Only the record for the template this upgrade targets changes;
+  // every other applied template's record is re-serialized untouched
+  // (ADR-0021 Confirmation (d): "the other applied template and its
+  // provenance record are unaffected").
+  const updatedRecords = existingRecords.map((record, index) =>
+    index === targetIndex ? { ...record, scaffoldedFromVersion: changeSet.targetVersion } : record,
+  );
+  await deps.writeProvenance(provPath, JSON.stringify(updatedRecords, null, 2));
```
tscbmstubp, line 214: Follow-up, not blocking: this `deps.writeProvenance` call is now the last write on the apply path that can still throw past the snapshot-restore boundary — it sits outside `inst-app-try`, so an fs error here escapes as a thrown exception after entries are already on disk, bypassing `inst-app-restore-on-error`/`inst-app-return-failure` by exactly the argument the precondition block above (:108–116) makes. Pre-existing, and this PR shrinks the exposure considerably; worth an issue so the apply path ends fully within the `ApplyResult` contract.

architecture/features/cli-scaffolding/FEATURE.md
```diff
@@ -257,6 +265,23 @@ The system **MUST** materialize a shared file co-owned by more than one applied
 - Component: `cpt-frontx-component-cli`, `cpt-frontx-component-cli-assembler`
 - Entities: `Template`, `OwnershipBoundary`, `Assembly`
 
+### Preserve Previously-Applied Regions Not Re-Contributed by This Assembly
+
+- [x] `p1` - **ID**: `cpt-frontx-dod-cli-scaffolding-preserve-applied-regions`
```
tscbmstubp, line 279: Scoping observation, fine if intentional: this DoD (and `inst-cs-if-unrecorded-block-owner`) guards `region-union` paths only. An unrecorded owner's content sitting at a path a contributor claims `exclusive` — including a marker block inside it — is still silently overwritten at materialization, since `inst-cs-write-single`/`inst-cs-foreach-single` never consult the on-disk file. That may be exactly the intended boundary (exclusive claims were arbitrated pre-flight against *recorded* occupants), but the same "the occupied picture was incomplete" evidence argument applies. Worth recording the decision — here or in a follow-up — so the asymmetry stays deliberate.

packages/cli/src/scaffold/compose-shared-files.ts
gs-layer: Not unreachable — confirmed and fixed in 4cd0a42b.

`locateAllMarkerBlocks` is a raw scanner: it neither deduplicates repeated `identity:key` pairs nor rejects nested/overlapping spans. So a hand-edited file with a duplicated block yielded two entries, the `Map` kept the last one, and both copies were composed.

Added `inst-cs-if-carried-block-conflict` / `inst-cs-return-carried-block-conflict`, run **before** any lookup keyed by `(identity, regionKey)` is built from the carried set, detecting both duplicate pairs and overlapping/nested spans. Kept as a distinct `reason: 'carried-block-conflict'` rather than folding into `carried-key-collision`: different failure class — evidence the target file was hand-edited or corrupted, not evidence the pre-flight check missed something. The user needs to know whether they are fixing their own file or reporting a bug.

Span comparison within the carried set is unconditionally valid, unlike the cross-buffer extracted case: every carried block came from the one buffer read at `inst-cs-read-existing-blocks`.

Also added direct unit tests for `locateAllMarkerBlocks` (duplicate keys, nested spans, overlapping spans, unterminated pair, malformed marker, empty file, marker inside an arbitrary comment style) — see the reply on the tests thread.

packages/cli/src/scaffold/materialize.ts
gs-layer: Fixed in d652d9d5. The chain was exactly as you traced it.

`MaterializeResult` now carries an optional `composeReason` (typed as `Extract<ComposeSharedFilesResult, {ok:false}>['reason']`), populated only when the failure came from compose and `undefined` for a real `writeProvenance` failure. A shared helper `isUserFixableMaterializeFailure` classifies:

- **user-fixable** → `unrecorded-owner`, `span-overlap`, `carried-block-conflict`
- **not** → `exclusive-contested`, `key-collision`, `carried-key-collision` (pre-flight should have refused; reaching materialization is a mechanism bug), or no `composeReason` at all

`addTemplate` / `seedRepository` report a new `reason: 'materialization-refused'` only in the first case; everything else still reports `provenance-failed` as before. `cli.ts` needed no change — its existing ternary lists the internal-error reasons explicitly, so the new one routes to `EXIT_USER_ERROR`.

The classification lives in one helper shared by both callers so it cannot drift between them.

New test in `cli.test.ts`: `add` against a repo whose shared file carries a block owned by an unrecorded template exits **1**, not 2, with the full refusal message on stderr. Verified it fails against the pre-fix code (exit 2).

packages/cli/src/scaffold/compose-shared-files.ts
gs-layer: Both added in 4cd0a42b, and pinning `carried-key-collision` turned up something the tests were hiding.

**`locateAllMarkerBlocks` unit tests** — duplicate `identity:key` pairs, nested spans, overlapping spans, unterminated pair (begin with no end), malformed marker (no `:` in the token), empty file, and a marker embedded in an arbitrary comment style. These are exactly the shapes the carried-block thread turned on, and they now pin the scanner's contract directly rather than through composition.

**`carried-key-collision` could not be pinned as written — it was structurally unreachable.** Not "unreachable if the pre-flight check worked", but unreachable for every possible input: `carriedBlocks` excludes every contributor identity by construction (`inst-cs-carry-forward-recorded-blocks`), and every extracted region's identity is drawn from that same contributor set, so the two sets could never intersect on identity.

Root cause was the comparison field. `inst-cc-if-region-key-clash` (`conflict.ts:105`) keys its clash purely on `regionKey` between two **distinct** templates — region keys are unique per path, not per identity, and FEATURE:158 says as much. `inst-cs-if-key-collision` for extracted regions is built correctly, keyed on `regionKey`. But `inst-cs-if-carried-key-collision` required `(identity, regionKey)` to match — while its own rationale (FEATURE:189) appeals to the occupied-boundary comparison, which compares differing identities on a shared key.

Fixed in 1ae41ec9: keyed on `regionKey` alone, refusal now names both identities, FEATURE step 8 reworded to drop "the same owning identity and". The guard is now a real invariant — provenance holds template A with key `k`, incoming B declares the same `k`, pre-flight should have refused via the occupied-boundary comparison, and materialization catches it if it did not. Pinned by a test that drives `composeSharedFiles` directly (the invariant-violation scenario), verified to fail without the fix by silently composing the colliding content.

Defect originated in this PR's own spec work, not in the code under review.

packages/cli/src/scaffold/compose-shared-files.ts
gs-layer: Half applied, half turned out to be a no-op — details on the second part.

**`TODO(#489)` added** at each site declaring the `async () => null` default: `compose-shared-files.ts`, `materialize.ts`, `commands/add-template.ts`, `commands/seed-repository.ts`. The parameter becomes required once #501 merges and `template-split.e2e.test.ts` (8 call sites there) can be updated in the same tree. `scaffold/composed.ts` deliberately has no TODO — its default is not blocked by that test, since nothing calls `scaffoldComposedProject` in production at all; noted in the docstring instead.

**Fixture 9 was not changed, because passing the real reader there changes nothing.** Verified empirically: swapped a throwing spy in as the 9th argument and the spy was called **zero** times. Both templates in that fixture declare `sharedFiles: []`, so the assembly resolves entirely through the whole-file `exclusive` branch and never reaches the `readProjectFileFn` read in `composeSharedFiles`. Passing a real reader would have made the test look stricter without exercising anything — worse than leaving it, since a future reader would trust it.

If you want the fixture to genuinely cover the seam, it needs a `region-union` shared file, which is a different fixture rather than an extra argument. The new `carried-block-conflict` and `carried-key-collision` tests in `compose-shared-files.test.ts` cover that path directly.

packages/cli/src/scaffold/compose-shared-files.ts
gs-layer: Both fixed in 5d355540.

**The overclaim** — verified by grep that `cli.ts` never dispatches to `scaffoldComposedProject` and its only caller is `composition.test.ts` with a null stub. The docstring no longer lists it among paths supplying real values; it now says outright that it has no production caller.

**"(see PR discussion)"** removed. The note is self-contained now: the reverse scanner splits the token on the first whitespace boundary, then the first `:`, so a region key containing whitespace, or a comment closer touching the token with no separating space, is outside what it resolves — and the reason is that the manifest contract places no charset restriction on a region key, `cpt-frontx-feature-template-manifest` owning that schema. Agreed a docstring is not the durable home for the underlying gap; filing it as an issue against the manifest contract rather than leaving a pointer to a thread.

packages/cli/src/adapters/provenance-io.ts
gs-layer: Done in 5d355540, though not by threading it onto the result — that path was bigger than it looks.

Widening `ReadProvenanceFn`'s return type reaches `upgrade/compute.ts` and `upgrade/flow.ts` (`ChangeSet` and `UpgradeFlowResult`, both FEATURE-governed and part of the `--json` handshake), plus ~19 mock sites across four test files. That is a rebuild of the upgrade engine's contract for a diagnostic line.

Smaller path taken instead: the diagnostic is **gone from the adapter entirely** — no `process.stderr.write` anywhere in `packages/cli/src` outside `cli.ts` now — and computed in `cli.ts`'s `upgrade` dispatch from the already-injected `readProvenanceRecordsFn`, then emitted like every other message and suppressed under `--json` (`stderr: jsonMode ? undefined : notice`). Adapter stays decision-free, `--json` stays clean, no seam type changed, and the line no longer leaks into test output.

packages/cli/src/upgrade/apply.ts
gs-layer: Confirmed and kept out of this PR, as you framed it.

Checked the boundaries: `inst-app-try` spans 142–171, the `catch` 172–190, and this `writeProvenance` call sits at 205 — outside both. An fs error there escapes as a thrown exception with entries already on disk, bypassing `inst-app-restore-on-error` and `inst-app-return-failure`, which is the same argument the precondition block at 108–116 makes.

Pre-existing, and this PR shrinks the exposure rather than adding to it — the provenance set is now parsed and the target record located before the first write, so the two failure modes that used to throw from inside the update step return `{ok:false}` with the project untouched. What remains is the fs write itself.

Filing an issue so the apply path ends fully within the `ApplyResult` contract. Not folding it in here: it needs the snapshot boundary extended past the provenance write, which is a change to `inst-app-try`'s scope and belongs in its own reviewable change.

architecture/features/cli-scaffolding/FEATURE.md
gs-layer: Intentional, and worth recording as you suggest — but recorded as a decision rather than closed by extending the guard.

Reading the on-disk file on an `exclusive` path pulls in the question this PR deliberately left out: what happens to content that no marker claims at all. `inst-cs-compose-union` is `orderedRegions.map(r => r.markerBlock).join('\n')`, so unmarked text is already dropped at **seed**, not just on `add` — the file model is de facto "a sequence of marker blocks", and no artifact records that as intentional. Whether a region-union file is a concatenation or a host document with anchors is a region-addressing question, and that schema belongs to `cpt-frontx-feature-template-manifest`, not to `cli-scaffolding`.

Your evidence argument does carry across, with one asymmetry that cuts the other way: for a `region-union` path the pre-flight arbitrated *keys* against recorded occupants, so an unrecorded block is evidence that comparison was incomplete. For an `exclusive` path the claim is whole-file by a single owner — there is no key to have been arbitrated, and the owner is entitled to the whole file by declaration. The gap is not that the check is missing; it is that "developer edits inside a template-owned file" has no defined semantics anywhere, `exclusive` and `region-union` alike.

Filing that as its own issue against the manifest contract, and recording the deliberate scoping in the FEATURE so the asymmetry does not read as an oversight later.

gs-layer:
Five findings from the latest review round are fixed. Four commits, `1bf44af1..38d52485`, tests 246 → **259**. Each finding was reproduced from the source before fixing, and each new test verified to fail against the pre-fix code.

**`1bf44af1` — unresolvable markers no longer discard content, and marker matching stopped using substrings.**

`locateAllMarkerBlocks` used to skip past a begin-marker it could not resolve — no `identity:key` separator, or no matching end marker. The file was then recomposed from what the scanner did return, so an unclosed block vanished without a word: silent data loss on `add`/`seed`. The scanner now reports those markers instead of dropping them, and materialization refuses on the first one — before the unrecorded-owner check and before carry-forward trusts the file's shape — naming the path, the 1-based line, and whether the marker is malformed or unterminated. It gets its own refusal reason rather than being folded into `carried-block-conflict`, because it fires earlier, before contributor identity is considered, and applies to any marker on the path.

Separately, end-markers were located with `line.includes(endMarker)`, so a region keyed `scripts` was closed by an end-marker for `scripts-dev` — truncating the block, then causing either a false span conflict or content loss. All containment checks are gone; a shared `parseMarkerLine` helper parses the `identity:key` token and compares fields. The same bug was present in `locateRegionSpan` on **both** its begin and end lookups, not only the end, and is fixed by the same helper — marker lines are now interpreted in exactly one place.

**`f0acbb5f` — two carried blocks sharing a region key across different identities are now refused.**

Region keys are unique per path, not per identity (`conflict.ts:105`, FEATURE:158), so two owners of `shared` on one path is a conflict. The check compared `(identity, regionKey)` pairs, so it passed and both blocks were written. Now keyed on `regionKey` alone, distinguishing a literal duplicate from two identities racing for one key; both refuse under `carried-block-conflict`, naming both identities. Audited the neighbouring checks: `carried-key-collision` and `key-collision` already key on the region key alone, and `span-overlap` compares spans irrespective of identity by design.

**`d81317f3` — provenance array elements are validated before use.**

Both readers only checked `Array.isArray` and then cast. Content of `[null]` therefore reached `findIndex(record => record.templateIdentity === …)` and threw `Cannot read properties of null`, escaping the `ApplyResult` contract. Each element is now validated against `ProvenanceRecord`'s required fields, reporting the offending index and field, in `upgrade/apply.ts` and `adapters/provenance-io.ts` alike. The existing pre-write guard catches it, so a malformed file aborts as `{ok: false}` with the project untouched.

**`38d52485` — `scaffoldComposedProject` reports a materialization refusal as one.**

It returned `reason: 'provenance-failed'` with the message "Scaffold completed but provenance write failed" for every failure — including compose refusals, which became reachable once this function received a real project reader, and where zero files have been written. It now uses the same classification helper as `seedRepository`/`addTemplate` and reports `materialization-refused`, with a message that matches what happened. `architecture/features/composed-provenance/FEATURE.md` needed widening as well: `inst-check-provenance-write-fail` / `inst-report-provenance-fail` named only the provenance-write failure, narrower than the behaviour now reachable.

### On the two follow-up issues suggested

Both confirmed independently. The second is worse than described: `materializeAssembly` never populates `occupiedOwnershipBoundary` at all, so the `?? '.'` fallback in `provenance/write.ts:39` applies on **every** scaffold rather than occasionally — and existing provenance files already carry `"."`, so a fix needs a migration story. The rollback-boundary one is confirmed too: `inst-app-try` spans 142–171, its `catch` 172–190, and the `writeProvenance` call sits at 205.

Filing is on hold pending a go-ahead; a batch from this review will go up together rather than piecemeal.

### Verification

`packages/cli` `test:unit` **259/259**; `type-check` and `type-check:test` clean; `scripts/test-architecture.ts` **12/12**; `cfs validate` 0 errors, coverage 119/119; `eslint --max-warnings 0` clean on every changed file.

The observation that the previous suite did not cover these scenarios was the more useful half of the round — these fixes add 13 tests, each pinned against the pre-fix behaviour.

gs-layer:
All five items fixed. Three commits, `ce4584c8..22157a02`, tests 259 → **271**. Each finding reproduced from the source first; each new test verified red against the pre-fix code.

**`ce4584c8` — orphan and malformed end markers no longer discard content.**

The scanner only looked for begin-marker prefixes; end markers were parsed solely while searching for a specific begin's partner. So a standalone `endregion`, or an end marker whose token had no `identity:key` separator, reached neither `blocks` nor `unlocatable` — and the file was recomposed without that line.

Reworked into two phases: classify every line as a candidate begin or end up front (catching malformed tokens on **both** sides), then match begins to ends nearest-first, with each end consumable by at most one begin. Reported through the existing refusal path as `kind: 'orphan-end'` under `malformed-marker-block`, reusing `inst-cs-if-malformed-marker` / `inst-cs-return-malformed-marker` with broadened wording rather than minting new instruction IDs.

The consumption model also closed a latent bug that was not in the report: two begins sharing one key with a single end previously produced two overlapping "found" blocks instead of one closed block plus one `unterminated` begin.

**`b7625482` — `malformed-marker-block` is now classified as user-fixable.**

This was a regression from the previous round: `1bf44af1` introduced the refusal, and it never reached `USER_FIXABLE_COMPOSE_REASONS`, so `add`/`seed` reported `provenance-failed` and exited as an internal error, and composed scaffold could claim "Scaffold completed but provenance write failed" with zero files written — exactly the defect `d652d9d5` had fixed for the other reasons in that same round.

Covered end-to-end for both `add` and `seed`. Cross-checked every remaining `ComposeSharedFilesResult` reason: `exclusive-contested`, `key-collision` and `carried-key-collision` stay unclassified deliberately — each is a pre-flight-check miss, a mechanism bug the repository owner did not cause and cannot fix by editing their tree. No further gap.

**`22157a02` — a present `occupiedOwnershipBoundary` must be a string.**

Both guards treated "optional" as "unchecked", with a comment saying so outright, and accepted a number or an object. Now validated as `undefined` or a string in both equivalent implementations (`adapters/provenance-io.ts` and the intentional duplicate in `upgrade/apply.ts`, which cannot import the adapter layer), reported by name when invalid, with the stale comments corrected. The field stays optional — that `materializeAssembly` never populates it is the separate defect noted earlier.

Same commit strengthens test `(e)` to assert the index and file location alongside the field name, matching `(d)`, and fixes the MD005 indentation at `FEATURE.md:193` — its parent is the two-digit `10.`, so the nested item needed seven spaces rather than six.

### Verification

`packages/cli` `test:unit` **271/271**; `type-check` and `type-check:test` clean; `scripts/test-architecture.ts` **12/12**; `cfs validate` 0 errors; `eslint --max-warnings 0` clean on every changed file. `markdownlint-cli2` is not wired into this project — absent from both `package.json` and `.pre-commit-config.yaml` — so the MD005 fix was verified by inspection against sibling indentation.

gs-layer:
Both fixed in `5f155677`. Tests 271 → **276**.

**Prefix boundary.** The false refusal was worse than a cosmetic misparse, and it was self-inflicted: before `1bf44af1`/`ce4584c8` a mis-recognised marker cost nothing, because malformed markers were silently skipped. Once they became a hard refusal, `// frontx:regional configuration` anywhere in a shared file started blocking `add`/`seed` outright. Reproduced exactly as reported — the token parses to `al`, which has no `:`, so it returned `'malformed'`.

`parseMarkerLine` now requires whitespace or end-of-line immediately after the prefix. The two cases stay distinct:

- `frontx:regional`, `frontx:endregionally` → prose, `undefined`, no marker and no refusal;
- `frontx:region` with a boundary but nothing parseable after it → still a marker, still `malformed`, refusal preserved.

Verified against the live scanner rather than only through tests: prose yields zero blocks and zero `unlocatable` entries, a legitimate `frontx:region template-a:key1` is recognised (reported `unterminated` when unclosed), and a bare `frontx:region` stays `malformed`.

**Ordering.** `localeCompare` replaced with code-unit `<`/`>` comparisons on both the region key and the template identity. The surrounding code promises byte-identical composition, which locale-sensitive collation cannot guarantee for non-ASCII identities. No other `localeCompare` remains anywhere in `packages/cli/src` — the only occurrences left are in the comment explaining why it is not used. Pinned by a test on a pair whose two orderings actually differ (`école` vs `zebra`), so the assertion fails if collation creeps back in.

No FEATURE change: `inst-cs-if-malformed-marker` specifies how an unparseable marker is classified, not what counts as a marker occurrence — the prefix-boundary rule sits one level below what the spec addresses.

### Verification

`packages/cli` `test:unit` **276/276**; `type-check` and `type-check:test` clean; `scripts/test-architecture.ts` **12/12**; `cfs validate` 0 errors; `eslint --max-warnings 0` clean on both changed files.

gs-layer:
Fixed in `08905291`. Tests 276 → **280**. This was a regression from `5f155677` in the previous round — the boundary guard was right, giving up after the first occurrence was not.

`parseMarkerLine` now walks every occurrence of the prefix in order and returns on the first one carrying a valid boundary, whether that yields a well-formed pair or `malformed`.

**The two-sided case is worse than the memo suggests.** Reproduced live before fixing:

```
// frontx:regional note; frontx:region prior:k
content
// frontx:endregionally note; frontx:endregion prior:k
```

→ `blocks: 0, unlocatable: []`. A well-formed block, invisible as both a block *and* a problem, so composition rewrote the file without it — silent data loss, the same class `ce4584c8` closed in round 3, returning through a different door. The one-sided forms are milder: they surface as `unterminated` / `orphan-end`, which still refuses rather than dropping content.

Verified live after the fix, all six shapes:

| input | result |
|---|---|
| begin after prose | `unterminated` |
| end after prose | `orphan-end` |
| full block after prose | **1 block, no problems** |
| `frontx:regional configuration` | nothing — prose, no refusal |
| bare bounded `frontx:region` | `malformed` |
| ordinary marker pair | 1 block |

Rows four and five are round 4's guarantees, confirmed intact.

**Two real markers on one line** — you did not raise it, but the new occurrence walk made it reachable, so it needed a decision rather than a silent default. The first bounded occurrence wins and the second is not recognised as a begin marker, keeping the function's single-result contract unchanged. Checked that this is not a silent-loss path: the second marker's unmatched end surfaces as `orphan-end`, so the scanner still names the line and materialization still refuses.

FEATURE specs untouched — neither `cli-scaffolding` nor `template-manifest` describes the per-line prefix search at this level, same conclusion as round 4.

### Verification

`packages/cli` `test:unit` **280/280**; `type-check` and `type-check:test` clean; `scripts/test-architecture.ts` **12/12**; `cfs validate` 0 errors; `eslint --max-warnings 0` clean on both changed files. The four new tests were confirmed red against the pre-fix code.

packages/cli/src/scaffold/compose-shared-files.ts
itechmeat: Thanks - the check itself is real (I verified the ordering and the new unit tests), but it turned out to be scoped to the wrong set. It compares spans only inside carriedBlocks, and the span-overlap check only inside extracted regions, so a carried block interleaved with a contributor's own stale on-disk block is compared by neither. I reproduced both shapes end-to-end: a nested pair composes ok and writes the region twice (the stale content comes back and survives every later add), and a crossing pair composes ok and writes a file with an unterminated marker that the next add then refuses as malformed. Both are the same hand-edited-file class this check was added for. Should the span comparison run over all existingBlocks (the filter at compose-shared-files.ts:673) rather than carriedBlocks only - two marker spans on one path can never legitimately overlap, whoever owns them? FEATURE step 6 would need the same widening.

packages/cli/src/scaffold/materialize.ts
```diff
@@ -138,7 +139,63 @@ function sameSourceAddress(left: string, right: string): boolean {
   return a.ok && b.ok && formatTemplateAddress(a.value) === formatTemplateAddress(b.value);
 }
 
-export type MaterializeResult = { ok: true } | { ok: false; message: string };
+// The subset of ComposeSharedFilesResult's discriminant reachable here —
+// derived rather than hand-duplicated, so a new compose refusal reason added
+// there is a type error here until this file (and its two callers,
+// addTemplate/seedRepository) decide which exit code it deserves, instead of
+// silently falling through to 'provenance-failed'.
```
itechmeat: The comment above ComposeFailureReason says a newly added compose refusal reason "is a type error here until ... decide which exit code it deserves" - but a ReadonlySet of a union is not exhaustiveness-checked: I added a member and tsc --noEmit --strict reports zero errors; the new reason silently classifies as not-user-fixable and exits as internal error. That exact sequence already happened inside this PR (malformed-marker-block, rounds 2-3). Either a Record<ComposeFailureReason, boolean> makes the claim true, or the sentence should go. The composed.ts comment listing three of the four user-fixable reasons has the same drift.

architecture/features/cli-scaffolding/FEATURE.md
```diff
@@ -164,25 +165,36 @@ Internal system functions and procedures called by actor flows above.
...
-   5. [x] - `p1` - Compose the repository file as the disjoint union of every contributor's extracted region(s), preserving each region's sentinel markers so a later boundary-scoped upgrade can re-locate it, in a deterministic order (by owning template identity, then region key). - `inst-cs-compose-union`
-   6. [x] - `p1` - Write the composed file to the target repository. - `inst-cs-write-composed`
-4. [x] - `p1` - **RETURN** the materialized repository files. - `inst-cs-return-materialized`
+   9. [x] - `p1` - **IF** a carried-forward block and an extracted region resolve the same declared region key on the path, regardless of whether they share an owning identity — mirroring the pre-flight conflict check's own region-key-clash comparison (`inst-cc-if-region-key-clash`), which is keyed on the region key alone, never on identity, since two different templates never share one - `inst-cs-if-carried-key-collision`
```
itechmeat, line 195: The fix rounds added two more user-facing refusals on top of unrecorded-owner (carried-block-conflict, malformed-marker-block), all three abort frontx add with exit 1, and seed refuses on a malformed marker too - but the Error Scenarios lists of both flows still name none of them, while the algorithm level and composed-provenance/FEATURE.md are fully updated. Could the two flow lists get their entries so the flow level matches the algorithm level?

packages/cli/src/upgrade/apply.ts
```diff
@@ -16,6 +16,90 @@ export type ApplyResult =
...
+// access only through injected deps (never through the adapters layer
+// directly), and the two callers read from different sources (a repo root
+// path there vs. an already-snapshotted string here). Guards each ELEMENT of
+// that array too (review #500 round 2, P2-1): `Array.isArray` alone let a
+// malformed element (e.g. `[null]`, or an object missing `templateIdentity`)
+// through uncast, and the very next line — `existingRecords.findIndex(record
+// => record.templateIdentity === ...)` — either threw a raw TypeError that
+// escaped `ApplyResult`'s `{ok:true}|{ok:false}` contract, or silently
+// misdiagnosed the malformed element as "record not found".
+function isRecordShaped(value: unknown): value is Record<string, unknown> {
```
itechmeat: [Not a blocker] The duplication grew from two functions in round 1 to five byte-identical constructs plus a repeated 14-line justification - and round 3 already had to patch occupiedOwnershipBoundary in both copies, which is the drift cost arriving inside this same PR. The layering argument rules out importing from adapters/, not a shared pure home in provenance/ - these are predicates over unknown with no fs. The duplicate ReadProjectFileFn in scaffold/types.ts and the barrel comment explaining its non-export would fold into the same move. If not now, could a TODO with an issue number sit at both copies?

packages/cli/src/commands/seed-repository.ts
```diff
@@ -95,11 +118,14 @@ export async function seedRepository(
     lookupFn,
     writeFileFn,
     provenanceWriteFn,
+    readProjectFileFn,
```
itechmeat, line 121: [not a blocker] seed now passes the real file reader but still hardcodes [] for existing provenance, so seeding into a directory that already holds recorded blocks refuses with advice to "record provenance and retry" for a record that already exists in a file seed never read - and after this PR that refusal exits 1, so it looks user-fixable while the suggested action changes nothing. If a non-empty target is out of contract, could seed either read provenance like add does, refuse on a non-empty target up front, or at least say that in the message?

itechmeat (changes requested):
Second round. The fix work here is genuinely strong - every thread from round 1 was addressed for real (I verified the fixes by reading and executing the code at head, 280/280 green), and the direction is exactly right. What keeps this at request-changes is one substantive finding plus mechanical catch-ups.

architecture/features/cli-scaffolding/FEATURE.md
gs-layer: Fixed in `2bedd591`, and checking reachability per flow turned the expected answer around.

I assumed `unrecorded-owner` was unreachable at seed. It is the opposite: `seedRepository` hardcodes `existingProvenance: []`, so `composeSharedFiles`' applied-identities set is always empty there — which makes any pre-existing foreign block **unconditionally** unrecorded. Now covered by a test. Meanwhile `carried-block-conflict` is genuinely unreachable at seed: it needs two blocks that already passed the recorded-in-provenance check, impossible when provenance is always empty.

So seed's Error Scenarios list `malformed-marker-block` and `unrecorded-owner`; add's lists all three, since `add` threads real provenance through. No new `@cpt-` markers — Error Scenarios in this file carry none, matching every neighbouring entry.

packages/cli/src/scaffold/materialize.ts
gs-layer: Fixed in `2bedd591`. The comment was making a promise the type system was not keeping, and the proof is already in this PR's own history.

`ReadonlySet<ComposeFailureReason>` replaced with `Record<ComposeFailureReason, boolean>`. Proven rather than asserted — adding a fake member to the union now produces:

```
src/scaffold/materialize.ts(193,7): error TS2741: Property '"TEMP-EXPERIMENT-…"'
is missing in type '{…}' but required in type 'Record<…>'
```

then reverted. Classification itself is unchanged: four user-fixable, three invariant.

That `malformed-marker-block` slipped through rounds 2-3 exactly this way is the part worth recording. The refusal was introduced, never added to the set, the compiler said nothing, and `add` exited with an internal-error code on a refusal whose own message tells the user which line to fix. A reviewer caught it, not the build.

The `composed.ts` enumeration is gone too — it now references the classifier instead of restating three of its four entries, so it cannot drift a second time.

packages/cli/src/upgrade/apply.ts
gs-layer: Done in `9f5d724e` — not deferred to a TODO.

You are right that the layering argument only ever ruled out importing from `adapters/`, and says nothing about a neutral home. These are predicates over `unknown` with no filesystem access, so they now live in `packages/cli/src/provenance/validate.ts` alongside `types.ts` and `contract.ts`, imported by both former copies. Verified the two were byte-identical before extracting, so no message wording moved; the "restore from version control" suffix stays adapter-only, since only that caller can suggest it.

The drift cost you point at was real and had already arrived: round 3 patched `occupiedOwnershipBoundary` in both copies.

`ReadProjectFileFn` I deliberately left duplicated. It is a one-line signature alias with no body — nothing in it can drift the way a predicate can — and `provenance/` is not its home either; it is a generic file-read seam, not a provenance concept. Folding it in would have moved it somewhere no more correct than where it is. Happy to revisit if you see it differently.

packages/cli/src/commands/seed-repository.ts
gs-layer: Fixed in `fbf03824`.

Checked the contract before choosing: the flow is specified as "Developer applies an installed template to an **empty** target directory", and `inst-seed-stage` says "against the **empty** target directory". So a non-empty target is outside seed's contract, not a supported case with a poor message — which rules out your first option (having seed read provenance like `add`).

Of the remaining two I took the message, not the early emptiness check: the check would add a new refusal path and a FEATURE amendment for no behavioural gain over saying the true thing. Seed's refusal now states that the operation expects an empty directory, that this one already holds applied-template content, and points at `frontx add`.

`composeSharedFiles` did not gain a seed/add flag — `MaterializeResult` optionally carries `{path, templateIdentity, regionKey}` for `unrecorded-owner`, and the calling command composes its own text. Also worth noting: this refusal is reachable at seed *unconditionally*, since the applied-identities set there is always empty — that reachability is now recorded in the flow's Error Scenarios as well.

packages/cli/src/scaffold/compose-shared-files.ts
```diff
@@ -141,29 +176,145 @@ export function extractOwnedRegion(content: string, templateName: string, region
...
+    const endIndex = lines.findIndex((line, index) => index > begin.lineIndex && line.includes(endMarker));
+    if (endIndex === -1) continue; // unterminated — no matching end marker to close the span
+    blocks.push({
+      identity: begin.identity,
+      regionKey: begin.regionKey,
+      span: { beginIndex: begin.lineIndex, endIndex },
+      text: lines.slice(begin.lineIndex, endIndex + 1).join('\n'),
+    });
+  }
+  return blocks;
```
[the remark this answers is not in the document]
gs-layer: Already fixed in `1bf44af1`, before this comment's revision.

The substring match is gone entirely. A shared `parseMarkerLine` helper parses the `identity:key` token and compares fields, so a key that is a prefix of a sibling no longer closes the wrong block. Your `scripts` / `scripts-dev` example is one of the regression tests.

Worth adding: the same bug was present in `locateRegionSpan` on **both** its begin and end lookups, not only on the end side — fixed by the same helper, so a marker line is now interpreted in exactly one place.

packages/cli/src/scaffold/compose-shared-files.ts
```diff
@@ -245,6 +504,52 @@ export async function composeSharedFiles(
...
+    // occupied-boundary derivation missed it.
+    const carriedByRegionKey = new Map(carriedBlocks.map((block) => [block.regionKey, block]));
+    let collidedCarried: { carriedBlock: LocatedBlock; extractedRegion: ExtractedRegion } | undefined;
+    for (const region of extracted) {
+      const carriedBlock = carriedByRegionKey.get(region.regionKey);
+      if (carriedBlock) {
+        collidedCarried = { carriedBlock, extractedRegion: region };
+        break;
+      }
+    }
```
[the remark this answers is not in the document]
gs-layer: Already fixed in `f0acbb5f`.

`carried-block-conflict` now keys on `regionKey` alone, so two carried blocks sharing a key across different identities are refused rather than silently composed. The refusal distinguishes `duplicate` (same identity — the file repeats a block) from `key-collision` (two identities racing for one key), and names both identities.

The root cause was the comparison unit: region keys are unique per path, not per identity (`conflict.ts:105`, FEATURE:158). The same mistake had already been fixed once in `1ae41ec9` for the carried-vs-extracted check and survived in this second one.

packages/cli/src/scaffold/composed.ts
```diff
@@ -187,6 +191,7 @@ export async function scaffoldComposedProject(
     lookupFn,
     writeFileFn,
     provenanceWriteFn,
+    readProjectFileFn,
```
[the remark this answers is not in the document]
gs-layer: Already fixed in `38d52485`.

`scaffoldComposedProject` now uses `isUserFixableMaterializeFailure` — the same helper `seedRepository` and `addTemplate` use — and reports `materialization-refused` for user-fixable compose refusals, keeping `provenance-failed` for a genuine provenance-write failure. The message no longer claims scaffolding completed when zero files were written.

`architecture/features/composed-provenance/FEATURE.md` needed widening too: `inst-check-provenance-write-fail` / `inst-report-provenance-fail` named only the provenance-write failure, narrower than what is now reachable.

architecture/features/cli-scaffolding/FEATURE.md
```diff
@@ -164,25 +165,36 @@ Internal system functions and procedures called by actor flows above.
...
+   10. [x] - `p1` - **IF** any two extracted regions on the path have overlapping actual on-disk marker spans — whether owned by different templates or by the same template declaring multiple region keys - `inst-cs-if-span-overlap`
       1. [x] - `p1` - **RETURN** a materialization conflict naming the contested file, the overlapping region markers, and the owning template(s); refuse the assembly and write no file. Actual marker-span overlap is a content-level property that neither pre-publish manifest validation (well-formed keys only) nor the pre-flight conflict check (declared keys only) can observe, so materialization — run per shared-file path regardless of contributor count — is where it is first detectable and refused, covering single-template self-overlap as well as cross-template overlap. - `inst-cs-return-span-overlap`
```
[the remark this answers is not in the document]
gs-layer: Already fixed in `1bf44af1`. The nested item's parent is the two-digit `10.`, so it needed seven spaces rather than six to align with the text after the marker.

Note `markdownlint-cli2` is not wired into this project — absent from both `package.json` and `.pre-commit-config.yaml` — so this was verified against sibling indentation rather than by running the linter.

packages/cli/src/adapters/__tests__/provenance-io.test.ts
```diff
@@ -80,4 +80,32 @@ describe('provenance-io', () => {
...
+    const location = provenancePath(root);
+    fs.mkdirSync(path.dirname(location), { recursive: true });
+    fs.writeFileSync(
+      location,
+      JSON.stringify([{ scaffoldedFromVersion: '1.0.0', sourceSpec: 'github:acme/x@v1.0.0' }]),
+      'utf-8',
+    );
+
+    await expect(readProvenanceRecords(root)).rejects.toThrow(/templateIdentity/);
+  });
```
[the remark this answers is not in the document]
gs-layer: Already fixed in `22157a02`. Test `(e)` now asserts `index 0` and the file `location` alongside the field name, matching `(d)`.

packages/cli/src/adapters/provenance-io.ts
```diff
@@ -31,6 +31,51 @@ export function createFsProvenanceWriteFn(): ProvenanceWriteFn {
...
+// avoid. The fields checked are exactly the `ProvenanceRecord` contract's own
+// required ones (provenance/types.ts) — `occupiedOwnershipBoundary` is the
+// one field declared optional there, so it is not checked here.
+const REQUIRED_PROVENANCE_RECORD_FIELDS = ['templateIdentity', 'scaffoldedFromVersion', 'sourceSpec'] as const;
+
+function isValidProvenanceRecord(value: unknown): value is ProvenanceRecord {
+  return (
+    isRecordShaped(value) &&
+    REQUIRED_PROVENANCE_RECORD_FIELDS.every((field) => typeof value[field] === 'string')
+  );
```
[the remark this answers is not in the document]
gs-layer: Already fixed in `22157a02`. Both guards now require `occupiedOwnershipBoundary` to be `undefined` or a string, and name the field when it is neither. Applied in `adapters/provenance-io.ts` and in `upgrade/apply.ts` — at the time two intentional copies, since consolidated into `provenance/validate.ts` (`9f5d724e`), so this invariant now has one home.

The field stays optional. That `materializeAssembly` never populates it at all — so `provenance/write.ts`'s `?? '.'` fallback applies on every scaffold — is a separate pre-existing defect, raised in review and queued as its own issue.

packages/cli/src/scaffold/compose-shared-files.ts
```diff
@@ -141,29 +238,181 @@ export function extractOwnedRegion(content: string, templateName: string, region
...
+    blocks.push({
+      identity: begin.identity,
+      regionKey: begin.regionKey,
+      span: { beginIndex: begin.lineIndex, endIndex },
+      text: lines.slice(begin.lineIndex, endIndex + 1).join('\n'),
+    });
+  }
+  unlocatable.sort((a, b) => a.lineIndex - b.lineIndex);
+  return { blocks, unlocatable };
+}
```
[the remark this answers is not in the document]
gs-layer: Already fixed in `ce4584c8` — extended, as you suggest, rather than documented as tolerated.

`UnlocatableMarker` gained an `orphan-end` kind, reported through the existing `inst-cs-if-malformed-marker` refusal path. The scanner was reworked into two phases: classify every line as a candidate begin or end first (which also catches a malformed token on the **end** side, previously invisible too), then match begins to ends with each end consumable by at most one begin.

That consumption model closed a defect neither of us had named: two begins sharing one key with a single end previously produced two overlapping "found" blocks instead of one closed block plus one `unterminated` begin.

packages/cli/src/scaffold/compose-shared-files.ts
itechmeat: Following up after the latest fixes: compose-shared-files.ts is untouched since the last round, so this one still stands - the span comparison runs only inside carriedBlocks, and a carried block interleaved with a contributor's own stale on-disk block still composes silently on the add flow (both reproduced shapes from the last round apply as-is). Running the comparison over all existingBlocks still looks like a one-identifier fix, with FEATURE step 6 widened to match. Is this planned here, or does it need to move to a tracked follow-up before merge?

itechmeat (approved):
The PR body still describes the pre-fix state: the old test count, "Two commits", the scaffoldComposedProject claim that this PR's own docstring fix corrected, and the stderr note about a diagnostic that has since moved to cli.ts and is suppressed under --json. Since the body is what the merge record keeps, could it get a holistic refresh in the same push as the remaining code fix?

GeraBart (approved):

PR #501, opened 2026-07-31
feat: template-mfe resolves in the monorepo and pins what it declares (#489)

### LINKED ISSUE #489: normalize MFE file: dependencies and undeclared mf-gts coupling in template-mfe

gs-layer:
Makes `template-mfe`'s packages resolve — both inside this monorepo, where today they resolve nothing at all, and in a seeded project, where what they *declare* is what gets installed.

Closes #489. Two commits: the harness, then the pins. Base `develop` @ `93c94747`.

## The problem

`template-mfe` is an add-only overlay: no root `package.json`, boundaries limited to `src-app/mfe_packages/` and `.frontx/ai/@gears-frontx/frontx-template-mfe/`. Two consequences, one per commit:

1. Inside this monorepo there is no workspace root above the MFE packages, so nothing can resolve `@gears-frontx/*` for them at all. `npm install` exits 0, `npm ls` reports invalid edges, `require.resolve` of the shell's `mf-gts` throws `MODULE_NOT_FOUND`. The MFEs do not build here — matching #490's own note that `template-mfe` is "intentionally not testable in place".
2. What they declared pointed nowhere: 24 `file:../../../packages/*` specifiers resolving to `template-mfe/packages/`, a directory that does not exist. Three of the four also imported `build/mf-gts` from the shell **without declaring it**, resolving through a `peerDependencies` entry belonging to the *other* template.

## Commit 1 — the harness

`template-mfe/package.json`: a workspace glob and nine `file:` overrides, nothing else. Deliberately **not** declared in `frontx-template.json`, so the assembler never copies it into a seeded project and the template claims ownership of no root `package.json` — that is `template-shell`'s exclusive subtree.

**Why this does not violate the ownership model.** ADR-0031's "Declared, not inferred" driver says presence must not create a claim; an undeclared file is the model's normal output. The violation would be the converse — inferring from this file that the template claims a root `package.json`. Two files already sit in exactly this category and are registered in the declaration-coverage test's allow-list: `frontx-template.json` and `template-mfe/README.md`. (`template-shell`, by contrast, *declares* its README and package.json, so it ships them.) ADR-0032 is untouched at three levels: pre-flight compares declarations, `assembler.ts:108` filters content by boundary before staging, and the honesty guard checks what a template *wrote*.

The allow-list gains `package.json`/`package-lock.json` **plus a positive assertion** that `package.json` must never appear in the manifest's boundaries — the allow-list only says "ignore", this says "and it must stay undeclared". Confirmed to fail when the invariant is broken.

## Commit 2 — the pins

All 24 escaping specifiers become exact published versions, matching the local sources they mirror:

```
api 0.3.0-alpha.0 · mfes 0.3.0-alpha.1 · gts-plugin 0.3.0-alpha.1
auth 0.2.0-alpha.1 · framework/i18n/react/state 0.2.0-alpha.0
frontx-template-shell 0.1.0-alpha.0
```

Exact pins rather than ranges or dist-tags, same rationale as #492: `latest` is poisoned for several of these (`@gears-frontx/api` still resolves to `0.1.0`), and a range would silently absorb the next alpha while the template is the ecosystem's reference consumer.

**The shell edge only became pinnable recently.** `@gears-frontx/frontx-template-shell` returned E404 when this issue was filed — that is why the issue body originally proposed standardizing a `file:` pattern instead. It is published at `0.1.0-alpha.0` today, so all four MFEs can finally declare the `mf-gts` import they have always used.

**The harness changes role rather than becoming redundant.** These pins would fetch registry tarballs inside this monorepo, hiding local edits to `packages/*` and `template-shell/packages/*`. Its overrides redirect each pinned name to its local source — they *redirect* rather than *re-declare*, so the installed versions still satisfy the pins. That is why `npm ls` comes back clean rather than merely quieter, and why every `@gears-frontx` entry in the tree is a symlink into local sources rather than a tarball.

## Verified

`npm install` exits 0 and **`npm ls` exits 0** — the `ELSPROBLEMS` goal this issue was filed for. Zero dangling symlinks under any `@gears-frontx` scope directory; `mf-gts` resolves from all four MFEs; `packages/cli` `test:unit` 231/231; `scripts/test-architecture.ts` 12/12. Rebased onto current `develop` and re-run after #492 and #496 landed.

`npm ls --all` still reports edges, all of them from `template-shell` pinning `gts-plugin`/`mfes` at `0.3.0-alpha.0` while local sources are at `0.3.0-alpha.1`. That is pre-existing drift in the shell and is fixed by #498, not here.

## Two notes for reviewers

**Interaction with #491.** Its `template-validate` job skips any `template-*/` without a `package.json`, which today silently skips `template-mfe` — this PR flips it to actually validated, which is the intent of #483. But that job calls `arch:deps`, `lint` and `test:unit` **without** `--if-present` (only the build steps carry the flag) and this harness deliberately carries no scripts. Whichever lands second turns the job red. Raised on #491; preference is adding `--if-present`, which keeps the harness minimal and version-free. Note #491 is currently `CONFLICTING` and still edits the deleted `template-standard/`, so this needs re-checking against its rebased form.

**Relationship to design fixture 8.** Not a substitute. The harness makes the monorepo tree honest (inner loop, IDE, language server); a CI job materializing shell+mfe proves the *assembled* tree works through the production resolution path. The harness introduces a second resolution scheme for the same edge, so a green monorepo stops proving a green seeded project — an argument *for* fixture 8, not against it.

## Summary by CodeRabbit

* **New Features**
  * Added a monorepo development harness with local package overrides for MFE development.
  * Improved template dependency discovery across repository and workspace packages.

* **Documentation**
  * Clarified that the MFE template is add-only and requires no host tooling.
  * Documented published dependency versions, runtime and build prerequisites, installation behavior, and harness limitations.

* **Bug Fixes**
  * Updated MFE examples to use published package versions instead of local file references.
  * Improved template ownership validation and package version drift detection.


packages/cli/src/__tests__/template-split.e2e.test.ts
```diff
@@ -334,12 +334,34 @@ describe('Fixture 6 — declaration coverage: every real file in template-shell/
...
+  // the harness depends on: template-mfe must never CLAIM a root package.json.
+  // Claiming it would collide with template-shell's exclusive subtree and turn
+  // a development-only file into an ownership assertion — reviving exactly the
+  // inferred-ownership model ADR-0031 rejected.
+  it('template-mfe/: the harness manifest stays undeclared — no claim on package.json', () => {
+    const { exclusiveSubtrees, sharedFiles } = readManifest(TEMPLATE_MFE_DIR).ownershipBoundaries;
+    expect(exclusiveSubtrees).not.toContain('package.json');
+    expect(exclusiveSubtrees).not.toContain('package-lock.json');
+    expect(sharedFiles.map((entry: { path: string }) => entry.path)).not.toContain('package.json');
+  });
```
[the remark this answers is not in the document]
gs-layer: Fixed in ca711f85 — the asymmetry was real: `exclusiveSubtrees` was checked against both files, `sharedFiles` against only one.

Extracted `sharedFilePaths` as suggested, added the matching assertion, and renamed the test to cover both root package files.

Verified the new assertion is load-bearing rather than decorative: temporarily added `{ "path": "package-lock.json", "mergeStrategy": "union", "ownedRegions": [] }` to `sharedFiles` in `template-mfe/frontx-template.json`, and the test failed exactly on the new line (`expected [ 'package-lock.json' ] to not include 'package-lock.json'`). Manifest reverted, `git diff` on it empty.

template-mfe/src-app/mfe_packages/demo-mfe/package.json
```diff
@@ -13,7 +13,7 @@
     "test:unit:watch": "vitest --watch"
   },
   "dependencies": {
-    "@gears-frontx/react": "file:../../../packages/react",
+    "@gears-frontx/react": "0.2.0-alpha.0",
```
itechmeat, line 16: The same in `widgets-fixture-a/package.json:26-31` and `widgets-fixture-b/package.json:26-31`

The PR body says all four MFEs can now declare the `mf-gts` import, but at head only `_blank-mfe` declares `@gears-frontx/frontx-template-shell`, and it already did before this PR - the diff just rewrites its specifier. `demo-mfe`, `widgets-fixture-a` and `widgets-fixture-b` all import `@gears-frontx/frontx-template-shell/build/mf-gts` in `vite.config.ts:6` without declaring it, so they resolve only by hoisting from `_blank-mfe` - the disposable scaffold the template's own add-mfe skill expects you to copy from and possibly remove. Could we add the declaration to those three, or adjust the body so the remaining gap is visible?

template-mfe/src-app/mfe_packages/_blank-mfe/package.json
```diff
@@ -34,14 +34,14 @@
...
-    "@gears-frontx/mfes": "file:../../../packages/mfes",
-    "@gears-frontx/gts-plugin": "file:../../../packages/gts-plugin",
-    "@gears-frontx/frontx-template-shell": "file:../../..",
-    "@gears-frontx/api": "file:../../../packages/api",
-    "@gears-frontx/i18n": "file:../../../packages/i18n",
+    "@gears-frontx/framework": "0.2.0-alpha.0",
+    "@gears-frontx/auth": "0.2.0-alpha.1",
+    "@gears-frontx/state": "0.2.0-alpha.0",
+    "@gears-frontx/mfes": "0.3.0-alpha.1",
+    "@gears-frontx/gts-plugin": "0.3.0-alpha.1",
```
itechmeat, line 41: The same in `widgets-fixture-a/package.json:28-29`, `widgets-fixture-b/package.json:28-29`.

This pins `mfes`/`gts-plugin` at `0.3.0-alpha.1` while `template-shell/package.json:70-71` still pins both at `0.3.0-alpha.0`. In a seeded project both layers install as one workspace tree, so npm resolves two exact conflicting pins on the same names by hoisting one and nesting the other with no error - which for the gts registry and MF singletons is a runtime hazard rather than untidiness. `scripts/link-template-ecosystem.mjs:64-70` also documents the `0.3.0-alpha.0` tarballs as unable to build the shell at all. Should the shell pins move in the same change, or is there a follow-up that lands them together?

template-mfe/package.json
```diff
@@ -0,0 +1,19 @@
...
+  "overrides": {
+    "@gears-frontx/frontx-template-shell": "file:../template-shell",
+    "@gears-frontx/auth": "file:../template-shell/packages/auth",
+    "@gears-frontx/framework": "file:../template-shell/packages/framework",
+    "@gears-frontx/i18n": "file:../template-shell/packages/i18n",
+    "@gears-frontx/react": "file:../template-shell/packages/react",
+    "@gears-frontx/state": "file:../template-shell/packages/state",
+    "@gears-frontx/api": "file:../packages/api",
+    "@gears-frontx/gts-plugin": "file:../packages/gts-plugin",
+    "@gears-frontx/mfes": "file:../packages/mfes"
```
itechmeat, line 17: The nine overrides cover the harness's own direct edges, but `build/mf-gts` lives inside `template-shell` and resolves its own `gts-plugin`/`mfes`/`api` from `template-shell/node_modules` - root overrides do not reach into a linked package's subtree, and `npm ls` stays green with both copies present (I reproduced this). So on the one path every MFE build actually uses, local edits to `packages/gts-plugin` and `packages/mfes` are still invisible, against a copy the link script documents as broken. Should the dev loop run `dev:template:link` first, or should the harness note that the plugin path is out of its reach?

template-mfe/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "//": "MONOREPO DEV HARNESS — NOT PART OF THIS TEMPLATE. This file is deliberately absent from frontx-template.json's ownershipBoundaries, so `frontx add` never copies it into a seeded project and this template claims ownership of no root package.json anywhere. The MFE packages below pin their @gears-frontx dependencies to exact published versions, which is what a seeded project installs. Inside this monorepo those pins would fetch registry tarballs, so local edits to packages/* and template-shell/packages/* would be invisible; the overrides here redirect each pinned name to its local source instead. Because they redirect rather than re-declare, the installed versions still satisfy the pins and `npm ls` stays clean. Keep this file free of version numbers: every entry below is a path or a glob, so the pins remain the single place a version is written down.",
```
itechmeat: `npm ls stays clean` holds only after `npm ci` at the repo root and in `template-shell` (plus a shell build, since its exports all point at `dist-lib`). On a fresh clone, a lone `npm install` here gives ELSPROBLEMS with thirteen missing edges, because npm does not install the dependencies of a `file:`-linked package. Could the comment and README spell out that install order, so the first person to try this does not read the clean-`npm ls` claim as unconditional?

packages/cli/src/__tests__/template-split.e2e.test.ts
```diff
@@ -334,12 +334,36 @@ describe('Fixture 6 — declaration coverage: every real file in template-shell/
...
+  // Claiming it would collide with template-shell's exclusive subtree and turn
+  // a development-only file into an ownership assertion — reviving exactly the
+  // inferred-ownership model ADR-0031 rejected.
+  it('template-mfe/: the harness manifest stays undeclared — no claim on package.json or package-lock.json', () => {
+    const { exclusiveSubtrees, sharedFiles } = readManifest(TEMPLATE_MFE_DIR).ownershipBoundaries;
+    const sharedFilePaths = sharedFiles.map((entry: { path: string }) => entry.path);
+    expect(exclusiveSubtrees).not.toContain('package.json');
+    expect(exclusiveSubtrees).not.toContain('package-lock.json');
+    expect(sharedFilePaths).not.toContain('package.json');
+    expect(sharedFilePaths).not.toContain('package-lock.json');
```
itechmeat, line 379: `not.toContain('package.json')` is an exact-string check, but the assembler decides membership by prefix (`assembler.ts:16`), so a manifest declaring `["package"]`, `[""]` or `["package-"]` claims the root files while this test stays green - three of the four ways to break the invariant the comment above describes. `isPathWithinExclusiveSubtrees` is already imported in this file at line 27 and used four times; using it here would close the gap in one call.

template-mfe/README.md
```diff
@@ -7,11 +7,24 @@ scaffold to copy for a new MFE), and two widget fixtures (`widgets-fixture-a`,
...
-tooling, and no `src-app/app/` host — those are owned by
+This template is **add-only**. It claims ownership of no root `package.json`, no
+build/test/lint tooling, and no `src-app/app/` host — those are owned by
 [`frontx-template-shell`](../template-shell/README.md). Seeding this template into
-an empty directory produces a non-functional repository (dangling `file:`
-dependencies into a shell that isn't there).
+an empty directory produces a repository with no host to mount its MFEs into.
+
+Each MFE pins its `@gears-frontx/*` dependencies to exact published versions, so a
+seeded project installs them from the registry like any other dependency.
```
itechmeat, line 16: This sentence is now contradicted by `src-app/mfe_packages/README.md:15-18`, which still says a package here "resolves its `file:` dependencies ... relative to the shell's `packages/` and root" - after this PR no MFE manifest has a `file:` specifier left. That file ships into seeded projects (it sits inside the declared `src-app/mfe_packages/` subtree), so the generated project would document a mechanism it no longer uses. You already made this exact correction one level up at lines 10 and 13; worth sweeping the shipped copy in the same PR?

itechmeat (changes requested):

template-mfe/src-app/mfe_packages/demo-mfe/package.json
gs-layer: Fixed in `2e38ee02`. All four now declare `"@gears-frontx/frontx-template-shell": "0.1.0-alpha.0"` in `devDependencies`, matching what `_blank-mfe` already carried.

You are right that the PR body overstated it — the diff only rewrote `_blank-mfe`'s specifier, and the other three were still resolving through hoisting from exactly the package the add-mfe skill tells you to copy and delete.

packages/cli/src/__tests__/template-split.e2e.test.ts
gs-layer: Fixed in `2e38ee02`, extended in `b67d3bda`.

The assertions now go through `isPathWithinExclusiveSubtrees` — as you note, it was already imported and used four times in the file.

The first commit added a negative control for `["package"]` only; `b67d3bda` covers all three shapes you named. `[""]` is the worst of them: `"".startsWith` holds for every path, so such a manifest claims the whole repository.

`["package-"]` turned out asymmetric — it captures `package-lock.json` but **not** `package.json`, which does not start with that prefix. My first attempt asserted both were captured and failed on that row, which was a useful catch: the control now states the expectation per file, so the asymmetry is recorded rather than papered over.

template-mfe/README.md
gs-layer: Swept in `2e38ee02`. `src-app/mfe_packages/README.md` no longer describes `file:` resolution — it states registry installation and what an applied shell actually supplies: the runtime host, and `frontxMfGts` at build/dev time.

The distinction you drew is the reason it mattered: that file sits inside the declared `src-app/mfe_packages/` subtree and travels into seeded projects, unlike `template-mfe/README.md` one level up, so it was the generated project documenting a mechanism it no longer uses.

CodeRabbit then caught a flaw in the replacement text — it grouped `react` under `@gears-frontx/*` while these packages depend on both `@gears-frontx/react` and ordinary `react`. Corrected in `b67d3bda`.

template-mfe/package.json
gs-layer: Fixed in `2e38ee02`, in both places.

The `"//"` comment and `template-mfe/README.md` now state the actual order: install and build `template-shell` first, then install the harness. Reproduced what you describe — a lone install here cannot work, because npm does not install a `file:`-linked package's own dependencies, and every shell export points into `dist-lib`, which is gitignored and only exists after a build.

The unconditional "npm ls stays clean" phrasing was mine and it was wrong; it now carries its precondition.

template-mfe/package.json
gs-layer: Confirmed, and the claim is narrowed in `2e38ee02` rather than the mechanism changed — of the two options you offered, the second.

Checked it: `template-shell/package.json` pins `api`/`mfes`/`gts-plugin` to published versions with no local override of its own, so edits to those packages are not visible through the prebuilt `mf-gts` on the path every MFE build actually takes. They are visible only through an MFE's own direct imports of those packages, and the harness comment now says exactly that instead of claiming local edits are visible generally.

Making `dev:template:link` a prerequisite of the harness would reach across into `template-shell`'s tree, which this PR deliberately does not touch while the pin alignment is in flight. Worth revisiting once #498 lands and the shell's own pins stop being the constraint.

template-mfe/src-app/mfe_packages/_blank-mfe/package.json
gs-layer: Not addressed here, deliberately — this needs #498, which bumps the shell to `alpha.1`.

Agreed on the hazard: in an assembled workspace npm resolves two exact conflicting pins by hoisting one and nesting the other without complaint, and for the GTS registry and Module Federation singletons that is a runtime problem, not untidiness. The `link-template-ecosystem.mjs:64-70` note you point at makes it sharper still — the `0.3.0-alpha.0` tarballs cannot build the shell at all.

Moving the shell pins here would mean editing `template-shell/**` from a PR scoped to `template-mfe/**`, and #498 already carries that change plus its lockfile. So the ordering constraint stands: **#498 before #501**, and I would rather have that recorded as a merge dependency than duplicate its bump.

template-mfe/src-app/mfe_packages/README.md
```diff
@@ -12,10 +12,15 @@ package here must satisfy).
...
 This directory is **add-only**. It ships with `frontx-template-mfe`, which has
 no root `package.json`, no build/test/lint tooling, and no host application —
-those are owned by `frontx-template-shell`. A package here resolves its
-`file:` dependencies (`@gears-frontx/react`, `@gears-frontx/framework`, the
-shell's own `build/mf-gts`, …) relative to the shell's `packages/` and root —
-without an applied shell those paths dangle.
+those are owned by `frontx-template-shell`. A package here declares exact
+published versions of its `@gears-frontx/*` dependencies (`react`,
+`framework`, `frontx-template-shell` itself, …) and installs them from the
+registry like any other dependency — there are no `file:` paths to dangle.
```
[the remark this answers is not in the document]
gs-layer: Fixed in `b67d3bda`. Correct catch — these packages depend on both `@gears-frontx/react` and ordinary `react@19.2.4`, so listing a bare `react` among the scoped examples was genuinely ambiguous.

The scoped names are now written in full and third-party dependencies mentioned separately.

gs-layer:
One item from the review summary has no inline thread, so answering it here: **the harness lockfile is not broken, and I did not change it beyond the three new devDependency entries.**

The report reads those `resolved` strings as filesystem paths. They are not. When a `file:` override targets a directory *outside* the project root — `../packages/api`, `../template-shell` — arborist cannot record a matching node in the lockfile's `packages` map, because nothing outside the tree is ever a package node there. It writes an in-tree-shaped placeholder instead. Rewriting those strings to the real relative paths makes `npm ci` fail outright with `EMISSINGTARGET`; that was tested and reverted.

The predicted symptoms do not reproduce. From a fully clean tree — recursive `node_modules` removal, since each workspace member gets its own:

- `npm ci` → exit 0
- `npm ls` → **exit 0**
- zero dangling symlinks under any `@gears-frontx` scope directory
- `require.resolve('@gears-frontx/frontx-template-shell/build/mf-gts')` resolves from **all four** MFEs

`npm ls --all` does still report the `mfes` / `gts-plugin` split between the shell's `alpha.0` pins and the MFEs' `alpha.1` — that is the #498 dependency discussed in its own thread, not a lockfile defect.

The commit message records this so the next reader does not mistake the placeholder form for drift. If you have a repro where these entries actually break an install, I would rather see it than leave the disagreement standing — but on this tree the gate passes.

Everything else from the review is fixed in `2e38ee02` and `b67d3bda`, answered per thread.

template-mfe/src-app/mfe_packages/_blank-mfe/package.json
itechmeat: The "#498 before #501" ordering answers the skew but boomerangs on this diff: #498's `policy:template-pin-drift` verifies `@gears-frontx/*` pins against `packages/*` manifests only, so the `frontx-template-shell`/`react`/`framework`/`state`/`i18n`/`auth` pins here have no truth entry and fail the check; and #498 bumps the shell to `0.1.0-alpha.1`, after which the four `0.1.0-alpha.0` pins stop matching the override target and `npm ls` here goes red. After the rebase, do the pins bump here, or should the check learn `template-shell/packages/*` as a truth source?

template-mfe/README.md
```diff
@@ -7,11 +7,73 @@ scaffold to copy for a new MFE), and two widget fixtures (`widgets-fixture-a`,
...
+The `package.json` next to this README is **not part of the template**: it is a
+monorepo dev harness, deliberately absent from `frontx-template.json`'s
+`ownershipBoundaries`, so `frontx add` never copies it anywhere. Inside this
+monorepo the pins above would fetch registry tarballs instead of resolving to
+local source, so the harness redirects each pinned name to its local source
+via `overrides`. It redirects rather than re-declares, so the installed
+versions still satisfy the pins.
+
+This makes local edits to `template-shell/packages/*` (`react`, `auth`,
+`framework`, `i18n`, `state`) visible immediately, since the MFE packages
```
itechmeat, line 27: "visible immediately" overstates it the same way the mf-gts claim did before the fix: all eight override targets resolve through `dist/`, so an edit shows up only after that package is rebuilt - which line 43 of this same README says. The package.json comment is narrower ("visible for MFEs' own direct imports") but also lacks the rebuild caveat. Could both say "after a rebuild of the edited package"?

template-mfe/src-app/mfe_packages/README.md
```diff
@@ -12,10 +12,16 @@ package here must satisfy).
 
 This directory is **add-only**. It ships with `frontx-template-mfe`, which has
 no root `package.json`, no build/test/lint tooling, and no host application —
-those are owned by `frontx-template-shell`. A package here resolves its
-`file:` dependencies (`@gears-frontx/react`, `@gears-frontx/framework`, the
-shell's own `build/mf-gts`, …) relative to the shell's `packages/` and root —
-without an applied shell those paths dangle.
+those are owned by `frontx-template-shell`. A package here declares exact
+published versions of its FrontX dependencies (`@gears-frontx/react`,
+`@gears-frontx/framework`, `@gears-frontx/frontx-template-shell` itself, …)
```
itechmeat, line 17: "A package here declares exact published versions of its FrontX dependencies" - but demo-mfe, the worked example this README points authors at, declares only `react` and the shell while its `vite.config.ts` shares seven `@gears-frontx/*` names; six resolve by hoisting alone, and `mf-gts` throws `Cannot resolve version` if hoisting ever stops covering one. Declare the six in demo-mfe, or soften the sentence?

itechmeat (approved):
Approved with comments

itechmeat (approved):

GeraBart (approved):

Issue #472, opened 2026-07-23
[BUG] `MfeHandlerMF.resolveManifest` does not resolve string manifest references via the type system, breaking CTI-based entries


dingoatemytokens:
## Summary

`MfeHandlerMF` requires the `manifest` field of an `MfeEntryMF` to be either an inline `MfManifest` object or a string ID that is already present in its internal manifest cache. It never attempts to resolve a string reference through the pluggable type system (`TypeSystemPlugin.getSchema`). This breaks the CTI contract, where `manifest` is defined as a `cti.reference` — a string pointing to a separately registered `MfManifest` instance.

## Steps to Reproduce

1. Register a manifest instance in the type system:

```typescript
registry.typeSystem.register({
  id: 'cti.a.ui.mfe.mf_manifest.v1.0~vendor.app.manifest.v1.0',
  name: 'myApp',
  metaData: {
    publicPath: 'https://cdn.example.com/myapp/',
    remoteEntry: { name: 'remoteEntry.js', path: '', type: 'module' },
    // ...
  },
  shared: [],
})
```

2. Register an entry with `manifest` as a string reference (correct per CTI contract):

```typescript
registry.typeSystem.register({
  id: 'cti.a.ui.mfe.entry.v1.0~a.ui.mfe.entry_mf.v1.0~vendor.app.entry.v1.0',
  manifest: 'cti.a.ui.mfe.mf_manifest.v1.0~vendor.app.manifest.v1.0',
  exposedModule: './lifecycle-app',
  // ...
})
```

3. Register an extension and trigger mounting. `MfeHandlerMF` fails:

```
MfeLoadError: Manifest 'cti.a.ui.mfe.mf_manifest.v1.0~vendor.app.manifest.v1.0' not found.
Provide manifest inline in MfeEntryMF or ensure another entry from the same remote was loaded first.
```

## Root Cause

Inside `MfeHandlerMF`, manifest resolution is handled by `resolveManifest`:

```typescript
private async resolveManifest(manifestRef: string | MfManifest): Promise<MfManifest> {
  if (typeof manifestRef === 'object' && manifestRef !== null) {
    this.manifestCache.cacheManifest(manifestRef)
    return manifestRef
  }
  if (typeof manifestRef === 'string') {
    const cached = this.manifestCache.getManifest(manifestRef)
    if (cached) return cached
    throw new MfeLoadError(
      `Manifest '${manifestRef}' not found. Provide manifest inline in MfeEntryMF ` +
[...]

PR #509, opened 2026-08-03
feat(mfes): resolve lifecycle stage ids via TypeSystemPlugin instead of hardcoded literals

### LINKED ISSUE #505: [ARCH] MFE runtime hardcodes its own GTS lifecycle-stage identifiers, defeating type-format agnosticism
## Description

The MFE runtime (`packages/mfes`) is declared type-format-agnostic: it must obtain type identifiers through the injected `TypeSystemPlugin` (`isTypeOf`, resolver methods), with `@gears-frontx/gts-plugin` as the GTS implementation owning the GTS nomenclature. In practice the runtime spells its own lifecycle-stage identifiers as GTS string literals in six places across three files (verified on develop @ `93c9474`):

| Location | Literal |
|---|---|
| `packages/mfes/src/runtime/default-extension-manager.ts:109` | `gts.frontx.mfes.lifecycle.stage.v1~frontx.mfes.lifecycle.destroyed.v1` |
| `packages/mfes/src/runtime/default-extension-manager.ts:193` | `gts.frontx.mfes.lifecycle.stage.v1~frontx.mfes.lifecycle.init.v1` |
| `packages/mfes/src/runtime/default-extension-manager.ts:214` | `gts.frontx.mfes.lifecycle.stage.v1~frontx.mfes.lifecycle.destroyed.v1` |
| `packages/mfes/src/runtime/default-mount-manager.ts:190` | `gts.frontx.mfes.lifecycle.stage.v1~frontx.mfes.lifecycle.activated.v1` |
| `packages/mfes/src/runtime/default-mount-manager.ts:216` | `gts.frontx.mfes.lifecycle.stage.v1~frontx.mfes.lifecycle.deactivated.v1` |
| `packages/mfes/src/runtime/DefaultMfeRegistry.ts:362` | `gts.frontx.mfes.lifecycle.stage.v1~frontx.mfes.lifecycle.init.v1` |

A consumer that plugs in a non-GTS type system and declares lifecycle stages with its own identifiers never gets matched: the runtime keeps dialing the GTS strings past the consumer's declarations.

```mermaid
flowchart LR
  c["consumer registers stages<br/>via non-GTS TypeSystemPlugin"] --> r["runtime"]
  r -->|"compares against hardcoded<br/>gts.frontx.mfes...init.v1"| m["no match:<br/>stages never fire"]
  style m stroke-dasharray: 5 5
```

This is an umbrella over two previously filed special cases of the same systemic defect:

- #471 (closed by PR #478): hardcoded GTS *action* ids - fixed by moving them into `gts-plugin`, which is the pattern to follow here.
- #472 (open): `MfeHandlerMF.resolveManifest` does not resolve string manifest refs via the type system.

Fixing #471 and #472 alone does not make a non-GTS consumer work: the six stage literals above remain. Conversely, the correct pattern already exists in the very same file - `DefaultMfeRegistry.ts` (~441-444) resolves the mount/unmount action ids via `this.typeSystem.resolveMountExtActionId()` / `resolveUnmountExtActionId()`, with a comment stating the runtime never spells a concrete type-format literal for these concepts. The stage ids simply never got the same treatment.

Why lint does not catch it: rule MFES-1 in the root `eslint.config.js` bans only solution-namespace literals (`screensets|framework|state|i18n|react`); the runtime's own `gts.frontx.mfes.*` namespace passes legally, so the constraint as enforced is weaker than the "no type-format literals in the MFE Runtime" contract documented in `type-substrate/index.ts`.

## Work

One PR covers the whole umbrella:

1. Add stage-id resolver methods to the `TypeSystemPlugin` contract following the #478 / `resolveMountExtActionId` pattern and replace all six literals; thread the plugin into the mount manager. Plugin fakes already exist in tests.
2. Resolve #472 in the same change: `MfeHandlerMF` resolves string manifest refs through the type system, and the registry supplies its type system to registered handlers - so the path is live for handlers constructed with no arguments, with no template changes involved. `handler/types.ts` contract wording updated in place: matching stays with the registry, the handler receives the type system for resolving its own manifest references.
3. Extend MFES-1 (or add a guard) so `gts.frontx.mfes.*` literals cannot silently reappear in `packages/mfes` runtime code. JSDoc mentions of the GTS format are fine and should stay allowed.
4. An end-to-end test on a fake non-GTS plugin proving both the stage path and the manifest path - that test is what proves the umbrella closed. Partial delivery is the trap: with only part of the sites converted, tests stay green while a non-GTS consumer still does not work.
5. Three adjacent defects surfaced by this work are folded in rather than deferred: (a) `DefaultMfeRegistry.unregisterDomain` detaches mediator handlers before the `destroyed` stage fires, so a teardown hook that dispatches an action always fails - reorder so destroyed hooks can dispatch, and upgrade the destroyed e2e case to a real hook; (b) `type-check:test` is declared in both `packages/mfes` and `packages/gts-plugin` but neither ships a `tsconfig.test.json` and neither is in the root aggregation - wire both so the tests this PR adds are gated, not hand-checked; (c) `DefaultMfeRegistryFactory.ts` carries stale `inst-1` markers matching no CDSL step - align them with the factory-build flow so the validator survives the first honest marking of that flow.

## Estimate

About two days including the folded defects: the resolver pattern and the test fakes exist; the handler wiring is the only new contract surface.

## Risks / open questions

- #472 was filed by another team and is unassigned - notify its authors that the fix lands here, to avoid doing the same work twice.
- Alternative resolution to record if the team prefers it: declare that the runtime owns its infrastructure nomenclature in GTS format, and non-GTS consumers must register mappings. Cheaper, but it must then be documented explicitly - today nothing states it.

Found during the templates gap review (finding Д10, 27 July); re-verified against develop @ `93c9474` on 3 August - inventory unchanged, no new literals, no open PR touching `packages/mfes/src/runtime/`.

itechmeat:
## Summary

Replaces the six hardcoded `gts.frontx.mfes.lifecycle.stage.v1~...` literals in the MFE runtime with stage-id resolvers on the injected `TypeSystemPlugin`, so the runtime never spells a concrete type-format literal for a lifecycle stage. Includes the #472 manifest-resolution fix as part of the same umbrella.

Closes #505
Refs #472

## Changes

**`@gears-frontx/gts-plugin`**
- `constants.ts`: add `FRONTX_LIFECYCLE_STAGE_INIT/ACTIVATED/DEACTIVATED/DESTROYED`
- `plugin.ts`: implement `resolveLifecycleStage{Init,Activated,Deactivated,Destroyed}Id()`

**`@gears-frontx/mfes`**
- `type-substrate/index.ts`: extend the `TypeSystemPlugin` contract with the four resolvers
- `default-extension-manager.ts`: init + destroyed stage ids resolved via plugin
- `DefaultMfeRegistry.ts`: init stage id resolved via plugin; `typeSystem` threaded into `DefaultMountManager` config
- `default-mount-manager.ts`: activated/deactivated stage ids resolved via plugin
- `MfeHandlerMF.ts` (#472): string manifest references fall back to `typeSystem.getSchema(ref)` after a cache miss (mirrors `DefaultExtensionManager.resolveEntry`), guarded by a structural `isMfManifest` check
- `handler/types.ts`: `MfeHandler` gains an optional `typeSystem` (backwards compatible)

**Lint**
- `eslint.config.js`: MFES-1 rule extended to the `mfes` namespace; test files excluded via `ignores` since they legitimately use the literals as format fixtures

**SDLC artifacts**
- `type-substrate-port/FEATURE.md`: instructions 8-11 (`inst-resolve-lifecycle-stage-*`)
- `gts-type-provider/FEATURE.md`: instructions 6-9 (`inst-tr-07..10`)

**Tests**
- `non-gts-lifecycle-stages.test.ts`: 4 scenarios covering all four lifecycle stages with a non-GTS `TypeSystemPlugin`
- `MfeHandlerMF.manifest-type-system.test.ts`: 3 scenarios for the #472 fallback (inline manifest, cache hit, getSchema fallback + fetch rejection)
- fakes updated: `DefaultMfeRegistry.test.ts`, `mount-strategies.test.ts`, `actions-chains-mediator.test.ts`

## Verification

- `cfs validate`: 53 artifacts, 0 errors/warnings, 118/118
- `npm run type-check:all`: clean
- `npx eslint . --max-warnings 0`: clean
- `npm run arch:check`: 12/12 passed
- `npm run arch:deps`: 0 errors (pre-existing warnings only)
- `npm run test:unit`: 796/796 across all workspaces

## Notes

- `template-shell/packages/react/src/mfe/components/__tests__/ExtensionDomainSlot.test.tsx:39` keeps an inline `TypeSystemPlugin` fake and is intentionally untouched: the file's owner is the templates team and it is out of scope for this runtime change (per git.md rule 15, flagging rather than silently modifying).

## Summary by CodeRabbit

* **New Features**
  * Added type-system support for resolving initialization, activation, deactivation, and destruction lifecycle stages.
  * Added support for resolving string-based manifest references through the configured type system.
  * Preserved existing behavior when no type system is configured.
* **Bug Fixes**
  * Improved validation and error handling for missing or invalid manifest references.
* **Tests**
  * Added coverage for custom lifecycle identifiers and manifest resolution across supported runtime scenarios.


itechmeat:
Closed on request: PR was created without explicit approval.

PR #510, opened 2026-08-03
refactor(telemetry): file the generic bus, timer and logger under utils

hello1101n:
Three files in `managers/` were not managers. Filed under `utils/`.

| moved | why |
|---|---|
| `scheduler.ts` | debounced timer, no telemetry knowledge |
| `logger.ts` | holds no state |
| `hooks.ts` | generic event bus — merged into `utils/hooks.ts`, where its own types already lived |

`managers/` now holds only what the name says: `events`, `session`, `plugins`, `userInfo`.

Relocation only, no logic touched. `scheduler.ts` moves with a zero-line diff; `logger.ts` and `events.ts` change one import each. Single non-import edit is `eslint.config.js:292` — the `ban-ts-comment` exemption following the bus to its new path, otherwise lint starts failing on the two `@ts-expect-error` lines.

## userInfo stays a manager
It resembles a plugin — `addHook('event')` to stamp a user id — but it holds state, and `identify()` is callable before `start()`. Plugin `setup()` runs *inside* `start()`, so converting it would silently drop early calls.

## Verify
lint, type-check (src/test/demo), unit 43/43, build, arch:check 12/12, arch:deps — all green.

Demo driven in Chrome, exercising each moved module: hook bus fires, the 5s flush debounce fires unprompted, `context_user_id` still stamped from a pre-`start()` `identify()`.

## Summary by CodeRabbit

- **Refactor**
  - Reorganized telemetry hooks, logging, scheduling, and type utilities without changing existing runtime behavior.
  - Improved telemetry scheduling controls, including immediate execution, delayed execution, and cancellation.
  - Telemetry hook processing now continues across individual failures while preserving error reporting.
  - Updated internal references and linting configuration to reflect the current telemetry structure.


packages/telemetry/src/utils/hooks.ts
```diff
@@ -26,3 +26,67 @@ export type HooksManager = {
...
+        // @ts-expect-error args are correct
+        item(...args);
+      } catch (e: unknown) {
+        error ??= e;
+      }
+    }
+
+    if (error) {
+      throw error instanceof Error ? error : new Error(String(error));
+    }
```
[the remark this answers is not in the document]
hello1101n: Real finding, but pre-existing — and the proposed diff regresses a case that works today.

**Pre-existing.** The `createHooks` body is byte-identical to `managers/hooks.ts` on `develop`; this PR only relocated the file. `diff` of the two function bodies is empty. The review flagged it as new because the path changed — the same run listed `managers/hooks.ts` under "files with no reviewable changes."

**The bug is real.** All five falsy throws are swallowed by `if (error)`:

```
hook throws undefined  -> rethrown: false
hook throws null       -> rethrown: false
hook throws false      -> rethrown: false
hook throws 0          -> rethrown: false
hook throws ''         -> rethrown: false
```

**But `hasError` alone makes things worse.** Making the first throw win unconditionally loses a real error hiding behind a nullish one. Today `error ??= e` falls through on nullish, so the real `Error` still surfaces:

```
current:       throw null, then throw new Error('REAL')  ->  "REAL"
proposed fix:  throw null, then throw new Error('REAL')  ->  "null"
```

A correct fix needs the flag *and* a preference for a real `Error` over a nullish captured value, plus a test covering the mixed case. That is a behavior change with its own edge cases, not a quick win.

Tracking it as a follow-up rather than folding it in here: this PR's reviewability rests on being a pure relocation with no logic touched. Same call as #496, which scoped around pre-existing defects surfaced by a move instead of fixing them in the migration PR.

packages/telemetry/src/utils/logger.ts
```diff
@@ -1,4 +1,4 @@
-import type { TelemetryConfigNormalized } from '../utils/types';
+import type { TelemetryConfigNormalized } from './types';
```
itechmeat, line 1: Agreed these aren't managers - but both relocated modules export `*Service` types, and the package already has a `services/` directory. Why did `utils/` win over `services/`? If the answer is that `services/` is the public SDK facade and these are internal plumbing, that works - though it might be worth renaming the `*Service` types in a follow-up so the names stop suggesting otherwise.

packages/telemetry/src/utils/types.ts
```diff
@@ -1,7 +1,6 @@
-import type { TelemetryHooksService } from '../managers/hooks';
-import type { TelemetryLoggerService } from '../managers/logger';
 import type { TelemetryLogEventParams, TelemetryRecord } from './eventTypes';
-import type { HooksManager } from './hooks';
+import type { HooksManager, TelemetryHooksService } from './hooks';
+import type { TelemetryLoggerService } from './logger';
```
itechmeat, line 3: Now that they live in one file, it's visible that `TelemetryHooksService` is just an alias of `HooksManager` (`createHooks(): HooksManager`). Pre-existing and out of scope for a pure move - but a follow-up collapsing them to one name would be a nice cleanup, possibly together with the `*Service` type renames.

eslint.config.js
```diff
@@ -321,7 +289,7 @@ export default [
   {
     files: [
       'packages/telemetry/src/managers/events.ts',
-      'packages/telemetry/src/managers/hooks.ts',
+      'packages/telemetry/src/utils/hooks.ts',
```
itechmeat, line 292: The comment above this block still says "the hooks manager". Since the whole point of this PR is that it isn't a manager, probably worth updating the wording here too - two words.

itechmeat (approved):

GeraBart (approved):

packages/telemetry/src/utils/logger.ts
hello1101n: thank you, I will review and prepare the next PR if I decide to change.

packages/telemetry/src/utils/types.ts
hello1101n: https://github.com/constructorfabric/gears-frontx/pull/510#discussion_r3702396331

PR #512, opened 2026-08-03
chore: add CODEOWNERS

hello1101n:
The repo has no CODEOWNERS, so nothing auto-requests a reviewer — PRs sit unassigned until someone is asked by hand.

`@GeraBart` is the default owner. Three paths with a clear specialist name them instead:

| path | owner |
|---|---|
| `/packages/ui-kit/` | @SysoevAndrey |
| `/packages/telemetry/` | @hello1101n |
| `/template-shell/`, `/template-mfe/` | @gs-layer @itechmeat @tscbmstubp |

Derived from commit history on `develop` and approvals on merged PRs. Deliberately minimal — a starting point to extend as ownership settles, not a full map of the repo.

Note that only the **last** matching pattern applies; ownership does not accumulate, so the paths above are owned by those names alone and not additionally by `@GeraBart`.

## Blocker: only @GeraBart has write access

GitHub's CODEOWNERS validation rejects every other owner in this file:

```
flagged: @SysoevAndrey  @hello1101n  @gs-layer  @itechmeat  @tscbmstubp
clean:   @GeraBart
```

Code owners need **explicit write access** — a stricter bar than being requestable as a reviewer, which only needs read. All five accounts exist, so it is the write-access half failing.

The practical effect is that every path-specific rule is inert and the file behaves as `* @GeraBart` until access is granted. It is harmless as-is — CODEOWNERS only auto-requests reviewers and gates nothing unless *Require review from Code Owners* is enabled in branch protection — and each line starts working the moment access lands, with no follow-up PR needed.

Happy to strip the inert lines instead if this PR should only contain what works today.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,6 +1,3 @@
-# Only the last matching pattern applies — ownership does not accumulate, so a
-# path listed below is owned by those names alone, not additionally by @GeraBart.
-
 *                        @GeraBart
 
 /packages/ui-kit/        @SysoevAndrey
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -6,5 +6,5 @@
 /packages/ui-kit/        @SysoevAndrey
 /packages/telemetry/     @hello1101n
 
-/template-shell/         @gs-layer @itechmeat
-/template-mfe/           @gs-layer @itechmeat
+/template-shell/         @gs-layer @itechmeat @tscbmstubp
+/template-mfe/           @gs-layer @itechmeat @tscbmstubp
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,10 @@
+# Only the last matching pattern applies — ownership does not accumulate, so a
+# path listed below is owned by those names alone, not additionally by @GeraBart.
+
+*                        @GeraBart
+
+/packages/ui-kit/        @SysoevAndrey
+/packages/telemetry/     @hello1101n
+
+/template-shell/         @gs-layer @itechmeat
+/template-mfe/           @gs-layer @itechmeat


itechmeat (approved):

.github/CODEOWNERS
```diff
@@ -0,0 +1,10 @@
+# Only the last matching pattern applies — ownership does not accumulate, so a
+# path listed below is owned by those names alone, not additionally by @GeraBart.
+
+*                        @GeraBart
+
+/packages/ui-kit/        @SysoevAndrey
+/packages/telemetry/     @hello1101n
+
+/template-shell/         @gs-layer @itechmeat
+/template-mfe/           @gs-layer @itechmeat
```
itechmeat: Let's add @tscbmstubp to `template-shell` and `template-mfe`

hello1101n: added

hello1101n: Added to both template paths in 42fa247b.

Worth noting the history backs this up, even though it does not look that way at first: `git log` on `template-shell/` and `template-mfe/` shows 0 commits from @tscbmstubp, because both directories are days old. Their work is in the content that moved into them — 217 file touches in `src/`, 76 in `packages/framework`, 66 in `packages/react`. I had dropped them from an earlier draft on exactly that bad signal, so thanks for the catch.

Separately, and more importantly — GitHub's CODEOWNERS validation currently rejects every owner in this file except @GeraBart:

```
flagged: @SysoevAndrey  @hello1101n  @gs-layer  @itechmeat  @tscbmstubp
clean:   @GeraBart
```

The check covers existence *and* write access, and all five of those accounts exist, so it is the write-access half failing. That is consistent with @GeraBart being the only person who merges PRs here.

The practical effect is that every path-specific rule is inert — the file behaves as `* @GeraBart` until write access is granted. It is harmless as-is (CODEOWNERS only auto-requests reviewers; it gates nothing unless *Require review from Code Owners* is enabled), and each line starts working the moment access lands, with no follow-up PR needed.

Flagging it because you would be better placed than me to sort out repo permissions. Happy to strip the inert lines instead if you would rather this PR only contain what works today.

GeraBart (approved):

PR #513, opened 2026-08-03
feat(mfes): type-format agnosticism restored - lifecycle stages and manifest refs resolve through the TypeSystemPlugin port (#505, #472)

### LINKED ISSUE #472: [BUG] `MfeHandlerMF.resolveManifest` does not resolve string manifest references via the type system, breaking CTI-based entries

### LINKED ISSUE #505: [ARCH] MFE runtime hardcodes its own GTS lifecycle-stage identifiers, defeating type-format agnosticism

itechmeat:
Closes #505, and resolves #472 within it - the umbrella lands whole in one PR.

Base `develop` @ `65ebb01d` (rebased 4 Aug: 24 develop commits absorbed, no conflicts, all branch commits replayed content-identical per range-diff). 15 commits, 35 files, +1486 / -109. Each commit type-checks and lints on its own.

## What changes

The MFE runtime is declared type-format-agnostic, yet it kept two kinds of private knowledge about the GTS format: it spelled its own lifecycle-stage identifiers as GTS string literals in six places across three files, and its Module Federation handler could not resolve a manifest named by id through the type system at all (#472). A consumer that plugs in a non-GTS type system and declares stages or manifests under its own notation never gets matched.

| Before | After |
|---|---|
| `default-extension-manager.ts` (3 literals), `default-mount-manager.ts` (2), `DefaultMfeRegistry.ts` (1) each spell the full GTS stage id inline | All six sites call the port: `typeSystem.resolveLifecycleStage{Init,Activated,Deactivated,Destroyed}Id()` |
| The GTS stage ids live in the runtime, next to a comment claiming the runtime never spells type-format literals | The ids live in `packages/gts-plugin/src/constants.ts` - the package that owns GTS nomenclature, same home PR #478 gave the action ids |
| `MfeHandlerMF.resolveManifest` throws `MfeLoadError` for any string manifest ref not already in its cache - the type system is never consulted (#472) | The handler asks the type system it received at registration; only when neither the cache nor the plugin holds the reference does it refuse, with a message naming both places it looked |
| Rule MFES-1 bans solution-namespace literals only, and only in single-quoted form; `gts.frontx.mfes.*` and any backtick string pass lint legally | The selectors cover the `mfes` namespace and `TemplateElement` nodes (raw and cooked, so escape sequences cannot spell the forbidden text); doc comments stay legal |
| No test exercises a non-GTS notation end to end | Two e2e suites on a fake plugin whose ids share no substring with GTS: every stage transition fires through the real registry pipeline, and a manifest named by id resolves through the plugin on a real `MfeHandlerMF` driven via `mfeHandlers` → `registerDomain` → `registerExtension` → mount |

## How the type system reaches the handler

The one design question in the PR. Handlers are constructed by consumers with zero arguments (`new MfeHandlerMF(FRONTX_MFE_ENTRY_MF)`) and arrive at the registry ready-made, so neither a constructor parameter (nobody would pass it) nor a template-side change (couples to publish sequencing) works. Instead the registry - which already owns the plugin - hands it to every handler it registers: the abstract `MfeHandler` base implements `attachTypeSystem()`, `DefaultMfeRegistry` calls it in its `mfeHandlers` registration loop. One injection path, no implementor needs a single line changed, and the contract wording in `handler/types.ts` now states the split precisely: hierarchy matching stays with the registry; the handler receives the same plugin for a different job - resolving the references its own load path owns.

The binding is exclusive: re-attaching the same plugin is a no-op, attaching a different one throws - one handler instance cannot be shared across registries (see the review round below for why).

```mermaid
flowchart LR
  subgraph before["Before"]
    r1["runtime"] -->|"inline gts.frontx.mfes...init.v1;<br/>manifest-by-id: cache or throw"| m1["non-GTS consumer:<br/>stages and manifests never match"]
  end
  subgraph after["After"]
    reg["DefaultMfeRegistry"] -->|"attachTypeSystem()<br/>at registration"| h["MfeHandlerMF"]
    r2["runtime"] -->|"resolveLifecycleStage*Id()"| p["TypeSystemPlugin"]
    h -->|"manifest ref lookup"| p
    p -.->|"GTS plugin or any other:<br/>same contract"| n["non-GTS consumer works"]
  end
```

The port grows four explicit stage methods rather than one string-keyed lookup, mirroring the existing `resolveMountExtActionId` / `resolveUnmountExtActionId` pair in the same interface: an explicit method per well-known concept is compile-time-enforced on every implementation, where a keyed lookup would silently resolve to nothing. No behaviour drift under the GTS plugin: the four constants are byte-identical to the literals they replace, protected by golden-literal tests in `packages/gts-plugin`'s first test file (all six resolvers covered).

## Three adjacent defects this work surfaced, fixed here rather than deferred

- **`unregisterDomain` fired the `destroyed` stage after the domain's mediator handlers were already detached**, so a teardown hook that dispatches an action always failed with "No handler found". The detach now runs after the stage completes (safety verified: the stage dispatch is awaited to completion and pending actions are untracked before it resolves, no serializer deadlock, no test or spec asserted the old order - which was in fact more exposed, detaching before extensions were torn down). The destroyed e2e case now uses the same real-hook shape as the other three stages.
- **`type-check:test` was declared in both `packages/mfes` and `packages/gts-plugin` with no `tsconfig.test.json` behind it and no place in the root aggregation** - their tests were type-checked nowhere. Both configs added (mfes's includes its out-of-`src` `__tests__/bridge/` tree, which `src/**` alone would have skipped) and both wired into `type-check:packages:test`, so the tests this PR adds are gated, not hand-verified.
- **`DefaultMfeRegistryFactory.ts` carried placeholder `inst-1` markers matching no CDSL step** - a mine that detonated the validator on the first honest marking of the factory-build flow. Markers realigned to the steps the code actually implements (`inst-flow-fb-01..03`, `inst-state-fc-01/02`), with each of the 9 checkbox flips verified against the code.
- **The concrete factory leaked into the public surface against ADR-0003.** Its own doc block said "not exported from the public barrel", the barrel exported it anyway, and the one barrel consumer (template-shell's framework package) was exactly the violation ADR-0003's confirmation check names ("no consumer outside the runtime package imports the concrete registry or factory classes"). The class export is removed; `createMfeRegistryFactory(): MfeRegistryFactory` (abstract return type) is the one creation path, the template-shell consumer is switched here, and three `@example` blocks that imported a never-exported `mfeRegistryFactory` singleton now name the real path. `declare class DefaultMfeRegistryFactory` is gone from the published d.ts.

## Breaking for plugin implementers

Two contract changes, both compile-time-loud:

- `TypeSystemPlugin` gains four required methods. Every implementation must supply them; the in-repo GTS plugin and all test fakes (including the template-shell one) are updated here.
- `DefaultMountManager`'s config gains a required `typeSystem`. `DefaultMfeRegistry` supplies it at the single in-repo construction site.

A third, surface-shrinking one: `DefaultMfeRegistryFactory` is no longer exported (see the folded defects above) - `@gears-frontx/mfes@0.3.0-alpha.1` did export it, so any out-of-tree consumer constructing it directly must switch to `createMfeRegistryFactory()`; the repo-wide inventory found exactly one such consumer, fixed here.

And a runtime-loud one from the review round: `MfeHandler.attachTypeSystem` rejects a second, different plugin while one is attached (same-plugin re-attach stays a no-op). Registering one handler instance into two registries was never a supported flow - it now fails at registration instead of silently re-binding.

`MfeHandler.attachTypeSystem` is implemented by the base class, so handler subclasses are NOT broken. `@gears-frontx/mfes` is pre-1.0 alpha; the breaks land with the next ecosystem publish and pin bump.

## Review round (CodeRabbit, 4 Aug)

All five findings from the first review are taken, in four commits on top of the reviewed head:

- **Factory config aliasing (Major).** `DefaultMfeRegistryFactory.build` cached the caller-owned `config` object, so mutating `config.typeSystem` after the first build made the mismatch check compare the mutated object against itself and silently return a registry bound to the original plugin. The factory now snapshots `config.typeSystem` at first build; a regression test mutates the built config and asserts the rebuild is rejected while the first registry stays bound to the original plugin.
- **Handler re-binding (Major).** `attachTypeSystem` silently overwrote the plugin, letting a second registry re-bind a handler the first registry could still drive, with the manifest cache keyed only by id. Now one handler - one type system, enforced as described above; regression test uses two plugins answering the same manifest id with different `publicPath` and proves resolution stays with the first plugin after the rejected second attach.
- The construction error and all doc comments name the public `createMfeRegistryFactory()` path (the old text recommended the removed export), including the `gts-plugin` usage example.
- The `mfe-registry` FEATURE factory-build flow now begins with the creation step via `createMfeRegistryFactory()` (`inst-flow-fb-create`, marker wiring verified with `cfs validate`).
- The non-GTS lifecycle suite additionally covers extension registration (`init` fires through the port) and unregistration ordering (`deactivated` strictly before `destroyed`), both checked against the real dispatch order in the runtime, not tuned to pass.

## Traceability

Stage resolution: instructions and `@cpt-` markers on both sides of the port (`type-substrate-port`, `gts-type-provider` FEATUREs), placed at the interface declarations - the same fragment the precedent action-resolver instructions mark. Manifest resolution: `mfe-isolation` FEATURE gains nested instructions (inline ref / by-id ref / unresolved-raise) with an Error Scenario and a new DoD; `mfe-registry` FEATURE gains the injection instruction on the handler-resolution algorithm and an extended DoD naming `attachTypeSystem`. `cfs validate`: 0 errors, 0 warnings, code coverage 121/121 - including the registry-factory flow and cache-state artifacts whose markers this PR realigned.

## Verification

At head, after the rebase and the review round: `npm run lint` (0 warnings), `npm run type-check:all`, `npm run test:unit` (77 files, 891 tests repo-wide; gts-plugin 0 → 7; template-shell suites green), `npm run arch:check` 12/12 - all green, independently re-run by a second pass rather than trusted from the first. mfes and gts-plugin test sources are now part of `type-check:all` (see the folded defects above) - no temporary configs involved. `grep -rn "gts\.frontx\.mfes" packages/mfes/src` outside `__tests__` returns doc comments only. Lint-guard hardening proven by stdin probes: single-quoted, backtick, interpolated and `\u`-escaped forms all flag; comment mentions stay clean. The template-shell fake compiles against both the pinned published `mfes` and the linked workspace source. Both review-round regression tests were additionally proven to fail against the reverted source fixes.

Test fixtures in `MfeHandlerMF.test.ts` and `cross-runtime-routing.test.ts` moved from GTS-shaped strings to mock notation - the code under test treats them as opaque keys, so nothing weakens and the widened lint guard needs no test exemption.

## Pre-existing gaps made visible, kept out of scope

- `template-shell/packages/react/tsconfig.test.json` sets no `jsx`, so that package's own `type-check:test` emits 148 config-noise errors on develop as much as here.
- ADR-0003 has gone stale in two places: it cites `packages/screensets/src/mfe/runtime/*` (deleted package) and its Confirmation still says the public surface exports an `mfeRegistryFactory` singleton, which was deliberately moved to the app layer. The ADR itself was not edited here - refreshing it is the architect's call. The same stale `mfeRegistryFactory` spelling also survives in `DESIGN.md`, `DECOMPOSITION.md` and `template-shell/.../types.ts` (which documents a nonexistent `factory.create`), all outside this PR's diff - follow-up material.
- `DefaultMfeRegistry` (the sibling concrete class) is still exported from the barrel with zero consumers outside `packages/mfes`, covered by the same ADR-0003 sentence enforced above - removable, but a second published-surface break, left for a deliberate decision.

## Summary by CodeRabbit

* **New Features**
  * Added lifecycle-stage identifier resolution for initialization, activation, deactivation, and destruction.
  * Added manifest lookup through the registered type system, including validation, caching, and diagnostic errors.
  * Added a supported factory creation API for MFE registries.
  * Improved lifecycle teardown sequencing and support for custom type identifiers.

* **Bug Fixes**
  * Prevented unresolved or invalid manifest references from proceeding to network requests.

* **Tests**
  * Expanded coverage for lifecycle resolution, manifest lookup, registry behavior, and factory usage.


itechmeat:
@coderabbitai review

GeraBart (approved):

Issue #495, opened 2026-07-30
[ARCH] Replace the three-pillar model with layered FrontX architecture and per-package CDSL ownership


SysoevAndrey:
## Topic

Replace the current three co-equal FrontX pillars with a layered architecture and define scoped CDSL ownership for every reusable package and template.

The direction was confirmed during the FrontX architecture sync on 2026-07-31: root FrontX architecture owns orchestration, while individual technologies and templates own their own architecture artifacts.

## Context

The current PRD/DESIGN model treats FrontX as three co-equal pillars:

| Current pillar | Responsibility | Examples |
|---|---|---|
| Core Framework | Runtime substrate | `mfes`, `gts-plugin`, `api` |
| CLI | Template lifecycle | `cli` |
| AI Tooling | Agent ecosystem | Cyber Pilot kit |

This no longer matches the intended product architecture. It also leaves no valid place for reusable runtime libraries consumed by templates—such as UI Kit and Telemetry—without contradicting the agnostic-core constraint.

Implementation work has already moved directly into `gears-frontx`:

- UI Kit: #499
- Telemetry: #496

These packages are still unpublished/interim. Temporary Constructor Studio ignores are acceptable during migration, but they are scaffolding—not the final ownership and traceability model.

## Architect direction (2026-07-31)

Replace the pillars with three layers:

| Layer | Responsibility |
|---|---|
| Technology | Reusable runtime technologies/libraries consumed by templates and applications |
| Product blueprints | Templates and other product-oriented starting points assembled from technologies |
| Orchestration | The FrontX lifecycle that composes, installs, upgrades, and coordinates technologies, templates, and AI tooling |

The root FrontX PRD/DESIGN should own the orchestration model, inter-layer contracts, and lifecycle. It should not absorb the internal architecture of every reusable library.

Each reusable library under `packages/` should own a scoped CDSL artifact set—at minimum PRD, DESIGN, and ADRs. The exact mandatory chain, including whether DECOMPOSITION and FEATURE artifacts are required for every library, must be standardized as part of this issue.

Templates should likewise own their product-specific specifications. This avoids growing one root DESIGN every time a reusable library is added.

## Current gaps

1. The PRD hardcodes co-equality and pillar-balance rules, but the new layers have different responsibilities and are not co-equal pillars.
2. DESIGN is declared frozen, so this change requires an explicit architecture revision/versioning decision.
3. #484 hardens the current pillar model; its guards and sequencing must be reconciled with this change.
[...]

PR #514, opened 2026-08-03
feat(ui-kit): forms batch on Base UI, AI docs layer, token polish

SysoevAndrey:
## What

Stage 2 of the UI Kit delivery plan (`packages/ui-kit/design-notes.md`): the forms component batch on `@base-ui/react`, the AI docs layer, and the theme-token polish. This is the work #499 announced in its "Next (separate PRs)" section. Related: #495 (the package remains under the interim `artifacts.toml` ignore, unchanged here).

One commit per component; each ships the component, its CSS Module, a colocated usage doc, an `llms.txt` entry, and render/interaction tests.

## Demo

All components exercised in the sandbox (packed tarball in a clean Vite app), light and dark:

https://github.com/user-attachments/assets/aba29629-1269-4347-b4d6-47112d6cc89b

## Contents

- **Button moves onto the Base UI primitive** (`@base-ui/react/button`, new dependency, pinned `1.6.0`): `render`-prop polymorphism per the kit's wrapping convention, `type="button"` default now comes from the primitive. Visuals untouched — same CSS Module and CVA map. Browser verification caught that non-`<button>` render targets need `nativeButton={false}`; the doc and tests pin that contract.
- **8 new form components** — `Input`, `Textarea`, `Label`, `Checkbox`, `Switch` (sizes `default|sm`), `RadioGroup`, `Select` (9 parts, portalled popup), `Field` (Base UI Field composition: automatic `htmlFor`/`aria-describedby` wiring and validation display; deliberately not shadcn's div-based `field.tsx`, which has no such wiring). Styles are base-vega translations to CSS Modules following the Button precedent: theme tokens only, no raw colors, shadows and `dark:`-only refinements dropped (the Select popup keeps a `color-mix`-derived border/shadow so it separates from the page). Check/chevron icons are inline lucide paths — the kit carries no icon dependency.
- **AI docs layer**: `llms.txt` at the package root (setup rules + component index) and a per-component usage doc (when to use, kit-level props, examples, anti-patterns) copied to `dist/docs/` at build. A docs test enforces that every component has a doc, is indexed, and documents every variant/size its CSS module defines; the consumer acceptance script asserts the docs ship inside the tarball.
- **Tokens polish**: `theme.css` exposes the shadcn v4 radius scale (`--radius-sm|md|lg|xl` derived from `--radius`); component modules consume the scale, so overriding `--radius` alone rebrands every corner. A tokens test guards the seam: modules may consume only theme-defined variables (Base UI runtime vars excepted) and no raw colors.
- **Select contracts documented and tested**: pass `items` on the root so the closed trigger renders labels; mouse selection commits on pointerdown-then-click (Base UI's `alignItemWithTrigger` guard); `SelectContent` forwards Base UI's `container` prop so subtree-scoped themes (`data-theme` on a section rather than `<html>`) reach the portalled popup.
- **Test infra**: `PointerEvent` polyfill as the vitest setup file (jsdom lacks the constructor; Base UI re-dispatches clicks through it), node types return to `tsconfig.test.json` (`@types/node`, pinned like telemetry) because the docs/tokens tests read files.

## Verified

Green on node 25.1.0: `type-check:all`, `lint` (incl. `lint:deps`), `test:unit` (78 tests / 11 files), `arch:check` 12/12, and `test:consumer` (tarball → clean Vite project → build → assert styles, class maps, and the AI docs in the package).

Beyond CI: every component was exercised in a real browser against the packed tarball (all states, light/dark). That run produced three fixes included here: the `nativeButton` contract for render-prop links, the Label color declaration that broke Field's invalid tinting via inheritance, and the portal container escape hatch for subtree themes.

## Next (separate PRs)

Structure batch (`card`, `tabs`, `badge`, `separator`, `skeleton`), overlays (`dialog`, `dropdown-menu`, `tooltip`, `toast`), `table`, composition recipes for the AI layer, kitchen-sink example app, then the #495 publication gates before `private` is lifted.

## Summary by CodeRabbit

* **New Features**
  * Added Checkbox, Field, Input, Label, Radio Group, Select, Switch, and Textarea components.
  * Added accessible form interactions, validation and disabled states, sizing options, and theme-based styling.
  * Enhanced Button support for accessible link-style usage.
* **Documentation**
  * Added comprehensive component usage guidance and published package documentation with an AI-friendly index.
* **Bug Fixes**
  * Standardized corner-radius styling with theme tokens.
* **Tests**
  * Expanded coverage for behavior, accessibility, styling, and documentation completeness.

### DIFF
--- a/packages/ui-kit/llms.txt
+++ b/packages/ui-kit/llms.txt
@@ -33,6 +33,10 @@ Rules for generated code:
 - [RadioGroup](dist/docs/radio-group.md): mutually exclusive options
   (RadioGroup + RadioGroupItem); `value`/`onValueChange` on the group;
   arrow-key navigation built in.
+- [Select](dist/docs/select.md): dropdown picking one value from a
+  list (Select/SelectTrigger/SelectValue/SelectContent/SelectGroup/
+  SelectItem); trigger sizes default|sm; keyboard and typeahead built
+  in.
 - [Switch](dist/docs/switch.md): immediate on/off toggle;
   `checked`/`onCheckedChange`, sizes default|sm; use checkbox for
   submit-time options.
--- a/packages/ui-kit/llms.txt
+++ b/packages/ui-kit/llms.txt
@@ -22,6 +22,9 @@ Rules for generated code:
 - [Button](dist/docs/button.md): action button; variants
   default|destructive|outline|secondary|ghost|link, sizes
   default|sm|lg|icon; render-prop polymorphism for rendering as a link.
+- [Input](dist/docs/input.md): single-line text field; `onValueChange`
+  for the string value, `aria-invalid` for the error state; wires into
+  Field automatically.
 
 ## Optional
 
--- a/packages/ui-kit/llms.txt
+++ b/packages/ui-kit/llms.txt
@@ -11,9 +11,10 @@ Rules for generated code:
   `import '@gears-frontx/ui-kit/theme.css';`
   `import '@gears-frontx/ui-kit/styles.css';`
 - Import components from the package root: `import { Button } from '@gears-frontx/ui-kit';`
-- Set the app font on `<body>` (or `<html>`), not on an inner container:
-  overlay components (Select) portal their popup to `<body>` and inherit
-  the font from there.
+- Set the app font AND `data-theme` on `<html>`/`<body>`, not on an inner
+  container: overlay components (Select) portal their popup to `<body>`
+  and inherit theme tokens and font from there. If a subtree must carry
+  its own theme, pass that element as `container` to `SelectContent`.
 - Brand/appearance changes go through theme tokens (CSS variables in
   `theme.css`, light/dark via `data-theme`), never through inline styles or
   overriding kit class names.
--- a/packages/ui-kit/llms.txt
+++ b/packages/ui-kit/llms.txt
@@ -11,6 +11,9 @@ Rules for generated code:
   `import '@gears-frontx/ui-kit/theme.css';`
   `import '@gears-frontx/ui-kit/styles.css';`
 - Import components from the package root: `import { Button } from '@gears-frontx/ui-kit';`
+- Set the app font on `<body>` (or `<html>`), not on an inner container:
+  overlay components (Select) portal their popup to `<body>` and inherit
+  the font from there.
 - Brand/appearance changes go through theme tokens (CSS variables in
   `theme.css`, light/dark via `data-theme`), never through inline styles or
   overriding kit class names.
--- a/packages/ui-kit/llms.txt
+++ b/packages/ui-kit/llms.txt
@@ -14,6 +14,8 @@ Rules for generated code:
 - Brand/appearance changes go through theme tokens (CSS variables in
   `theme.css`, light/dark via `data-theme`), never through inline styles or
   overriding kit class names.
+- Rounding is a scale derived from `--radius` (`--radius-sm|md|lg|xl`);
+  override `--radius` alone and every component follows.
 - Per-component docs below list the kit-level props; everything else is the
   underlying native element's props.
 
--- a/packages/ui-kit/llms.txt
+++ b/packages/ui-kit/llms.txt
@@ -1,0 +1,28 @@
+# @gears-frontx/ui-kit
+
+> Standard React component base for FrontX templates. Behavior comes from
+> Base UI primitives, styling is compiled CSS on semantic design tokens,
+> variants via a `variant`/`size` prop pair. Consumers need no CSS framework
+> or build plugins.
+
+Rules for generated code:
+
+- Import styles exactly once at the app entry, theme first:
+  `import '@gears-frontx/ui-kit/theme.css';`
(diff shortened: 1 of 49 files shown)


packages/ui-kit/src/components/button/button.md
```diff
@@ -0,0 +1,67 @@
...
+An action button. Wraps the Base UI Button primitive: render-prop
+polymorphism, correct disabled/focus behavior, `type="button"` by default
+(pass `type="submit"` explicitly for form submission).
+
+## When to use
+
+- Any click-triggered action: submit, delete, open a dialog, toggle a panel.
+- Navigation styled as a button:
+  `<Button render={<a href="..." />} nativeButton={false}>` — a real anchor
+  underneath (crawlable, cmd-clickable) with button semantics on top.
```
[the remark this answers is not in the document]
SysoevAndrey: Resolved in `c68fc3d1` by taking the second option from this comment

packages/ui-kit/src/components/textarea/textarea.tsx
```diff
@@ -0,0 +1,10 @@
+import { cx } from 'class-variance-authority';
+import type { ComponentProps } from 'react';
+
+import styles from './textarea.module.css';
+
+export type TextareaProps = ComponentProps<'textarea'>;
+
+export function Textarea({ className, ...props }: TextareaProps) {
+  return <textarea className={cx(styles.textarea, className)} {...props} />;
```
itechmeat: Textarea seems to be the only control left outside the Field wiring: Input is Base UI's Field.Control, so label/aria-describedby/invalid land automatically, while a textarea inside a Field silently gets none of that. Base UI's Field.Control accepts render, so `<Field.Control render={<textarea />} />` would give the same styled tag with the wiring. Was leaving it out deliberate? If so, worth a line in field.md; if not, looks like a two-line fix.

packages/ui-kit/llms.txt
```diff
@@ -0,0 +1,59 @@
...
+- Rounding is a scale derived from `--radius` (`--radius-sm|md|lg|xl`);
+  override `--radius` alone and every component follows.
+- Per-component docs below list the kit-level props; everything else is the
+  underlying native element's props.
+
+## Components
+
+- [Button](dist/docs/button.md): action button; variants
+  default|destructive|outline|secondary|ghost|link, sizes
+  default|sm|lg|icon; render-prop polymorphism for rendering as a link.
```
itechmeat: llms.txt line 30 still sells the Button render prop as "for rendering as a link" - the exact framing c68fc3d1 removed from button.md after the review thread. Since llms.txt is the entry point an agent reads first, the index now contradicts the doc it points to. Same one-line sweep as the button.md fix?

packages/ui-kit/src/components/button/button.md
```diff
@@ -0,0 +1,70 @@
...
+  as a button (`role="button"`), while the real anchor underneath keeps
+  cmd-click/middle-click and crawlability. Not a substitute for a link: for
+  plain navigation use the consumer app's link component.
+
+## When not to use
+
+- Plain in-text navigation — use the consumer app's link component, not
+  `variant="link"`; that variant exists for actions that visually read as
+  links.
+- Toggling on/off state — use `switch` or `checkbox` (planned) instead.
```
itechmeat: [Not a blocker] This "(planned)" outlived the sweep - Switch and Checkbox ship in this very PR. Same one-word fix as the ones already applied to input.md/label.md.

packages/ui-kit/src/components/checkbox/checkbox.module.css
```diff
@@ -0,0 +1,92 @@
...
+  position: relative;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  flex-shrink: 0;
+  width: 1rem;
+  height: 1rem;
+  padding: 0;
+  border: 1px solid var(--input);
+  border-radius: 4px;
```
itechmeat: This PR's whole token story is "override --radius and every corner follows", and checkbox is the one corner that won't - the 4px is raw. If the shadcn reference look matters here, maybe derive it (--radius-xs or calc(var(--radius) - 6px)) so rebranding doesn't skip it?

itechmeat (approved):

packages/ui-kit/src/components/textarea/textarea.tsx
SysoevAndrey: Fixed, thanks

packages/ui-kit/llms.txt
SysoevAndrey: Fixed, thanks

packages/ui-kit/src/components/button/button.md
SysoevAndrey: Fixed, thanks

packages/ui-kit/src/components/checkbox/checkbox.module.css
SysoevAndrey: Fixed, thanks

GeraBart (approved):

Issue #502, opened 2026-07-31
Extract the duplicated bridge-resync and RTL direction logic into demo-mfe shared hooks


tscbmstubp:
Follow-up from [@itechmeat's review of #491](https://github.com/constructorfabric/gears-frontx/pull/491#discussion_r3681298809).

## What is duplicated

#491 added an RTL/bridge-resync block to five screens. Four pieces are pasted verbatim in each:

- the `RTL_LANGUAGES` const
- the `readBridgeProperty` helper
- the `prevBridge` render-time re-read block (theme + language resync on bridge swap)
- the `dir` effect that syncs the Shadow DOM host's text direction to the active language

Copies:

| package | screen |
|---|---|
| demo-mfe | `src/screens/uikit/UIKitElementsScreen.tsx` |
| demo-mfe | `src/screens/profile/ProfileScreen.tsx` |
| demo-mfe | `src/screens/theme/CurrentThemeScreen.tsx` |
| demo-mfe | `src/screens/helloworld/HelloWorldScreen.tsx` |
| _blank-mfe | `src/screens/home/HomeScreen.tsx` |

Only `HomeScreen` got a test for the new bridge-swap behaviour, and it is the one copy that cannot be deduplicated (see below).

## Proposed shape

Extract into `demo-mfe/src/shared/`, per @itechmeat's suggestion:

- `useBridgeProperty(bridge, propertyId, fallback)` — the read + subscribe + resync-on-bridge-swap behaviour
- `useHostDirection(containerRef, language)` — the `dir` effect and `RTL_LANGUAGES`

Move the bridge-swap test onto the hook, so the behaviour is pinned once instead of per-screen.

## The _blank-mfe copy stays

`no-cross-mfe-imports` is an error-severity rule in `template-shell/.dependency-cruiser.cjs`, so `_blank-mfe` cannot import from `demo-mfe`. Per-MFE `src/shared/` duplicated across packages is the existing convention — `useScreenTranslations.ts` and its test already live verbatim in both `demo-mfe/src/shared/` and `_blank-mfe/src/shared/`, and `demo-mfe/src/shared/README.md` scopes that directory to "reuse between the 4 screen entries" of that package.

So this is a 4-screen change inside demo-mfe. `_blank-mfe` keeps its copy, which is defensible anyway — it is a blank starter people copy from.

[...]

PR #520, opened 2026-08-04
refactor(react): extract useLatest hook, fix render-time ref writes (#515)

### LINKED ISSUE #515: Extract a shared useLatest helper for the three deps-less ref-write effects
Follow-up from [@itechmeat's review of #491](https://github.com/constructorfabric/gears-frontx/pull/491#discussion_r3701119254).

## What is duplicated

Three places write a "latest value" ref inside a deps-less effect (runs after every render, during commit — refs must not be mutated while rendering), with the same rationale sentence pasted in each:

| file | ref | what it holds |
|---|---|---|
| `template-shell/packages/react/src/hooks/useApiStream.ts:79-82` | `descriptorRef` | the `StreamDescriptor` value |
| `template-shell/packages/react/src/mfe/components/ExtensionDomainSlot.tsx:96-101` | `onAttachedRef` / `onDetachedRef` | two callback props |
| `template-shell/packages/react/src/queryClient.tsx:444-447` | `latestRef` | `{ options, callbackCtx }` |

There is no `useLatest`-style helper anywhere in `template-shell` or `packages/` today. #502 covers the five MFE-screen copies of a related pattern; these three are outside that scope and need their own pointer.

`useEffectEvent` (React 19.2's purpose-built answer for the callback case) doesn't fit cleanly here even where it would apply: `@types/react` is pinned at `^19.0.8`, below the version that exports it, and `useApiStream`/`queryClient` hold *values*, not callbacks, which `useEffectEvent` doesn't wrap anyway.

## Attempted fix and why it didn't land in #491

A generic `useLatest<T>(value: T)` hook (a plain `useRef` + deps-less effect, same as today, just factored out) builds and type-checks cleanly when substituted at all three sites. It does not lint cleanly:

`eslint-plugin-react-hooks` (currently `7.1.1`, `react-hooks/exhaustive-deps`) only recognizes a ref as an omittable-from-deps stable value when it is declared with a literal `useRef(...)` call in the same scope. Once the ref comes back from a custom hook instead:

- Every place the ref is read now gets a "missing dependency" error. Adding the ref to the array is harmless (ref identity never changes) and *does* clear this one.
- Wherever the ref is read **inside a returned effect-cleanup closure**, a second, different error fires — `"The ref value 'x.current' will likely have changed by the time this effect cleanup function runs"` — and adding the ref to deps does not clear it. Its suggested fix (copy `.current` into a local variable at effect-setup time) is wrong for this pattern specifically: the entire point of a "latest" ref is to read the freshest value *at cleanup time*, not a stale snapshot from setup.

`latestRef` (`queryClient.tsx`) and `onDetachedRef` (`ExtensionDomainSlot.tsx`) both hit the second case — they're read inside their effect's cleanup function. `descriptorRef` and `onAttachedRef` don't, and convert cleanly.

There are zero existing `eslint-disable` comments for `react-hooks/exhaustive-deps` anywhere in this codebase, so accepting them here would be a first, for 2 of the 4 ref instances — and splitting `onAttachedRef`/`onDetachedRef` (written together, mirrored props) between a shared hook and a lint-suppressed inline ref reads worse than the current consistent triplication.

## Open question for whoever picks this up

Pick one:
1. Accept `eslint-disable-next-line react-hooks/exhaustive-deps` at the two cleanup-reading sites, with a comment explaining why (this doc's second bullet).
2. Only extract `useLatest` for the two sites that lint cleanly (`descriptorRef`, `onAttachedRef`), leave `latestRef`/`onDetachedRef` inline — accepts an inconsistent split within `ExtensionDomainSlot.tsx`.
3. Something else — e.g. bump `@types/react` far enough to get `useEffectEvent` for the callback case, and handle the two value-holding refs (`descriptorRef`, `latestRef`) with a `useLatest` restricted to non-cleanup reads.

## Why not in #491

Behavior-neutral refactor with a real tooling wrinkle to work through — not worth blocking the conflict-resolution merge on. Not a blocker per @itechmeat's review either.

## Verification

Whichever shape lands: `npm run build && npm run type-check && npm run lint && npm run test:unit` inside `template-shell` must stay green, same as the rest of #491's fixes.

tscbmstubp:
Closes #515.

## What

Extracts a shared `useLatest<T>(value)` hook (`template-shell/packages/react/src/hooks/useLatest.ts`) and uses it at all four "latest value" ref sites the issue lists:

- `descriptorRef` in `hooks/useApiStream.ts`
- `onAttachedRef` / `onDetachedRef` in `mfe/components/ExtensionDomainSlot.tsx`
- `latestRef` in `queryClient.tsx`

The hook writes the ref in a deps-less effect (commit phase), not during render.

## Which option from the issue landed

Option 1: one shared helper at all four sites, with exactly **two** `eslint-disable-next-line react-hooks/exhaustive-deps` comments — only at the two cleanup-closure read sites the issue predicted (`ExtensionDomainSlot.tsx`, `queryClient.tsx`), each with a comment explaining why the rule's suggested copy-at-setup fix is wrong for this pattern (the point is reading the freshest value at cleanup time). The other conversions lint cleanly with the ref added to the deps array. `useEffectEvent` was ruled out per the issue (`@types/react` pinned below it; two sites hold values, not callbacks).

## Bonus: fixes 4 live lint errors on develop

On `develop` these refs are written **during render**, and `eslint-plugin-react-hooks` 7.1.1's `react-hooks/refs` rule hard-errors on that ("Cannot access refs during render") — 4 errors that this PR removes. Lint goes 24 → 20 errors; the remaining 20 are pre-existing and out of scope (gitignored `dist-lib/` artifacts plus `react-hooks/set-state-in-effect` in `useCanAccess.ts`, `useScreenTranslations.ts`, and two untouched `useApiStream.ts` lines).

One deliberate semantic note: in `ExtensionDomainSlot`, if a parent changes `onDetached` in the same commit as `domainId`/`registry`, the cleanup now fires the last *committed* handler (one render behind) rather than the same-commit one — the handler that was in effect for the domain actually being torn down. The code comment states this precisely.

## Verification

From `template-shell/`: `npm run build`, `npm run type-check`, `npm run test:unit` (138 passed, 3 new tests for `useLatest`) all green; `npm run lint` strictly improves vs the develop baseline as above. `npm run arch:deps` clean. Independently QA-verified against origin/develop baseline via stdin-linting the base blobs.

## Scope note

Issue #502's five MFE-screen copies of the related pattern remain out of scope, as the issue specifies. `useLatest` is internal to the react package (not barrel-/publicly exported).

## Summary by CodeRabbit

* **New Features**
  * Added a reusable React hook for accessing the latest value in callbacks and effects.

* **Tests**
  * Added coverage for initial values, updates after rerendering, and stable reference identity.

* **Refactor**
  * Improved callback and lifecycle handling across streaming, extension slot, and mutation functionality while preserving existing behavior.

### DIFF
--- a/template-shell/packages/react/src/mfe/components/ExtensionDomainSlot.tsx
+++ b/template-shell/packages/react/src/mfe/components/ExtensionDomainSlot.tsx
@@ -17,7 +17,7 @@
 // @cpt-state:cpt-frontx-state-react-bindings-extension-slot:p1
 // @cpt-dod:cpt-frontx-dod-react-bindings-extension-slot:p1
 
-import React, { useEffect, useRef, useState } from 'react';
+import React, { useEffect, useEffectEvent, useRef, useState } from 'react';
 import type { MfeRegistry } from '@gears-frontx/framework';
 
 /**
@@ -89,12 +89,15 @@ export function ExtensionDomainSlot(props: ExtensionDomainSlotProps): React.Reac
 
   const rootRef = useRef<HTMLDivElement | null>(null);
 
-  // Callbacks in refs so effect deps stay minimal — inline parent handlers do not
-  // re-run the attach/detach effect on every parent render.
-  const onAttachedRef = useRef(onAttached);
-  const onDetachedRef = useRef(onDetached);
-  onAttachedRef.current = onAttached;
-  onDetachedRef.current = onDetached;
+  // Effect events so effect deps stay minimal — inline parent handlers do not
+  // re-run the attach/detach effect on every parent render, yet each call
+  // sees the latest committed onAttached/onDetached prop.
+  const emitAttached = useEffectEvent((root: Element) => {
+    onAttached?.(root);
+  });
+  const emitDetached = useEffectEvent(() => {
+    onDetached?.();
+  });
 
   // @cpt-begin:cpt-frontx-state-react-bindings-extension-slot:p1:inst-slot-attached
   // Tracks whether the mounter has been attached to the DOM root.
@@ -119,7 +122,7 @@ export function ExtensionDomainSlot(props: ExtensionDomainSlotProps): React.Reac
     setAttached(true);
     // @cpt-end:cpt-frontx-state-react-bindings-extension-slot:p1:inst-slot-attached
 
-    onAttachedRef.current?.(root);
+    emitAttached(root);
 
     // @cpt-begin:cpt-frontx-flow-react-bindings-extension-domain-slot:p1:inst-detach-root
     return () => {
@@ -132,12 +135,18 @@ export function ExtensionDomainSlot(props: ExtensionDomainSlotProps): React.Reac
       setAttached(false);
       // @cpt-end:cpt-frontx-state-react-bindings-extension-slot:p1:inst-slot-detached
 
-      onDetachedRef.current?.();
+      // emitDetached is an effect event: it always calls the onDetached prop
+      // from the latest committed render, not whatever prop was current when
+      // this effect instance was originally set up.
+      emitDetached();
     };
     // @cpt-end:cpt-frontx-flow-react-bindings-extension-domain-slot:p1:inst-detach-root
     // @cpt-begin:cpt-frontx-state-react-bindings-extension-slot:p2:inst-slot-stable-rerender
     // ATTACHED → ATTACHED (no-op) when parent supplies the same domainId and
     // registry on re-render: useEffect dep equality short-circuits the attach/detach pair.
+    // emitAttached/emitDetached are effect events — non-reactive by contract,
+    // so they are excluded from deps and the effect does not re-run when the
+    // parent passes new inline handlers.
   }, [registry, domainId]);
     // @cpt-end:cpt-frontx-state-react-bindings-extension-slot:p2:inst-slot-stable-rerender
 
--- a/template-shell/package-lock.json
+++ b/template-shell/package-lock.json
@@ -43,7 +43,7 @@
         "@testing-library/react": "16.3.2",
         "@testing-library/user-event": "^14.6.1",
         "@types/lodash": "4.17.20",
-        "@types/react": "19.0.8",
+        "@types/react": "19.2.18",
         "@types/react-dom": "19.0.3",
         "@vitejs/plugin-react": "4.3.4",
         "@vitest/coverage-v8": "4.1.4",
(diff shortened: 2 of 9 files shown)


template-shell/packages/react/src/mfe/components/ExtensionDomainSlot.tsx
```diff
@@ -132,13 +131,24 @@ export function ExtensionDomainSlot(props: ExtensionDomainSlotProps): React.Reac
...
       // @cpt-end:cpt-frontx-state-react-bindings-extension-slot:p1:inst-slot-detached
 
+      // onDetachedRef (from useLatest) is deliberately read at cleanup time,
+      // not copied at setup time. React runs this cleanup before running
+      // useLatest's own deps-less effect for the commit that triggered it, so
+      // the value read here is the last *committed* onDetached — one render
+      // behind if onDetached changes in the same commit as domainId/registry
+      // — rather than whatever prop was current when this effect instance
+      // was originally set up.
+      // eslint-disable-next-line react-hooks/exhaustive-deps
```
itechmeat: This adds the first `react-hooks/exhaustive-deps` disables in the repo, and our style guide says never to disable that rule. I understand why the rule's own fix does not work here - snapshotting `.current` at setup would defeat a "latest value" ref - so I am fine with it, but I want to sign off on it explicitly since it becomes the precedent everyone else points at. Could you keep these two as the only disables and make sure any future one has a rationale at least this strong?

itechmeat (approved):

template-shell/packages/react/src/mfe/components/ExtensionDomainSlot.tsx
tscbmstubp: Went one better and removed both disables in 3bb125e1: the runtime pin was already React ^19.2.4, and 19.2 ships `useEffectEvent` — the purpose-built API for calling the latest committed callback from an effect without making it reactive. eslint-plugin-react-hooks v7 understands it natively, so both cleanup-time reads (this `onDetached` emit and `useFrontXMutation`'s `abortOnUnmount` check) now lint clean with zero `eslint-disable`s in the repo, and no precedent gets set. `useLatest` stays for the useCallback-body reads, where the rule never objected.

Two side effects worth noting: `@types/react` moves 19.0.8 → 19.2.18 for the stable `useEffectEvent` typings (lockfile patched surgically per the #524 remediation — a naive `npm install` reproduced the self-link drop on the first try), and the branch is merged up with develop since the pre-#535 pins made it unbuildable locally. One behavioral nuance: the old comment documented the cleanup reading a one-render-behind `onDetached` when it changed in the same commit as `domainId`/`registry`; an effect event reads the value from the latest committed render, which is the React-blessed semantics and strictly closer to 'the latest handler'.

GeraBart (approved):

PR #517, opened 2026-08-03
feat(arch): federate the SDLC artifacts — members own their chains, root keeps the layers (#495)

tscbmstubp:
**The federation refactoring, as one PR** — per the maintainer's direction in [this comment](https://github.com/constructorfabric/gears-frontx/pull/517#issuecomment-5177483406): no new ADR, the refactoring itself. Root DESIGN and DECOMPOSITION shrink to layer-level content; every lib and orchestrator owns the SDLC artifacts that describe it, in its package folder. (The ADR this PR previously carried is archived on the fork at `arch/495-adr-archive`; the outdated review threads below refer to it.)

**The PRD is federated** — per the maintainer's answers on #495 ([1](https://github.com/constructorfabric/gears-frontx/issues/495#issuecomment-5187380480), [2](https://github.com/constructorfabric/gears-frontx/issues/495#issuecomment-5187390530)): the root PRD describes the 3-layer approach (libs, templates, orchestration) and keeps only the requirements binding every member equally; every member owns the PRD explaining its own reqs; and the CLI's artifacts own the contract to templates. Five member PRDs join telemetry's — mfes, gts-plugin, api, cli, cyber-pilot-kit-frontx — with requirement, interface, contract and use-case elements moved verbatim, identifiers unchanged, so every marker and DECOMPOSITION citation resolves as before (the same measured pattern as the DESIGN federation). LAYER-2 now requires PRD+DESIGN+FEATURE for every member and the registration gate enforces all three.

> **History note:** after review the branch was rebased onto develop and its working commits squashed into five logically grouped ones: telemetry as the federation pilot (`683df307`), the published-libraries member artifacts (`e1cb7a06`), the projects-orchestration member artifacts (`3972139d`), the root artifacts at layer altitude (`f8d47134`), and tooling/guards/docs (`9fdb6725`). Commit hashes cited in older review threads refer to pre-squash history and no longer resolve; the narrative below is kept as the record of how the PR got here.

### Progress

- [x] **telemetry** — full member chain (PRD, DESIGN, 2 FEATUREs), registered as its own system, source files marked (`683df307`)
- [x] **gts-plugin** — first *moved* member: `gts-type-provider` FEATURE relocated from root, member registered (`e1cb7a06`)
- [x] **api** — 1 FEATURE (`e1cb7a06`)
- [x] **mfes** — 5 FEATUREs + `type-substrate-port` (spans mfes/gts-plugin; filed with mfes, which defines the port) (`e1cb7a06`)
[...]

PR #521, opened 2026-08-04
docs(react): document useApiStream status lifecycle at code altitude (#516)

### LINKED ISSUE #516: Document useApiStream's status lifecycle (or walk back its @cpt- markers)
Follow-up from [@itechmeat's review of #491](https://github.com/constructorfabric/gears-frontx/pull/491#discussion_r3701127718).

## What changed

`useApiStream`'s `status` used to initialize to `'idle'` unconditionally. It now initializes to `'connecting'` whenever `enabled` (which defaults to `true`):

```ts
const [status, setStatus] = useState<StreamStatus>(() => (enabled ? 'connecting' : 'idle'));
// ...
return { data, events, status: enabled ? status : 'idle', error, disconnect };
```

Nothing ever writes `'idle'` back into `status` once a connection starts, so for the default configuration (`enabled: true`), `'idle'` is now unobservable — the first committed frame changed from `'idle'` to `'connecting'`. That's a real, user-visible contract change on a hook that carries `@cpt-dod:cpt-frontx-dod-request-lifecycle-use-api-stream:p2` / `@cpt-flow:cpt-frontx-flow-request-lifecycle-use-api-stream:p2` markers.

## Why this needs a decision, not just a fix

`use-api-stream` doesn't appear in any file under `architecture/`, nor in any of the 18 FEATURE specs — so nothing currently documents (and nothing currently contradicts) this hook's status lifecycle. The `@cpt-` markers point at instructions that don't exist yet.

Per this repo's convention (`architecture/` as the authority on what the system is and why, with `@cpt-` markers tracing code back to it), this status lifecycle should get written down somewhere before behavior like this becomes tribal knowledge — but which FEATURE it belongs under, and whether it's this PR's job to create that FEATURE from scratch versus a follow-up, is an architecture-authoring call, not a code-review fix.

## Options for whoever picks this up

1. Add a FEATURE spec (or extend an existing one) documenting `useApiStream`'s status state machine, with the `@cpt-dod`/`@cpt-flow` markers in the code pointing at real instructions.
2. Decide the markers were premature/aspirational and remove them until a FEATURE exists to justify them.
3. Something else — e.g. a lighter-weight doc comment in the hook itself, if a full FEATURE spec is overkill for this one behavior.

## Why not in #491

Behavior-neutral from #491's own perspective (the lifecycle change happened earlier in this PR's history, not in the conflict-resolution work), and writing architecture artifacts is a bigger, more deliberate undertaking than the mechanical fixes this PR's conflict-resolution pass covered. Not a blocker per @itechmeat's review either.

tscbmstubp:
Closes #516.

## The decision (ADR-0033)

Issue #516 asked whether `useApiStream`'s status lifecycle should get a FEATURE spec or whether its `@cpt-` markers should be walked back. Investigation showed the markers are **residue from the archived pre-redesign `feature-request-lifecycle` FEATURE** (moved to `.archive-pre-ecosystem-redesign/`, then deleted) — and `template-shell/**` is already deliberately excluded from `cfs validate`'s universe in `.cf-studio/config/artifacts.toml`. Reviving one archived feature for one hook would create a specified island inside otherwise-unspecified template payload.

So this PR records **ADR-0033 (template-territory traceability)**: the ecosystem artifact tree does not specify template payload; residue `@cpt-` markers there bind nothing, are removed when a file is otherwise modified, and no new ones are added; consumer-visible template behaviour is documented **at code altitude**. A template-owned artifact tree is explicitly deferred, not rejected. (Originally numbered 0034 while the #495 branch carried an ADR-0033; that branch archived its ADR, so this one took the freed number.)

## Changes

- `architecture/ADR/0033-template-territory-traceability.md` — the decision; `architecture/DESIGN.md` — one line in the §1.2 ADR inventory
- `useApiStream.ts` — residue `@cpt-dod`/`@cpt-flow`/`@cpt-FEATURE`/`@cpt-begin/end` markers removed; full status-lifecycle TSDoc added (per-value semantics, all transitions, `enabled` gating, data/events persistence across a disable); header corrected from "EventSource lifecycle" to "connection lifecycle" (the transport is the descriptor's detail)
- `useApiStream.test.tsx` — residue markers removed, tests untouched
- `template-shell/packages/react/CLAUDE.md` — returns list, `enabled` subsection, gating advice, and a pointer to the TSDoc, which is the single owner of the state machine (see the review response)
- `.cf-studio/config/artifacts.toml` — template ignore reason now cites `cpt-frontx-adr-template-territory-traceability`
- `.cf-studio/config/AGENTS.md`, root `CLAUDE.md` — the marker-preservation instructions gained the template-territory carve-out so agents and FX-002 review enforce the ADR instead of contradicting it
- `packages/api/src/types.ts` — the last two archived-generation markers in the scanned universe removed (comment-only)
- `scripts/template-discovery.test.mjs` — sync guard: the artifacts.toml template-territory ignore globs must cover exactly the manifest-discovered template directories, so the literal config cannot drift from the manifest rule the prose surfaces state

## Response to the review (squashed into the single head commit)

All 13 threads addressed:

- **ADR renumbered 0034 → 0033** — the skip rationale went stale when #517 archived its ADR (references use the ID slug, so only the filename moved)
- **The removal rule is stated once**: whole file — when a template file is otherwise modified, every residue marker it carries leaves; the ADR asks only that marker-only deletions remain easy to distinguish from substantive changes in review, without prescribing a commit topology
- **Instruction surfaces taught the carve-out** (AGENTS.md ×2, root CLAUDE.md) — an agent following them no longer preserves what the ADR retires
- **Mechanism-agnostic deferral, not "superseded"** — the decision is revisited through the repository's ADR lifecycle once template ownership settles, preserving ADR-0027's immutability premise; the prompting-instance narration dropped (the Decision Outcome keeps the example in timeless form)
- **The scanned universe is actually clean now** — `packages/api/src/types.ts` carried the last two dead markers; a sweep found no others, so the ADR's consequence holds
- **The TSDoc owns the state machine**; CLAUDE.md's duplicate table/transitions (already drifted within this PR) replaced by a pointer plus gating advice
- **The example models the gating advice** (`status === 'connecting'` alone) instead of contradicting it behind two caveats
- **TSDoc truth fixes**: cleanup-path rationale corrected (dep array is `[descriptorKey, enabled, mode]`; unmount arrives via unmount semantics); disconnect-while-disabled exception to the `'idle'` invariant recorded; the completion-during-connect adoption quirk documented (a short-lived stream can show `'disconnected'` → `'connected'` with no consumer action — behavioural fix left to #491's derived-status rewrite); `data` no longer claims "always set"

## Verification

- Stripping comments and blank lines, all three touched TS files are **identical to develop** — zero logic change (verified at the rebased head)
- `cfs validate` at the rebased head: 0 errors, 0 warnings, coverage 124/124; `validate-toc` / `check-language` pass
- react package: type-check clean, 135/135 tests pass (17/17 in `useApiStream.test.tsx`) — unchanged since only comments and docs moved
- Every documented lifecycle claim was verified line-by-line against the hook implementation (QA + architect review)

## ⚠️ Coordination with #491

These docs are grounded in **develop's** `useState<StreamStatus>('idle')` initializer. The unmerged #491 changes initialization to `enabled ? 'connecting' : 'idle'` and derives the returned status. The in-file reminder was removed in round 3 (maintainer-only text does not belong in shipped template payload), so **this section is the coordination record**: whichever PR lands second must update the TSDoc's initial-state claims. CLAUDE.md no longer carries initial-state claims at all. #491's derived-status rewrite is also the right home for fixing the completion-during-connect adoption quirk and the stray-`'connected'`-after-`disconnect()` sibling documented in the TSDoc.

## Summary by CodeRabbit

- **Documentation**
  - Clarified streaming connection lifecycles, status transitions, loading and error states, disconnect behavior, and enabling or disabling streams.
  - Added guidance distinguishing product code from reusable template content and documenting template maintenance practices.

- **Maintenance**
  - Added a formal architecture decision record for template traceability.
  - Removed obsolete internal annotations and updated configuration and review guidance to reflect template boundaries.

### DIFF
--- a/.cf-studio/config/AGENTS.md
+++ b/.cf-studio/config/AGENTS.md
@@ -59,14 +59,16 @@ WHEN:
 DO:
   - LOAD `architecture/DESIGN.md` §1.3 (Architecture Layers) and §2.2 (Constraints) before editing any code
   - LOAD `packages/cyber-pilot-kit-frontx/guidelines/ecosystem-boundaries.md` and apply the boundary constraints for every touched package
-  - LOAD the governing FEATURE.md — each lives with the package it describes, under `packages/*/architecture/features/` — before changing code that carries its `@cpt-` markers
-  - RUN keep `@cpt-` traceability markers intact and aligned with the FEATURE instructions they cite
+  - LOAD `architecture/ADR/0033-template-territory-traceability.md` before changing template territory (any top-level directory carrying a `frontx-template.json` manifest)
+  - LOAD the governing FEATURE.md — each lives with the package it describes, under `packages/*/architecture/features/` — before changing ecosystem code that carries its `@cpt-` markers
+  - RUN keep `@cpt-` traceability markers intact and aligned with the FEATURE instructions they cite — in ecosystem code only
+  - RUN apply the template-territory marker policy from `cpt-frontx-adr-template-territory-traceability` to template code changes
 
 RULES:
   - ALWAYS read the named DESIGN constraint (MFES-*, GTS-PLUGIN-*, API-*, CLI-*, KIT-*) before editing code it governs; NEVER code from assumed rules
   - ALWAYS stop and ask when a change would cross a DESIGN §2.2 constraint or a package boundary from `ecosystem-boundaries.md`
-  - ALWAYS stop and ask when no FEATURE covers behaviour the change introduces
-  - NEVER restate architecture rules in artifacts or prompts; reference `architecture/DESIGN.md` and the kit guideline instead
+  - ALWAYS stop and ask when no FEATURE covers behaviour the change introduces in ecosystem code; no ecosystem FEATURE covers template territory by decision, so template consumer-visible behaviour is governed by `cpt-frontx-adr-template-territory-traceability`
+  - NEVER restate architecture rules in artifacts or prompts; reference `architecture/DESIGN.md`, ADRs, and the kit guideline instead
 ```
 
 ```pdsl
@@ -81,9 +83,9 @@ WHEN:
   - REQUIRE reviewing code changes in this repository
 
 DO:
-  - LOAD `architecture/DESIGN.md` §2.2 and `packages/cyber-pilot-kit-frontx/guidelines/ecosystem-boundaries.md`; NEVER review from memory of their rules
+  - LOAD `architecture/DESIGN.md` §2.2, `packages/cyber-pilot-kit-frontx/guidelines/ecosystem-boundaries.md`, and `architecture/ADR/0033-template-territory-traceability.md`; NEVER review from memory of their rules
   - RUN FX-001 Constraints: every changed package complies with its named DESIGN §2.2 constraints and its package boundary in `ecosystem-boundaries.md`; ambiguous ownership was raised, not guessed
-  - RUN FX-002 Traceability: changed code keeps `@cpt-` markers consistent with the governing FEATURE, and `cfs validate` passes
+  - RUN FX-002 Traceability: changed ecosystem code keeps `@cpt-` markers consistent with the governing FEATURE, and `cfs validate` passes; changed template territory (any top-level directory carrying a `frontx-template.json` manifest) follows the marker policy in `cpt-frontx-adr-template-territory-traceability`
   - RUN FX-003 Architecture checks: `npm run arch:check` passes for the changed packages
 
 RULES:
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -8,6 +8,6 @@ ALWAYS resolve and enforce prerequisites of skills/workflows/commands BEFORE app
 
 AI tooling and FrontX development guidelines are provided by the Constructor Studio kit under `.cf-studio/`.
 
-Treat `architecture/` as the authority on what the system is and why: PRD for intent, DESIGN for structure, ADRs for decisions, and each FEATURE for the behaviour its numbered instructions specify. Code carries `@cpt-` markers back to those instructions; `cfs validate` checks that the chain holds.
+Treat `architecture/` as the authority on what the system is and why: PRD for intent, DESIGN for structure, ADRs for decisions, and each FEATURE for the behaviour its numbered instructions specify. Ecosystem code carries `@cpt-` markers back to those instructions; `cfs validate` checks that the chain holds. Template territory — any top-level directory carrying a `frontx-template.json` manifest (how `scripts/template-discovery.mjs` finds templates) — sits outside this chain: markers found there are non-authoritative residue being retired as files change (`cpt-frontx-adr-template-territory-traceability`).
 
 ALL user requests MUST be handled by the Orchestrator agent.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,6 +8,6 @@ ALWAYS resolve and enforce prerequisites of skills/workflows/commands BEFORE app
 
 AI tooling and FrontX development guidelines are provided by the Constructor Studio kit under `.cf-studio/`.
 
-Treat `architecture/` as the authority on what the system is and why: PRD for intent, DESIGN for structure, ADRs for decisions, and each FEATURE for the behaviour its numbered instructions specify. Code carries `@cpt-` markers back to those instructions; `cfs validate` checks that the chain holds.
+Treat `architecture/` as the authority on what the system is and why: PRD for intent, DESIGN for structure, ADRs for decisions, and each FEATURE for the behaviour its numbered instructions specify. Ecosystem code carries `@cpt-` markers back to those instructions; `cfs validate` checks that the chain holds. Template territory — any top-level directory carrying a `frontx-template.json` manifest (how `scripts/template-discovery.mjs` finds templates) — sits outside this chain: markers found there are non-authoritative residue being retired as files change (`cpt-frontx-adr-template-territory-traceability`).
 
 ALL user requests MUST be handled by the Orchestrator agent.
--- a/architecture/ADR/0033-template-territory-traceability.md
+++ b/architecture/ADR/0033-template-territory-traceability.md
@@ -1,0 +1,126 @@
+---
+status: accepted
+date: 2026-08-04
+---
+
+# Which Artifact Tree Specifies Template-Territory Code, and What Its Traceability Markers Mean
+
+<!-- toc -->
+
+- [Context and Problem Statement](#context-and-problem-statement)
+- [Decision Drivers](#decision-drivers)
+- [Considered Options](#considered-options)
+- [Decision Outcome](#decision-outcome)
+  - [Consequences](#consequences)
+  - [Confirmation](#confirmation)
+- [Pros and Cons of the Options](#pros-and-cons-of-the-options)
+  - [Template territory is outside the ecosystem artifact universe; markers there are non-authoritative residue, removed as files are touched](#template-territory-is-outside-the-ecosystem-artifact-universe-markers-there-are-non-authoritative-residue-removed-as-files-are-touched)
+  - [Extend the ecosystem artifact tree to cover template internals](#extend-the-ecosystem-artifact-tree-to-cover-template-internals)
+  - [Keep the residue markers and rely on the scan exclusion alone](#keep-the-residue-markers-and-rely-on-the-scan-exclusion-alone)
+  - [Give template territory its own artifact tree now](#give-template-territory-its-own-artifact-tree-now)
+- [More Information](#more-information)
+- [Traceability](#traceability)
+
+<!-- /toc -->
+
+**ID**: `cpt-frontx-adr-template-territory-traceability`
+
+## Context and Problem Statement
+
+The repository's architecture artifacts describe the FrontX ecosystem — the published packages, the CLI, and the AI tooling kit — while the template subtrees — the top-level directories carrying a `frontx-template.json` manifest, which is how the repository defines a template — hold template payload the CLI resolves and applies into a developer's repository, not ecosystem contract. That payload still carries `@cpt-` traceability markers from the pre-ecosystem-redesign artifact generation, whose FEATUREs (`feature-request-lifecycle`, `feature-react-bindings`, `feature-state-management` and their peers) were archived and then deleted, so the markers name instruction IDs no artifact defines. Reviewers reading a marked template file reasonably infer a specified contract behind it and open a gap report when they cannot find one. Which artifact tree, if any, specifies template-territory code, and what status do the `@cpt-` markers already sitting in it have?
+
+## Decision Drivers
+
+* **One universe per artifact tree** — the ecosystem's PRD, DESIGN, and DECOMPOSITION describe the ecosystem's own artifacts; admitting template payload into that same tree would give one artifact set two unrelated subjects and two reasons to change.
+* **Template content is not ecosystem content** — the ecosystem's tooling deliberately carries no bundled template or solution content (`cpt-frontx-principle-template-agnostic-tooling`); its specification should draw the same line, or the decoupling holds in code but not in the artifacts.
+* **A marker is a claim** — a `@cpt-` marker asserts that a numbered instruction specifies the marked region; a marker that names nothing is a false claim to every reader, whether or not any validator reads it.
+* **Silence must be legible** — that template internals are unspecified is a deliberate position, and it must be discoverable from the artifacts rather than inferred from a validator's exclusion list.
+* **No premature specification** — writing FEATUREs for template internals now would fix contracts for code whose ownership is still moving (the template split), and it is the larger, harder-to-undo commitment of the available options.
+
+## Considered Options
+
+* **Template territory is outside the ecosystem artifact universe; markers there are non-authoritative residue, removed as files are touched** — the ecosystem artifact tree's subject is the ecosystem's own artifacts; template subtrees are payload it does not specify. Existing `@cpt-` markers in those subtrees are declared residue of the archived generation: they bind nothing, no new ones are added, and the ones present are removed from a file when that file is otherwise modified. Behaviour of template code is documented at code altitude, where the code lives.
+* **Extend the ecosystem artifact tree to cover template internals** — add DECOMPOSITION entries and FEATUREs (starting with a revived request-lifecycle feature) so every existing marker in template territory resolves to a live instruction.
+* **Keep the residue markers and rely on the scan exclusion alone** — change nothing in the artifacts or the payload; the registry's exclusion of the template subtrees is the whole answer.
+* **Give template territory its own artifact tree now** — author a separate, nested artifact tree owned by the template, and re-point every marker in the payload at IDs from that tree.
+
+## Decision Outcome
+
+Chosen option: **Template territory is outside the ecosystem artifact universe; markers there are non-authoritative residue, removed as files are touched**, because it is the only option that keeps one subject per artifact tree, retires the false claims the stale markers make, and commits to no specification the template's ownership is not yet settled enough to support.
+
+The **ecosystem artifact tree specifies the ecosystem's own artifacts** — the published packages, the CLI, and the AI tooling kit. Template subtrees are payload the CLI resolves, applies, and upgrades through the uniform mechanism (`cpt-frontx-adr-uniform-template-mechanism`); what they contain internally is the template's own business, bounded by the ownership declaration it publishes (`cpt-frontx-adr-template-ownership-boundary-declaration`), not an ecosystem contract. Consequently **no ecosystem FEATURE is authored to back template-internal code**, and no artifact in this tree is extended to cover it.
+
+The `@cpt-` markers presently in template territory are **residue of the archived pre-redesign generation and bind nothing**: they are not authoritative, they are not evidence that a contract was specified, and they must not be read as traceability. **No new markers are added to template territory**, and when a template file is otherwise modified, **every residue marker that file carries is removed** — the rule is uniform, needs no separate sweep, and leaves no half-covered file behind. The strip is mechanical, and review should keep marker-only deletion easy to distinguish from substantive change. Where template code has behaviour a consumer depends on — a hook's observable status lifecycle, for instance — that behaviour is documented **at code altitude, in the code and its package documentation**, which is the reader's location and carries no claim of upstream specification.
+
+The scope of this decision is which artifact tree specifies template-territory code and the standing of `@cpt-` markers found there. It does not decide the boundary a template declares (`cpt-frontx-adr-template-ownership-boundary-declaration`), how the CLI applies or upgrades a template (`cpt-frontx-adr-uniform-template-mechanism`, `cpt-frontx-adr-project-upgrade-mechanism`), nor whether the template will eventually carry an artifact tree of its own — that is deliberately deferred (see More Information).
+
+### Consequences
+
+* Good, because each artifact tree keeps one subject: the ecosystem's artifacts are described by the ecosystem's artifacts, and template payload is not smuggled into them.
+* Good, because a `@cpt-` marker recovers its meaning inside the specified universe: there it names a live instruction; residue in paths the registry ignores is retired by the rule above.
+* Good, because the position is now stated in the artifacts, so the next reviewer who finds a marked-but-unspecified template file reads a decision instead of filing a gap.
+* Good, because it commits to no specification of code whose ownership is still moving, leaving the template's own artifact tree available as a later, deliberate step.
+* Bad, because template territory keeps residue markers until each file is otherwise touched, so the payload stays in a mixed state for as long as that takes.
+* Bad, because behaviour documented only at code altitude is easier to change without review than behaviour fixed by a numbered instruction, so a consumer-visible change in template payload rests on code review alone.
+
+### Confirmation
+
+Compliance is confirmed by the artifact registry and by review. The registry admits only the ecosystem's own source paths to traceability scanning and excludes the template subtrees, so no template file is scanned for markers and no marker there is validated; `cfs validate` passing therefore never depends on template payload. Review confirms that no FEATURE in this tree names a path under a template subtree, that a change adding a `@cpt-` marker to template territory is refused, and that a change modifying an already-marked template file removes every residue marker that file carries while keeping the mechanical marker deletion easy to distinguish from substantive change and documenting any consumer-visible behaviour at code altitude in the same change.
+
+## Pros and Cons of the Options
+
+### Template territory is outside the ecosystem artifact universe; markers there are non-authoritative residue, removed as files are touched
+
+The ecosystem tree specifies ecosystem artifacts only; template payload is unspecified by it, residue markers bind nothing and are removed as files are modified, and template behaviour is documented in the code.
+
+* Good, because one artifact tree keeps one subject, matching the tooling's own template-agnostic separation.
+* Good, because it retires false traceability claims rather than preserving them.
+* Good, because the removal rule is uniform and needs no repository-wide sweep to be consistent.
+* Neutral, because it leaves the template's own artifact tree as an open, deferred option.
+* Bad, because the payload stays mixed until files are touched, and code-altitude documentation is weaker protection than a numbered instruction.
+
+### Extend the ecosystem artifact tree to cover template internals
+
+Revive the archived feature areas as live FEATUREs, with DECOMPOSITION entries and DESIGN components, so every existing marker resolves.
+
+* Good, because every marker in the repository would resolve to a live instruction.
+* Good, because consumer-visible template behaviour would be fixed by numbered instructions.
+* Bad, because it re-imports template payload into an artifact tree whose subject is the ecosystem, giving one tree two subjects and two reasons to change.
+* Bad, because it is the largest and least reversible option, fixing contracts for code whose ownership is still in motion.
+* Bad, because doing it for one hook, rather than for every archived feature area at once, produces an arbitrary island of specified code inside unspecified payload.
+
+### Keep the residue markers and rely on the scan exclusion alone
+
+Change nothing; the registry's exclusion of the template subtrees is the whole answer.
+
+* Good, because it costs nothing and breaks no validation, which already ignores those paths.
+* Bad, because the false claims stay in the payload and keep misleading readers — the exclusion is invisible from the file a reviewer is reading.
+* Bad, because it leaves the position recorded only in tool configuration, where an architectural decision is not discoverable.
+
+### Give template territory its own artifact tree now
+
+Author a separate nested artifact tree owned by the template and re-point the payload's markers at its IDs.
+
+* Good, because it would specify template behaviour without mixing subjects, and would give the markers a legitimate home.
+* Neutral, because it is the natural destination if template internals ever need specified contracts.
+* Bad, because it is premature while template ownership is still moving, and it is a large authoring commitment ahead of any demonstrated need.
+
+## More Information
+
+**Deferred deliberately**: whether template territory gets an artifact tree of its own is **not decided here**. It is deferred until template ownership has settled and template internals present a contract a consumer depends on across versions — at which point this decision must be revisited through the repository's ADR lifecycle, markers may be reintroduced against IDs from that tree, and the registry's scanning scope is revisited in the same step. Deciding it now would fix contracts for code whose ownership is in motion.
+
+The uniform mechanism by which the CLI treats any template is decided in `cpt-frontx-adr-uniform-template-mechanism`; the boundary a template declares over what it owns in `cpt-frontx-adr-template-ownership-boundary-declaration`. These are non-binding pointers and do not form part of this decision's durable identity.
+
+Applicability of the remaining checklist categories: **REL** — Not applicable, because this decision changes no runtime behaviour and introduces no failure mode; it fixes what the artifacts specify and what a marker asserts. **PERF** — Not applicable, because no latency or throughput budget is bound to artifact scope. **SEC** — Not applicable, because no secret material or authentication surface is introduced. **DATA** — Not applicable, because no schema is fixed. **INT** — addressed: the integration surface a consumer relies on is the template's declared ownership boundary and its published payload, not an ecosystem instruction for template internals. **OPS** — Not applicable, because there is no running service. **MAINT** — addressed: markers regain a single meaning, and the touch-it-remove-it rule keeps residue retirement uniform without a repository-wide sweep. **COMPL** — Not applicable. **UX** — addressed for the developer-as-reader: a marked file inside the specified universe resolves to a live instruction, and template payload no longer implies specification it does not have. **BIZ** — Not applicable, because product requirements live in the PRD and are cited here by ID.
+
+## Traceability
+
+- **PRD**: [PRD.md](../PRD.md)
+- **DESIGN**: [DESIGN.md](../DESIGN.md)
+
+This decision directly addresses the following requirements and design elements:
+
+* `cpt-frontx-principle-template-agnostic-tooling` — The ecosystem's tooling bundles no template or solution content; this decision draws the same line through the artifacts, keeping template payload outside the specification of the ecosystem's own artifacts.
+* `cpt-frontx-principle-ownership-bounded-composition` — What a template contains internally falls inside the boundary that template declares and owns; this decision keeps that content out of the ecosystem tree rather than specifying it from outside the owner.
+* `cpt-frontx-adr-uniform-template-mechanism` — Templates are payload the tooling resolves, applies, and upgrades uniformly without branching on type; this decision is the artifact-side consequence: the ecosystem specifies the mechanism, not any template's internals.
+* `cpt-frontx-nfr-evolvability` — Retiring markers that bind nothing, and declining to fix contracts for code whose ownership is still moving, keeps the artifacts an accurate description of what is specified and leaves the template free to evolve on its own cadence.
--- a/template-shell/packages/react/CLAUDE.md
+++ b/template-shell/packages/react/CLAUDE.md
@@ -98,7 +98,7 @@ endpoint descriptors directly (e.g., `queryCache.get(service.getCurrentUser)`).
 
 ### SSE Streaming with Stream Descriptors
 
-Services declare SSE endpoints as stream descriptors. Components consume them via `useApiStream`, which manages the EventSource lifecycle automatically (connect on mount, disconnect on unmount).
+Services declare SSE endpoints as stream descriptors. Components consume them via `useApiStream`, which manages the connection lifecycle automatically (connect on mount, disconnect on unmount).
 
 ```tsx
 import { useApiStream, apiRegistry } from '@gears-frontx/react';
@@ -124,18 +124,38 @@ function ChatStream() {
   );
 
   if (status === 'connecting') return <Loading />;
-  if (error) return <Error error={error} />;
+  if (status === 'error') return <Error error={error} />;
 
   return <div>{data?.text}</div>;
 }
 ```
 
 `useApiStream` returns `{ data, events, status, error, disconnect }`:
-- `data` — latest event payload (always set in both modes)
+- `data` — latest event payload (set in both modes)
 - `events` — all received events when `mode: 'accumulate'`; empty array in `'latest'` mode
-- `status` — `'idle' | 'connecting' | 'connected' | 'disconnected' | 'error'`
-- `error` — connection error if any
-- `disconnect()` — manually close the connection
+- `status` — connection lifecycle state, `StreamStatus` (see below)
+- `error` — the last connect attempt's error; retained until the next connect attempt resets it, so check `status` rather than `error` truthiness
+- `disconnect()` — close the connection manually
+
+#### Status lifecycle
+
+`status` is a `StreamStatus`: `'idle' | 'connecting' | 'connected' | 'disconnected' | 'error'`.
+
+The TSDoc block on `useApiStream` in `src/hooks/useApiStream.ts` is the detailed and authoritative state-machine reference, including teardown edge cases. This guide summarizes consumer-facing usage and gating.
+
+`disconnect()` closes the connection and does not reconnect. There is no reconnect function: to open a new connection, change the descriptor key, change `mode`, or toggle `enabled` back to true.
+
+#### The `enabled` option
+
+`enabled` (default `true`) gates whether the hook connects at all. While it is false the hook never calls the descriptor's `connect()` and ordinarily holds `status: 'idle'` (calling `disconnect()` while disabled is the one exception — see the TSDoc). Flipping it to true re-runs the connect effect and starts a fresh connection; flipping it back to false tears the connection down and returns `status` to `'idle'`.
+
+Tearing down via `enabled` does not clear `data` or `events` — a descriptor-key or `mode` change, or re-enabling, does. A component reading `data` while the stream is disabled still sees the last event from the previous connection.
+
+Gating advice, since `'idle'` covers both "not started yet" and "deferred by `enabled`":
+
+- Gate the loading state on `'connecting'` alone. Treating `'idle'` as loading renders a permanent spinner for a stream deliberately held at `enabled: false`.
+- Handle `'idle'` as "not started / deferred". A component that both defers and shows loading should key its spinner off its own active flag, not off `'idle'`.
+- Check `status === 'error'` rather than the truthiness of `error` to render failures.
 
 ### Available Hooks
 
--- a/template-shell/packages/react/src/hooks/useApiStream.ts
+++ b/template-shell/packages/react/src/hooks/useApiStream.ts
@@ -1,10 +1,9 @@
 /**
  * useApiStream - Declarative SSE streaming hook
  *
- * Accepts a StreamDescriptor from @gears-frontx/api and manages the EventSource
- * lifecycle: connects on mount, disconnects on unmount/descriptor change.
- * Returns the latest event, accumulated events, connection status, and
- * a manual disconnect function.
+ * Accepts a StreamDescriptor from @gears-frontx/api and returns the latest
+ * event, accumulated events, connection status, and a manual disconnect
+ * function.
  *
  * @example
  * ```tsx
@@ -17,10 +16,6 @@
  * const { events, status } = useApiStream(service.messageStream, { mode: 'accumulate' });
  * ```
  */
-
-// @cpt-dod:cpt-frontx-dod-request-lifecycle-use-api-stream:p2
-// @cpt-flow:cpt-frontx-flow-request-lifecycle-use-api-stream:p2
-// @cpt-FEATURE:cpt-frontx-fr-sse-stream-descriptors:p3
 
 import { useEffect, useMemo, useRef, useState, useCallback } from 'react';
 import type { StreamDescriptor, StreamStatus } from '@gears-frontx/framework';
@@ -39,19 +34,76 @@ export interface ApiStreamOptions {
 
 /** Return type of useApiStream. */
 export interface ApiStreamResult<TEvent> {
-  /** Latest event payload (always set in both modes). */
+  /** Latest event payload, set in both modes — `undefined` until the first event and after each connect attempt resets it. */
   data: TEvent | undefined;
   /** All received events when `mode: 'accumulate'`; empty array in `'latest'` mode. */
   events: TEvent[];
   /** Connection lifecycle status. */
   status: StreamStatus;
-  /** Error if the connection failed. */
+  /** Last connect error — set when a connect attempt rejects, cleared when the next attempt starts. */
   error: Error | null;
   /** Manually close the connection. */
   disconnect: () => void;
 }
 
-// @cpt-begin:cpt-frontx-flow-request-lifecycle-use-api-stream:p2:inst-use-api-stream
+/**
+ * Manages the status lifecycle for a single SSE stream connection.
+ *
+ * **Status values**
+ * - `'idle'` — no connection is being attempted. This is the initial status
+ *   before the first connect effect runs, and it is ordinarily the status
+ *   while `enabled` is `false`. Consumers must not assume `'idle'` means "never
+ *   connected" — a stream that is disabled after having connected returns to
+ *   `'idle'` too, and `data`/`events` from the earlier connection are not
+ *   cleared by that transition alone (they are cleared by mode/key changes,
+ *   see below).
+ * - `'connecting'` — `descriptor.connect()` has been called and its promise
+ *   has not yet settled.
+ * - `'connected'` — set both when `connect()` resolves (unless a disconnect
+ *   was requested while it was pending) and, independently, whenever the
+ *   `onEvent` callback delivers a new event. Because every event re-asserts
+ *   `'connected'`, this status also serves as "still receiving events."
+ * - `'disconnected'` — reached via any of: the descriptor's `onComplete`
+ *   callback firing (the stream ended normally), `connect()` rejecting while
+ *   a disconnect was requested concurrently, or the manual `disconnect()`
+ *   function being called. `disconnect()` has no `enabled` guard: called
+ *   while the stream is disabled it still sets `'disconnected'`, which
+ *   persists until the next connect-effect run.
+ * - `'error'` — `connect()` rejected and no disconnect was requested while it
+ *   was pending. `error` holds the rejection, coerced to an `Error` if it
+ *   wasn't one already.
+ *
+ * **Transitions**
+ * - Mount / `enabled` becomes `true` / `descriptor.key` or `mode` changes:
+ *   `data`, `events`, and `error` are reset, then status goes to
+ *   `'connecting'` and `connect()` is called.
+ * - `enabled` becomes `false` (or starts `false`): status is forced to
+ *   `'idle'` and `connect()` is never called; no other field is reset by
+ *   this transition alone.
+ * - Unmount, `descriptor.key` change, `mode` change, or `enabled` flipping to
+ *   `false` while connected or connecting: the previous connection is torn
+ *   down via `descriptor.disconnect()` once its `connect()` promise settles
+ *   (or immediately skipped if it rejected). All four share the same cleanup
+ *   path: the effect's cleanup runs on every dependency change and on unmount.
+ * - Calling `disconnect()`: if a connection id is already resolved, it is
+ *   disconnected immediately and status becomes `'disconnected'`. If
+ *   `connect()` is still pending, the request is recorded and honored when
+ *   the promise settles — the newly resolved connection is torn down instead
+ *   of being adopted, and status is `'disconnected'`. One caveat:
+ *   `disconnect()` does not cancel the event callback, so a buffered or
+ *   in-flight event landing after the call re-asserts `'connected'` — and in
+ *   the pending-connect case the settle handler tears the connection down
+ *   without touching status, so that stray `'connected'` persists.
+ * - Stream completion while `connect()` is still pending: the current
+ *   implementation lets `onComplete` set `'disconnected'`, then the settle
+ *   handler adopts the connection id and re-asserts `'connected'` anyway.
+ *   This is known current behavior, not a durable guarantee: a pending
+ *   upstream derived-status rewrite is expected to remove this transition.
+ *
+ * `enabled` (default `true`) is the only way to defer connecting past mount;
+ * flipping it back to `true` re-runs the same connect sequence as a fresh
+ * mount would.
+ */
 export function useApiStream<TEvent>(
   descriptor: StreamDescriptor<TEvent>,
   options?: ApiStreamOptions,
@@ -169,4 +221,3 @@ export function useApiStream<TEvent>(
 
   return { data, events, status, error, disconnect };
 }
-// @cpt-end:cpt-frontx-flow-request-lifecycle-use-api-stream:p2:inst-use-api-stream
--- a/.cf-studio/config/artifacts.toml
+++ b/.cf-studio/config/artifacts.toml
@@ -21,7 +21,7 @@ reason = "In-package demos exercise a package's public surface rather than imple
 patterns = ["*/demo/*"]
 
 [[ignore]]
-reason = "template-shell/template-mfe are template territory (Phase 11 template-move, split in issue #470); code carries pre-redesign markers that reference archived IDs — no active FEATURE artifact backs these packages"
+reason = "These globs cover template territory — top-level directories carrying a frontx-template.json manifest — outside the ecosystem artifact universe (cpt-frontx-adr-template-territory-traceability); residue markers bind nothing and are removed as files are touched — no active FEATURE artifact backs these packages"
 patterns = ["template-shell/**", "template-mfe/**"]
 
 [[ignore]]
--- a/architecture/DESIGN.md
+++ b/architecture/DESIGN.md
@@ -82,6 +82,7 @@ Foundational:
 * `cpt-frontx-adr-artifact-versioning-and-distribution` — Distributes the ecosystem as independently published, per-concern, independently versioned artifacts.
 * `cpt-frontx-adr-core-package-boundaries` — Partitions the published libraries into boundary-governed concerns (runtime, type-system provider, protocol surface).
 * `cpt-frontx-adr-contract-schema-ownership` — Ends the circular DESIGN↔ADR schema deferral by assigning each owned contract's role to DESIGN, its decision rationale to the ADR, and its concrete field-level schema to the owning FEATURE.
+* `cpt-frontx-adr-template-territory-traceability` — Fixes this artifact tree's subject as the ecosystem's own artifacts, leaves template payload unspecified by it, and declares the `@cpt-` markers surviving in template territory to be non-authoritative residue removed as files are touched.
 
 Published libraries:
 
--- a/packages/api/src/types.ts
+++ b/packages/api/src/types.ts
@@ -442,7 +442,6 @@ export interface RestRequestContext {
   readonly signal?: AbortSignal;
 }
 
-// @cpt-FEATURE:cpt-frontx-dod-request-lifecycle-abort-signal:p1
 /**
  * REST Request Options
  * Per-request options for REST protocol HTTP methods.
@@ -925,7 +924,6 @@ export interface MutationDescriptor<TData, TVariables> {
 // Stream Descriptor Types
 // ============================================================================
 
-// @cpt-FEATURE:cpt-frontx-fr-sse-stream-descriptors:p2
 /**
  * Stream connection status.
  */
--- a/scripts/template-discovery.test.mjs
+++ b/scripts/template-discovery.test.mjs
@@ -95,3 +95,28 @@ describe('MANIFEST_FILENAME sync guard', () => {
     expect(MANIFEST_FILENAME).toBe(match[1]);
   });
 });
+
+// Same drift class, different surface: template territory is defined by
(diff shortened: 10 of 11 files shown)


tscbmstubp:
## Coordination note: relationship to #495 / PR #517 (ADR-0033)

**Numbering.** This PR adds ADR-**0034**, deliberately skipping 0033: that number is claimed by PR #517's `arch/495` branch (`0033-ecosystem-layer-partition-and-artifact-federation.md`). The gap on `develop` closes when #517 merges. Durable identity is the `cpt-frontx-adr-*` ID, so the filename gap is harmless either way.

**Merge collision (one line).** Both this PR and #517 insert one bullet into the same "Foundational" ADR list in `architecture/DESIGN.md` §1.2. Whichever merges second will likely hit a trivial one-line conflict there — resolution is "keep both lines."

**Substance — no contradiction, by design.** ADR-0033 makes templates a layer whose members own their own artifacts; ADR-0034 records that *today* the ecosystem tree does not specify template internals and no template-owned tree exists yet. These are the target state and the interim state of the same position:

- ADR-0033 itself says the member artifact chain binds only once a member's registry node declares its required artifact kinds — `template-shell` has no such node yet, so ADR-0034 accurately describes the current state.
- ADR-0034 explicitly **defers** (does not reject) a template-owned artifact tree, and states it is *superseded rather than amended* once template ownership settles — at which point markers may be reintroduced against IDs from that tree and the registry's scanning scope revisited. That is exactly the slot #495's "templates own their product-specific specifications" fills.
- The `@cpt-` markers removed here pointed at archived *ecosystem* IDs; under ADR-0033's model they would be re-pointed at the template's own tree anyway, so removing them now loses nothing.

architecture/ADR/0033-template-territory-traceability.md
```diff
@@ -0,0 +1,126 @@
...
+  - [Template territory is outside the ecosystem artifact universe; markers there are non-authoritative residue, removed as files are touched](#template-territory-is-outside-the-ecosystem-artifact-universe-markers-there-are-non-authoritative-residue-removed-as-files-are-touched)
+  - [Extend the ecosystem artifact tree to cover template internals](#extend-the-ecosystem-artifact-tree-to-cover-template-internals)
+  - [Keep the residue markers and rely on the scan exclusion alone](#keep-the-residue-markers-and-rely-on-the-scan-exclusion-alone)
+  - [Give template territory its own artifact tree now](#give-template-territory-its-own-artifact-tree-now)
+- [More Information](#more-information)
+- [Traceability](#traceability)
+
+<!-- /toc -->
+
+**ID**: `cpt-frontx-adr-template-territory-traceability`
```
itechmeat, line 26: Gera's direction on #517 (issuecomment-5177483406) was "we need not a new ADR, but rather a full refactoring of PRD, DESIGN, DECOMPOSITION and FEATURE documents", and #517 responded by archiving its ADR-0033 - which also makes this PR's "0033 is claimed by the in-flight #495 branch" numbering rationale stale. Should this decision be folded into the #495 artifact-federation restructuring instead of minting a new root ADR? Worth aligning with @GeraBart before merge - the decision content itself reads like the interim state of exactly what #495 is building.

architecture/ADR/0033-template-territory-traceability.md
```diff
@@ -0,0 +1,126 @@
...
+
+The `@cpt-` markers presently in template territory are **residue of the archived pre-redesign generation and bind nothing**: they are not authoritative, they are not evidence that a contract was specified, and they must not be read as traceability. **No new markers are added to template territory**, and residue markers are **removed from a file when that file is otherwise modified** — the rule is uniform, needs no separate sweep, and leaves no half-covered island. Where template code has behaviour a consumer depends on — a hook's observable status lifecycle, for instance — that behaviour is documented **at code altitude, in the code and its package documentation**, which is the reader's location and carries no claim of upstream specification.
```
itechmeat: This rule conflicts with two live instruction surfaces that were not updated: `.cf-studio/config/AGENTS.md` still says "keep @cpt- traceability markers intact" (line 33) and FX-002 at line 56 treats marker removal as a HIGH review finding, with no template carve-out, and the root `CLAUDE.md` line 11 states the marker chain unqualified. An agent editing a template file would preserve the markers this ADR orders removed, and a reviewer would flag their removal. Should those surfaces gain the template-territory exception in this same PR?

architecture/ADR/0033-template-territory-traceability.md
```diff
@@ -0,0 +1,126 @@
...
+* Bad, because template territory keeps residue markers until each file is otherwise touched, so the payload stays in a mixed state for as long as that takes.
+* Bad, because behaviour documented only at code altitude is easier to change without review than behaviour fixed by a numbered instruction, so a consumer-visible change in template payload rests on code review alone.
+
+### Confirmation
+
+Compliance is confirmed by the artifact registry and by review. The registry admits only the ecosystem's own source paths to traceability scanning and excludes the template subtrees, so no template file is scanned for markers and no marker there is validated; `cfs validate` passing therefore never depends on template payload. Review confirms that no FEATURE in this tree names a path under a template subtree, that a change adding a `@cpt-` marker to template territory is refused, and that a change modifying an already-marked template file removes the residue markers in the region it touches and documents any consumer-visible behaviour at code altitude in the same change.
```
itechmeat: The removal rule is stated two incompatible ways: the Decision Outcome (line 53, same as the option text at line 42) removes markers "from a file when that file is otherwise modified", while this Confirmation line scopes removal to "the region it touches". On the first partial edit of a heavily marked file (auth.ts carries 69 residue markers) the two readings produce opposite review outcomes - which one is the decision?

architecture/ADR/0033-template-territory-traceability.md
```diff
@@ -0,0 +1,126 @@
...
+
+Author a separate nested artifact tree owned by the template and re-point the payload's markers at its IDs.
+
+* Good, because it would specify template behaviour without mixing subjects, and would give the markers a legitimate home.
+* Neutral, because it is the natural destination if template internals ever need specified contracts.
+* Bad, because it is premature while template ownership is still moving, and it is a large authoring commitment ahead of any demonstrated need.
+
+## More Information
+
+**Deferred deliberately**: whether template territory gets an artifact tree of its own is **not decided here**. It is deferred until template ownership has settled and template internals present a contract a consumer depends on across versions — at which point this decision is superseded rather than amended, markers may be reintroduced against IDs from that tree, and the registry's scanning scope is revisited in the same step. Deciding it now would fix contracts for code whose ownership is in motion.
```
itechmeat: "Superseded rather than amended" pre-commits to the mechanism the house artifact rules forbid (never supersede an ADR; a changed decision is rewritten in place), and the next paragraph's "the immediate instance that prompted the record" sentence is change history inside an artifact. Could both be reworded - "rewritten in place once template ownership settles", and the prompting-instance sentence dropped? The deferral itself is worth keeping as is.

architecture/ADR/0033-template-territory-traceability.md
```diff
@@ -0,0 +1,126 @@
...
+
+The scope of this decision is which artifact tree specifies template-territory code and the standing of `@cpt-` markers found there. It does not decide the boundary a template declares (`cpt-frontx-adr-template-ownership-boundary-declaration`), how the CLI applies or upgrades a template (`cpt-frontx-adr-uniform-template-mechanism`, `cpt-frontx-adr-project-upgrade-mechanism`), nor whether the template will eventually carry an artifact tree of its own — that is deliberately deferred (see More Information).
+
+### Consequences
+
+* Good, because each artifact tree keeps one subject: the ecosystem's artifacts are described by the ecosystem's artifacts, and template payload is not smuggled into them.
+* Good, because a `@cpt-` marker recovers its meaning — inside the specified universe it names a live instruction, and it no longer appears in a universe where it can name nothing.
```
itechmeat: This claim is not true at head: `packages/api/src/types.ts` still carries `@cpt-FEATURE` markers naming `cpt-frontx-dod-request-lifecycle-abort-signal` (line 445) and `cpt-frontx-fr-sse-stream-descriptors` (line 928) - the same archived request-lifecycle generation - inside the scanned universe, not ignored, with `cfs validate` green. Either those two comment lines leave in this PR, or this sentence needs softening plus a follow-up pointer - which do you prefer?

template-shell/packages/react/CLAUDE.md
```diff
@@ -123,19 +123,61 @@ function ChatStream() {
...
+- `data` — latest event payload (set in both modes)
 - `events` — all received events when `mode: 'accumulate'`; empty array in `'latest'` mode
-- `status` — `'idle' | 'connecting' | 'connected' | 'disconnected' | 'error'`
-- `error` — connection error if any
-- `disconnect()` — manually close the connection
+- `status` — connection lifecycle state, `StreamStatus` (see below)
+- `error` — the error from a failed connect attempt; `null` otherwise
+- `disconnect()` — close the connection manually
+
+#### Status lifecycle
```
itechmeat, line 140: [Not a blocker] The status lifecycle now lives twice - this section and the new TSDoc in `useApiStream.ts` - and the two already disagree within this same diff: the TSDoc lists "connect() rejecting while a disconnect was requested" as a route into `'disconnected'`, while this list's `'connecting'` → `'disconnected'` bullet covers only the promise-resolve path. Could one surface own the state machine and the other point to it (e.g. contract in the TSDoc, gating advice here)?

template-shell/packages/react/CLAUDE.md
```diff
@@ -123,19 +123,61 @@ function ChatStream() {
...
+
+`status` is a `StreamStatus`: `'idle' | 'connecting' | 'connected' | 'disconnected' | 'error'`.
+
+| Status | Meaning |
+|--------|---------|
+| `'idle'` | No connect attempt is in flight — the initial state, and the state while `enabled` is false, including after a previous connection was torn down |
```
itechmeat: [Not a blocker] Everything "initial state is idle" here becomes wrong the moment #491 lands (it initializes to `enabled ? 'connecting' : 'idle'` and derives the returned status), and the only warning lives in this PR's description. Would an ordering agreement with #491, or a short in-file note next to this section, be safer than a description-level promise?

template-shell/packages/react/CLAUDE.md
```diff
@@ -123,19 +123,61 @@ function ChatStream() {
     { enabled: active }
   );
 
-  if (status === 'connecting') return <Loading />;
-  if (error) return <Error error={error} />;
+  // This stream is always enabled, so 'idle' only occurs on the first render
+  // pass — safe to fold into the loading branch. See the gating note below for
+  // streams that defer via `enabled`.
+  if (status === 'idle' || status === 'connecting') return <Loading />;
```
itechmeat: [Not a blocker] Before this PR the example gated loading on `'connecting'` alone - already compliant with the gating advice this same PR adds at line 176 - and now it demonstrates the discouraged pattern and needs two caveats (lines 126-128 and 180) telling readers not to copy it. Reverting this guard to `status === 'connecting'` deletes both caveats and the contradiction; the `status === 'error'` fix on the next line is a clear improvement and worth keeping.

template-shell/packages/react/src/hooks/useApiStream.ts
```diff
@@ -50,7 +46,53 @@ export interface ApiStreamResult<TEvent> {
...
+ *   `data`, `events`, and `error` are reset, then status goes to
+ *   `'connecting'` and `connect()` is called.
+ * - `enabled` becomes `false` (or starts `false`): status is forced to
+ *   `'idle'` and `connect()` is never called; no other field is reset by
+ *   this transition alone.
+ * - Unmount, `descriptor.key` change, `mode` change, or `enabled` flipping to
+ *   `false` while connected or connecting: the previous connection is torn
+ *   down via `descriptor.disconnect()` once its `connect()` promise settles
+ *   (or immediately skipped if it rejected). All four share the same cleanup
+ *   path because all four are the effect's dependencies.
```
itechmeat: [Not a blocker] "All four are the effect's dependencies" is not accurate - the dependency array is `[descriptorKey, enabled, mode]`; unmount reaches the same cleanup through React's unmount semantics, not the dep list. The observable claim (all four tear down identically) is the useful part - suggest dropping the false "because" clause.

template-shell/packages/react/CLAUDE.md
```diff
@@ -123,19 +123,61 @@ function ChatStream() {
...
+- **`'connected'` → `'disconnected'`** — the stream invokes its close callback, or the consumer calls `disconnect()`.
+- **any → `'idle'`** — `enabled` flips to false.
+
+`disconnect()` sets `'disconnected'` and does not reconnect. There is no reconnect function: to open a new connection, change the descriptor key, change `mode`, or toggle `enabled` back to true.
+
+Each connect attempt clears `data`, `events`, and `error` before setting `'connecting'`. `data` and `events` are also cleared when the descriptor key or `mode` changes.
+
+#### The `enabled` option
+
+`enabled` (default `true`) gates whether the hook connects at all. While it is false the hook holds `status: 'idle'` and never calls the descriptor's `connect()`. Flipping it to true re-runs the connect effect and starts a fresh connection; flipping it back to false tears the connection down and returns `status` to `'idle'`.
```
itechmeat: [Not a blocker] "While it is false the hook holds `status: 'idle'`" is stated as an invariant, but `disconnect()` has no `enabled` guard - calling it while the stream is disabled sets `'disconnected'`, which then persists because the connect effect does not re-run while disabled. Narrow edge, but worth softer wording here (and in the TSDoc's "it is also the status while enabled is false") or an explicit note.

template-shell/packages/react/CLAUDE.md
```diff
@@ -123,19 +123,61 @@ function ChatStream() {
...
+| `'disconnected'` | The connection was closed — by the stream itself, or by calling `disconnect()` |
+| `'error'` | The connect attempt rejected; `error` holds the reason |
+
+Transitions:
+
+- **`'idle'` → `'connecting'`** — the connect effect runs. With `enabled: true` this happens right after mount, so `'idle'` is only observable for the first render pass. It also re-runs whenever the descriptor key, `mode`, or `enabled` changes.
+- **`'connecting'` → `'connected'`** — the connect promise resolves, or the first event arrives.
+- **`'connecting'` → `'error'`** — the connect promise rejects.
+- **`'connecting'` → `'disconnected'`** — `disconnect()` is called while the connect promise is still pending; the connection is torn down as soon as it resolves.
+- **`'connected'` → `'disconnected'`** — the stream invokes its close callback, or the consumer calls `disconnect()`.
```
itechmeat: [Not a blocker] Nothing documents leaving `'disconnected'` without a key/mode/enabled change, but the code allows it spontaneously: if the stream completes (`onComplete`) before `connect()` resolves, the promise `.then` still sets `'connected'` and adopts the id (the adoption refusal is keyed to a requested `disconnect()`, which `onComplete` is not), and any late event re-asserts `'connected'`. Is that ordering excluded by the descriptor contract, or a transition the lifecycle docs should carry?

template-shell/packages/react/CLAUDE.md
```diff
@@ -123,19 +123,61 @@ function ChatStream() {
...
+|--------|---------|
+| `'idle'` | No connect attempt is in flight — the initial state, and the state while `enabled` is false, including after a previous connection was torn down |
+| `'connecting'` | A connect attempt is in flight; no events received yet |
+| `'connected'` | The connection is open — set when the connect promise resolves or when the first event arrives, whichever happens first, and re-asserted by every subsequent event |
+| `'disconnected'` | The connection was closed — by the stream itself, or by calling `disconnect()` |
+| `'error'` | The connect attempt rejected; `error` holds the reason |
+
+Transitions:
+
+- **`'idle'` → `'connecting'`** — the connect effect runs. With `enabled: true` this happens right after mount, so `'idle'` is only observable for the first render pass. It also re-runs whenever the descriptor key, `mode`, or `enabled` changes.
```
itechmeat: [Not a blocker] The trailing sentence folds descriptor-key/mode re-runs into the `'idle'` → `'connecting'` row, but when those change while `'connected'`, `'disconnected'` or `'error'`, the observed transition starts from that state, not from `'idle'` - three transitions the list does not carry. Maybe a separate "any active state → `'connecting'` on key/mode change" bullet?

template-shell/packages/react/CLAUDE.md
```diff
@@ -123,19 +123,61 @@ function ChatStream() {
...
+  if (status === 'idle' || status === 'connecting') return <Loading />;
+  if (status === 'error') return <Error error={error} />;
 
   return <div>{data?.text}</div>;
 }
 ```
 
 `useApiStream` returns `{ data, events, status, error, disconnect }`:
-- `data` — latest event payload (always set in both modes)
+- `data` — latest event payload (set in both modes)
```
itechmeat, line 134: [Not a blocker] CLAUDE.md drops "always" from "latest event payload", but the `ApiStreamResult.data` member doc in the same file this PR rewrites (useApiStream.ts line 37) still says "always set in both modes" - false until the first event and after every reconnect reset. Same PR, same field, two wordings; worth fixing the member doc too.

itechmeat (changes requested):
Careful, well-verified work - the zero-logic-change claim holds mechanically and the lifecycle documentation is accurate on almost every point I checked. My concern is the vehicle rather than the content: this mints a new root ADR right after the maintainer asked on #517 for no new ADRs and for the artifact refactoring instead, and the ADR text itself needs a consistency pass - its removal rule is stated two incompatible ways, and two live instruction surfaces still say the opposite of what it decides. With the direction question settled and a handful of small doc fixes, this is a solid merge.

architecture/ADR/0033-template-territory-traceability.md
tscbmstubp: Renumbered to 0033 — the skip rationale died when #517 archived its ADR; references use the ID slug, so only the filename moved. On folding into #495: kept standalone for now — #517 deliberately carries no ADR, and this decision (markers-as-residue, code-altitude docs) is not made anywhere in it; it records the interim state whose target #517's LAYER constraints build. @GeraBart — the shape question (standalone ADR vs folding into the #495 restructuring) is yours to confirm; happy to move it if you prefer the fold.

architecture/ADR/0033-template-territory-traceability.md
tscbmstubp: Done — both AGENTS.md lines and the root CLAUDE.md now carry the template-territory carve-out: add no markers there, and a change modifying a template file strips every marker it carries as its own commit; FX-002 inverts in template territory (residue surviving a modified file is the defect, not its removal).

architecture/ADR/0033-template-territory-traceability.md
tscbmstubp: Whole file is the decision. The Confirmation now matches the Decision Outcome, plus one addition: the strip travels as its own commit inside the modifying change, so the mechanical deletions review separately from the substantive diff. For the auth.ts case: the first edit takes all 69, in a dedicated commit.

architecture/ADR/0033-template-territory-traceability.md
tscbmstubp: Both done: 'rewritten in place' replaces the superseding mechanism, and the prompting-instance sentence is dropped — the Decision Outcome keeps the hook-lifecycle example in timeless form.

architecture/ADR/0033-template-territory-traceability.md
tscbmstubp: Took the first option — both comment lines are gone (comment-only diff). Also swept the scanned universe for the archived generation: types.ts was the last carrier, so the consequence now holds as written.

template-shell/packages/react/CLAUDE.md
tscbmstubp: Done — the TSDoc is the single owner of the state machine; CLAUDE.md keeps the returns list, the enabled section, and the gating advice, and points at the TSDoc instead of restating the machine. The 'disconnected'-route disagreement dissolved with the duplicate.

template-shell/packages/react/CLAUDE.md
tscbmstubp: Went with the in-file note, pinned on the exact initializer line #491 rewrites — whichever PR lands second, the rebase collides with the reminder at the moment it applies. After the example fix, CLAUDE.md carries no initial-state claims at all, so the reminder points only at the TSDoc.

template-shell/packages/react/CLAUDE.md
tscbmstubp: Reverted — the guard is `status === 'connecting'` again, both caveats deleted, the `status === 'error'` line kept. Bonus: the caveats carried CLAUDE.md's last 'idle only occurs on the first render pass' claims, so this also shrank #491's blast radius.

template-shell/packages/react/src/hooks/useApiStream.ts
tscbmstubp: Fixed — kept the observable claim, corrected the rationale: the effect's cleanup runs on every dependency change and on unmount.

template-shell/packages/react/CLAUDE.md
tscbmstubp: Documented: the TSDoc's 'idle' bullet softens to 'ordinarily', the 'disconnected' bullet names the disconnect-while-disabled route and its persistence, and CLAUDE.md's enabled paragraph drops the invariant phrasing. Behaviour left as-is — a guard belongs to #491's derived-status rewrite, not a docs PR.

template-shell/packages/react/CLAUDE.md
tscbmstubp: Not excluded — StreamDescriptor.connect() documents onComplete with no ordering guarantee against the promise, so it is a real transition of today's implementation. Documented in the TSDoc's transitions ('disconnected' → 'connected' with no consumer action on a short-lived stream). The behavioural fix (refuse adoption after completion) fits #491's derived-status rewrite.

template-shell/packages/react/CLAUDE.md
tscbmstubp: Resolved structurally by the ownership change from the thread above: the from-state-labelled list this targeted is deleted, and the surviving TSDoc states the rule trigger-based (mount / enabled true / key or mode change → 'connecting'), which holds from any starting state.

template-shell/packages/react/CLAUDE.md
tscbmstubp: Fixed — 'Latest event payload, set in both modes — undefined until the first event and after each connect attempt resets it.'

architecture/ADR/0033-template-territory-traceability.md
```diff
@@ -0,0 +1,126 @@
...
+
+### Confirmation
+
+Compliance is confirmed by the artifact registry and by review. The registry admits only the ecosystem's own source paths to traceability scanning and excludes the template subtrees, so no template file is scanned for markers and no marker there is validated; `cfs validate` passing therefore never depends on template payload. Review confirms that no FEATURE in this tree names a path under a template subtree, that a change adding a `@cpt-` marker to template territory is refused, and that a change modifying an already-marked template file removes the residue markers in the region it touches and documents any consumer-visible behaviour at code altitude in the same change.
+
+## Pros and Cons of the Options
+
+### Template territory is outside the ecosystem artifact universe; markers there are non-authoritative residue, removed as files are touched
+
+The ecosystem tree specifies ecosystem artifacts only; template payload is unspecified by it, residue markers bind nothing and leave as files are touched, and template behaviour is documented in the code.
```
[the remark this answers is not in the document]
tscbmstubp: Reworded to match the decided whole-file rule as well: 'residue markers bind nothing and are removed as files are modified'.

template-shell/packages/react/CLAUDE.md
```diff
@@ -123,19 +123,61 @@ function ChatStream() {
...
 - `events` — all received events when `mode: 'accumulate'`; empty array in `'latest'` mode
-- `status` — `'idle' | 'connecting' | 'connected' | 'disconnected' | 'error'`
-- `error` — connection error if any
-- `disconnect()` — manually close the connection
+- `status` — connection lifecycle state, `StreamStatus` (see below)
+- `error` — the error from a failed connect attempt; `null` otherwise
```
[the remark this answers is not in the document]
tscbmstubp: Fixed in both surfaces: CLAUDE.md now describes `error` as the last connect attempt's error, retained until the next attempt resets it (check `status`, not truthiness), and the `ApiStreamResult.error` member doc says when it is set and when it clears.

.cf-studio/config/AGENTS.md
```diff
@@ -30,7 +30,8 @@ DO:
   - LOAD `architecture/DESIGN.md` §1.3 (Architecture Layers) and §2.2 (Constraints) before editing any code
   - LOAD `packages/cyber-pilot-kit-frontx/guidelines/ecosystem-boundaries.md` and apply the boundary constraints for every touched package
   - LOAD the governing `architecture/features/*/FEATURE.md` before changing code that carries its `@cpt-` markers
-  - RUN keep `@cpt-` traceability markers intact and aligned with the FEATURE instructions they cite
+  - RUN keep `@cpt-` traceability markers intact and aligned with the FEATURE instructions they cite — in ecosystem code only
+  - RUN treat `@cpt-` markers in template territory (`template-shell/`, `template-mfe/`) as non-authoritative residue: add none, and when a change modifies a template file, remove every marker that file carries as its own commit inside the change (`cpt-frontx-adr-template-territory-traceability`)
```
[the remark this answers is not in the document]
tscbmstubp: Both lines qualified: the FEATURE-loading prerequisite is scoped to ecosystem code, and the stop-and-ask rule now carries the template-territory exception — no FEATUREs there by decision, consumer-visible behaviour documented at code altitude (`cpt-frontx-adr-template-territory-traceability`).

architecture/ADR/0033-template-territory-traceability.md
```diff
@@ -0,0 +1,126 @@
...
+
+The `@cpt-` markers presently in template territory are **residue of the archived pre-redesign generation and bind nothing**: they are not authoritative, they are not evidence that a contract was specified, and they must not be read as traceability. **No new markers are added to template territory**, and when a template file is otherwise modified, **every residue marker that file carries is removed** — the rule is uniform, needs no separate sweep, and leaves no half-covered file behind. The strip is mechanical, so it travels as its own commit inside the modifying change, keeping the marker deletions separable from the substantive diff in review. Where template code has behaviour a consumer depends on — a hook's observable status lifecycle, for instance — that behaviour is documented **at code altitude, in the code and its package documentation**, which is the reader's location and carries no claim of upstream specification.
```
itechmeat: This new commit-shape rule seems to work against itself: AGENTS.md line 34 renders it as "remove every marker that file carries as its own commit inside the change", which the agents executing that file can read as one commit per marker; this PR itself lands the strips and the TSDoc rewrite in one squashed commit; and FX-002, which the Confirmation delegates enforcement to, never checks commit separation. Drop it to a review-time preference, or keep it with unambiguous wording ("as a single strip commit") plus an FX-002 leg?

architecture/ADR/0033-template-territory-traceability.md
```diff
@@ -0,0 +1,126 @@
...
+
+Author a separate nested artifact tree owned by the template and re-point the payload's markers at its IDs.
+
+* Good, because it would specify template behaviour without mixing subjects, and would give the markers a legitimate home.
+* Neutral, because it is the natural destination if template internals ever need specified contracts.
+* Bad, because it is premature while template ownership is still moving, and it is a large authoring commitment ahead of any demonstrated need.
+
+## More Information
+
+**Deferred deliberately**: whether template territory gets an artifact tree of its own is **not decided here**. It is deferred until template ownership has settled and template internals present a contract a consumer depends on across versions — at which point this decision is rewritten in place, markers may be reintroduced against IDs from that tree, and the registry's scanning scope is revisited in the same step. Deciding it now would fix contracts for code whose ownership is in motion.
```
itechmeat: "Rewritten in place" - my own round-1 ask, so this one is on me - now contradicts ADR-0027, which argues normatively from "an accepted ADR is immutable" (0027 lines 47 and 90). Two accepted ADRs codify opposite mutation models after this PR. Could this sentence be worded so 0027 survives, e.g. "this decision is revisited together with the template-tree decision", or a reconciling note added?

template-shell/packages/react/CLAUDE.md
```diff
@@ -124,18 +124,38 @@ function ChatStream() {
...
-- `disconnect()` — manually close the connection
+- `status` — connection lifecycle state, `StreamStatus` (see below)
+- `error` — the last connect attempt's error; retained until the next connect attempt resets it, so check `status` rather than `error` truthiness
+- `disconnect()` — close the connection manually
+
+#### Status lifecycle
+
+`status` is a `StreamStatus`: `'idle' | 'connecting' | 'connected' | 'disconnected' | 'error'`.
+
+The state machine — what each status means and every transition between them, including the teardown edge cases — is specified in the TSDoc block on `useApiStream` in `src/hooks/useApiStream.ts`. That TSDoc is the contract and the single owner; this file does not restate the machine and gives only the gating advice below.
```
itechmeat: [Not a blocker] "This file does not restate the machine and gives only the gating advice below" is contradicted a few lines down: 146, 150 and 152 restate the reconnect rule, three `enabled` transitions and the field-persistence behaviour the TSDoc already owns. Either trim those to pointers or drop the "only" claim - as written the sentence is false about its own file.

.cf-studio/config/AGENTS.md
```diff
@@ -53,7 +54,7 @@ WHEN:
 DO:
   - LOAD `architecture/DESIGN.md` §2.2 and `packages/cyber-pilot-kit-frontx/guidelines/ecosystem-boundaries.md`; NEVER review from memory of their rules
   - RUN FX-001 Constraints: every changed package complies with its named DESIGN §2.2 constraints and its package boundary in `ecosystem-boundaries.md`; ambiguous ownership was raised, not guessed
-  - RUN FX-002 Traceability: changed code keeps `@cpt-` markers consistent with the governing FEATURE, and `cfs validate` passes
+  - RUN FX-002 Traceability: changed ecosystem code keeps `@cpt-` markers consistent with the governing FEATURE, and `cfs validate` passes; in template territory (`template-shell/`, `template-mfe/`) the finding inverts — residue markers surviving in a modified file are the defect, not their removal (`cpt-frontx-adr-template-territory-traceability`)
```
itechmeat: [Not a blocker] This line re-derives the ADR's inversion, line 34 restates its removal rule and line 39 its no-FEATUREs decision - all a few lines above this file's own "NEVER restate architecture rules in artifacts or prompts" (line 40). Could these point at `cpt-frontx-adr-template-territory-traceability` with one clause each instead of re-deriving it? Seven surfaces now carry this rule in prose.

architecture/ADR/0033-template-territory-traceability.md
```diff
@@ -0,0 +1,126 @@
...
+
+The scope of this decision is which artifact tree specifies template-territory code and the standing of `@cpt-` markers found there. It does not decide the boundary a template declares (`cpt-frontx-adr-template-ownership-boundary-declaration`), how the CLI applies or upgrades a template (`cpt-frontx-adr-uniform-template-mechanism`, `cpt-frontx-adr-project-upgrade-mechanism`), nor whether the template will eventually carry an artifact tree of its own — that is deliberately deferred (see More Information).
+
+### Consequences
+
+* Good, because each artifact tree keeps one subject: the ecosystem's artifacts are described by the ecosystem's artifacts, and template payload is not smuggled into them.
+* Good, because a `@cpt-` marker recovers its meaning — inside the specified universe it names a live instruction, and it no longer appears in a universe where it can name nothing.
```
itechmeat: [Not a blocker] All six surviving dangling IDs live outside the scanned universe (test-file and build-config ignore patterns), so the thread's "types.ts was the last carrier" holds as you scoped it - the unscoped sentence is this one. `cpt-frontx-dod-request-lifecycle-abort-signal`, the exact ID stripped here, survives 13 times in `packages/api/src/__tests__/abortSignal.test.ts`, and this ADR's own driver says a marker naming nothing is a false claim "whether or not any validator reads it". Scope this bullet to scanned source, or file a follow-up to retire the six?

template-shell/packages/react/src/hooks/useApiStream.ts
```diff
@@ -38,19 +34,73 @@ export interface ApiStreamOptions {
...
+ *   `false` while connected or connecting: the previous connection is torn
+ *   down via `descriptor.disconnect()` once its `connect()` promise settles
+ *   (or immediately skipped if it rejected). All four share the same cleanup
+ *   path: the effect's cleanup runs on every dependency change and on unmount.
+ * - Calling `disconnect()`: if a connection id is already resolved, it is
+ *   disconnected immediately and status becomes `'disconnected'`. If
+ *   `connect()` is still pending, the request is recorded and honored when
+ *   the promise settles — the newly resolved connection is torn down instead
+ *   of being adopted, and status is `'disconnected'`.
+ * - Stream completion while `connect()` is still pending: `onComplete` sets
```
itechmeat: [Not a blocker] This bullet documents what reads like a defect: a completed stream re-asserting `'connected'`. Is documenting it the decision, or the interim until adoption is fixed (a completed-while-pending flag in the settle handler, the way `disconnectRequestedRef` already works)? If it is known-bug-until-#491, an issue link in this bullet would keep it from fossilizing into contract.

.cf-studio/config/AGENTS.md
```diff
@@ -29,13 +29,14 @@ WHEN:
...
+  - LOAD the governing `architecture/features/*/FEATURE.md` before changing ecosystem code that carries its `@cpt-` markers
+  - RUN keep `@cpt-` traceability markers intact and aligned with the FEATURE instructions they cite — in ecosystem code only
+  - RUN treat `@cpt-` markers in template territory (`template-shell/`, `template-mfe/`) as non-authoritative residue: add none, and when a change modifies a template file, remove every marker that file carries as its own commit inside the change (`cpt-frontx-adr-template-territory-traceability`)
 
 RULES:
   - ALWAYS read the named DESIGN constraint (MFES-*, GTS-PLUGIN-*, API-*, CLI-*, KIT-*) before editing code it governs; NEVER code from assumed rules
   - ALWAYS stop and ask when a change would cross a DESIGN §2.2 constraint or a package boundary from `ecosystem-boundaries.md`
-  - ALWAYS stop and ask when no FEATURE covers behaviour the change introduces
+  - ALWAYS stop and ask when no FEATURE covers behaviour the change introduces in ecosystem code — template territory has no FEATUREs by decision; document consumer-visible behaviour at code altitude instead (`cpt-frontx-adr-template-territory-traceability`)
```
itechmeat: [Not a blocker] "Template territory has no FEATUREs by decision" overstates the ADR, which forbids only ecosystem FEATUREs and explicitly leaves a template-owned artifact tree open. "No ecosystem FEATURE covers template territory by decision" matches it.

template-shell/packages/react/src/hooks/useApiStream.ts
```diff
@@ -60,6 +110,8 @@ export function useApiStream<TEvent>(
 
   const [data, setData] = useState<TEvent | undefined>(undefined);
   const [events, setEvents] = useState<TEvent[]>([]);
+  // #491 replaces this initializer with `enabled ? 'connecting' : 'idle'` — when it
```
itechmeat: [Not a blocker] This file is template payload - the comment ships into every scaffolded repository, whose owners cannot act on gears-frontx#491. Maybe phrase the constraint without the issue number ("a pending upstream change replaces this initializer with `enabled ? 'connecting' : 'idle'` - update the TSDoc's initial-state claims with it") so it survives the copy?

CLAUDE.md
```diff
@@ -8,6 +8,6 @@ ALWAYS resolve and enforce prerequisites of skills/workflows/commands BEFORE app
 
 AI tooling and FrontX development guidelines are provided by the Constructor Studio kit under `.cf-studio/`.
 
-Treat `architecture/` as the authority on what the system is and why: PRD for intent, DESIGN for structure, ADRs for decisions, and each FEATURE for the behaviour its numbered instructions specify. Code carries `@cpt-` markers back to those instructions; `cfs validate` checks that the chain holds.
+Treat `architecture/` as the authority on what the system is and why: PRD for intent, DESIGN for structure, ADRs for decisions, and each FEATURE for the behaviour its numbered instructions specify. Ecosystem code carries `@cpt-` markers back to those instructions; `cfs validate` checks that the chain holds. Template subtrees (`template-shell/`, `template-mfe/`) sit outside this chain: markers found there are non-authoritative residue being retired as files change (ADR 0033).
```
itechmeat: [Not a blocker] "(ADR 0033)" is the only number-shaped ADR citation this change set leaves - every other surface cites `cpt-frontx-adr-template-territory-traceability` - and the number already moved once inside this very PR. Worth using the slug here too.

template-shell/packages/react/src/hooks/useApiStream.ts
```diff
@@ -1,7 +1,7 @@
 /**
  * useApiStream - Declarative SSE streaming hook
  *
- * Accepts a StreamDescriptor from @gears-frontx/api and manages the EventSource
+ * Accepts a StreamDescriptor from @gears-frontx/api and manages the connection
```
itechmeat: [Not a blocker] The module banner (lines 1-19) and the new TSDoc now both summarize the connection lifecycle. If the TSDoc is the single owner, the banner could shrink to the module role and the usage example, dropping its own lifecycle summary.

itechmeat (changes requested):
Strong follow-through - all 12 fix threads verified genuinely landed at head, the renumbering is complete, and every lifecycle claim in the TSDoc checks out against the implementation. Two sentence-level items keep me from approving: the new commit-shape rule needs one more decision (ambiguous as written in AGENTS.md, this PR itself does not follow it, and nothing enforces it), and one consistency question my own round-1 ask created ("rewritten in place" vs ADR-0027's immutability premise). The direction question stays with Gera as agreed.

architecture/ADR/0033-template-territory-traceability.md
tscbmstubp: Addressed in ae015508: I dropped the hard commit-shape requirement. The ADR now states the review-time outcome directly: marker-only deletions should remain easy to distinguish from substantive changes, without prescribing a commit topology that a squash would invalidate.

architecture/ADR/0033-template-territory-traceability.md
tscbmstubp: Addressed in ae015508: the ADR no longer says it will be rewritten in place. It now says the decision must be revisited through the repository ADR lifecycle, preserving ADR-0027 while leaving the future template-tree decision open.

template-shell/packages/react/CLAUDE.md
tscbmstubp: Addressed in ae015508: removed the false only claim. The guide now calls the hook TSDoc the detailed, authoritative state-machine reference and accurately describes this section as a consumer-facing usage and gating summary.

.cf-studio/config/AGENTS.md
tscbmstubp: Addressed in ae015508: the instruction surfaces now load the ADR and refer to its marker policy instead of re-deriving the rule. FX-002 likewise delegates template-territory review to cpt-frontx-adr-template-territory-traceability.

architecture/ADR/0033-template-territory-traceability.md
tscbmstubp: Addressed in ae015508: scoped the consequence to the scanned specified universe and explicitly acknowledged ignored test/build residue outside that universe, avoiding the broader claim that all dangling markers are gone.

template-shell/packages/react/src/hooks/useApiStream.ts
tscbmstubp: Addressed in ae015508: the TSDoc now labels this as known current behavior rather than a durable guarantee and says a pending upstream derived-status rewrite is expected to remove it, without embedding a repository issue link in shipped template payload.

.cf-studio/config/AGENTS.md
tscbmstubp: Addressed in ae015508: changed the wording to no ecosystem FEATURE covers template territory by decision, matching the ADR while leaving a future template-owned artifact tree open.

template-shell/packages/react/src/hooks/useApiStream.ts
tscbmstubp: Addressed in ae015508: removed the gears-frontx issue number from shipped template payload. The comment now describes the pending upstream initializer change and the required TSDoc update without a repository-local reference.

CLAUDE.md
tscbmstubp: Addressed in ae015508: replaced the numeric ADR citation with the durable cpt-frontx-adr-template-territory-traceability slug.

template-shell/packages/react/src/hooks/useApiStream.ts
tscbmstubp: Addressed in ae015508: trimmed the module banner to the hook role and returned values; lifecycle ownership now stays with the detailed hook TSDoc.

architecture/ADR/0033-template-territory-traceability.md
```diff
@@ -0,0 +1,126 @@
...
+
+The `@cpt-` markers presently in template territory are **residue of the archived pre-redesign generation and bind nothing**: they are not authoritative, they are not evidence that a contract was specified, and they must not be read as traceability. **No new markers are added to template territory**, and when a template file is otherwise modified, **every residue marker that file carries is removed** — the rule is uniform, needs no separate sweep, and leaves no half-covered file behind. The strip is mechanical, and review should keep marker-only deletion easy to distinguish from substantive change. Where template code has behaviour a consumer depends on — a hook's observable status lifecycle, for instance — that behaviour is documented **at code altitude, in the code and its package documentation**, which is the reader's location and carries no claim of upstream specification.
```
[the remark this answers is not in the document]
tscbmstubp: Fixed in 01b29c23. Root AGENTS.md now states the exception at the authority surface itself: ecosystem code remains in the FEATURE traceability chain, while template-shell/ and template-mfe/ sit outside that chain and are governed by cpt-frontx-adr-template-territory-traceability plus code-altitude/package documentation.

AGENTS.md
```diff
@@ -8,6 +8,6 @@ ALWAYS resolve and enforce prerequisites of skills/workflows/commands BEFORE app
 
 AI tooling and FrontX development guidelines are provided by the Constructor Studio kit under `.cf-studio/`.
 
-Treat `architecture/` as the authority on what the system is and why: PRD for intent, DESIGN for structure, ADRs for decisions, and each FEATURE for the behaviour its numbered instructions specify. Code carries `@cpt-` markers back to those instructions; `cfs validate` checks that the chain holds.
+Treat `architecture/` as the authority on what the system is and why: PRD for intent, DESIGN for structure, ADRs for decisions, and each FEATURE for the behaviour its numbered instructions specify. Ecosystem code carries `@cpt-` markers back to those instructions, and `cfs validate` checks that chain; `template-shell/` and `template-mfe/` are outside that FEATURE traceability chain and are governed instead by `cpt-frontx-adr-template-territory-traceability` plus code-altitude/package documentation.
```
itechmeat: [Not a blocker] At the merge-base these two root files were byte-identical; after this round they describe the same carve-out in two different sentences, and this one drops the actionable half - an agent reading only AGENTS.md learns template markers are outside the chain, but not that a touched file must shed them. One line: make the two line-11s identical, either wording.

architecture/ADR/0033-template-territory-traceability.md
```diff
@@ -0,0 +1,126 @@
...
+
+The scope of this decision is which artifact tree specifies template-territory code and the standing of `@cpt-` markers found there. It does not decide the boundary a template declares (`cpt-frontx-adr-template-ownership-boundary-declaration`), how the CLI applies or upgrades a template (`cpt-frontx-adr-uniform-template-mechanism`, `cpt-frontx-adr-project-upgrade-mechanism`), nor whether the template will eventually carry an artifact tree of its own — that is deliberately deferred (see More Information).
+
+### Consequences
+
+* Good, because each artifact tree keeps one subject: the ecosystem's artifacts are described by the ecosystem's artifacts, and template payload is not smuggled into them.
+* Good, because a `@cpt-` marker recovers its meaning inside the scanned specified universe: there it names a live instruction, and ignored test/build residue outside that universe cannot imply a validated contract.
```
itechmeat: [Not a blocker] The new second half defends surviving residue by the validator's scan scope - the exact reasoning this ADR rejects at line 36 ("whether or not any validator reads it") and in the rejected keep-and-rely-on-exclusion option (line 97). Maybe keep the first half and neutralize the appeal to validation, e.g. "...there it names a live instruction; residue in paths the registry ignores is retired by the rule above"?

template-shell/packages/react/src/hooks/useApiStream.ts
```diff
@@ -60,6 +108,8 @@ export function useApiStream<TEvent>(
 
   const [data, setData] = useState<TEvent | undefined>(undefined);
   const [events, setEvents] = useState<TEvent[]>([]);
+  // A pending upstream change replaces this initializer with `enabled ? 'connecting' : 'idle'`;
```
itechmeat: [Not a blocker] Residual thought on my own round-1 ask: this file ships into consumer repositories where there is no "upstream" and nobody there can perform the TSDoc update, so the note reads as a maintainer TODO aimed at no one in the copy. Fine to keep for develop's sake - just flagging that the trade-off is now placement, not the issue number.

itechmeat (commented):
Every ask from both rounds has landed and verified - the commit-shape rule is gone from every surface, the revisit wording is mechanism-agnostic, the instruction surfaces point instead of restating, and the code is still provably untouched (comment-stripped md5 identical across all three revisions). Three light observations this round, none blocking. The only thing between this PR and my approval is the open direction thread - once Gera confirms the vehicle either way, I am ready to approve.

AGENTS.md
tscbmstubp: Addressed in 5e6f541: `AGENTS.md` line 11 now matches `CLAUDE.md` byte-for-byte, including the actionable marker-retirement wording.

architecture/ADR/0033-template-territory-traceability.md
tscbmstubp: Addressed in 5e6f541: the consequence no longer relies on validator scan scope. It now says a marker names a live instruction inside the specified universe and that registry-ignored residue is retired by the policy above.

template-shell/packages/react/src/hooks/useApiStream.ts
tscbmstubp: Addressed in 5e6f541: removed the two maintainer-only lines from the shipped template payload. The `StreamStatus` initializer remains `idle`, so the change is comment-only.

template-shell/packages/react/CLAUDE.md
```diff
@@ -124,18 +124,38 @@ function ChatStream() {
...
+
+The TSDoc block on `useApiStream` in `src/hooks/useApiStream.ts` is the detailed and authoritative state-machine reference, including teardown edge cases. This guide summarizes consumer-facing usage and gating.
+
+`disconnect()` closes the connection and does not reconnect. There is no reconnect function: to open a new connection, change the descriptor key, change `mode`, or toggle `enabled` back to true.
+
+#### The `enabled` option
+
+`enabled` (default `true`) gates whether the hook connects at all. While it is false the hook never calls the descriptor's `connect()` and ordinarily holds `status: 'idle'` (calling `disconnect()` while disabled is the one exception — see the TSDoc). Flipping it to true re-runs the connect effect and starts a fresh connection; flipping it back to false tears the connection down and returns `status` to `'idle'`.
+
+Tearing down via `enabled` does not clear `data` or `events` — only a descriptor-key or `mode` change does. A component reading `data` while the stream is disabled still sees the last event from the previous connection.
```
itechmeat: 🟡 [Not a blocker] The TSDoc this guide defers to says data/events/error are reset when `enabled` becomes true (lines 76-78), and the code resets them unconditionally at the top of the connect effect - so a false-to-true flip clears them too, and "only a descriptor-key or `mode` change does" overstates. Maybe "a descriptor-key or `mode` change - or re-enabling - does"?

template-shell/packages/react/src/hooks/useApiStream.ts
```diff
@@ -38,19 +33,72 @@ export interface ApiStreamOptions {
...
+ *   `'connecting'` and `connect()` is called.
+ * - `enabled` becomes `false` (or starts `false`): status is forced to
+ *   `'idle'` and `connect()` is never called; no other field is reset by
+ *   this transition alone.
+ * - Unmount, `descriptor.key` change, `mode` change, or `enabled` flipping to
+ *   `false` while connected or connecting: the previous connection is torn
+ *   down via `descriptor.disconnect()` once its `connect()` promise settles
+ *   (or immediately skipped if it rejected). All four share the same cleanup
+ *   path: the effect's cleanup runs on every dependency change and on unmount.
+ * - Calling `disconnect()`: if a connection id is already resolved, it is
```
itechmeat, line 88: 🟡 [Not a blocker] Sibling of the completion quirk documented just below: `disconnect()` never cancels the `onEvent` closure (only the effect cleanup flips `cancelled`), so a buffered or in-flight event landing after the call re-asserts `'connected'` - and in the pending-connect case the settle handler tears down and returns without touching status, so the stray `'connected'` sticks. Worth one clause in this bullet?

AGENTS.md
```diff
@@ -8,6 +8,6 @@ ALWAYS resolve and enforce prerequisites of skills/workflows/commands BEFORE app
 
 AI tooling and FrontX development guidelines are provided by the Constructor Studio kit under `.cf-studio/`.
 
-Treat `architecture/` as the authority on what the system is and why: PRD for intent, DESIGN for structure, ADRs for decisions, and each FEATURE for the behaviour its numbered instructions specify. Code carries `@cpt-` markers back to those instructions; `cfs validate` checks that the chain holds.
+Treat `architecture/` as the authority on what the system is and why: PRD for intent, DESIGN for structure, ADRs for decisions, and each FEATURE for the behaviour its numbered instructions specify. Ecosystem code carries `@cpt-` markers back to those instructions; `cfs validate` checks that the chain holds. Template subtrees (`template-shell/`, `template-mfe/`) sit outside this chain: markers found there are non-authoritative residue being retired as files change (`cpt-frontx-adr-template-territory-traceability`).
```
itechmeat: 🟡 [Not a blocker] `scripts/template-discovery.mjs` opens with "A template IS its manifest (ADR-0018)" and cites the #470 rename as why a name-based rule quietly stops covering anything - should this sentence lead with the manifest rule and name the two directories as today's instances? The CodeRabbit thread above proposing exactly that is marked "Addressed in 22e9d2b", but the wording still enumerates directories, so it looks auto-closed by the byte-identity fix rather than answered. The toml globs stay literal either way.

template-shell/packages/react/src/hooks/useApiStream.ts
```diff
@@ -38,19 +33,72 @@ export interface ApiStreamOptions {
...
+ *   disconnected immediately and status becomes `'disconnected'`. If
+ *   `connect()` is still pending, the request is recorded and honored when
+ *   the promise settles — the newly resolved connection is torn down instead
+ *   of being adopted, and status is `'disconnected'`.
+ * - Stream completion while `connect()` is still pending: the current
+ *   implementation lets `onComplete` set `'disconnected'`, then the settle
+ *   handler adopts the connection id and re-asserts `'connected'` anyway.
+ *   This is known current behavior, not a durable guarantee: a pending
```
itechmeat, line 100: 🔵 [Not a blocker] The PR body drifted behind the tree in three places: it says the #491 reminder is "pinned in-file on the initializer line" (both lines were removed this round - this sentence is the only surviving in-file trace), it still describes the removal rule "as its own commit inside the change" (dropped in round 2; the word "commit" appears in no changed file), and it quotes "rewritten in place" where the ADR now reads "revisited through the repository's ADR lifecycle". Worth a body pass before merge - it is what the second-lander will read.

itechmeat (commented):
Fourth pass. Every prior thread is fixed in the tree and verified - the root files are byte-identical again, ADR:60 no longer leans on validator scan scope, the maintainer note left the shipped payload, and comment-stripped md5 is still identical to develop. Four light observations this round, none blocking: the guide's teardown line contradicts the TSDoc it defers to, the disconnect() bullet misses a sibling of the documented completion quirk, the carve-out defines templates by directory names where the repo defines them by manifest, and the PR body describes round-2 text in three places. The merge gate is unchanged - the direction thread still awaits @GeraBart; after four days of silence a short call may close it faster than a fifth async round.

itechmeat (approved):
Approving - this supersedes my earlier request for changes. All substantive asks from four rounds are fixed and verified in the tree: zero logic change throughout (comment-stripped md5 identical to develop), the docs now match the code, and the four observations from my last pass are non-blocking wording notes. The direction thread (standalone ADR vs folding into #495) stays open for @GeraBart - the ADR's ID is slug-stable, so if he prefers the fold the cost is a file move, not a rewrite. Content-wise this is ready.

architecture/ADR/0033-template-territory-traceability.md
```diff
@@ -0,0 +1,126 @@
...
+- [More Information](#more-information)
+- [Traceability](#traceability)
+
+<!-- /toc -->
+
+**ID**: `cpt-frontx-adr-template-territory-traceability`
+
+## Context and Problem Statement
+
+The repository's architecture artifacts describe the FrontX ecosystem — the published packages, the CLI, and the AI tooling kit — while the template subtrees (`template-shell/`, `template-mfe/`) hold template payload the CLI resolves and applies into a developer's repository, not ecosystem contract. That payload still carries `@cpt-` traceability markers from the pre-ecosystem-redesign artifact generation, whose FEATUREs (`feature-request-lifecycle`, `feature-react-bindings`, `feature-state-management` and their peers) were archived and then deleted, so the markers name instruction IDs no artifact defines. Reviewers reading a marked template file reasonably infer a specified contract behind it and open a gap report when they cannot find one. Which artifact tree, if any, specifies template-territory code, and what status do the `@cpt-` markers already sitting in it have?
```
GeraBart: This problem statement is very current-project-state-specific. In general, I think that artifacts structure should be treated as a part of the project design, and thus needs an ADR. Otherwise SDLC specs become recursive.

gs-layer (approved):
Superseded by follow-up review on ccc1d829: current verdict is COMMENTED due to the AGENTS.md:11 policy/enforcement mismatch.

template-shell/packages/react/CLAUDE.md
tscbmstubp: Fixed in ccc1d82 — the sentence now reads 'a descriptor-key or `mode` change, or re-enabling, does', matching the TSDoc and the unconditional reset at the top of the connect effect.

template-shell/packages/react/src/hooks/useApiStream.ts
tscbmstubp: Documented in ccc1d82 — the `disconnect()` bullet gained the clause: the event callback is not cancelled, so a buffered or in-flight event landing after the call re-asserts 'connected', and in the pending-connect case the settle handler tears down without touching status, so the stray 'connected' persists. Behavioural fix stays with #491's derived-status rewrite, alongside the completion quirk.

AGENTS.md
tscbmstubp: Reworded in ccc1d82 — both root line-11s now lead with the manifest rule ('any top-level directory carrying a `frontx-template.json` manifest, how `scripts/template-discovery.mjs` finds templates') and name `template-shell/` and `template-mfe/` as today's instances. The two files remain byte-identical.

template-shell/packages/react/src/hooks/useApiStream.ts
tscbmstubp: Body pass done — all three drift points corrected: the #491 section now states the in-file reminder was removed in round 3 and that the section itself is the coordination record; the removal-rule bullet drops the commit-shape phrasing for the review-time-distinguishability wording; the deferral bullet now quotes the ADR-lifecycle mechanism instead of 'rewritten in place'.

architecture/ADR/0033-template-territory-traceability.md
tscbmstubp: Thanks — taking that as the direction call: artifact structure is part of the project design, so the decision stays a standalone root ADR rather than folding into the #495 restructuring. That was the last open merge gate from the review (itechmeat approved content-wise pending exactly this confirmation). The problem-statement specificity is acknowledged in the ADR itself — the decision is scoped to be revisited through the ADR lifecycle once template ownership settles.

architecture/ADR/0033-template-territory-traceability.md
tscbmstubp: Resolved by Gera's comment on the ADR (2026-08-11): artifact structure is part of the project design and warrants an ADR — the standalone vehicle stands, no fold into #495.

AGENTS.md
```diff
@@ -8,6 +8,6 @@ ALWAYS resolve and enforce prerequisites of skills/workflows/commands BEFORE app
 
 AI tooling and FrontX development guidelines are provided by the Constructor Studio kit under `.cf-studio/`.
 
-Treat `architecture/` as the authority on what the system is and why: PRD for intent, DESIGN for structure, ADRs for decisions, and each FEATURE for the behaviour its numbered instructions specify. Code carries `@cpt-` markers back to those instructions; `cfs validate` checks that the chain holds.
+Treat `architecture/` as the authority on what the system is and why: PRD for intent, DESIGN for structure, ADRs for decisions, and each FEATURE for the behaviour its numbered instructions specify. Ecosystem code carries `@cpt-` markers back to those instructions; `cfs validate` checks that the chain holds. Template territory — any top-level directory carrying a `frontx-template.json` manifest (how `scripts/template-discovery.mjs` finds templates; today `template-shell/` and `template-mfe/`) — sits outside this chain: markers found there are non-authoritative residue being retired as files change (`cpt-frontx-adr-template-territory-traceability`).
```
gs-layer: This sentence generalizes template territory to any top-level directory with `frontx-template.json`, but the enforcement/config surfaces in this PR are still hardcoded to `template-shell/` and `template-mfe/` (`.cf-studio/config/AGENTS.md`, `.cf-studio/config/artifacts.toml`). That creates a policy split: a future manifest-backed template dir would be outside the `@cpt-` chain per this line, while `cfs validate` and FX-002 would still treat it as in-universe. Please either scope this wording back to the current two directories or make the ignore/review surfaces manifest-driven in the same PR.

AGENTS.md
tscbmstubp: Follow-up in 00296f4: dropped the directory names entirely — a prose-level directory enumeration rots the same way the #470 name-glob would have, so the manifest rule is now the whole definition. Both root line-11s remain byte-identical.

AGENTS.md
tscbmstubp: Took the second option, in fa73c7a: the review surfaces in `.cf-studio/config/AGENTS.md` now state the manifest rule directly (any top-level directory carrying `frontx-template.json`), and the toml globs — which must stay literal, the registry can't evaluate a manifest — gained a sync guard in `scripts/template-discovery.test.mjs` asserting they cover exactly the manifest-discovered template directories. Same drift class and same fix as the MANIFEST_FILENAME guard from #492: a future manifest-backed template dir added without updating the ignore entry now fails the unit-test suite instead of silently splitting the policy.

GeraBart (approved):

PR #522, opened 2026-08-04
ci: add template-drift job — relink and check templates against the working tree (#518)

### LINKED ISSUE #518: CI: nothing goes red when packages/ drifts from the templates' registry pins
Spun out of https://github.com/constructorfabric/gears-frontx/pull/491#discussion_r3703509644.

Since #485 the templates pin `@gears-frontx/*` to exact registry versions, so the `template-validate` job resolves published tarballs and never this repo's `packages/`. That is deliberate (it validates what a seeded project actually installs), but it also means the coverage the old `file:` links provided incidentally is gone: a change to `packages/*` that breaks the template surface produces nothing red until the next publish-and-pin-bump cycle. CONTRIBUTING.md documents the silence as expected for the local dev loop.

Proposed cover: a periodic (or `packages/**`-triggered) job that relinks via `dev:template:link` and runs the template type-check against the working tree — red means the working tree and the templates have drifted, without changing what `template-validate` itself validates.

tscbmstubp:
## Summary

Implements the guard proposed in #518: since #485 the templates pin `@gears-frontx/*` to exact registry versions, so template validation resolves published tarballs and never this repo's `packages/`. A change to `packages/*` that breaks the template surface therefore produces nothing red until the next publish-and-pin-bump cycle — the coverage the old `file:` links provided incidentally.

This adds `.github/workflows/template-drift.yml`, which covers the reverse direction **without changing what template validation itself validates**:

1. root `npm ci` + `npm run build:packages` — the link script refuses an unbuilt package, since consumers resolve through `dist/`
2. `npm ci` inside `template-shell` — the pinned tree must exist before it can be repointed
3. `npm run dev:template:link` — repoint the three installed ecosystem directories at the working tree
4. `npm run build` + `npm run type-check` inside `template-shell`

Red means the working tree and the templates have drifted; the pins are not consulted.

Triggers: `packages/**`, `template-shell/**`, the root `package.json`/`package-lock.json` (the job runs the root's scripts and dependency tree), the link script, and the workflow itself, on push/PR to `main`/`develop`, plus `workflow_dispatch`. No schedule: everything the job reads is in-repo, so path triggers cover it completely.

Also adds one sentence to CONTRIBUTING.md's known-issue section: the post-#485 relink silence is local only, CI now catches the drift.

Deliberately a separate workflow file rather than a job in `main.yml`: PR #491 (which spun out #518 from [this discussion](https://github.com/constructorfabric/gears-frontx/pull/491#discussion_r3703509644)) rewrites `main.yml` heavily, and this guard shares no steps with its `template-validate` job — the two validate opposite directions of the same pin contract.

## Verification

The full sequence was run locally on `develop` and comes out green; the PR's own first CI run confirmed it — "Template vs Working Tree" passed in 1m42s.

Closes #518

### DIFF
--- a/.github/workflows/template-drift.yml
+++ b/.github/workflows/template-drift.yml
@@ -1,0 +1,102 @@
+# Reverse-direction guard for the template registry pins (#518).
+#
+# Since #485 template-shell pins @gears-frontx/* to exact registry versions, so
+# template validation resolves published tarballs and never this repo's
+# packages/. That is deliberate — it validates what a seeded project actually
+# installs — but it also means a change to packages/* that breaks the template
+# surface produces nothing red until the next publish-and-pin-bump cycle;
+# CONTRIBUTING.md documents the silence as expected for the local dev loop.
+#
+# This job covers that gap without changing what template validation itself
+# validates: it relinks the template's installed ecosystem packages at the
+# working tree via `dev:template:link` and runs the template's build and
+# type-check against it. Red here means the working tree and template-shell
+# have drifted — either side may have moved; the pins are not consulted.
+name: Template Drift
+
+# Everything this job reads is in-repo, so path triggers cover it completely —
+# no schedule needed. The root manifest, lockfile, npm config, internal
+# build-time packages, and scripts are in the list because the job's first half
+# runs the root's scripts and dependency tree (`build:packages`,
+# `dev:template:link`): a root or internal dependency bump can change what the
+# linked packages build to without touching packages/** itself.
+on:
+  push:
+    branches: [ main, develop ]
+    paths:
+      - '.npmrc'
+      - 'package.json'
+      - 'package-lock.json'
+      - 'internal/**'
+      - 'packages/**'
+      - 'scripts/**'
+      - 'template-shell/**'
+      - '.github/workflows/template-drift.yml'
+  pull_request:
+    branches: [ main, develop ]
+    paths:
+      - '.npmrc'
+      - 'package.json'
+      - 'package-lock.json'
+      - 'internal/**'
+      - 'packages/**'
+      - 'scripts/**'
+      - 'template-shell/**'
+      - '.github/workflows/template-drift.yml'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  template-drift:
+    name: Template vs Working Tree
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v5
+      with:
+        persist-credentials: false
+
+    - name: Setup Node.js
+      uses: actions/setup-node@v5
+      with:
+        node-version: '25.x'
+        cache: 'npm'
+        cache-dependency-path: |
+          package-lock.json
+          template-shell/package-lock.json
+
+    - name: Enable Corepack
+      run: corepack enable
+
+    - name: Install dependencies
+      run: npm ci
+
+    # The link script refuses an unbuilt package: what the template resolves
+    # through each link is the package's dist/, so the working-tree surface
+    # under test only exists after this build.
+    - name: Build workspace packages
+      run: npm run build:packages
+
+    # `dev:template:link` replaces installed directories inside the template's
+    # node_modules — the pinned tree has to exist before it can be repointed.
+    - name: Install template from its lockfile
+      working-directory: template-shell
+      run: npm ci
+
+    - name: Relink template at the working tree
+      run: npm run dev:template:link
+
+    # Full `build`, not just the compile steps: vite bundling resolves the
+    # linked dist/ too, and type-check:app expects the generated MFE manifests.
+    # `build` does not type-check the app (vite strips types), so the explicit
+    # type-check after it is not redundant.
+    - name: Build template against the working tree
+      working-directory: template-shell
+      run: npm run build
+
+    - name: Type-check template against the working tree
+      working-directory: template-shell
+      run: npm run type-check
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -118,6 +118,8 @@ cd template-shell && npm run dev
 
 > **Forgetting to relink is silent.** The template builds, type-checks and tests green against the published pins on its own, so a missing relink produces no error anywhere - it just means the `packages/*` edits you are testing are not the code being run. Nothing warns you. Relink after every ecosystem change, and when a template-side result contradicts a change you just made in `packages/*`, suspect the link first.
 
+The silence is local only: the [Template Drift workflow](.github/workflows/template-drift.yml) runs this same relink-and-check sequence in CI for the in-repo inputs it watches, so a working tree that has drifted from `template-shell` goes red there even when every local check passes ([#518](https://github.com/constructorfabric/gears-frontx/issues/518)).
+
 This was not always the steady state. Until the pins reached `0.3.0-alpha.1` ([#485](https://github.com/constructorfabric/gears-frontx/issues/485)), the published `0.3.0-alpha.0` tarballs predated the move of `FRONTX_ACTION_*` into `@gears-frontx/gts-plugin` and the addition of `DomainContext.typeSystem` to `@gears-frontx/mfes`, so `type-check` and `build:packages` failed against the pins alone and linking was mandatory rather than an optimisation. That took two merges to clear - one to publish the fixed surfaces, one to move the pins onto them, because a lockfile cannot resolve a tarball that does not exist yet - and both have landed. A seeded project now builds from its pins with no monorepo present.
 
 To go back to the pinned registry versions, run `npm ci` inside `template-shell`. There is no `--unlink`: the links replace published tarball *content*, which only npm can restore.
--- a/template-shell/package-lock.json
+++ b/template-shell/package-lock.json
@@ -13,7 +13,7 @@
         "src-app/mfe_packages/*"
       ],
       "dependencies": {
-        "@gears-frontx/api": "0.3.0-alpha.0",
+        "@gears-frontx/api": "0.3.0-alpha.1",
         "@gears-frontx/gts-plugin": "0.3.0-alpha.1",
         "@gears-frontx/mfes": "0.3.0-alpha.1",
         "@hookform/resolvers": "5.2.2",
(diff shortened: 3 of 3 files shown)


.github/workflows/template-drift.yml
```diff
@@ -0,0 +1,89 @@
...
+name: Template Drift
+
+on:
+  push:
+    branches: [ main, develop ]
+    paths:
+      - 'packages/**'
+      - 'template-shell/**'
+      - 'scripts/link-template-ecosystem.mjs'
+      - '.github/workflows/template-drift.yml'
```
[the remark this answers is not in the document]
tscbmstubp: Fixed in c5ea7586 — root `package.json` and `package-lock.json` added to both path filters (and the weekly schedule dropped, since with a complete path list everything this job reads is in-repo).

.github/workflows/template-drift.yml
```diff
@@ -0,0 +1,89 @@
...
+  workflow_dispatch:
+
+jobs:
+  template-drift:
+    name: Template vs Working Tree
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v5
```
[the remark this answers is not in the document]
tscbmstubp: Fixed in c5ea7586 — `persist-credentials: false` set on the checkout.

.github/workflows/template-drift.yml
```diff
@@ -0,0 +1,89 @@
...
+      uses: actions/setup-node@v5
+      with:
+        node-version: '25.x'
+        cache: 'npm'
+        cache-dependency-path: |
+          package-lock.json
+          template-shell/package-lock.json
+
+    - name: Enable Corepack
+      run: corepack enable
```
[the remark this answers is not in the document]
tscbmstubp: Not applied — empirically refuted: this job's first run passed on `25.x` including the `corepack enable` step (["Template vs Working Tree", 1m42s](https://github.com/constructorfabric/gears-frontx/actions/runs/30889362697/job/91927610789)), and `main.yml`'s `validate (25.x)` matrix leg has been running the same `corepack enable` green on every push. `actions/setup-node` installs the toolchain this repo already relies on at 25.x; changing the version here would diverge from the rest of CI to fix a failure that does not occur.

tscbmstubp:
The first run of the new job on the rebased branch went red — on a real finding, not a workflow bug: develop's 309d77fc bumped `template-shell/package.json`'s api pin to `0.3.0-alpha.1` without regenerating the lockfile, so `npm ci` inside `template-shell` fails on develop itself. This PR now carries the sync (26bf0abf).

The lockfile bump is surgical — just the api requirement string and the `node_modules/@gears-frontx/api` version/resolved/integrity entries — because a full `npm install` regeneration under npm 11.7 drops the template's `file:.` self-link entry from the lockfile it writes, producing a lockfile its own `npm ci` then rejects. That npm quirk may deserve its own issue if it bites again.

.github/workflows/template-drift.yml
```diff
@@ -0,0 +1,94 @@
...
+# packages build to without touching packages/** itself.
+on:
+  push:
+    branches: [ main, develop ]
+    paths:
+      - 'package.json'
+      - 'package-lock.json'
+      - 'packages/**'
+      - 'template-shell/**'
+      - 'scripts/link-template-ecosystem.mjs'
```
itechmeat: [Not a blocker] The comment at line 17 says path triggers cover everything the job reads, but two inputs are missing: `scripts/template-ecosystem-packages.mjs` (imported by the link script at its line 53 - the module that derives which directories get relinked) and `internal/**` (`build:packages` starts with the eslint-config workspace build). Widening this entry to `scripts/**` plus adding `internal/**` would make the completeness claim true - or the comment could be softened.

.github/workflows/template-drift.yml
```diff
@@ -0,0 +1,94 @@
...
+
+    - name: Relink template at the working tree
+      run: npm run dev:template:link
+
+    # Full `build`, not just the compile steps: vite bundling resolves the
+    # linked dist/ too, and type-check:app expects the generated MFE manifests.
+    # `build` does not type-check the app (vite strips types), so the explicit
+    # type-check after it is not redundant.
+    - name: Build template against the working tree
+      working-directory: template-shell
```
itechmeat, line 97: [Not a blocker] The body says this guard "shares no steps with its template-validate job", but #491's job at its current head runs the same `npm ci` → `npm run build` → `npm run type-check` loop inside the template - this file inserts a relink before the same three steps, hardcoded to one template where #491 discovers them from manifests. Separate file now is fine given the main.yml rewrite in flight; could the description drop that claim, and a follow-up be recorded to fold this into (or reuse the classification of) template-validate once #491 lands?

CONTRIBUTING.md
```diff
@@ -118,6 +118,8 @@ cd template-shell && npm run dev
 
 > **Forgetting to relink is silent.** The template builds, type-checks and tests green against the published pins on its own, so a missing relink produces no error anywhere - it just means the `packages/*` edits you are testing are not the code being run. Nothing warns you. Relink after every ecosystem change, and when a template-side result contradicts a change you just made in `packages/*`, suspect the link first.
 
+The silence is local only: the [Template Drift workflow](.github/workflows/template-drift.yml) runs this same relink-and-check sequence in CI whenever `packages/**`, `template-shell/**`, or the root manifest/lockfile change, so a working tree that has drifted from the templates goes red there even when every local check passes ([#518](https://github.com/constructorfabric/gears-frontx/issues/518)).
```
itechmeat: [Not a blocker] This sentence re-enumerates the trigger list that lives in the workflow (the first path edit there makes this stale), and "drifted from the templates" is plural while only template-shell is linked, built and type-checked (template-mfe has no root manifest and only `file:` pins, which the pin detector rejects). Pointing at the workflow without repeating its paths, and singularizing to the shell template - here and in the workflow header - keeps the doc from drifting.

.github/workflows/template-drift.yml
```diff
@@ -0,0 +1,94 @@
...
+    paths:
+      - 'package.json'
+      - 'package-lock.json'
+      - 'packages/**'
+      - 'template-shell/**'
+      - 'scripts/link-template-ecosystem.mjs'
+      - '.github/workflows/template-drift.yml'
+  workflow_dispatch:
+
+jobs:
```
itechmeat, line 51: [Not a blocker] This is now the only workflow with a hardened checkout (`persist-credentials: false`) but no `permissions:` block - docs.yml and publish-packages.yml declare one, and #491 adds a workflow-level `contents: read` to main.yml with the same rationale. Two lines here would complete the pattern.

template-shell/package-lock.json
```diff
@@ -13,7 +13,7 @@
         "src-app/mfe_packages/*"
       ],
       "dependencies": {
-        "@gears-frontx/api": "0.3.0-alpha.0",
+        "@gears-frontx/api": "0.3.0-alpha.1",
```
itechmeat, line 16: [Not a blocker] The surgical lockfile sync itself is verified (integrity matches the registry byte-for-byte, and develop's `npm ci` in template-shell is genuinely broken since 309d77fc) - but the npm 11.7 quirk that forced hand-editing lives only in a PR comment with "may deserve its own issue if it bites again". The next local `npm install` in template-shell regenerates a lockfile its own `npm ci` rejects, with nothing written down; worth filing that issue now? (CI itself runs npm 10.9.4 via corepack, so the hazard is local-machine only.)

.github/workflows/template-drift.yml
```diff
@@ -0,0 +1,94 @@
...
+        persist-credentials: false
+
+    - name: Setup Node.js
+      uses: actions/setup-node@v5
+      with:
+        node-version: '25.x'
+        cache: 'npm'
+        cache-dependency-path: |
+          package-lock.json
+          template-shell/package-lock.json
```
itechmeat, line 69: [Not a blocker] The cache path (and the three working-directory values) hardcode the template name where #491's parallel job uses `template-*/package-lock.json` and manifest-driven discovery - understood that `link-template-ecosystem.mjs` itself hardcodes `templateDirName = 'template-shell'`, so this job cannot cover a second template until the script can. Is generalizing both together the plan?

.github/workflows/template-drift.yml
```diff
@@ -0,0 +1,94 @@
...
+    - name: Relink template at the working tree
+      run: npm run dev:template:link
+
+    # Full `build`, not just the compile steps: vite bundling resolves the
+    # linked dist/ too, and type-check:app expects the generated MFE manifests.
+    # `build` does not type-check the app (vite strips types), so the explicit
+    # type-check after it is not redundant.
+    - name: Build template against the working tree
+      working-directory: template-shell
+      run: npm run build
```
itechmeat, line 98: [Not a blocker] The drift check is not vacuous - the shell's own packages import api/mfes/gts-plugin heavily - but the composed tree a seeded project actually runs (shell + template-mfe overlay) is out of reach until #491's composition step exists: `template-shell/src-app/mfe_packages` is empty in-repo. Worth a follow-up note to point this job at the composed tree once that lands?

itechmeat (approved):
Solid, well-aimed guard - it closes the exact hole #518 names, the first run already caught a real lockfile desync on develop, and the hand-edited lockfile entry verifies byte-for-byte against the registry. Everything I have is polish: trigger-path completeness vs the comment that claims it, wording that overstates scope (the "shares no steps" claim and the plural "templates"), and small hardening/coordination touches around #491. None of it blocks approval.

.github/workflows/template-drift.yml
tscbmstubp: Fixed in 922978b2: both trigger lists now cover scripts/** and internal/**, and I also added .npmrc because both npm ci steps read it implicitly. The completeness comment now names those inputs.

.github/workflows/template-drift.yml
tscbmstubp: Agreed: the install, build, and type-check steps overlap with #491. This job stays separate only to avoid coupling this focused guard to the in-flight main.yml rewrite. Consolidating it with the manifest-driven template validation path should be tracked after #491 lands; the PR description claim that there are no shared steps is too strong.

template-shell/package-lock.json
tscbmstubp: Agreed that the npm 11.7 regeneration quirk should be tracked separately because it can recreate a lockfile that npm ci rejects locally. No further lockfile change is needed here: the surgical sync is verified and CI is pinned to npm 10.9.4.

CONTRIBUTING.md
tscbmstubp: Fixed in 922978b2. CONTRIBUTING now points to the workflow without duplicating its path list, and both the workflow header and contributor wording describe the current template-shell-only scope.

.github/workflows/template-drift.yml
tscbmstubp: Agreed. The current guard intentionally verifies the reachable template-shell surface. Once #491 provides composed shell plus MFE materialization, this job should be folded into or extended through that composition path so it exercises the seeded tree rather than the empty in-repo overlay location.

.github/workflows/template-drift.yml
tscbmstubp: Yes. The workflow and link script should be generalized together rather than making only the cache and working-directory paths generic. This PR intentionally remains template-shell-specific; generalization belongs in a coordinated follow-up with the manifest-driven work from #491.

.github/workflows/template-drift.yml
tscbmstubp: Fixed in 922978b2 by adding workflow-level permissions with contents: read. Checkout still keeps persist-credentials disabled.

GeraBart (approved):

PR #525, opened 2026-08-04
ci: guard the template lockfile self-link entry npm reconciliation drops (#524)

### LINKED ISSUE #524: npm regenerates template-shell's lockfile without its file:. self-link entry, producing a lockfile npm ci rejects
Found while fixing the lockfile desync in PR #522 (spun out of [this comment](https://github.com/constructorfabric/gears-frontx/pull/522#issuecomment-5176443728)).

## What the quirk is

`template-shell/package.json` depends on itself — `"@gears-frontx/frontx-template-shell": "file:."` — so the template's own library build is resolvable through `node_modules` by the app and by the workspace packages (`packages/framework` reaches it as `file:../..`). The committed lockfile carries the entry npm needs for that:

```json
"node_modules/@gears-frontx/frontx-template-shell": {
  "resolved": "",
  "link": true
}
```

When `npm install` has to *reconcile* a manifest change (i.e. the lockfile is out of sync with `package.json`), the lockfile it writes back **omits this entry**. A subsequent `npm ci` then fails against the very lockfile `npm install` just produced:

```
npm error Missing: @gears-frontx/frontx-template-shell@0.1.0-alpha.2 from lock file
```

## Why this is a trap, not just a curiosity

`npm ci`'s error message for an out-of-sync lockfile says, verbatim: *"Please update your lock file with `npm install` before continuing."* Following that instruction inside `template-shell` produces another lockfile `npm ci` rejects — for a different reason, with no hint that the self-link entry is what went missing. Every future dependency change in the template walks into this: regenerating the lockfile is the documented workflow, and CI (`template-validate` from #491, `template-drift` from #518/#522) runs `npm ci` and goes red on the result.

This is exactly how #522 hit it: 309d77fc bumped the api pin in the manifest without the lockfile, `npm ci` broke, and the natural `npm install` remediation produced a second broken lockfile. The fix that shipped was a hand edit of the api entries with the self-link entry left untouched.

## Reproduction (npm 11.7.0, node v25.4.0)

Steady state is fine — from a clean tree with a *consistent* lockfile, the entry survives:

```bash
cd template-shell
rm -rf node_modules && npm install   # entry still present, npm ci passes
```

The drop needs a lockfile that npm has to reconcile. Recreate the state 309d77fc left (manifest pins `@gears-frontx/api@0.3.0-alpha.1`, lockfile still resolves `alpha.0`):

```bash
cd template-shell
git show 309d77fc^:template-shell/package-lock.json > package-lock.json   # pre-bump lockfile
git show 309d77fc:template-shell/package.json > package.json              # post-bump manifest
rm -rf node_modules
npm install    # reconciles the api bump; the written lockfile has NO frontx-template-shell entry
npm ci         # -> npm error Missing: @gears-frontx/frontx-template-shell@... from lock file
```

## Workaround until fixed

When a template dependency change requires regenerating the lockfile, either make the lockfile edit surgically (version/resolved/integrity of the changed package only, as in #522), or diff the regenerated lockfile and restore the `node_modules/@gears-frontx/frontx-template-shell` entry by hand before committing.

## Possible directions

1. **Report upstream** to npm/cli with a minimal reproduction (self-referential `file:.` dependency + workspaces + a reconciling install) — this looks like an ideal-tree bug, and the repo is on the receiving end of it, not the cause.
2. **Guard in CI**: the template-drift/template-validate jobs already catch the broken lockfile after the fact; a cheaper pre-commit check could assert the self-link entry exists in `template-shell/package-lock.json` whenever it changes, so the failure names the actual problem instead of surfacing as a generic `npm ci` red.
3. **Design out the self-link** if it is no longer load-bearing — though it currently is: the app and the workspace packages resolve the template's `dist-lib` through it, and the dev-link script's docs (`scripts/link-template-ecosystem.mjs`) rely on npm preserving it.

Grounded in the #522 investigation; reproduced twice from clean `node_modules` on npm 11.7.0 / node v25.4.0 (the same major pair `setup-node@v5` with `node-version: '25.x'` gives CI).

tscbmstubp:
## Summary

Implements direction 2 from #524: a guard that fails **by name** when a template's lockfile loses the `file:.` self-link entry, instead of the loss surfacing later as a generic `npm ci` red in template CI.

The quirk it guards (#524, observed on npm 11.7.0 / node 25.4.0): when `npm install` has to *reconcile* an out-of-sync lockfile — the state a pin bump that skips the lockfile leaves behind, as `309d77fc` did — the lockfile it writes back omits

```json
"node_modules/@gears-frontx/frontx-template-shell": { "resolved": "", "link": true }
```

and `npm ci` then rejects that lockfile (`Missing: ... from lock file`). Reconciling with `npm install` is exactly what `npm ci`'s out-of-sync error message instructs, so the documented remediation produces a second broken lockfile with nothing naming what vanished.

## What's in the change

- **`scripts/template-lockfile-selflink-check.mjs`** — in the mold of `template-pin-drift-check.mjs`: templates discovered by `frontx-template.json` presence via the shared `template-discovery.mjs` (zero templates found is a hard failure, never a vacuous pass); what to assert is read from each template's own root manifest — self-link `file:.` specs in the installed dependency fields **and in `overrides`**, which is where the real tree declares it (`packages/framework` depends on the template by name, the root override pins the resolution to `file:.`). Overlay templates without a root `package.json`/lockfile (e.g. `template-mfe/`) are skipped — no `npm ci` contract to protect. Unreadable/malformed JSON fails closed. The failure message quotes the missing entry, links #524, and warns that re-running `npm install` is what removed it.
- **`scripts/template-lockfile-selflink-check.test.mjs`** — 20 tests in the sibling suites' fixture style, plus two against the committed repo tree. One of those is a **vacuity guard** asserting the guard sees at least one self-link in the real `template-shell/package.json` — the first draft of this script scanned only dependency fields, missed the `overrides` declaration, and passed on a lockfile with the entry deleted; that test makes this failure shape impossible to reship.
- **Wiring** — `npm run policy:template-lockfile-selflink`, plus a `main.yml` step beside the pin-drift check (no build-order dependency: reads JSON directly).

## Verification

- All 289 tests of the `repo-scripts` vitest project pass (20 new).
- The guard passes on the committed tree and was verified end-to-end against a copy of `template-shell` with the entry deleted — it fails with the message naming the entry.

Closes #524

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation to ensure installable templates retain valid lockfile self-link entries.
  * Improved diagnostics for missing, invalid, or unreadable template lockfiles.

* **Tests**
  * Added automated coverage for template discovery, dependency overrides, malformed files, and validation outcomes.

* **Chores**
  * Integrated template lockfile validation into continuous integration checks to catch issues earlier.

### DIFF
--- a/scripts/template-lockfile-selflink-check.mjs
+++ b/scripts/template-lockfile-selflink-check.mjs
@@ -1,0 +1,323 @@
+/**
+ * Template Lockfile Self-Link Guard (#524).
+ *
+ * A self-contained template depends on itself: the workspace packages depend
+ * on the template by name (`packages/framework` declares
+ * `"@gears-frontx/frontx-template-shell": "file:../.."`), and the template's
+ * root manifest pins that name's resolution to its own directory via
+ * `"overrides": { "@gears-frontx/frontx-template-shell": "file:." }`, so the
+ * template's own library build is resolvable through `node_modules`. npm
+ * records that resolution in the lockfile as
+ *
+ *   "node_modules/@gears-frontx/frontx-template-shell": {
+ *     "resolved": "",
+ *     "link": true
+ *   }
+ *
+ * and `npm ci` refuses to install without it ("Missing: <name>@<version> from
+ * lock file").
+ *
+ * The trap (#524): when `npm install` has to RECONCILE a manifest change — a
+ * lockfile out of sync with `package.json`, exactly the state a pin bump that
+ * skipped the lockfile leaves behind — the lockfile it writes back omits this
+ * entry. And reconciling with `npm install` is what `npm ci`'s own out-of-sync
+ * error message tells the developer to do, so the documented remediation
+ * produces a second broken lockfile, red in every job that runs `npm ci` in
+ * the template, with nothing naming the entry that vanished. This guard is
+ * that name: it fails on the exact missing lockfile entry, before the commit
+ * lands (observed on npm 11.7.0 / node 25.4.0, the pair CI's `25.x` jobs get).
+ *
+ * Discovery mirrors the sibling guards (`template-pin-drift-check.mjs`,
+ * `validate-templates.mjs`): a template is a top-level directory carrying
+ * `frontx-template.json` (ADR-0018 manifest presence, never a `template-*`
+ * name guess), and zero templates found — or zero templates actually checked —
+ * is a hard failure, never a vacuous pass. WHAT to assert is read from each
+ * template's own root manifest — every
+ * dependency declared at the `file:.` self-path — so a renamed template or a
+ * second self-linking template is covered with no change here. An overlay
+ * template (no root `package.json`, e.g. `template-mfe/`) or one without a
+ * lockfile has no `npm ci` contract to protect and is skipped.
+ *
+ * CLI entry: `node scripts/template-lockfile-selflink-check.mjs` (exit 0 on
+ * success). Core logic is exported for unit tests in
+ * `scripts/template-lockfile-selflink-check.test.mjs`.
+ */
+import fs from 'node:fs';
+import path from 'node:path';
+import process from 'node:process';
+import { pathToFileURL } from 'node:url';
+import { MANIFEST_FILENAME, findTemplateDirs } from './template-discovery.mjs';
+
+// Re-exported for the zero-templates failure message, same as the pin-drift
+// guard: the filename is owned by `template-discovery.mjs`.
+export { MANIFEST_FILENAME };
+
+/**
+ * The dependency fields npm materialises into the install tree. `peerDependencies`
+ * is deliberately absent: a peer declaration alone produces no `node_modules`
+ * entry of its own to lose. `overrides` is scanned separately
+ * (`selfLinkDependenciesOf`) because its values can be nested objects, and it
+ * is where the real tree declares its self-link today: `template-shell`'s root
+ * manifest carries `"overrides": { "@gears-frontx/frontx-template-shell": "file:." }`,
+ * redirecting the workspace packages' dependency on that name
+ * (`packages/framework` declares it at `file:../..`) to the template root.
+ */
+const INSTALLED_DEPENDENCY_FIELDS = ['dependencies', 'devDependencies', 'optionalDependencies'];
+
+/**
+ * A `file:` specifier that resolves to the manifest's own directory — the
+ * self-link shape npm's reconciliation drops from the lockfile (#524).
+ * `file:.` is what the repo declares; the normalised variants are accepted so
+ * a formatting change in the manifest cannot silently take the dependency out
+ * of the guard's view.
+ *
+ * @param {string} spec
+ * @returns {boolean}
+ */
+export function isSelfLinkSpec(spec) {
+  if (typeof spec !== 'string' || !spec.startsWith('file:')) return false;
+  const target = spec.slice('file:'.length);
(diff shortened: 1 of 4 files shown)


tscbmstubp:
Addressed (squashed into df1d6b5b): `check()` now hard-fails when `checkedCount === 0` (every discovered template skipped), with a message naming what was skipped and a test pinning the case — same rule as the zero-discovered guard, extended to zero-checked.

gs-layer (changes requested):
### Code review

Found 1 issue:

1. Vacuous pass when every discovered template is skipped (`checkedCount` never checked against zero)

`check()` hard-fails when `findTemplateDirs()` finds zero template directories, but does not verify that at least one template was actually *checked*. Each discovered template lacking a root `package.json` or `package-lock.json` is silently skipped via `continue`, and `checkedCount` is only used for the human-readable success message — never compared against zero. If every discovered template ends up skipped (e.g. a template's lockfile is missing, renamed, or moved in a future refactor), `findings` stays empty and the script prints `"Template lockfile self-link check passed: 0 installable template(s) of N discovered"` and exits 0.

This is the exact defect class this guard's own docblock says it's designed to prevent for the "zero templates found" case (a comment two lines above the check literally states "a vacuous pass would hide both"), just not extended to the "zero templates checked" case — and it mirrors a defect CodeRabbit previously flagged on the sibling script `template-pin-drift-check.mjs` in review on PR #501 ("An empty `packages/*` result disarms the entire guard and still exits 0").

https://github.com/constructorfabric/gears-frontx/blob/330d10e6ef151bc2890712c8e6553a4fd14d730c/scripts/template-lockfile-selflink-check.mjs#L204-L271

tscbmstubp:
Re: [review](https://github.com/constructorfabric/gears-frontx/pull/525#pullrequestreview-4860749014) — fixed in df1d6b5b (current head).

`check()` now hard-fails when zero templates were actually *checked*, not just when zero were discovered: after the per-template loop, `checkedCount === 0` exits 1 with an error explaining that every discovered template was skipped (missing root `package.json` or `package-lock.json`), so a lockfile rename/move can no longer disarm the guard while it reports success (`scripts/template-lockfile-selflink-check.mjs:248`). The docblock comment above the check was extended to cover this case as well.

GeraBart (approved):

scripts/template-lockfile-selflink-check.mjs
```diff
@@ -0,0 +1,323 @@
...
+      typeof packages === 'object' && packages !== null
+        ? /** @type {Record<string, unknown>} */ (packages)[`node_modules/${dependency.name}`]
+        : undefined;
+
+    if (entry === undefined) {
+      findings.push({ templateName, dependency, lockfilePath, problem: 'entry-missing' });
+    } else if (
+      typeof entry !== 'object' ||
+      entry === null ||
+      /** @type {Record<string, unknown>} */ (entry).link !== true
```
gs-layer, line 169: The guard only checks that the lockfile package entry is an object with `link === true`, but the contract documented above and in the remediation is the full self-link shape: `{ "resolved": "", "link": true }`. As written, a corrupted entry like `{ "resolved": "../other", "link": true }` still passes even though it no longer represents the template-root self-link this check is meant to protect. Please also assert `entry.resolved === ""` and add a regression case for wrong/missing `resolved`.

gs-layer (approved):

PR #526, opened 2026-08-04
fix(arch): match doNotFollow node_modules at any depth, not just the root

### LINKED ISSUE #523: arch: root depcruise configs anchor doNotFollow to '^node_modules', missing nested node_modules
Follow-up to the Template Validate OOM found on #491 (see https://github.com/constructorfabric/gears-frontx/pull/491#issuecomment-5174363656).

`template-shell/.dependency-cruiser.cjs` bounded traversal with `doNotFollow: '^node_modules'`, which is anchored to the path root. When the CI composition step overlaid template-mfe onto the shell, `npm install` nested a `node_modules` under each `src-app/mfe_packages/*` workspace whose pins conflict with the shell's; those paths don't match the anchored pattern, so dependency-cruiser traversed the entire installed dependency tree and died on the default ~4 GB heap. Fixed there by matching at any depth: `(^|/)node_modules/`.

Two configs on the ecosystem side carry the same anchored pattern and the same latent behavior:

- `.dependency-cruiser.cjs:304` — `doNotFollow: { path: ['^node_modules', '^packages/[^/]+/dist'] }`
- `internal/depcruise-config/base.cjs:22` — `doNotFollow: '^node_modules'`

They pass today only because the ecosystem tree happens to have few/no nested `node_modules`, so the graph stays under the heap. Any future workspace whose pins force npm to nest one (or a per-package `arch:deps:core` run against a tree that has them) silently grows the graph, and the failure mode is an OOM in CI rather than a rule violation — the same way #491 hit it.

Proposed fix: the same one-line hardening as `template-shell/.dependency-cruiser.cjs` — `(^|/)node_modules/` — in both files, keeping the `^packages/[^/]+/dist` entry anchored (it is deliberately workspace-root-relative).

tscbmstubp:
Closes #523. Follow-up to the Template Validate OOM found on #491.

## Problem

Two ecosystem-side depcruise configs anchored `doNotFollow` to the path root:

- `.dependency-cruiser.cjs` — `doNotFollow: { path: ['^node_modules', '^packages/[^/]+/dist'] }`
- `internal/depcruise-config/base.cjs` — `doNotFollow: '^node_modules'`

npm nests a `node_modules` under a workspace whenever its pins conflict with the root's; those paths don't match the anchored pattern, so dependency-cruiser follows them and the graph silently grows until it OOMs in CI — the same failure mode #491 hit on the template composition step, instead of a rule violation.

## Fix

Match `(^|/)node_modules/` at any depth. The `^packages/[^/]+/dist` entry stays anchored — it is deliberately workspace-root-relative.

Scope boundary: this PR covers only the two **ecosystem-side** configs. `template-shell/.dependency-cruiser.cjs` — the config the OOM actually came from — is still anchored on develop; its identical hardening rides the still-open #491 and lands whenever that merges. The two PRs are independent, so merge order doesn't matter, but until both land one side stays exposed.

## Review follow-ups

- `scripts/verify-guard-configs.ts` now asserts the `(^|/)node_modules/` pattern literally in both configs (including the root config, which lives outside `internal/`) — previously it checked rule names only, so a regression back to `^node_modules` would stay silent until the next CI OOM.
- Dropped `packages/api/.dependency-cruiser.cjs`: it carried a third `doNotFollow` spelling (unanchored substring) and nothing runs it — `arch:deps:core` points at `core.cjs` directly, `test-architecture.ts` passes the root config, and the api package has no depcruise script.
- The nested-`node_modules` rationale now lives once, in `base.cjs`; the root config's comment points there.

## Verification

- `npm run arch:deps` — 0 errors. The cruise shrinks from **770 modules / 2195 dependencies** (anchored) to **338 / 1163** (fixed, after dropping the dead api config): the anchored pattern was already traversing nested `node_modules` on the current tree, so this isn't just latent.
- `npm run arch:deps:core` — no violations (133 modules, 453 dependencies).
- `npm run arch:guards` — 22/22 passed, including the two new `doNotFollow` depth assertions.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved dependency analysis by consistently excluding nested `node_modules` directories, reducing irrelevant dependency reports.
* **Chores**
  * Consolidated dependency-analysis configuration across packages.
  * Added automated validation to detect missing or incorrectly configured dependency exclusions and report configuration errors clearly.


.dependency-cruiser.cjs
```diff
@@ -301,7 +301,12 @@ module.exports = {
     // excluding `dist` deleted the very node those rules needed to match.
     // `doNotFollow` keeps the node visible as an un-traversed leaf, which is
     // all the noise reduction that was ever wanted.
-    doNotFollow: { path: ['^node_modules', '^packages/[^/]+/dist'] },
+    // `node_modules` matches at any depth: npm nests a `node_modules` under a
+    // workspace whenever its pins conflict with the root's, and an anchored
+    // `^node_modules` lets traversal descend into those — the graph then
+    // silently grows until dependency-cruiser OOMs (#523). The `dist` entry
+    // stays anchored: it is deliberately workspace-root-relative.
+    doNotFollow: { path: ['(^|/)node_modules/', '^packages/[^/]+/dist'] },
```
itechmeat, line 312: [Not a blocker] One clause worth adding to the body: template-shell's own config is still anchored on develop (`template-shell/.dependency-cruiser.cjs:200`) - the identical fix rides the still-open #491, so "the same hardening as template-shell on #491" reads as already landed when it has not yet. Worth naming that boundary and the merge order explicitly, since the template config is the one the OOM actually came from.

internal/depcruise-config/base.cjs
```diff
@@ -19,7 +19,11 @@ module.exports = {
     // It would flag legitimate standalone files like templates and test fixtures
   ],
   options: {
-    doNotFollow: '^node_modules',
+    // Matches at any depth, not just the root: npm nests a `node_modules`
+    // under a workspace whenever its pins conflict with the root's, and an
+    // anchored `^node_modules` lets traversal descend into those until
+    // dependency-cruiser OOMs (#523).
+    doNotFollow: '(^|/)node_modules/',
```
itechmeat, line 26: [Not a blocker] `verify-guard-configs.ts` asserts config structure and rule names only, never `doNotFollow` - the 20/20 would pass identically with the anchored pattern, and the root `.dependency-cruiser.cjs` is outside the script's reach entirely. A regression back to `^node_modules` stays silent until the next CI OOM; worth a follow-up assertion pinning the pattern in both configs?

.dependency-cruiser.cjs
```diff
@@ -301,7 +301,12 @@ module.exports = {
     // excluding `dist` deleted the very node those rules needed to match.
     // `doNotFollow` keeps the node visible as an un-traversed leaf, which is
     // all the noise reduction that was ever wanted.
-    doNotFollow: { path: ['^node_modules', '^packages/[^/]+/dist'] },
+    // `node_modules` matches at any depth: npm nests a `node_modules` under a
```
itechmeat: [Not a blocker] Formatting nit: this paragraph is stacked straight onto the existing eight-line `doNotFollow` block with no bare `//` separator, so lines 296-308 read as one paragraph carrying two distinct claims - #491's identical edit in template-shell does insert the separator. And the same constraint now lives in three independently worded comments (root, base.cjs, template-shell once #491 lands); one of them could own the story and the others point at it.

internal/depcruise-config/base.cjs
itechmeat, line 26: [Not a blocker] `packages/api/.dependency-cruiser.cjs:14` overrides the inherited options with a third spelling (`{ path: 'node_modules' }` - unanchored substring, no OOM risk, just looser), so the value fixed here never reaches it - and nothing seems to run that config at all: `arch:deps:core` points at `core.cjs` directly and the api package has no depcruise script. If it is dead, dropping it removes a spelling; if it is alive, inheriting would.

itechmeat (approved):
Correct, minimal, and well-evidenced fix. I verified the regex semantics and the boundary-rule behaviour independently and found no enforcement loss: matched node_modules stay visible as un-traversed leaves, so every boundary rule still fires on the first hop, and the new pattern is even stricter at the root than the old one. The remaining notes are coordination clarity (template-shell's own copy of this fix rides the still-open #491) and durability (nothing pins the new pattern, so a regression to the anchored form would be silent again). Approving.

.dependency-cruiser.cjs
tscbmstubp: Done — the PR body now names the boundary explicitly: this PR covers only the two ecosystem-side configs, `template-shell/.dependency-cruiser.cjs` (where the OOM actually surfaced) is still anchored on develop, and its identical hardening rides the still-open #491.

internal/depcruise-config/base.cjs
tscbmstubp: Added in 868b00f8, in this PR rather than a follow-up: `verifyDoNotFollowPatterns()` asserts the literal `(^|/)node_modules/` pattern in both `base.cjs` and the root `.dependency-cruiser.cjs` (the root lives outside `internal/`, but it is a boundary guard all the same). A regression back to `^node_modules` now fails `arch:guards` instead of waiting for the next CI OOM. 22/22 passing.

.dependency-cruiser.cjs
tscbmstubp: Both done in 868b00f8: the bare `//` separator now splits the two claims, and the nested-`node_modules` story is owned by `base.cjs` — the root comment points there instead of restating it. Once #491 lands, template-shell's copy is the remaining candidate to point at `base.cjs` too, but that belongs on that branch.

internal/depcruise-config/base.cjs
tscbmstubp: Dead confirmed, dropped in 868b00f8. Nothing runs it: `arch:deps:core` points at `core.cjs` directly, both `test-architecture.ts` invocations pass `--config .dependency-cruiser.cjs` from the root, no CI step or package script references it, and depcruise is never invoked with `packages/api` as cwd (which is the only way it would auto-discover). Its four modules leaving the cruise is the 342→338 delta in the updated verification numbers.

gs-layer (approved):
Approved. The ecosystem-side depcruise configs now match node_modules at any depth, the exact pattern is pinned by arch:guards, and the dead API-specific config was removed with a clear verification trail. Merge-order context: this does not need to wait for #491; #491 carries the separate template-shell copy, so until both land one side remains exposed but the order is not material.

gs-layer (approved):

GeraBart (approved):

Issue #503, opened 2026-07-31
[BUG] CJS TypeScript consumers resolve ESM declarations for every ecosystem package


SysoevAndrey:
## Description

Every dual-format FrontX package advertises a CommonJS runtime entry but exposes only its ESM declaration file through the top-level `types` export condition. A TypeScript CommonJS consumer using `moduleResolution: Node16` or `NodeNext` therefore rejects the package even though runtime `require()` and the generated `.d.cts` file exist.

This is a repo-wide package-contract defect, not a ui-kit-specific issue. It currently reproduces for:

- `@gears-frontx/mfes`
- `@gears-frontx/gts-plugin`
- `@gears-frontx/api` (including the same shape on `./types`)
- `@gears-frontx/cli`
- `@gears-frontx/ui-kit`

Related architecture decision: #462. That issue decides whether FrontX should retain dual ESM/CJS support or become ESM-only; this issue tracks making the published package metadata and TypeScript declarations correct for the chosen outcome.

## Environment

- **FrontX package versions**: `0.3.0-alpha.x`
- **TypeScript**: `5.4.2`
- **Module resolution**: `Node16` / `NodeNext`
- **Node.js**: reproduced on `v24.13.0`
- **OS**: macOS
- **Packages affected**: all five packages listed above

## Steps to Reproduce

1. Build the workspace packages so both `dist/index.d.ts` and `dist/index.d.cts` exist.
2. Create a CommonJS TypeScript consumer such as `consumer.cts`:

```ts
import * as frontx from '@gears-frontx/mfes';

void frontx;
```

3. Type-check it with a config equivalent to:

```json
{
  "compilerOptions": {
    "module": "Node16",
    "moduleResolution": "Node16",
    "target": "ES2022",
    "noEmit": true,
    "strict": true
  }
}
```

4. Repeat with `gts-plugin`, `api`, `cli`, and `ui-kit`.

## Expected Behavior

If the package advertises a `require` condition, TypeScript CommonJS consumers resolve the generated `.d.cts` declaration and type-check successfully.

## Actual Behavior

All five packages fail with the same diagnostic:

```text
The current file is a CommonJS module whose imports will produce 'require' calls; however, the referenced file is an ECMAScript module and cannot be imported with 'require'.
```

The manifests currently place a single ESM declaration before both runtime branches:

```json
{
  "type": "module",
  "exports": {
    ".": {
      "types": "./dist/index.d.ts",
      "import": "./dist/index.js",
[...]

Issue #462, opened 2026-07-06
[ARCH] ESM-first module-format constraint conflicts with Publishing Pipeline's dual CJS/ESM package metadata contract


sulasen:
## Topic
`cpt-hai3-constraint-esm-first-module-format` — DESIGN's ESM-only constraint contradicts the Publishing Pipeline feature's build contract.

## Context
[DESIGN.md ("ESM-First Module Format")](https://github.com/constructorfabric/gears-frontx/blob/f69fdab7508a683e0b15c96ff2d29a120e1e1653/architecture/DESIGN.md#L288-L296):
> All packages output ESM as the primary module format... **CJS is not supported.**

[feature-publishing-pipeline/FEATURE.md ("Package Metadata Contract")](https://github.com/constructorfabric/gears-frontx/blob/f69fdab7508a683e0b15c96ff2d29a120e1e1653/architecture/features/feature-publishing-pipeline/FEATURE.md#L269-L285), which explicitly claims to cover this same constraint:
> - `"exports"` must expose both ESM (`import`) and CJS (`require`) conditions
> - `"main"` points to `.cjs` entry
> - `dist/` must contain `index.cjs` (CJS)

The feature spec requires a CJS build artifact the DESIGN constraint says is unsupported.

## Proposed Changes
Either:
1. Drop the CJS requirements from the Package Metadata Contract to align with ESM-only, or
2. Update the DESIGN constraint (and its ADR) to allow dual CJS/ESM output if that's now the real target.

## Affected Components
- [x] Multiple packages

## References
`cpt-hai3-constraint-esm-first-module-format` — [DESIGN.md](https://github.com/constructorfabric/gears-frontx/blob/f69fdab7508a683e0b15c96ff2d29a120e1e1653/architecture/DESIGN.md#L288-L296) ("ESM-First Module Format"), [feature-publishing-pipeline/FEATURE.md](https://github.com/constructorfabric/gears-frontx/blob/f69fdab7508a683e0b15c96ff2d29a120e1e1653/architecture/features/feature-publishing-pipeline/FEATURE.md#L269-L285) ("Package Metadata Contract"). Surfaced by an automated traceability check.

PR #529, opened 2026-08-04
feat(ui-kit): complete the MVP component set, then fix what clicking through found

SysoevAndrey:
Completes the MVP component set for `@gears-frontx/ui-kit` (stage 3 of the delivery plan in `packages/ui-kit/design-notes.md`), then fixes what clicking through the result exposed.

Twenty commits, one component or one concern each. The package stays `private: true` — publication is gated on #495, not on this PR.

## Components added — 10, taking the set to 19

| Group | Added |
|---|---|
| Overlays | `Dialog`, `DropdownMenu`, `Tooltip`, `Toast` |
| Structure | `Card`, `Tabs`, `Badge`, `Separator`, `Skeleton` |
| Data | `Table` (primitive markup only) |

Behaviour comes from Base UI primitives, variants from CVA, styles from CSS Modules translated from shadcn's `base-vega` registry variant (the Base UI one — the Radix-based styles are not this kit's source). Each component ships a unit test and a usage doc under `dist/docs/` for agents to read from `node_modules`.

Every overlay portals its popup to `<body>` and takes a `container` prop, so a theme scoped to a subtree can pull the popup into it — the failure mode Select hit earlier, where a portalled popup lost the theme.

## Defects found by clicking through, not by tests

Each of these was invisible to the 216 unit tests and to CSS review; all were measured in a browser against the packed tarball installed into a clean Vite consumer.

- **The kit assumed a Tailwind preflight it does not ship.** Styles were translated from Tailwind, which sets `border-box` globally; without it every declared size collided with the kit's own padding and border. Input rendered 46px instead of 36, Textarea 82 instead of 64, Dialog's popup 498 instead of 448. One case was functional rather than cosmetic: Select's item overflowed its popup, and because the popup clips overflow, **the checkmark showing which option was selected was pushed off-screen entirely**.
- **`theme.css` defined tokens and painted nothing.** Under `prefers-color-scheme: dark` the tokens flipped, but `<body>` stayed transparent with UA-black text — components that paint their own surface went dark while the page behind them did not. `[data-theme='light']` was not a working scope at all: the light values lived only on `:root`, so a light-forced subtree inherited dark tokens.
- **Dialog stretched to fit wide content.** Its grid column was `auto`, so a Table inside it grew the track past the popup's own `max-width` — the table's scroll wrapper rendered 957px and never scrolled, because an auto track gave it no width to scroll against.
- **The destructive palette failed WCAG AA in three places** — the solid button in dark mode at 2.77:1, the same button on hover in light mode at 4.34:1, and DropdownMenu's destructive item while highlighted at 3.97:1. All three now pass in both themes.
- **Toast's card was 2px taller than the height Base UI measured for it**, and that height feeds the stack's own positioning arithmetic, so the gap between stacked cards drifted.
- **A consumer importing one component got the whole kit.** See below.

## Tree shaking

The kit shipped one `dist/index.js` and one `dist/index.css`. Now: Vite library mode, one entry per component, each chunk importing its own stylesheet, plus `./*` subpath exports.

Measured on a Button-only consumer page, React externalized, built against the packed tarball:

| Bundler | JS before | JS after |
|---|---|---|
| raw esbuild | 233,672 B | 9,383 B |
| webpack 5 | 62,924 B | 8,581 B |
| Vite / Rollup | already eliminated | unchanged |

CSS, under Vite: **33.06 kB → 4.38 kB** (5.17 → 1.08 kB gzipped).

Two things worth knowing, because both contradict the assumption this started from:

1. What the old build retained was **Base UI's own submodules** — `SelectRoot`, `MenuRoot`, `ToastRoot`, `FloatingFocusManager`, `@floating-ui/dom`, roughly 290 files — none reachable from `Button`. Rollup cascades "this wrapper is unused" through to "so is its exclusive import"; esbuild's bundler does not. Annotating the `cva()` calls with `/* @__PURE__ */`, which this was first attributed to, produces **byte-identical output** and fixes none of it.
2. **esbuild does not tree-shake CSS at all.** A barrel import under esbuild drops the unused components' JS but still ships every component's stylesheet, with `sideEffects` set to `["**/*.css"]`, `["*.css"]` or `false` alike. On esbuild, the `./button` subpath is the only way to get per-component CSS. `README.md` and `llms.txt` now say which bundler makes which claim true.

### Public contract changes

- `./styles.css` is **removed** — component CSS arrives with the component. `./theme.css` is unchanged and still imported once.
- **CJS dropped, ESM only**, as the sibling `react-kit` does. Note for the standing issues: dropping `main`/`require` resolves ui-kit's line in #503 (a CJS TS consumer can no longer resolve this package's ESM declarations through a `require` condition, because there isn't one) and is a data point for the ESM-only side of #462 — those issues should be updated against this rather than drift.
- New `./*` subpath exports; `./components/*` and `./__test-utils__/*` explicitly denied, so a deep path that type-checks cannot resolve to a file that was never built.

Not adopted from `react-kit`: `publishConfig.directory` with a generated `package.json`. Exports point into `dist/` from the real package root so the existing publish workflow keeps working.

## Acceptance

`scripts/verify-consumer.sh` grew from one leg to four, because a Vite-only gate proves the CSS half of the tree-shaking fix and silently skips the JS half:

| Leg | Asserts |
|---|---|
| `[vite]` | the ecosystem's own tooling; JS and CSS, both directions |
| `[esbuild-barrel]` | JS absence — where the 25× regression lived |
| `[esbuild-subpath]` | CSS absence — only assertable via a subpath under esbuild |
| `[types]` | `tsc` under both `nodenext` and `bundler`, root plus three subpaths |

Every assertion was verified by breaking it. The two that matter most were proven against the real pre-repackaging tarball rather than a synthetic mutation, so the gate is known to catch the regression it exists for. Two earlier bugs in the gate itself were fixed on the way: a probe guard that was unreachable under `set -euo pipefail` (the script died at the assignment, so its own diagnostic never printed) and absence checks that passed silently when their glob matched no file.

## Also in here

- **`prefers-reduced-motion` handled by what actually moves.** Transforms and geometry are suppressed; colour and opacity crossfades stay, because a fade is not motion and removing it costs the UI something for no accessibility gain. Toast's loading spinner slows to a quarter speed rather than freezing — it is the only cue that a loading toast is still pending.
- **Base UI's `data-instant` honoured** on the two primitives that actually emit it (Menu and Tooltip, verified in the pinned 1.6.0 source), and deliberately not on Dialog, Select or Toast, which never do.
- **Corrected three false statements about the build**: nothing was ever bundled — `@base-ui/react` and CVA are external imports — which inverted `toast.md`'s note about whether a consumer's own Base UI can share the kit's Toast context. It can, when the versions dedupe to one copy.
- **The build tooling now type-checks**, under its own tsconfig. It was checked by nothing before, which is how a downlevel-iteration error sat in it unnoticed.
- **`design-notes.md`** records the ten components `insight-front` needs that the kit lacks (`toggle`, `toggle-group`, `popover`, `sheet`, `preview-card`, `alert`, `avatar`, `collapsible`, `empty`, `spinner`) as stage 4, ahead of publication, plus why `calendar`, `chart` and `sidebar` stay out of scope.

## Verification

216 tests / 21 files · `type-check`, `type-check:tools`, `type-check:test` · `eslint --max-warnings 0` · `lint:deps` · `arch:check` 12/12 · `arch:deps` 0 errors · `verify-consumer.sh`, all four legs.

`arch:deps` also reports ~6.1k warnings, of which ~342 are new here. They carry no signal: the rules producing them (`mfes-opaque-schema-surface` and the other `STUB:` entries in `.dependency-cruiser.cjs`) are declared with empty `from`/`to` matchers, so each one matches every dependency edge in the repo — the new count is just the new `public.ts` entry files adding edges.

The `audit-high` pre-commit hook is bypassed on these commits: it fails on a pre-existing repo-wide advisory unrelated to this package.

## Not in scope

Publication. `private: true` stays until #495 approves the package's architecture ownership and version policy and the temporary `artifacts.toml` ignore is replaced with real CDSL artifacts. The kitchen-sink example app and the three composition recipes are stage 5.

## Summary by CodeRabbit

* **New Features**
  * Added Badge, Card, Dialog, Dropdown Menu, Separator, Skeleton, Table, Tabs, Toast, and Tooltip components.
  * Expanded public access to UI-kit components and their types.
  * Added accessible interactions, variants, portal support, responsive layouts, and reduced-motion behavior.
  * Added Select positioning and collision controls.
  * Added automatic dark mode, scoped theming, page surface colors, and expanded design tokens.
  * Component styles load independently, improving tree-shaking and reducing unused CSS.

* **Documentation**
  * Added comprehensive usage, accessibility, theming, portal, and composition guidance.
  * Updated package usage and build instructions.


packages/ui-kit/design-notes.md
```diff
@@ -113,8 +168,17 @@ confirmation dialog) once the components they compose exist.
...
 2. Tokens polish + first component batch (forms) on `@base-ui/react`.
-3. Remaining components.
-4. AI docs + kitchen-sink example app; satisfy the #495 publication gates,
+3. Remaining components — done, the 19-component MVP set exists.
+4. The ten `insight-front` gap components. Before the release step, not after:
+   they are part of the MVP set the `private` gate names, and the kitchen-sink
+   app and composition recipes should cover the whole set once rather than be
+   extended right after publication.
+5. AI docs + kitchen-sink example app; satisfy the #495 publication gates,
    remove the temporary artifact ignore, then flip `private` and publish.
```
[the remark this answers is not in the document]
SysoevAndrey: Fixed in f1854554 — the intro now says steps 1–3 are done, 4–5 remain. One correction to the finding: the PR does not claim a 29-component MVP; it ships 19 and names the ten `insight-front` gap components as stage 4, explicitly future work. The statuses themselves were right — only the "log of completed phases" framing was wrong.

packages/ui-kit/src/components/card/card.md
```diff
@@ -0,0 +1,132 @@
...
+  an item in a list of similar items) with a clear visual boundary.
+- A summary or preview that pairs a title/description with an action
+  (`CardAction`) or a set of actions (`CardFooter`).
+
+## When not to use
+
+- A focused task that should block the rest of the page — use `dialog`.
+- Purely tabular data — a plain `<table>` (or a future kit table
+  component) fits rows/columns better than a boundary-and-padding
+  container.
```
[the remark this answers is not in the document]
SysoevAndrey: Fixed in f1854554 — points at the kit's `table` component now.

packages/ui-kit/src/components/card/card.module.css
```diff
@@ -0,0 +1,139 @@
...
+   * through *any* matching ancestor regardless of depth, but a custom
+   * property's used value is whatever a descendant's nearest ancestor (or
+   * itself) declares — and every `.card` unconditionally re-declares these,
+   * so a size="default" Card nested inside a size="sm" one asserts its own
+   * defaults instead of inheriting the outer card's.
+   */
+  --card-spacing: 1.5rem;
+  --card-title-font-size: 1rem;
+  --card-title-line-height: 1.5rem;
+  gap: var(--card-spacing);
```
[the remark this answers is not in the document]
SysoevAndrey: Skipping. This repo runs no Stylelint (lint is `eslint . --max-warnings 0`, see the root `package.json`), so nothing blocks. And the empty `.sizeDefault` block is load-bearing, per the comment right above it: CSS Modules only emits a class mapping for classes that appear in a rule, and cva's `defaultVariants` targets `styles.sizeDefault` — delete the block and `size="default"` loses its DOM-visible class.

packages/ui-kit/src/components/dialog/dialog.module.css
```diff
@@ -0,0 +1,163 @@
...
+   * Substitutes for the preflight the kit lacks: without it, content-box
+   * (this kit's ambient default) piles the padding and border on top of
+   * the declared width, rendering 498px instead of 448px at the sm+
+   * breakpoint below — same acute case as tabs.module.css/
+   * badge.module.css, scoped the same way (this class only).
+   */
+  box-sizing: border-box;
+  position: fixed;
+  top: 50%;
+  left: 50%;
```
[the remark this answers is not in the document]
SysoevAndrey: Fixed in 59bdc788 — `max-height: calc(100% - 2rem)` mirroring the max-width cap (both resolve against the viewport on a fixed element), plus `overflow-y: auto` so the popup becomes its own scroll container. Known trade-off, noted in the CSS: the absolutely-positioned close button scrolls with the content.

packages/ui-kit/src/components/dialog/dialog.module.css
```diff
@@ -0,0 +1,163 @@
...
+  line-height: 1.25rem;
+}
+
+.srOnly {
+  position: absolute;
+  width: 1px;
+  height: 1px;
+  padding: 0;
+  overflow: hidden;
+  clip: rect(0, 0, 0, 0);
```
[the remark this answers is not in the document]
SysoevAndrey: Fixed in 59bdc788 — `clip-path: inset(50%)`. The kit ships ESM-only to evergreen browsers, so the legacy `clip` fallback isn't carrying anything.

packages/ui-kit/src/components/dialog/dialog.tsx
```diff
@@ -0,0 +1,109 @@
...
+        {children}
+        {showCloseButton && (
+          <DialogPrimitive.Close
+            className={styles.closeButton}
+            render={<Button variant="ghost" size="icon" />}
+          >
+            <CloseIcon />
+            <span className={styles.srOnly}>Close</span>
+          </DialogPrimitive.Close>
+        )}
```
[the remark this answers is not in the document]
SysoevAndrey: Keeping the opt-out; test upgraded in 59bdc788. `DialogClose` is exported precisely for consumer-composed closes (a footer Cancel button is the documented example), and dialog.md already states the contract: only pass `showCloseButton={false}` when composing your own `DialogClose` inside the popup. Removing the prop would force two close buttons on every dialog with a Cancel action. The absence-only test is replaced with the sanctioned shape: a consumer-composed `DialogClose` inside the popup that actually closes the dialog.

packages/ui-kit/src/components/input/input.module.css
```diff
@@ -18,6 +26,8 @@
   font-size: 1rem;
   line-height: 1.5rem;
   outline: none;
+  /* Colour-only transition, no transform/geometry — not motion under
+     WCAG 2.3.3, so deliberately left unguarded under prefers-reduced-motion. */
```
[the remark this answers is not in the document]
SysoevAndrey: Fixed in f1854554 — and in all ten files carrying that comment (button, input, textarea, select, radio-group, checkbox, switch, tabs, badge, table), not just the three named. The declarations are unchanged: a focus ring's box-shadow crossfade is no more motion under WCAG 2.3.3 than the colour ones, so leaving them unguarded under prefers-reduced-motion stays correct — the comments just now name what's actually in the list.

packages/ui-kit/src/components/tabs/tabs.module.css
```diff
@@ -0,0 +1,244 @@
...
+.variantLine .trigger[data-active]::after {
+  opacity: 1;
+}
+
+.content {
+  flex: 1 1 0%;
+  font-size: 0.875rem;
+  line-height: 1.25rem;
+  outline: none;
+}
```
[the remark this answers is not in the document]
SysoevAndrey: Fixed in feba1920 — and the premise is stronger than the finding states: the panel is focusable out of the box, not only via a consumer's `tabIndex`. Base UI puts the open panel in the tab order itself (`tabIndex: open ? 0 : -1`, TabsPanel.js:83), so keyboard users land on it right after the tablist. Added `.content:focus-visible` with the kit's ring idiom plus the forced-colors fallback every other ring in the kit has, and a regression test guarding the focusability premise (the ring itself is CSS, which jsdom doesn't apply).

packages/ui-kit/src/components/toast/toast.module.css
```diff
@@ -0,0 +1,343 @@
...
+ * (Dialog/DropdownMenu/Tooltip all use 50) — a toast should surface over an
+ * open modal, matching the source and typical toast UX.
+ */
+.toast {
+  --gap: 0.75rem;
+  --height: var(--toast-frontmost-height, var(--toast-height));
+  --offset-y: calc(
+    (var(--toast-offset-y) * -1) + (var(--toast-index) * var(--gap) * -1) +
+      var(--toast-swipe-movement-y)
+  );
```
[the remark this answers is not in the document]
SysoevAndrey: Skipping — this repo runs no Stylelint (lint is `eslint . --max-warnings 0`) and no Prettier, so there is no `scss/operator-no-newline-after` to fail and no formatter to align. The wrapping matches the file's existing style.

packages/ui-kit/src/components/tooltip/tooltip.module.css
```diff
@@ -0,0 +1,135 @@
...
+  translate: 0 calc(-50% - 2px);
+}
+
+.arrow[data-side='top'] {
+  bottom: -0.625rem;
+}
+
+.arrow[data-side='bottom'] {
+  top: 0.25rem;
+}
```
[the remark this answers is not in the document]
SysoevAndrey: Checked the arithmetic — the asymmetry is superficial; the rendered result is already the mirror the finding asks for. Arrow is 10px, shared translate is `calc(-50% - 2px)` = −7px. Side `top`: `bottom: -0.625rem` puts it at [H, H+10] relative to the popup, translated to [H−7, H+3] — 3px past the bottom edge, 7px overlapping. Side `bottom`: `top: 0.25rem` puts it at [4, 14], translated to [−3, +7] — 3px past the top edge, 7px overlapping. Identical protrusion and seam on both sides; the offset values differ because the same upward translate tucks one side in and pushes the other side out. Bottom-side placement was also verified in-browser on this branch (flipped positioning with collision avoidance).

packages/ui-kit/src/components/tooltip/tooltip.test.tsx
```diff
@@ -0,0 +1,110 @@
...
+          Hover me
+        </TooltipTrigger>
+        <TooltipContent>Saved successfully</TooltipContent>
+      </Tooltip>,
+    );
+    fireEvent.mouseEnter(screen.getByRole('button', { name: 'Hover me' }));
+    // Absence can't be `waitFor`-ed positively; give the (skipped) open delay
+    // a moment to have fired if the disabled check were broken.
+    await new Promise((resolve) => setTimeout(resolve, 20));
+    expect(screen.queryByText('Saved successfully')).toBeNull();
```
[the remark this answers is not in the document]
SysoevAndrey: Skipping. Pinning closed-at-599/open-at-600 couples the test to when Base UI arms its hover-intent rest timer (mousemove-driven, see the comment in the test) — that boundary is Base UI's contract, not this kit's. The kit's contract is the design call the test names: mounting `TooltipProvider` must not silently speed up opening — which the closed-at-200ms / open-by-800ms shape asserts with margin on both sides. For the disabled case: that test runs `delay={0}`, so a broken disabled check would open essentially immediately; 20ms of real time is enough to catch it.

packages/ui-kit/src/tokens.test.ts
```diff
@@ -42,16 +51,106 @@ describe('theme tokens', () => {
...
+  // runs back to the previous `}`) swallows whatever comment precedes the
+  // rule, and prose mentioning a dotted filename (`theme.css`,
+  // `tokens.test.ts`) parses as a bogus leading class.
+  function extractRules(css: string) {
+    const withoutComments = css.replace(/\/\*[\s\S]*?\*\//g, ' ');
+    return Array.from(withoutComments.matchAll(/([^{}]+)\{([^{}]*)\}/g), (match) => ({
+      selector: (match[1] ?? '').trim(),
+      body: match[2] ?? '',
+    }));
+  }
```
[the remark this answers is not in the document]
SysoevAndrey: Fixed both halves in 198767f5. `definedTokens` is now built from comment-stripped CSS (theme.css's prose does quote tokens with values, so the false-positive was real). `tokenMap` now captures every declaration rather than only custom properties, which puts `color-scheme` inside both the dark-block identity check and the light↔dark parity check.

SysoevAndrey:
Review addressed in 59bdc788…f1854554 — replies on each thread. Eight findings fixed (Dialog viewport cap + clip-path, Tabs panel focus ring, Toast `closeLabel` — the nitpick, taken: it is the card's only kit-authored string — token-guard hardening, and the prose drift), four declined with reasoning on the thread (the two Stylelint findings — this repo runs no Stylelint, and one flagged block is load-bearing for CSS Modules; the tooltip arrow, whose sides already mirror once the shared translate is applied; and the fake-timer rewrite, which would pin Base UI's internal timing rather than the kit's contract). 218 tests, type-check and eslint clean.

packages/ui-kit/package.json
```diff
@@ -3,17 +3,19 @@
...
   "type": "module",
-  "main": "./dist/index.cjs",
-  "module": "./dist/index.js",
   "types": "./dist/index.d.ts",
   "exports": {
     ".": {
       "types": "./dist/index.d.ts",
-      "import": "./dist/index.js",
-      "require": "./dist/index.cjs"
+      "import": "./dist/index.js"
```
itechmeat, line 10: Dropping `main`/`require` here quietly resolves ui-kit's line in #503, which itself asks not to fix packages one by one. Could you add a one-line note to the PR body referencing #503/#462 so those issues get updated instead of drifting?

packages/ui-kit/src/components/dialog/dialog.tsx
```diff
@@ -0,0 +1,109 @@
...
+      <DialogPrimitive.Backdrop className={styles.backdrop} />
+      <DialogPrimitive.Popup className={cx(styles.popup, className)} {...props}>
+        {children}
+        {showCloseButton && (
+          <DialogPrimitive.Close
+            className={styles.closeButton}
+            render={<Button variant="ghost" size="icon" />}
+          >
+            <CloseIcon />
+            <span className={styles.srOnly}>Close</span>
```
itechmeat: You added `closeLabel` to Toaster in this same PR for exactly this reason - the one kit-authored string a non-English app needs to replace. Dialog's built-in close button has the same hardcoded "Close" but no prop. Should they be symmetric before the package goes public?

packages/ui-kit/src/components/skeleton/skeleton.md
```diff
@@ -0,0 +1,143 @@
...
+// does — a hardcoded `aria-busy="true"` would never announce completion.
+function LoadingCard() {
+  const [loading, setLoading] = useState(true);
+  // ...fetch real data, then setLoading(false) when it arrives.
+
+  return (
+    <div role="status" aria-busy={loading}>
+      {/* Visually hidden, inlined: matches the kit's own .srOnly recipe
+          (dialog.module.css), which is module-local to Dialog and not
+          exported for reuse here. */}
```
itechmeat, line 101: The inlined sr-only snippet in skeleton.md still uses the deprecated `clip: rect(...)` form that this same PR replaced with `clip-path: inset(50%)` in dialog's CSS. Since agents will copy this doc verbatim, could the snippet match the kit's current recipe?

packages/ui-kit/design-notes.md
```diff
@@ -40,36 +40,65 @@ generate screens consistently.
...
+  `prefers-color-scheme` with no attribute and no CSS of its own.
 - Publishing: version-gated like every ecosystem package. `private: true`
   remains in place until both the MVP component set lands and #495 approves
   the package's architecture ownership, traceability, and version policy. The
   required CDSL artifacts must then replace the temporary `artifacts.toml`
   ignore. Only after all of those gates pass is flipping `private` the release
   act.
 
-## Component set (MVP, 19 components)
+## Component set (MVP, 29 components)
```
itechmeat, line 95: The component table reads as 29 shipped components, but 10 of them are the planned insight-front gap set - the marker lives only in the prose around it. Could the table itself flag the unbuilt ten so a grep-level reader gets the right answer?

packages/ui-kit/src/components/tooltip/tooltip.test.tsx
```diff
@@ -0,0 +1,110 @@
...
+    const trigger = screen.getByRole('button', { name: 'Hover me' });
+    fireEvent.mouseEnter(trigger);
+    // With a real (non-zero) open delay, Base UI defers to its "hover
+    // intent" rest timer, which only arms on a mousemove after entry — a
+    // bare mouseenter never opens it (unlike the delay={0} tests above).
+    fireEvent.mouseMove(trigger);
+    // The design call here (see tooltip.tsx) is that mounting TooltipProvider
+    // must not silently speed up opening — Base UI's per-trigger 600ms
+    // default should still apply unless a delay is passed explicitly.
+    await new Promise((resolve) => setTimeout(resolve, 200));
```
itechmeat, line 111: This assertion races a real 200ms sleep against Base UI's 600ms open delay - on a loaded CI runner the sleep can overshoot and the "not yet open" check turns flaky. Would fake timers (or a wider margin) for this one case be worth it?

packages/ui-kit/src/components/toast/toast.tsx
```diff
@@ -0,0 +1,273 @@
...
+      strokeLinecap="round"
+      strokeLinejoin="round"
+      aria-hidden="true"
+    >
+      <path d="M18 6 6 18M6 6l12 12" />
+    </svg>
+  );
+}
+
+function ToastTypeIcon({ type }: { type: string | undefined }) {
```
itechmeat: House style asks for `type?: string` over `type: string | undefined` in property position - as written, a call site omitting the prop gets a type error instead of the no-icon default the switch already implements.

packages/ui-kit/src/components/dropdown-menu/public.ts
```diff
@@ -0,0 +1,28 @@
+export {
+  DropdownMenu,
```
itechmeat, line 2: The root pass-throughs (`DropdownMenu`, `Dialog`, `Tooltip`, `Select`) export no props type, so a consumer writing a typed wrapper has nothing to import - Base UI is the kit's dependency, not theirs. Would re-exporting `Root.Props` aliases from each `public.ts` close that?

packages/ui-kit/src/components/toast/toast.tsx
```diff
@@ -0,0 +1,273 @@
...
+      strokeLinecap="round"
+      strokeLinejoin="round"
+      aria-hidden="true"
+    >
+      <path d="M21 12a9 9 0 1 1-6.219-8.56" />
+    </svg>
+  );
+}
+
+function CloseIcon({ className }: { className?: string }) {
```
itechmeat, line 104: `CloseIcon` is byte-identical in dialog and toast, and the `.svgIcon` block repeats in four modules. If the concern is per-entry chunking, has a shared internal module actually been measured to regress the flat dist layout, or is that assumption worth verifying once?

packages/ui-kit/src/components/dialog/dialog.module.css
```diff
@@ -0,0 +1,174 @@
+/*
+ * Dialog styles, translated from shadcn/ui (MIT) to CSS Modules.
+ * Consumes only theme.css variables — no raw colors.
+ */
+
+.backdrop {
+  position: fixed;
+  inset: 0;
+  isolation: isolate;
+  z-index: 50;
```
itechmeat, line 10: `z-index: 50` appears in five modules and Toast jumps to 1000 - a consumer with their own header/drawer stack has no seam to slot the kit into. Would `--z-overlay` / `--z-toast` tokens next to `--radius` be worth the two lines, or is that deliberately deferred until a real consumer hits it?

itechmeat (approved):
Solid, careful work - the component set does what the PR promises, the fixes are backed by actually clicking through and measuring rather than assumption, and the reasoning behind every non-obvious choice is written down where the next editor will find it. I found nothing that blocks approval: CI is green and the earlier review threads are all genuinely settled, including the declined ones, which hold up under checking. My comments are consistency and polish items - mostly small parity and doc drifts the PR itself introduced along the way.

packages/ui-kit/package.json
SysoevAndrey: Added to the PR body's public-contract section: dropping `main`/`require` resolves ui-kit's line in #503 (no `require` condition left for a CJS TS consumer to resolve ESM declarations through) and is a data point for the ESM-only side of #462.

packages/ui-kit/src/components/dialog/dialog.tsx
SysoevAndrey: Fixed in 2d4e6f40 — `DialogContent` now takes `closeLabel` (default `'Close'`), same contract and doc wording as `Toaster`'s. Symmetric before the package goes public, as you said.

packages/ui-kit/src/components/skeleton/skeleton.md
SysoevAndrey: Fixed in b8d1d274 — the snippet now uses `clipPath: 'inset(50%)'`, matching the recipe this PR moved Dialog's CSS to.

packages/ui-kit/design-notes.md
SysoevAndrey: Fixed in b8d1d274 — the ten planned entries are ⏳-marked in the table itself now, with a one-line legend above it, so a grep-level reader gets 19 built / 10 planned without the surrounding prose.

packages/ui-kit/src/components/tooltip/tooltip.test.tsx
SysoevAndrey: Fixed in b19dd28b — took the wider-margin direction rather than fake timers: the 'still closed' assertion is now guarded on measured elapsed time, so it's only made while the test is actually inside the 600ms window it reasons about. On an overshooting runner the check degrades to the (still asserted) 'opens by 800ms' half instead of failing falsely. Fake timers were avoided for the reason on the earlier CodeRabbit thread — they'd couple the test to when Base UI's hover-intent rest timer arms.

packages/ui-kit/src/components/toast/toast.tsx
SysoevAndrey: Fixed in a73c36f1 — `type?: string`. For the record the call site couldn't actually hit the type error (internal component, single caller always passes `item.type`), but house style is house style.

packages/ui-kit/src/components/dropdown-menu/public.ts
SysoevAndrey: Fixed in da91256d — `DialogProps`, `DropdownMenuProps` (and `DropdownMenuSubProps`, which had the same gap), `TooltipProps` and `SelectProps` are exported as `Root.Props` aliases, the idiom `TooltipProviderProps` already established. `SelectProps` stays generic (`<Value, Multiple>`) like the root itself.

packages/ui-kit/src/components/toast/toast.tsx
SysoevAndrey: Not measured, and you're right that the assumption deserves the check — for the record Rollup would put a shared internal module into `dist/chunks/` exactly like the sibling react-kit's output (86 shared chunks there), so there's no reason to expect a regression. Deferring the dedup itself: it's a refactor touching four components with zero behaviour change, and this PR is already at 30+ commits. Filed mentally against stage 4, where new components would otherwise copy the same icon a fifth time.

packages/ui-kit/src/components/dialog/dialog.module.css
SysoevAndrey: Deliberately deferred until a real consumer hits it, as you guessed — and one is scheduled: stage 4 is the insight-front gap set, and insight-front has exactly the header/drawer stack that would exercise a `--z-overlay`/`--z-toast` seam. Doing it then means the tokens get designed against a real stacking context instead of a guessed one. The five hardcoded 50s are at least uniform, so the later migration is mechanical.

gs-layer (changes requested):
**Verdict by area:**

| Area | Verdict | Blockers |
|---|---|---|
| Component logic | Request changes (low bar) | 2 medium |
| CSS / theming | Approve | 0 |
| Build / tree-shaking | Comment | 0 critical/high, 4 medium |
| Security | Clean | 0 |
| Test coverage | Good overall, some gaps | 2 high (coverage, not code) |

All of the bugs claimed as fixed in the PR description (preflight/border-box, theme painting, Dialog+Table overflow, destructive-palette WCAG contrast, Toast's 2px height mismatch, tree-shaking) were independently verified as genuinely fixed, with no new regressions introduced.

---

### 🟠 Worth fixing before merge

**1. `Dialog modal={false}` doesn't actually produce a non-modal dialog**
`packages/ui-kit/src/components/dialog/dialog.tsx:79`
`DialogContent` always renders `<DialogPrimitive.Backdrop>` regardless of `modal`. `dialog.md` documents `modal` as a working root prop, but there's no way to get an actual non-modal dialog — the backdrop is always `position:fixed; inset:0` with no `pointer-events:none`, and clicking it still closes the dialog. The doc also contradicts itself elsewhere by calling non-modal "out of scope for MVP".
→ Either drop `modal` from the documented prop table, or add a `showBackdrop?: boolean` (mirroring the existing `showCloseButton`).

**2. The toast region's `aria-label="Notifications"` can't be localized**
`packages/ui-kit/src/components/toast/toast.tsx:240-244`
The docstring claims `closeLabel` is "the one string" a non-English app needs to translate. In reality Base UI puts `role="region" aria-label="Notifications"` on the whole `Viewport`, and the kit gives no way to override it — `ToasterProps` extends `Provider`'s props, not `Viewport`'s. A screen-reader user will hear the English landmark even after the close button is translated.
→ Add `label?: string` (`@default 'Notifications'`) to `ToasterProps` and pass it as `aria-label` to `Viewport` (Base UI merges the consumer prop after the default, so this is a small change).

**3. `sideEffects` in package.json isn't guarded by any check — silent style loss under webpack**
`packages/ui-kit/package.json:63-65`
A mutation test against a real webpack build shows that if `sideEffects` is ever "simplified" to `false` (a plausible-looking future edit), *all* webpack consumers silently lose every component's styles — while all 4 legs of `verify-consumer.sh` stay green, because none of them exercise webpack. This is exactly the class of regression the verify script exists to catch, just not for this bundler.
→ Add a 5th webpack leg to `verify-consumer.sh` (measured cost: ~280ms build time).

**4. Popup positioner (`DropdownMenu`/`Tooltip`/`Select`) has no escape hatch**
`packages/ui-kit/src/components/dropdown-menu/dropdown-menu.tsx:59-96` (same pattern in tooltip.tsx, select.tsx)
Only `Pick<...Positioner.Props, 'align'|'alignOffset'|'side'|'sideOffset'>` is exposed. Inside a container with `transform`/`filter` (common for animated panels), the standard fix — `positionMethod="fixed"` — isn't reachable, nor are `collisionBoundary`/`collisionPadding`. A consumer hitting this has to bypass the kit's component entirely.
→ At minimum extend the `Pick<>`; better, add a single `positionerProps?: Omit<...Positioner.Props, 'className'>`.

**5. `container` isn't inherited by `DropdownMenuSubContent` (nested submenu)**
`packages/ui-kit/src/components/dropdown-menu/dropdown-menu.tsx:221-241`
This breaks exactly the scenario `container` was built for (a theme scoped to a subtree) — silently, one level deep: the submenu portals to `<body>` with the root theme. The existing test only covers the top-level popup, so this regression wouldn't be caught.

**6. `./*` in exports opens `dist/chunks/*` and breaks `package.json` resolution**
`packages/ui-kit/package.json:14-17`
Verified via `import.meta.resolve`: `pkg/package.json` now resolves to a non-existent `dist/package.json.js` (`ERR_MODULE_NOT_FOUND` instead of a clean `ERR_PACKAGE_PATH_NOT_EXPORTED`) — a path many bundler/resolver tools touch. Also, once a shared chunk appears between components (`buildPlugin.ts` is already configured for `chunks/[name].js`), it becomes importable despite the explicit `./components/*` denial.
→ Add `"./package.json": "./package.json"` and `"./chunks/*": null` to `exports`.

### 🟡 Worth considering (non-blocking)

- **No test for multiple toasts stacked at once** (`toast.test.tsx`) — the stack arithmetic was the actual source of the 2px bug this PR fixes, but there's no regression test for the stack itself.
- **DropdownMenu keyboard navigation isn't tested** — the whole test suite uses `fireEvent.click`, even though `dropdown-menu.md` documents keyboard nav as part of the contract, and the pattern for testing it already exists in `tabs.test.tsx` from this same PR.
- **Two mounted `Toaster`s on the shared singleton duplicate every toast** (`toast.tsx:181-200`) — a silent failure mode plausible in this monorepo's shell+MFE architecture. Worth documenting (`createToastManager()` for the second viewport), and optionally a dev-only `console.warn`.
- **Base UI internal-symbol probes in `verify-consumer.sh` have no positive control** (`verify-consumer.sh:316-325`) — if `@base-ui/react` renames its internals, the absence assertions would pass vacuously forever without anyone noticing the leg stopped checking anything.
- **`tabIndex={0}` on the Table wrapper has no accessible name** (`table.tsx:38-44`) — a focus stop that announces nothing to screen readers; the documented workaround (wrap it yourself with `aria-label`) doesn't actually fix this, since the name ends up on a non-focusable element.
- **`Dialog showCloseButton={false}` has no runtime safeguard** — the contract ("only do this if you compose your own `DialogClose`") isn't checked; a consumer who forgets traps VoiceOver/TalkBack users in the dialog, undetected by CI.
- A few `buildPlugin.ts` fragility findings around Windows path handling (`endsWith('/public.d.ts')`, `replace(/^components\//, '')`) and paths resolved from `process.cwd()` instead of `import.meta.dirname` — harmless today (CI is POSIX-only, cwd matches the package root), but the invariant holds by accident.
- README/llms.txt make a webpack claim that no automated check backs up (see finding 3 above) — true today, unprotected against drifting false.

### ⚪ Nits
- `tokens.test.ts`'s raw-color guard regex doesn't catch named colors (`white`, `black`) or `hwb()/lab()`.
- `tooltip.test.tsx:110` — the negative test relies on real `Date.now()` timing (600ms) and can silently no-op on a slow runner; `vi.useFakeTimers()` would be more deterministic.
- The portal test is duplicated almost verbatim across 4 files — a candidate for a shared test helper.
- Focus-ring/invalid-ring CSS patterns are duplicated across 8-12 module.css files — looks like a deliberate tradeoff for module independence, not a blocker.
- `select.tsx`: `SelectScrollUpButton`/`SelectScrollDownButton` types aren't named-exported, unlike the rest of the public API.
- Breaking change to the public contract (dropped `main`/`require`/`./styles.css`) without a version bump — not a blocker since the package is `private: true` with no in-repo consumers.

SysoevAndrey:
@gs-layer Review addressed in 1b9dabf8…9d47ba42 (7 commits). All six pre-merge items fixed, four of the eight considerations taken, plus two nits.

**Fixed — the six:**

1. **`modal={false}`** → `showBackdrop` on `DialogContent` (default `true`, mirroring `showCloseButton`); dialog.md now documents the pairing instead of contradicting itself. (1b9dabf8)
2. **Toast region landmark** → `Toaster` takes `label` (default `'Notifications'`, defaulted kit-side so an explicit `undefined` can't wipe the name in the prop merge). The `closeLabel` docstring's "one string" claim corrected to two. (43e057be)
3. **`sideEffects` unguarded** → fifth `[webpack]` leg in `verify-consumer.sh`, production mode, barrel import. Re-ran your mutation locally: with `sideEffects: false` webpack emits **no CSS file at all**, so the leg dies at its first assertion with a named diagnostic while the other four legs stay green — exactly the blind spot you described. This also puts an automated check behind the README/llms.txt webpack claims (your penultimate consideration). (4682076b)
4. **Positioner escape hatch** → took the extend-the-`Pick` shape rather than a `positionerProps` object, to keep the kit's existing convention: all three components now expose `positionMethod`/`collisionBoundary`/`collisionPadding`, forwarded to the positioner and documented. (64c4e922)
5. **Submenu `container`** → inherited from the nearest `Content` via context (an explicit `container` on `SubContent` still wins), with a regression test asserting the submenu popup lands inside the themed container. (64c4e922)
6. **`./*` holes** → `"./package.json": "./package.json"` and `"./chunks/*": null`; verified via `import.meta.resolve` — clean resolution for the former, `ERR_PACKAGE_PATH_NOT_EXPORTED` for the latter. (0597bae6)

**Considerations taken:** toast stack test (three cards indexed newest-first via `--toast-index`, overflow past `limit` staying mounted as `data-limited`); DropdownMenu keyboard-nav test (with one contract correction along the way: Base UI focuses disabled items rather than skipping them — the WAI-ARIA discoverability pattern — so the test asserts focusable-but-not-activatable); double-`Toaster` failure mode documented in toast.md and the singleton docstring, pointing at `createToastManager()`; Table wrapper `label` → `role="region"` + `aria-label` on the focusable element itself — including your observation that the documented wrap-it-yourself workaround never worked, which the doc now explains.

**Nits taken:** raw-color guard now catches `hwb()`/`lab()`/`lch()`/`color()` and `white`/`black` (with lookarounds so `white-space` doesn't trip); `SelectScrollUpButtonProps`/`SelectScrollDownButtonProps` named and exported.

**Declined / deferred, with reasons:**
- *Fake timers in tooltip.test* — declined twice on this PR already (CodeRabbit and itechmeat threads): pinning the boundary couples the test to when Base UI's hover-intent rest timer arms. The silent no-op on a slow runner you noted is the deliberate degraded mode after b19dd28b: the guard makes the "still closed" claim only while measurably inside the 600ms window, and the "opens by 800ms" half still always asserts.
- *`showCloseButton` runtime safeguard* — a dev-only warning would have to detect a composed `DialogClose` anywhere in an arbitrary children tree; a reliable check isn't cheap, an unreliable one trains consumers to ignore it. The contract stays documented (and the sanctioned shape is tested).
- *Positive control for the Base UI probes, CloseIcon/svgIcon dedup, buildPlugin Windows hardening, portal-test helper, focus-ring dedup* — agreed on all as real but none block a `private: true` package; queued behind stage 4, where new components would otherwise copy the same patterns again.
- *Version bump for the contract break* — noted and agreed: `private: true`, no in-repo consumers; the version story lands with #495.

227 tests, type-check, eslint, and all five `verify-consumer.sh` legs green.

gs-layer (approved):
## Re-review notes

Approve with comments

### 🟠 Worth fixing

**1. `positionMethod`/`collisionBoundary`/`collisionPadding` aren't inherited by `DropdownMenuSubContent`, unlike `container`**
`packages/ui-kit/src/components/dropdown-menu/dropdown-menu.tsx:107-108`
The two fixes from this round interact badly at their intersection. A submenu now portals into the *same* `container` as its parent (fix #5), but if that container has a `transform`/`filter` — exactly the case `positionMethod="fixed"` (fix #4) exists for — that prop doesn't travel with it. The parent menu positions correctly; the submenu silently drifts, and the consumer has to repeat `positionMethod` on every `DropdownMenuSubContent` by hand.
→ Either extend the inheritance context to carry the positioner props alongside `container`, or document the asymmetry explicitly next to the `container` inheritance note.

**2. Documented `collisionBoundary` default doesn't match Base UI's actual default**
`packages/ui-kit/src/components/dropdown-menu/dropdown-menu.md:47`, `packages/ui-kit/src/components/tooltip/tooltip.md:64`
Both prop tables say the default is `viewport`. The real default (`useAnchorPositioning.d.ts:144`, `MenuPositioner.js`, `TooltipPositioner.js`) is `'clipping-ancestors'`. This isn't cosmetic — `clipping-ancestors` accounts for scrolling ancestor elements, `viewport` doesn't, and that's exactly the scenario (nested scroll container) this doc section is written for.
→ Fix both tables to say `clipping-ancestors`.

**3. `table.md` now contradicts itself after the Table `label` fix**
`packages/ui-kit/src/components/table/table.md:88-91`
The "scroll wrapper" section was updated to describe the new `role="region"` behavior when `label` is set, but the separate "Semantics and accessibility" section wasn't touched and still flatly states the wrapper "carries no role of its own." Same class of defect as last round's `closeLabel` docstring ("one string" instead of two): the doc makes an a11y claim the code no longer backs up.
→ Reconcile the two sections.

### 🟡 Worth considering

- **`container={null}` on `SubContent` can't opt back into `<body>`** (`dropdown-menu.tsx:108`) — `container ?? inheritedContainer` treats `null` as "not set," but Base UI treats `null` as a legal, meaningful value (portal to `<body>`). A consumer who deliberately wants a submenu to escape a themed container back to `<body>` has no way to express that except `document.body`, which breaks under SSR.
- **`Table label=""` produces an unnamed `role="region"`** (`table.tsx:57`) — the check is `label === undefined` rather than falsy, so an empty string (e.g. from an API field) still sets `role="region" aria-label=""`, which is exactly the "useless unnamed landmark" the fix was meant to avoid.
- **Weak assertion in the toast-limit test** (`toast.test.tsx:123`) — `toBeGreaterThanOrEqual(2)` can't actually fail on a real regression; the neighboring `data-limited` check is doing the real work. Tighten to `toHaveLength(3)`.
- **DropdownMenu keyboard-nav test lacks `waitFor` around its one synchronous assertion, and doesn't spy on the disabled item's `onClick`** (`dropdown-menu.test.tsx:234`) — the rest of the file (and its own comment) acknowledges these transitions happen via queued microtasks, so a regression that closes the menu asynchronously could slip past this specific check. "Doesn't activate" is currently inferred from "menu stayed open" rather than asserted directly on the handler.
- **`tokens.test.ts` raw-color guard still only catches `white`/`black`** (`tokens.test.ts:168`) — expanded from zero to two words, but `red`, `gray`, `navy`, `teal`, and most other CSS named colors still pass through unchecked. Not currently exploited anywhere in the codebase, but the guard remains structurally leaky for the class of bug it's meant to catch.

### ⚪ Nits
- `llms.txt` wasn't updated for `showBackdrop`, `closeLabel`, or the positioner escape hatch — it still describes the pre-fix prop surface for Dialog/DropdownMenu.
- `showBackdrop`'s documentation doesn't cover `modal="trap-focus"` — with the default `showBackdrop={true}`, that combination yields a dialog that's visually blocked but not scroll-locked, which the docs only warn about for `modal={false}`.
- `showBackdrop` is all-or-nothing (boolean) — there's no way to swap in a custom backdrop without bypassing the kit and reaching into `@base-ui/react` directly.
- `SelectScrollUpButtonProps`/`SelectScrollDownButtonProps` are publicly exported but undocumented and untested; `SelectContent` already renders both internally, so a consumer who finds the export and drops it into `children` gets a duplicate pair of arrows.
- The test covering `Toaster`'s kit-side `label` default doesn't actually pin the default — it would still pass if the `= 'Notifications'` default were removed from the code, since Base UI's own default would take over invisibly.
- The comment above the toast-stack test overpromises: it claims coverage of `--toast-height`/`--toast-offset-y` drift, but only `--toast-index` is actually asserted.

GeraBart (approved):

PR #534, opened 2026-08-05
Feature/function resolver proposal

### LINKED ISSUE #304: Support UI elements dynamic filtering
In 0.3.0 will be implemented as function resolver

GeraBart:


gs-layer:
@GeraBart thanks!

I’ve got a few observations and questions that might help me understand better:

### 1. Could the base GTS IDs be invalid?

I tried validating the two base IDs locally with the repository’s GTS CLI:

```text
gts validate-id -i gts.frontx.func.params.v1~
gts validate-id -i gts.frontx.func.result.v1~
```

Both returned an error similar to:

```text
Invalid GTS segment: Too few tokens
```

It appears that `frontx.func.params.v1` has four segments, while the parser expects at least five for a regular segment.

Could we double-check whether these IDs should include an additional namespace segment, consistent with the existing GTS ID hierarchy?

### 2. Could the resolver-domain target references be adjusted?

`action_function_execute.v1.json` and `action_function_return.v1.json` use the target pattern:

```text
...functions_resolver.v1~*
```

while the added domain instance ID ends with:

```text
...functions_resolver.v1
```

without the trailing `~`.

Could we confirm that this pattern matches the concrete domain instance as intended? It may be safer to use the exact domain ID, or a pattern that follows the instance-ID shape.

### 3. Could we recheck the direction of `run` and `return`?

`function.run` is currently declared in `extensionsActions`, while `function.return` is declared in `actions`.

Could we verify this mapping against `validateContract()` and the intended message flow?

- the resolver domain sends `function.run` to an extension;
- the extension sends `function.return` back to the resolver domain.

It would be helpful to ensure that these edges are declared in the sets expected by the current MFES contract validator.

### 4. Should we introduce a dedicated `requestId`?

`timestamp` currently acts as the correlation value between execute and return/result. It may be worth considering whether a timestamp is sufficient for concurrent calls, retries, or duplicate messages.

Also, since `caller` is supplied in the payload, should result routing be bound to runtime-derived caller identity instead?

One possible option would be a host-generated `requestId`, propagated through:

```text
execute → run → return → result
```

and internally bound to the original sender, selected function, and result destination.

### 5. Are we sure that `signature` and payload values should remain optional?

At the moment, an extension can declare `signature: {}`, while execute/run may omit `params` and return/result may omit `result`.

Is that intentional? Since the proposal focuses on typed function calls, perhaps it would be clearer to require:

- `signature.params` and `signature.result`;
- `params` in execute/run;
- `result` in return/result.

No-argument and void-result functions could still use an explicit empty schema or `{}`, rather than omitting the field.

### 6. Should we add `mount_ext` / `unmount_ext` to the domain?

The comments describe mounting a function extension for execution and unmounting it after completion.

Would it make sense to add `mount_ext` and `unmount_ext` to the domain contract as well? It may also be worth considering whether resolver extensions need the complete lifecycle set, including `init` and `destroyed`.

### 7. Could we add a typed failure result later?

For the eventual runtime implementation, it may be useful to consider a structured failure response. That would let callers distinguish a timeout, validation error, or execution exception from a missing response.

template-shell/src/gts/schemas/action_function_result.v1.json
```diff
@@ -0,0 +1,40 @@
+{
+  "$id": "gts://gts.frontx.mfes.comm.action.v1~frontx.mfes.function.result.v1~",
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$comment": "Action for returning the result of a single function resolver extension execution. Upon receiving the return action from a function resolver extension, the function resolver extensions domain validates the returned value against the schema declared in the corresponding extension function signature, internally unmounts that function extension, and sends this result action to the caller from the original execute request.",
+  "allOf": [
+    { "$ref": "gts://gts.frontx.mfes.comm.action.v1~" }
+  ],
+  "type": "object",
+  "properties": {
+    "target": {
```
itechmeat, line 10: The response leg has no contractual home: an extension caller must declare `function.result` in its entry `actions` or the mediator rejects the dispatch, and for a domain caller nothing checks inbound declarations at all - this would be the platform's first response-style action, with no convention to lean on. Where should a caller declare "I accept function results"?

template-shell/packages/framework/src/plugins/microfrontends/gts/frontx.screensets/instances/domains/functions_resolver.v1.json
```diff
@@ -0,0 +1,24 @@
+{
+  "id": "gts.frontx.mfes.ext.domain.v1~frontx.screensets.layout.functions_resolver.v1",
+  "sharedProperties": [
+    "gts.frontx.mfes.comm.shared_property.v1~frontx.mfes.comm.theme.v1~",
+    "gts.frontx.mfes.comm.shared_property.v1~frontx.mfes.comm.language.v1~"
+  ],
+  "actions": [
```
itechmeat, line 7: On point 3 of @gs-layer 's review (recheck the direction of `run` and `return`): verified against `validateContract` and the mediator - the mapping here is correct as declared, `run` is what the domain requires of extensions and `return` is what it accepts. What is actually inverted are the $comments in the platform's own entry.v1.json / domain.v1.json schemas: they describe both sets backwards, which is what made the mapping look wrong. One gap the question does expose: admission enforces `run` handling but not a declared `return`, so an under-declared function would fail only at runtime.

template-shell/src/gts/schemas/action_function_execute.v1.json
```diff
@@ -0,0 +1,44 @@
+{
+  "$id": "gts://gts.frontx.mfes.comm.action.v1~frontx.mfes.function.execute.v1~",
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$comment": "Action for triggering a single function resolver extension execution. Upon receiving this action, the function resolver extensions domain validates the params value against the schema declared in the corresponding extension function signature, internally mounts that function extension, and sends a run action to the extension named in 'func'. The caller must support receiving the result of the execution via the result action.",
```
itechmeat, line 4: Which execution model is the design - long-lived activated functions or per-call mount/destroy? The flow leaves open what a second concurrent `execute` does (queue, second mount, reject?) and what hot small functions pay per call; and the granted lifecycle (`activated`/`deactivated` only) would reject exactly the `init`/`destroyed` hooks a per-call extension lives through - which connects to @gs-layer's point 6 on `mount_ext`/`unmount_ext` and the complete lifecycle set.

template-shell/src/gts/schemas/action_function_execute.v1.json
```diff
@@ -0,0 +1,44 @@
...
+          "type": "string",
+          "format": "date-time",
+          "$comment": "Timestamp of the function execution request; it identifies a function execution request and its response"
+        },
+        "params": {
+          "type": "object",
+          "$ref": "gts://gts.frontx.func.params.v1~",
+          "$comment": "Function run params, validated by the domain against the schema declared in the corresponding extension function signature"
+        },
+        "caller": {
```
itechmeat, line 31: On @gs-layer's point 4 (a dedicated `requestId`): the current `timestamp` is minted by the caller, so it is not unique even statistically across different callers - and a caller-minted request id would inherit that. Separately, with no sender field in the action envelope, `caller` is self-declared payload data, so result routing today trusts whatever the caller wrote.

template-shell/src/gts/schemas/action_function_execute.v1.json
```diff
@@ -0,0 +1,44 @@
...
+          "oneOf": [
+            { "x-gts-ref": "gts.frontx.mfes.ext.domain.v1~*" },
+            { "x-gts-ref": "gts.frontx.mfes.ext.extension.v1~*" }
+          ],
```
itechmeat, line 35: On @gs-layer's point 7 (a typed failure result "later"): later has a cost - timeout, validation rejection and a crash are the same silence to the caller, and an optional `result` (point 5) cannot double as the failure signal, so v1 consumers will bake in the silent-failure assumption. Should v1 already say how a failure reaches the caller, and which leg does `defaultActionTimeout` bound - each hop or the whole round trip?

template-shell/packages/framework/src/plugins/microfrontends/gts/frontx.screensets/instances/domains/functions_resolver.v1.json
```diff
@@ -0,0 +1,24 @@
...
+    "gts.frontx.mfes.comm.action.v1~frontx.mfes.ext.load_ext.v1~",
+    "gts.frontx.mfes.comm.action.v1~frontx.mfes.function.execute.v1~",
+    "gts.frontx.mfes.comm.action.v1~frontx.mfes.function.return.v1~"
+  ],
+  "extensionsActions": [
+    "gts.frontx.mfes.comm.action.v1~frontx.mfes.function.run.v1~"
+  ],
+  "extensionsTypeId": "gts.frontx.mfes.ext.extension.v1~frontx.screensets.layout.functions_resolver.v1~",
+  "defaultActionTimeout": 30000,
```
itechmeat, line 16: Can a function extension send `execute` while handling `run`? If chains are allowed, what stops A -> B -> A cycles, and is the timeout budget per hop or per chain (the platform's mediator already knows action chains with a shared timeout)? If not, the restriction seems worth stating in the schema comments.

template-shell/src/gts/schemas/action_function_execute.v1.json
```diff
@@ -0,0 +1,44 @@
+{
+  "$id": "gts://gts.frontx.mfes.comm.action.v1~frontx.mfes.function.execute.v1~",
```
itechmeat, line 2: Is this a platform capability or a screensets feature? The action IDs sit in the platform namespace next to load_ext/mount_ext while the schemas live in the shell template - and the answer also decides the missing segment for the invalid base IDs @gs-layer flagged in point 1.

template-shell/src/gts/schemas/action_function_execute.v1.json
```diff
@@ -0,0 +1,44 @@
...
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$comment": "Action for triggering a single function resolver extension execution. Upon receiving this action, the function resolver extensions domain validates the params value against the schema declared in the corresponding extension function signature, internally mounts that function extension, and sends a run action to the extension named in 'func'. The caller must support receiving the result of the execution via the result action.",
+  "allOf": [
+    { "$ref": "gts://gts.frontx.mfes.comm.action.v1~" }
+  ],
+  "type": "object",
+  "properties": {
+    "target": {
+      "x-gts-ref": "gts.frontx.mfes.ext.domain.v1~frontx.screensets.layout.functions_resolver.v1~*",
```
itechmeat, line 11: Beyond the pattern issue @gs-layer flagged in point 2: mount_ext/load_ext target an open domain wildcard and name the extension in payload.subject, while execute pins the target to the resolver family. The fix shape matters either way - an instance-shaped pattern would admit future derived resolvers, an exact ID would not - so which shape is intended?

template-shell/src/gts/schemas/action_function_result.v1.json
GeraBart: Yes, correct. Caller must declare support of the function result action. Caller is either an extensions domain or an MFE entry. They both have capability of declaring actions support. `frontx-gears/mfes` already supports validating that action target has declared support of the action type during each action routing.

GeraBart:
> ### 1. Could the base GTS IDs be invalid?

it is valid against the gts-spec version that is used in `gears-frontx/gts-plugin`. Though it's quite likely that GTS spec may have evolved. We need an issue to release `gts-ts`, which is based on the latest GTS spec, bump `gts-ts` in `gears-frontx/gts-plugin`, and update all our GTS types and well-known instances.

GeraBart:
> 2. Could the resolver-domain target references be adjusted?

Yes, I could have made a mistake here. Each extensions domain is indeed a well-know instance, and thus can't be further extended.

GeraBart:
> 3. Could we recheck the direction of `run` and `return`?

The direction is correct. In extensions domain definition `actions` property refers to the list of actions that this extensions domain is capable of executing, while `extensionsActions` property refers to a list of actions that must be supported by MFE entries that implement this extensions domain extensions.

GeraBart:
> 4. Should we introduce a dedicated `requestId`?

Probably. But then we need to provide a convenient way for each caller to generate this request id. I was thinking that in the current proposal each function execution request can be identified by a pair of function id + timestamp. I think that it's extremely unlikely that the same function can be requested for execution at exactly the same moment. But maybe I'm wrong. Let's ask each contributor what he/she thinks on this matter.

GeraBart:
> 5. Are we sure that `signature` and payload values should remain optional?

This is intentional. Function can have zero parameters, or can return `void`. Function resolver domain will treat missing params or return as `void`, and will validate against `void`.

GeraBart:
> Should we add `mount_ext` / `unmount_ext` to the domain?

I have omitted them intentionally. My logic is the following. If we support `mount_ext` action in the function resolver extensions domain, then any caller can mount a function resolver extension and send `run` action to this extension directly, and thus bypass params schema validation. Same for `unmount_ext`.

template-shell/packages/framework/src/plugins/microfrontends/gts/frontx.screensets/instances/domains/functions_resolver.v1.json
GeraBart: Good catch. Yes, correct, comments in the base types are inverted.

template-shell/src/gts/schemas/action_function_execute.v1.json
GeraBart: Mounting and unmounting decisions are private to the function resolver extensions domain. I think, implementation-wise it will be a strategy of mounting on the first execution request

template-shell/src/gts/schemas/action_function_execute.v1.json
GeraBart: Real communication happens with actions chains, not actions. Action chain has `next` and `fallback` properties. So caller of a function defines what should happen upon failure in the `fallback` property.

template-shell/packages/framework/src/plugins/microfrontends/gts/frontx.screensets/instances/domains/functions_resolver.v1.json
GeraBart: yes, functions composition is a must

template-shell/src/gts/schemas/action_function_execute.v1.json
GeraBart: Current file structure is outdated. This is not a framework capability, but is a part of shell layout.

GeraBart:
> > ### 1. Could the base GTS IDs be invalid?
> 
> it is valid against the gts-spec version that is used in `gears-frontx/gts-plugin`. Though it's quite likely that GTS spec may have evolved. We need an issue to release `gts-ts`, which is based on the latest GTS spec, bump `gts-ts` in `gears-frontx/gts-plugin`, and update all our GTS types and well-known instances.

Issue created: https://github.com/constructorfabric/gears-frontx/issues/538

PR #536, opened 2026-08-05
feat(ui-kit): adopt the Studio design — tokens, reskin, typography, metric guard

SysoevAndrey:
## What

Moves `@gears-frontx/ui-kit` from shadcn's neutral defaults onto the Studio design (the F-mockups Figma file, page `00 · Foundations`, variable collection "Studio / shadcn") — delivery-plan step 4 in `design-notes.md`, 17 commits, plus a 10-commit review-fix series (see below).

### Tokens (`theme.css`)
- Full Studio palette in all three blocks (light, `[data-theme='dark']`, `prefers-color-scheme` mirror): violet primary, surface layering (`surface` / `surface-elevated` / `card` / `card-hover`), status colors with `-soft` fills, sidebar group, unprefixed shadcn names.
- New invariant scales: spacing (`--space-1..8`), control heights (`--control-height-sm/md/lg` = 32/36/40 per the drawn specimens), icon sizes, border widths, radius scale re-derived from `--radius`.
- Typography: `--font-sans`/`--font-mono` (families only — no font files shipped) plus the seven-role Studio type ramp as `--text-<role>-{size,line-height,weight,tracking}`.
- Values no mockup node draws are marked `derived:` with their derivation reasoning inline — pinning them is an open designer question.
- Review-fix additions: `--link-foreground` (per-theme), `--popover-border` + `--popover-shadow` (theme-invariant elevation recipe shared by dialog/dropdown/select/toast/card), and `--ring-inset` (the inward thickness that lets a 2px ring thicken without changing an element's box — eight modules had spelled the same `calc()` out by hand).

### Components (19 existing, reskinned)
- **Button**: new API — `icon` render slot (icons never in children), auto icon-only when no children (`aria-label` required), `loading` with stable width; `size="icon"` removed. Text is Studio/Label, the one component text bound to a ramp style in the design file. `loading` keeps the button focusable (`aria-disabled`, not native `disabled`) so `aria-busy` is actually announced.
- **Badge**: semantic `variant` (`success|warning|info|danger|muted`) × `shape` (`pill|dot`); shadcn variant list retired. The values are states, never paint jobs — the semantic-only rule lives in the value names and `badge.md`, while the prop names stay the kit-wide ones.
- **Fields** (Input/Textarea/Select): unified Field look — 40px, raised on `--surface-elevated`; `type="search"` grows the magnifier; Select gains the `filter` trigger variant (36px toolbar chip).
- **Tabs**: trackless default variant; segment styling deliberately left to the future `toggle-group` (per the design file's own component description).
- **Table**: carded rows (separated-borders model, ring/fill/radius assembled from cells), `density="compact"`, `data-state="selected|stale|restricted"` row hooks.
- **Focus idiom** everywhere: 2px ring with geometry-stable borders (recolored 1px border + inset shadow). The ring supports two independently-colored layers; `default`/`destructive` use a two-tone ring so both adjacent surfaces (fill and page background) clear the WCAG 3:1 floor in both themes.
- Checkbox/radio/switch/label/dialog/toast/dropdown/tooltip aligned to their drawn specimens.

### Metric rulings (recorded in `design-notes.md` and CSS comments)
1. Drawn specimens beat the design file's doc text for control heights (32/36/40).
2. The token system beats hand-set specimen values for metrics: all spacing snapped onto the `--space-*` grid, all text metrics onto the `--text-*` ramp (specimens' 13/18 → Label 13/16, 12/17 → Meta 12/16, the 10px table header → Meta at the lg control height). A `tokens.test.ts` guard rejects literal metrics in spacing/type declarations (including `border-spacing`, `scroll-margin`, `font-weight`, and semicolon-less final declarations); the five surviving exception entries carry inline reasons under three rationales (the fields' 16px iOS anti-zoom floor for font-size, a WCAG 1.4.12-based line-height, the switch thumb's 2px inset geometry).
3. Drawn colors that fail WCAG floors get corrected, not shipped: the corrected values and their measured ratios are recorded in `design-notes.md`, with open questions to pin them in the design file.

### Review fixes (10 commits)

A full review pass (3 static reviewers + QA + Figma cross-check + architecture review), then two rounds of PR review, landed:

- **WCAG contrast floors cleared and guarded**: two-tone Button focus rings (every variant ≥3:1 against both fill and page background, both themes), dark link text via `--link-foreground` (7.23:1, was 3.78:1), table header via corrected `--subtle-foreground` (5.10:1 / 4.88:1, was 2.56:1 / 3.74:1). A new contrast-guard test parses the live hex values out of `theme.css` and resolves each Button variant's actual ring declarations from the CSS, so re-pointing a token can't silently regress.
- **Button behavior**: icon-only detection handles falsy children (`{cond && 'Save'}`); `loading` no longer blurs the focused button; derived attributes (`aria-busy`, `data-loading`, `data-icon-only`) win over conflicting forwarded props.
- **Guard hardening + elevation tokens**: the metric guard's regex holes closed; the duplicated popup shadow/border recipe extracted into `--popover-border`/`--popover-shadow`.
- **Spec alignments from the Figma specimens**: disabled dim unified at the drawn 0.42, Ghost rest text on `--muted-foreground` (verified ≥4.5:1 both themes), Outline on `surface-elevated`/`border-strong` per the drawn bordered button.
- **Docs sweep**: design-notes (two-page demo, type-ramp groups, dated ruling block with consumer-facing changes for insight-front), table.md row-state rewrite, llms.txt preamble, stale comments.
- **Demo**: hash router no longer traps the Back button (normalization via `replaceState`).

Two rounds of PR review then landed:

- **Button's icon-only detection, again**: `Children.toArray` keeps empty strings (`toArray('')` is length 1, verified on this repo's React), so an empty label rendered a wide pill wrapping an empty span instead of squaring up. The predicate filters them; the falsy-label test covers `{''}` alongside `{cond && '...'}`.
- **`--subtle-foreground`'s contrast note was misattributed**: 3.74:1 belongs to `#667085`, the previous *dark* value — `#94a3b8` reads 7.25:1 on the dark surface and fails in light only. Each mode's drawn value failed against its own fill, which is what the notes now say.
- **One API shape**: Badge's `intent`/`form` renamed (see above). `form` also shadowed a real HTML attribute for anyone rendering a form-associated element via `render`. `docs.test.ts` now resolves styling axes from a named list, which closed a gap it had — Table's `density` values were never checked for documentation.
- **Comment provenance stripped**: dates, author names, review-round IDs and "an earlier version of this comment claimed" self-history are gone from shipped comments, keeping every constraint and reason. Button's and Badge's WCAG ratio tables moved to `design-notes.md` — `button.test.tsx` recomputes its six variants from the live CSS, while Badge's are `color-mix()` results resolved at runtime that no test can reach.
- **Deduplication**: `--ring-inset` (see above); the metric guard and its fixture share one `GUARDED_PROP` and one declaration splitter (`declarations()` in `__test-utils__`), which also absorbed a `tokenMap` copy that had drifted to a `;`-only terminator.

## Verification

- Full monorepo suite green across all 8 projects, 1380 tests (ui-kit: 21 files / 253 tests, incl. the metric guard, the WCAG contrast guards, theme-parity and raw-color guards). `arch:check` 12/12, `arch:deps` 0 errors, and `test:consumer` passing across vite + esbuild barrel/subpath + webpack + types.
- `npm run type-check` (+ tools/test/demo tsconfigs) and `eslint` clean.
- Clicked through both demo pages in light/dark; QA drove the demo in headless Chrome end-to-end: two-tone focus rings verified from token to painted pixel in both themes, loading buttons keep focus (`aria-disabled`), Back-button history verified, icon-only buttons measure square.

## Open designer questions (non-blocking, tracked in design-notes)

- Pin the `derived:` token values in the design file (popover pair, light `primary-hover`, dark `muted`, light `warning-soft` + dark `danger-soft`, light `blue`, dark sidebar trio).
- Pin the review-fix color corrections: `--subtle-foreground` (was below AA as drawn), `--link-foreground`, and the two-tone focus-ring inner colors (no drawn counterpart).
- The Outline button's rest-state border (`--border-strong` on `--surface-elevated`) measures 1.48:1 in light mode — pre-existing, left as drawn, flagged.
- The drawn Overlay options' muted/active color language — queued for the component phase.

## Summary by CodeRabbit

- **New Features**
  - Added an interactive UI Kit demo with Tokens and Components pages, theme switching, hash-based navigation, and responsive layouts.
  - Added token showcases for colors, typography, spacing, radii, controls, icons, and borders.
  - Added Select filter styling, compact Table density, semantic Badge intents and shapes, Button loading and icon-only states, and enhanced search inputs.

- **Enhancements**
  - Refreshed theme styling, accessibility contrast, focus indicators, disabled states, and component presentation across light and dark themes.

- **Documentation**
  - Expanded UI Kit usage, typography, component API, demo, and design guidance.


packages/ui-kit/src/components/button/button.tsx
```diff
@@ -29,8 +29,100 @@ const buttonVariants = cva(styles.button, {
...
+}: ButtonProps) {
+  // One predicate for "does this button have a visible label", driving
+  // BOTH the icon-only derivation below and the label-render guard at the
+  // bottom: `children == null` only catches a missing prop, not a falsy
+  // one — `<Button icon={x}>{cond && 'Save'}</Button>` with cond=false
+  // passes `false` as children, which is neither null nor renderable, so
+  // checking each spot with its own logic could (and did) disagree about
+  // whether the button has a label. Children.toArray drops exactly the
+  // values React itself would render nothing for (false/null/undefined/
+  // ''/[]), so "the array is empty" is the one true test.
```
itechmeat: `Children.toArray([''])` returns `['']` (length 1) - React drops `false`/`null`/`undefined`/`[]`, but keeps empty strings (verified on the repo's React 19.2.8). So `<Button icon={x}>{label}</Button>` with `label === ''` gets `hasLabel === true` and renders a wide pill with an empty label span instead of going icon-only - the exact bug class the "squares up for a falsy label" test exists for, which only exercises `false`. Either filter empty strings out of the predicate (and add the `{''}` case to that test), or fix the comment - as written, "the one true test" claims something React does not do.

packages/ui-kit/src/styles/theme.css
```diff
@@ -36,40 +50,241 @@
...
+  /* derived (dark only): light muted equals light secondary in the
+   * collection, so dark muted follows dark secondary the same way. */
+  --muted: #f1f5f9;
+  --muted-foreground: #64748b;
+  /*
+   * Table's header row is the only consumer (tokens.test.ts's leading-class
+   * guard doesn't reach theme.css itself, but a grep confirms it — see
+   * table.module.css). Corrected 2026-08-06 (Andrey): the drawn value
+   * (#94a3b8, slate-400) measured 2.56:1 light / 3.74:1 dark against the
+   * header's --surface fill — under the 4.5:1 AA floor in both themes, not
```
itechmeat: This attribution is garbled: `#94a3b8` vs the dark `--surface` (#0f131b) computes to 7.25:1 - a comfortable AA pass. The 3.74:1 figure belongs to `#667085`, the previous DARK value of the token (design-notes.md names the light/dark pair correctly), so "under the 4.5:1 AA floor in both themes" is wrong for dark as written - and the same sentence ships a second time in table.module.css. Since this comment exists to correct "an earlier, wrongly-backdropped measurement" and feeds the open designer question, could the attribution be fixed in both places (or the narrative dropped in favour of design-notes)?

packages/ui-kit/src/components/button/button.module.css
```diff
@@ -112,44 +177,174 @@
...
+  color: var(--link-foreground);
   text-underline-offset: 4px;
 }
 
 .variantLink:hover {
   text-decoration: underline;
 }
 
+/*
+ * Focus colors, verified against WCAG 1.4.11's >=3:1 floor for BOTH
```
itechmeat: [Not a blocker] button.module.css's comments grew 4.6x against 1.6x code growth, and this 30-line WCAG ratio table restates numbers that `button.test.tsx` already computes from the live CSS - whose own comment argues against exactly this kind of hand-maintained copy. Could the CSS keep the one-line constraint ("two tones because no single hue clears both adjacent surfaces") and the ratio tables live in design-notes, with the tests as the record of the numbers? Same shape in badge.module.css and table.module.css.

packages/ui-kit/src/styles/theme.css
```diff
@@ -36,40 +50,241 @@
...
+  --input: #cbd5e1;
+  /* Focus color. Light from the MVP Foundations frame; dark is drawn by
+   * the Field focus specimens (frame 193:383 strokes focus with the dark
+   * primary value), so neither mode is derived. */
+  --ring: #8b5cf6;
+
+  /* Sidebar Navigation block. --sidebar/--sidebar-accent are each drawn by
+   * their own dark-mode mockup node; the other three are not — named
+   * individually below, per this file's derived-marker convention. */
+  --sidebar: #f1f5f9;
```
itechmeat, line 311: [Not a blocker] Nothing under `src/` reads the five `--sidebar*` tokens, `--code-*`, `--blue`, or any of the four `*-soft` fills - and the file's own header excludes workflow tokens as "screen-level tokens for one consumer's workflow surfaces", which describes the sidebar group verbatim (design-notes lists app layout as a non-goal). The `*-soft` comment names "Alert/Badge/table-state backgrounds" as consumers, but Alert doesn't exist, Badge deliberately declines them, and table states are ring-only. If these are step-5 provisions, fair - but should the header's exclusion rationale and the `*-soft` comment stop describing consumers that don't exist?

packages/ui-kit/src/tokens.test.ts
```diff
@@ -160,6 +153,94 @@ describe('theme tokens', () => {
...
+
+  // Synthetic fixture for the `[;}]` terminator fix above: a rule whose
+  // LAST declaration has no trailing semicolon used to be invisible to the
+  // `;`-only version of this guard (no `;` in the source meant no match at
+  // all), letting a literal metric slip through undetected. Exercises the
+  // guard's own regex directly rather than the whole test, since the guard
+  // reads real component files and this fixture is neither.
+  it('does not let a semicolon-less final declaration bypass the metric guard', () => {
+    const GUARDED_PROP =
+      /^(?:padding|margin|scroll-margin)(?:-[a-z]+)*$|^(?:gap|row-gap|column-gap|font-size|line-height|letter-spacing|font-weight|border-spacing)$/;
```
itechmeat: [Not a blocker] The fixture re-declares `GUARDED_PROP` and the declaration regex as verbatim copies of the real guard's (lines 166-167 and 202) - if someone edits the guard's regex, this test keeps passing against yesterday's copy. Hoisting both to module scope (or into `__test-utils__` next to `extractRules`) lets the guard and its fixture share one definition. This is CodeRabbit's nitpick (b), still open.

packages/ui-kit/src/components/input/input.module.css
```diff
@@ -52,13 +65,21 @@
...
 }
 
 .input:focus-visible {
+  /* The mockups' Field focus: a 2px --ring border, no halo (frame
+   * 193:383). Drawn as recolored 1px border + inset ring rather than a
+   * real width switch, so the text never shifts on focus/blur (see
+   * button.module.css). */
   border-color: var(--ring);
-  box-shadow: 0 0 0 3px color-mix(in oklab, var(--ring) 50%, transparent);
+  box-shadow: inset 0 0 0 calc(var(--border-width-focus) - var(--border-width))
```
itechmeat: [Not a blocker] `calc(var(--border-width-focus) - var(--border-width))` appears 12 times across 8 modules - the same idiom with the same reason - while this same PR solved the identical duplication for popup chrome with `--popover-border`/`--popover-shadow`. The ring colors differ per variant, so the full recipe can't be shared, but a `--focus-ring-inset` width token would give the geometry-stable trick one definition and one place to fix.

packages/ui-kit/src/components/badge/badge.tsx
```diff
@@ -5,38 +5,44 @@ import { cva, type VariantProps } from 'class-variance-authority';
...
 const badgeVariants = cva(styles.badge, {
   variants: {
-    variant: {
-      default: styles.variantDefault,
-      secondary: styles.variantSecondary,
-      destructive: styles.variantDestructive,
-      outline: styles.variantOutline,
-      ghost: styles.variantGhost,
-      link: styles.variantLink,
+    intent: {
```
itechmeat: [Not a blocker] Every other component takes `variant`/`size`; Badge now takes `intent`/`form`, and the cost shows immediately - llms.txt gained a "some, like Badge, take other prop names" disclaimer and docs.test.ts had to learn `.intentXxx`/`.formXxx` selectors. The mockups' point (semantic intent, not paint) is real - but is the naming worth breaking the kit's one API shape, versus `variant="success"` with the semantics carried by the docs?

packages/ui-kit/demo/main.tsx
```diff
@@ -0,0 +1,138 @@
...
+      <header
+        style={{
+          display: 'flex',
+          justifyContent: 'space-between',
+          alignItems: 'center',
+          gap: 'var(--space-4)',
+          flexWrap: 'wrap',
+        }}
+      >
+        <h1 style={{ margin: 0, fontSize: 28, fontWeight: 600 }}>ui-kit demo</h1>
```
itechmeat: [Not a blocker] `28/600` here is `--text-display-*`, and `20/600` in shared.tsx is `--text-heading-1-*` - while this PR's own llms.txt tells consumers "Use the ramp for template text instead of inventing sizes". The demo is outside the metric guard, which is exactly where the anti-pattern gets copied from. The swatch geometry in tokens-page is specimen furniture and fine; the two headings (and the `fontSize: 11` labels) deserve the ramp they showcase.

packages/ui-kit/src/components/button/button.module.css
```diff
@@ -92,9 +150,16 @@
   background-color: color-mix(in oklab, var(--destructive) 90%, var(--foreground) 10%);
 }
 
+/*
+ * The drawn bordered button ("secondary" in the Figma component list,
+ * mapped to this kit's `outline` per button.md) sits on --surface-elevated
+ * with a --border-strong border, not the flat page background + hairline
+ * border the shadcn translation started from (Andrey, 2026-08-06 —
```
itechmeat: [Not a blocker] Shipped comments carry dated attributions "(Andrey, 2026-08-05/06)" in six places, review-round item IDs unresolvable from the repo ("B3", "D2/D3", "the 2026-08-06 fix series"), and "an earlier version of this comment claimed" self-history in four. The constraints in these comments are worth keeping - the provenance is git's and design-notes' job (this very line already says "recorded in design-notes.md" while still carrying the date and name). Could the dates, names, round IDs and self-history be stripped, keeping the reasons?

packages/ui-kit/design-notes.md
```diff
@@ -171,18 +202,129 @@ left is writing them.
...
+   table's 6px → 4), the specimens' 13/18 and 12/17 text normalized onto
+   Label/Meta, and the 10px table header onto Meta at the lg control
+   height — and a tokens.test.ts guard now rejects literal metrics in
+   spacing and type declarations (documented exceptions: the fields' 16px
+   iOS anti-zoom floor, the switch thumb's 2px inset geometry). Still
+   open: the token values marked `derived:` in theme.css; and the drawn
+   Overlay options' muted/active color language (a component-phase item,
+   not a token one).
+
+   **2026-08-06 accessibility + spec-alignment pass (Andrey).** A review
```
itechmeat: [Not a blocker] Step 4 now carries a 45-line dated pass-log with review provenance ("QA's follow-up review flagged the razor-thin margin, so..."). The decisions belong here; the review transcript doesn't - state the ruling ("light --link-foreground takes --primary-hover's value: 4.98:1, real headroom") and drop who flagged it in which round.

packages/ui-kit/src/__test-utils__/css-rules.ts
```diff
@@ -0,0 +1,51 @@
...
+ * Every declaration in a rule's body as a plain name->value map — custom
+ * properties (`--foo: ...`) and ordinary properties (`color: ...`) alike,
+ * since `[a-z-]` as the first character class matches both a letter and
+ * the leading `-` of `--foo`. `[;}]` as the terminator, not `;` alone, for
+ * the same last-declaration reason extractRules keeps the closing brace.
+ * Last declaration of a given name wins (a plain object literal via
+ * Array.from -> Map already does this), matching how the cascade would
+ * resolve two declarations of the same property in one rule.
+ */
+export function declarationMap(body: string): Map<string, string> {
```
itechmeat, line 70: [Not a blocker] `tokens.test.ts` still keeps its own `tokenMap` doing this exact job - with the `;`-only terminator, i.e. without the `[;}]` fix this util exists for, so the two diverge on a semicolon-less final declaration. Harmless today (theme.css is authored with trailing semicolons), but one of the two copies is the one that was fixed - worth pointing `tokenMap` at `declarationMap`?

packages/ui-kit/src/components/input/input.test.tsx
```diff
@@ -37,9 +52,15 @@ describe('Input', () => {
...
-    render(<Input disabled onValueChange={onValueChange} />);
+  it('marks the input disabled', () => {
+    // Renamed from "does not accept input when disabled": jsdom's
+    // fireEvent dispatches a change event directly, bypassing the browser
+    // input pipeline that a real disabled attribute blocks — it fires the
+    // handler regardless, so asserting "the spy was not called" here would
+    // pass or fail for reasons unrelated to Input's own behavior. What this
+    // component controls, and what a real browser enforces the rest of the
+    // way, is the native `disabled` attribute itself.
+    render(<Input disabled onValueChange={vi.fn()} />);
```
itechmeat: [Not a blocker] The comment above explains why nothing asserts on this spy anymore - which is the argument for deleting the prop, not keeping it. CodeRabbit's nitpick (c), still open.

packages/ui-kit/src/components/select/select.test.tsx
```diff
@@ -105,4 +105,23 @@ describe('Select', () => {
...
+        </SelectTrigger>
+        <SelectContent>
+          <SelectGroup>
+            <SelectItem value="a">A</SelectItem>
+          </SelectGroup>
+        </SelectContent>
+      </Select>,
+    );
+    const trigger = screen.getByRole('combobox', { name: 'Status filter' });
+    expect(trigger.className).toContain(styles.variantFilter);
```
itechmeat, line 128: [Not a blocker] This asserts the positive; the default-trigger test (lines 38-45, untouched) still has no `not.toContain(styles.variantFilter)` proving the variant doesn't leak into the default - the negative-case pattern table.test.tsx already uses. CodeRabbit's nitpick (a), still open.

packages/ui-kit/src/tokens.test.ts
```diff
@@ -160,6 +153,94 @@ describe('theme tokens', () => {
...
+      for (const rule of extractRules(readFileSync(file, 'utf8'))) {
+        // `[;}]`, not just `;`: a rule's last declaration is valid CSS
+        // without a trailing semicolon (`padding: 12px }`), and a
+        // `;`-only terminator never matches that declaration at all —
+        // silently exempting it from every check below. `body` always
+        // ends in `}` (see extractRules), so this terminator always finds
+        // the end of the last declaration whether or not the source had a
+        // semicolon there. The value capture excludes both terminators so
+        // a `}`-terminated value doesn't carry the brace into `leftover`.
+        for (const decl of Array.from(rule.body.matchAll(/([a-z-]+)\s*:\s*([^;}]+)[;}]/g))) {
```
itechmeat: [Not a blocker] A literal metric can be laundered through a local custom property: `--foo: 12px` matches `[a-z-]+` as the property name so `GUARDED_PROP` never fires, and `padding: var(--foo)` is then stripped clean. Not exploited today (both local metric literals feed unguarded properties), but the guard already keeps a "Known blind spot" note - this one belongs in it.

packages/ui-kit/src/components/table/table.md
```diff
@@ -103,33 +103,43 @@ wider than 100.
 
 ## Row states
 
-`TableRow` reads two attributes for its background tint, forwarded like any
-other prop rather than driven by a kit-specific prop:
-
-- `data-state="selected"` — a solid `--muted` fill, for a row the user has
-  selected (e.g. via a leading checkbox column).
+`TableRow` reads attributes for its background tint, forwarded like any
```
itechmeat: [Not a blocker] "reads attributes for its background tint" contradicts line 128's correct "colored border, not a tinted surface" - one leftover clause from the old model inside an otherwise-updated section.

itechmeat (changes requested):
Impressive, careful work - the token seam, the geometry-stable focus idiom, and the guards that parse live CSS values (metric scale, WCAG contrast, dark-block parity) are exactly the right machinery, and nearly every number I recomputed checked out. Two small things hold my approve: a factually wrong `Children.toArray` comment hiding a real empty-string edge in Button's icon-only detection, and a garbled contrast-measurement attribution shipped into two CSS comments. Everything else is non-blocking - mostly the comment-volume pattern from the previous round grown further, plus a few of CodeRabbit's smaller nitpicks still open. All cheap fixes; happy to approve after.

packages/ui-kit/src/components/button/button.tsx
SysoevAndrey: Verified — you are right, and I had it backwards. Probe on this repo's React: `Children.toArray('')` and `toArray([''])` both length 1; `false` is 0. So the "one true test" claim was false for exactly the shape it needed to cover.

Fixed at the predicate, not the comment: `hasLabel` is now `Children.toArray(children).some((child) => child !== '')`, and there is a dedicated `{''}` test asserting both `data-icon-only` and the absence of a `.label` span. `0` stays unfiltered — React renders it.

packages/ui-kit/src/styles/theme.css
SysoevAndrey: Confirmed, and the error was worse than garbled — I recomputed all three:

| pair | ratio |
|---|---|
| `#94a3b8` / light `--surface` `#ffffff` | 2.56:1 ❌ |
| `#94a3b8` / dark `--surface` `#0f131b` | **7.25:1 ✅** |
| `#667085` (previous dark) / `#0f131b` | 3.74:1 ❌ |

So 3.74 was the previous DARK value's number, and `#94a3b8` fails in light only. The narrative now says what actually happened — each mode's own drawn value failed against that mode's own fill — and I took your second option for the duplicate: `table.module.css` no longer restates it, it states the constraint (this bar's fill is the backdrop the token must clear) and points at theme.css/design-notes. design-notes carries the per-mode numbers plus the note that `#94a3b8` passes on dark, so the designer question is not fed a wrong premise.

packages/ui-kit/src/tokens.test.ts
SysoevAndrey: Done — hoisted `GUARDED_PROP` to module scope so the guard and the fixture share one definition, and the declaration regex moved into `__test-utils__/css-rules.ts` as `declarations()` next to `extractRules`. Both call sites now use it, so editing either the pattern or the splitter shows up in the fixture.

packages/ui-kit/src/components/select/select.test.tsx
SysoevAndrey: Added, in the default-trigger test rather than as a new one, with the reason inline: `.variantFilter` overrides `.sizeDefault`'s height by source order, so an unconditional class would drop every default trigger 40px → 36px with the positive test still green.

packages/ui-kit/src/components/input/input.test.tsx
SysoevAndrey: Agreed — deleted the prop. The comment now explains why the assertion is the attribute and not a spy, ending on "so there is no spy to pass" instead of leaving one lying there.

packages/ui-kit/src/__test-utils__/css-rules.ts
SysoevAndrey: Pointed it at the util. `declarationMap` is now built on the new `declarations()` array (array is the primitive: a value-scanning guard has to see both declarations of a twice-declared property, which a map would collapse), and `tokenMap` is gone — so the `[;}]` terminator is the only one left.

packages/ui-kit/src/components/input/input.module.css
SysoevAndrey: Added as `--ring-inset`, replacing all 12 occurrences across the 8 modules.

One deviation from your suggestion: not `--focus-ring-inset`. Five of the twelve are not focus rings — the four `aria-invalid` rings and Table's `data-state="selected"` row — so a "focus" name would misdescribe them. The token comment covers all three uses and says why the inset exists rather than a width switch (text shift on a control, every row below on a table).

packages/ui-kit/src/components/badge/badge.tsx
SysoevAndrey: Renamed — Badge takes `variant`/`shape` now, and the kit is one API shape again. The semantic-intent point survives where you suggested it should: in the value names (there is no `primary`/`outline`/`ghost` to pick from) and in badge.md, which spells out that choosing `warning` because you want orange is a misuse rather than a style choice. `llms.txt` drops the "some, like Badge, take other prop names" disclaimer.

The second axis went to `shape`, not `size`: the values are pill vs. bare dot and Badge has no size axis at all (the pill is a fixed 24px). And a second reason not to keep `form` that I had not clocked before — `form` is a real HTML attribute, so a styling prop by that name shadows it for anyone rendering a form-associated element via `render`. `badge.test.tsx` was already asserting the prop does not reach the DOM, which is exactly that collision.

`docs.test.ts` now resolves axes from a named `AXES` list instead of a hardcoded alternation, which closed a gap it had: `density` was never in that regex, so Table's density values were never checked for documentation.

packages/ui-kit/src/styles/theme.css
SysoevAndrey: You are right on all three counts — I grepped: nothing in `src/` reads the five `--sidebar*`, `--code-*`, `--blue`, or any `*-soft`.

Kept the tokens, fixed the claims. The header no longer pretends the exclusion line is "screen-level vs kit semantics" — the real line is the namespace (Figma exports workflow/scenario-canvas under `--studio-*`), and it now states outright that some groups here are unread by any component, why they are carried, and that app layout stays a non-goal so the sidebar tokens are colors for a consumer's own chrome rather than a promise of a Sidebar component. Each group repeats it at its own declaration.

The `*-soft` comment no longer names Alert. It says no component reads them, and why: Badge and the table states both went to a colored ring on the drawn surface, so nothing paints a tinted status surface. `badge.module.css` had the same stale "(those belong to Alert)" clause — also fixed.

packages/ui-kit/src/tokens.test.ts
SysoevAndrey: Added to the guard's blind-spot note. Also recorded why it stays unfixed rather than guarded: custom properties legitimately hold literal geometry (Toast's stacked-card math, Card's spacing arithmetic) that this guard deliberately leaves alone for dimensions, so catching the laundering case means resolving each local var to its value before scanning, not widening `GUARDED_PROP`.

packages/ui-kit/src/components/table/table.md
SysoevAndrey: Fixed — it now reads "plain `data-*` attributes for its state", which agrees with line 128.

packages/ui-kit/src/components/button/button.module.css
SysoevAndrey: Stripped, kit-wide — dates, names, round IDs (`B3`, `A1`, `D2/D3`, "the 2026-08-06 fix series") and the four "an earlier version of this comment claimed" passages are gone, in `theme.css`, `button.module.css`, `button.tsx`, `button.test.tsx`, `table.module.css` and `tokens.test.ts`. Every constraint and reason stayed; where a comment leaned on the provenance to justify itself, it now states the standing ruling instead and points at design-notes.

packages/ui-kit/src/components/button/button.module.css
SysoevAndrey: Taken. `button.module.css` keeps the constraint — two tones because no single hue clears both surfaces the ring's two pixels touch, plus why `--info` is the outer tone and why `outline` falls through to `--ring` — and the 30-line ratio table is gone, with an explicit note that `button.test.tsx` recomputes all six variants against both surfaces in both themes from that file's live values, so there is nothing to hand-maintain.

The numbers landed in design-notes as a table rather than being dropped. Same treatment in `badge.module.css`, with one difference worth flagging: Badge's ratios are `color-mix()` results the browser resolves at runtime, so unlike Button's rings **no test can recompute them** — design-notes is their only record, and both the CSS comment and the notes say so.

`table.module.css`'s duplicated `--subtle-foreground` narrative is gone too. Its two remaining one-line measurements (`--foreground` on `--surface`, the caption's `--muted-foreground`) I left as-is — a single number beside the color it justifies reads like the one-line constraint you are asking for, not a table.

packages/ui-kit/design-notes.md
SysoevAndrey: Trimmed. The heading is no longer dated-and-signed, "QA's follow-up review flagged the razor-thin margin, so..." is gone, and the link-variant entry states the ruling directly: light takes `--primary-hover`'s value at 4.98:1 rather than `--primary`'s own, which clears the floor by 0.005 and would fail again on any one-step nudge to `--background`. Same for the typography ruling above it.

design-notes did grow, though — deliberately, and I want to flag it rather than have it read as ignoring you: the ratio tables from `button.module.css` and `badge.module.css` moved in here per your other comment, so the section is longer while carrying less transcript.

packages/ui-kit/demo/main.tsx
SysoevAndrey: Fixed — `main.tsx`'s `h1` takes `--text-display-*` and `shared.tsx`'s `h2` takes `--text-heading-1-*`, both whole (size/line-height/weight/tracking). The eight `fontSize: 11` captions in `tokens-page` now read `--text-meta-size`.

The swatch geometry stays, with a comment drawing the line you drew: 148/44/56 and the 2px grid gaps are specimen furniture with no token to name them, whereas the caption text is ordinary UI text and takes the role it is there to showcase.

eslint.config.js
```diff
@@ -306,6 +306,16 @@ export default [
     },
   },
 
+  // ui-kit's demo consumes the package by name (the workspace-linked dist), same as the
+  // telemetry demo — that one is exempt because the telemetry block is scoped to src/;
+  // ui-kit's block above spans the whole package, so the demo needs the explicit carve-out.
+  {
+    files: ['packages/ui-kit/demo/**/*.ts', 'packages/ui-kit/demo/**/*.tsx'],
+    rules: {
+      '@typescript-eslint/no-restricted-imports': 'off',
```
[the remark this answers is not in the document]
SysoevAndrey: Already addressed in bf79184, which is why this thread has no "addressed" marker but the code does not match the finding any more. The demo override no longer turns the rule off: it re-declares it with only the intra-ecosystem group lifted, so `@gears-frontx/*/src/**` and `@/*` stay banned and the demo keeps validating the public surface (eslint.config.js, the demo block).

SysoevAndrey:
All 13 review comments addressed in 1286147 — replies are in each thread. Both blockers were real:

- **Button** — `Children.toArray` keeps empty strings (verified: `toArray('')` is length 1), so an empty label rendered a wide pill instead of squaring up. Predicate fixed, test added for that shape.
- **Contrast attribution** — 3.74:1 was the previous *dark* value `#667085`; `#94a3b8` reads 7.25:1 on the dark surface and fails in light only. Corrected in `theme.css`, `table.module.css` and design-notes, with the duplicate narrative dropped in favour of the constraint plus a pointer.

Three deviations worth naming, since none is quite what was suggested:

1. The shared inset token is **`--ring-inset`**, not `--focus-ring-inset` — 5 of the 12 sites are `aria-invalid` rings and Table's selected row, which are not focus.
2. Badge went to **`variant`/`shape`**, not `variant`/`size`: the values are pill vs. bare dot and Badge has no size axis. `form` also turned out to shadow a real HTML attribute under `render`, which `badge.test.tsx` was already asserting against.
3. **design-notes got longer**, not shorter — the review transcript came out, but Button's and Badge's ratio tables moved in from CSS.

One incidental fix: `docs.test.ts` matched axes by a hardcoded `variant|size|intent|form` alternation, so Table's `density` values were never checked for documentation. It now resolves from a named `AXES` list.

Verification: `npm run lint` clean, `type-check:all` clean (the demo needed a `dist` rebuild after the Badge rename), full monorepo suite 8/8 projects green (1380 tests, ui-kit 253), `arch:check` 12/12, `arch:deps` 0 errors, and `test:consumer` passes across vite + esbuild barrel/subpath + webpack + types.

packages/ui-kit/src/styles/theme.css
```diff
@@ -36,40 +60,266 @@
...
+   * fail it again) and 3.78:1 dark (a clear fail) against the page
+   * background the borderless variant sits on. derived: light takes
+   * --primary-hover's value instead of --primary's own (still the same
+   * violet family, one step darker) for real headroom — 4.98:1, not a
+   * borderline pass; dark takes a lighter tint of the same hue.
+   */
+  --link-foreground: #7c4fde;
+
+  /* Status colors with their -soft fills. The status colors themselves are
+   * consumed (Badge's ring, Table's stale/restricted row rings); the four
```
itechmeat: 🔵 [Not a blocker] "Badge's ring" is not a thing the status colors paint: Badge's status colors go to the dot (`--badge-accent`) and the label text (`--badge-text`), while its actual rings are `--ring` on focus and `--destructive` on `aria-invalid` - never a status color (badge.module.css says so itself). Table's half is right. Same defect class as the "(those belong to Alert)" clause this very round removed - worth "Badge's dot and label text" here.

packages/ui-kit/src/components/badge/badge.module.css
```diff
@@ -1,32 +1,54 @@
...
  */
 
 .badge {
   /*
-   * The registry relies on Tailwind's global preflight (border-box on every
-   * element) to make `height` include this padding and border. This kit has
-   * no reset, so content-box would inflate the box past 1.25rem — same acute
-   * case as tabs.module.css, scoped the same way (this class only).
+   * Safe fallbacks for --badge-accent/--badge-text — the muted recipe
+   * (cva's defaultVariants sets `intent: 'muted'`, so this is what every
```
itechmeat: ⚪ [Not a blocker] `badge.tsx` sets `variant: 'muted'` now - the next line of this same comment was updated in the rename, this token wasn't. The only `intent` left in the package.

packages/ui-kit/src/tokens.test.ts
```diff
@@ -287,5 +425,63 @@ describe('theme tokens', () => {
...
+        expect(value, `${name} missing from the theme block being checked`).toBeDefined();
+        return value as string;
+      }
+
+      const themes: Array<[string, Map<string, string>]> = [
+        ['light', lightTokens],
+        ['dark', darkAttrTokens],
+      ];
+
+      // A2: Button's link variant text against --background specifically —
```
itechmeat: ⚪ [Not a blocker] The provenance strip got everything else (zero `(Andrey`/dates/`B3`/`D2/D3` left package-wide), but `A2` here and `A3` at line 475 survived. Same reason as the rest: round-item IDs are unresolvable from the repo, and the sentences stand fine without them.

packages/ui-kit/src/tokens.test.ts
```diff
@@ -160,6 +159,91 @@ describe('theme tokens', () => {
...
+  // intrinsic sizes no token names (a 24px badge, a 38x22 switch track).
+  // Exceptions carry their reason inline; add one only with a constraint
+  // the token system genuinely cannot express.
+  //
+  // Known blind spot: a literal metric can be laundered through a local
+  // custom property. `--foo: 12px` is itself a declaration whose "property"
+  // is `--foo`, which GUARDED_PROP never matches, and the `padding:
+  // var(--foo)` that reads it strips to nothing. Guarding custom properties
+  // directly is not the fix — plenty of them legitimately hold literal
+  // geometry (Toast's stacked-card math, Card's spacing arithmetic) that
```
itechmeat: ⚪ [Not a blocker] Card's local custom properties hold token references only (`--card-spacing: var(--space-6)` etc.) - no literal geometry, so it doesn't illustrate this sentence; only Toast does. And "plenty of them legitimately hold literal geometry" contradicts "both local metric literals in the kit" three lines below - the kit has exactly two. Drop Card from the example and reconcile "plenty" with "both".

packages/ui-kit/design-notes.md
```diff
@@ -171,18 +202,167 @@ left is writing them.
...
+     colorable inset shadow), `outline` gave up on `--border-strong`
+     entirely and now falls back to the kit-wide `--ring`. Ratios as
+     measured (light/dark), with button.test.tsx recomputing them from
+     the live CSS on every run:
+
+     | variant | ring vs page bg | inner tone vs fill |
+     |---|---|---|
+     | `default` | `--info` 3.91 / 10.89 | `--foreground` 3.79 / 4.86 |
+     | `destructive` | `--primary-hover` 4.98 / 4.18 | `--destructive-foreground` 4.70 / 4.62 |
+     | `ghost`, `link`, `secondary` | `--primary-hover` 4.76+ / 3.10+ | single-tone |
```
itechmeat: ⚪ [Not a blocker] Under the "ring vs page bg" header these three variants measure 4.98 light / 4.18 dark (`--primary-hover` vs `--background`); the printed 4.76+/3.10+ are the fill-side worst cases (`--primary-hover` vs `--secondary`). Honest as a floor across both surfaces, but not the column they sit in - either rename the column ("worst adjacent surface") or print 4.98/4.18.

itechmeat (approved):
Thorough fix round - both blockers are genuinely fixed (I re-verified the predicate and the `{''}` test, and recomputed every cell of the relocated ratio tables), all three deviations you called out check out (`--ring-inset` naming, Badge's `variant`/`shape` with the accurate `form`-shadowing rationale, the deliberate design-notes growth), and the provenance strip is near-total. What's left is five small comment/doc accuracy nits - one of them a fresh claim of the same class this round was fixing - and a PR body that now lags the branch in four places. None of it blocks; approving.

packages/ui-kit/src/styles/theme.css
SysoevAndrey: Caught me writing the exact defect I had just removed. Confirmed from the CSS: `--badge-accent` paints `.badge::before` (the dot) and `--badge-text` is the `color` — the only `border-color`s in the file are `--ring` on `:focus-visible` and `--destructive` on `aria-invalid`. Now reads "Badge's dot and label text, Table's stale/restricted row rings", and the sentence after it no longer says Badge "went to a colored ring" either — it draws its status on the dot and label over a neutral fill.

packages/ui-kit/src/components/badge/badge.module.css
SysoevAndrey: Fixed. Also `per-intent class` on line 21, which names the `.variantX` classes — those two were the last prop-name references in the package.

packages/ui-kit/src/tokens.test.ts
SysoevAndrey: Both gone. The sentences read the same without them.

packages/ui-kit/src/tokens.test.ts
SysoevAndrey: Right on both halves. Card's locals are all token references (`--card-spacing: var(--space-6)`, the title pair off the ramp) — no literal in any of them. Grepped for locals holding a literal kit-wide and there are exactly two: Toast's `--peek: 0.75rem` and Table's `--table-row-ring-inset: 0px`.

So the note now names those two instead of "plenty", drops Card, and says what each feeds — a transform and a box-shadow, both unguarded, which is what makes the blind spot unexploited rather than merely unexploited-so-far.

packages/ui-kit/design-notes.md
SysoevAndrey: Recomputed: `--primary-hover` vs `--background` is 4.98 / 4.18, vs `--secondary` is 4.76 / 3.10. You are right that I filed the fill-side pair under the page-bg column.

Split rather than renamed, since the two cases differ: `secondary` gets its own row carrying 4.98/4.18 outer and 4.76/3.10 against its own fill, and `ghost`/`link` get a row saying they have no fill of their own so the page background is both surfaces. Columns now name the surface explicitly, and 3.10 is called out as the tightest number in the table and the one to watch if `--secondary` or `--primary-hover` moves.

SysoevAndrey:
All five nits fixed in 26045f01, and the PR body is caught up to the branch.

The theme.css one was the same defect class this round was fixing, as you said — status colors go to Badge's dot and label text, and its only rings are `--ring`/`--destructive`. The other four: `intent: 'muted'` and `per-intent class` in badge.module.css (the last prop-name references anywhere in the package), `A2`/`A3` in tokens.test.ts, and the blind-spot note, which cited Card — whose locals are all token references. Grepped kit-wide: exactly two locals hold a literal (Toast's `--peek`, Table's `--table-row-ring-inset`), so "plenty" is now those two by name, with what each feeds.

The ratio table I split rather than relabelled: `--primary-hover` is 4.98/4.18 vs `--background` and 4.76/3.10 vs `--secondary`, so `secondary` gets a row with both and `ghost`/`link` get one noting they have no fill, with 3.10 flagged as the tightest cell.

PR body corrected in five places: commit count (17 + 10, was 17 + 6), Badge's `variant`/`shape`, `--ring-inset` in the token additions, the test numbers (1380 across 8 projects; ui-kit 253, was 252), plus a stale "2026-08-06" pointing at a design-notes heading that no longer carries a date. Added a short round-2 section so the description covers what the last two commits did.

Re-verified after the changes: lint clean, type-check clean, ui-kit 253 green.

GeraBart (approved):

Issue #374, opened 2026-06-04
[FEATURE] I want to be able to install frontx into existing repo, not only as new project/repo


il10241024:
> 🔗 **Mirrored from** [cyberfabric/cyberware-frontx#261](https://github.com/cyberfabric/cyberware-frontx/issues/261)
> Originally opened by **Artifizer** on 2026-04-01T23:03:20Z

---

## Description
I want to install frontx into existing repo, not as a new project only. I may have already my repo initialized, and also it may have already some files (README, PRD, ADR, etc) 

## Problem Statement
Right now, the README.md tells me I can create a frontx project, but it doesn't explain me how to install it into existing repo

## Proposed Solution
I'd like to have some CLI example explaining how to install it

## Affected Package(s)
Which HAI3 package(s) would this impact?
- [ ] &#219;hai3/cli

## Priority
- [ ] Low - Nice to have
- [x] Medium - Improves workflow
- [ ] High - Blocking or critical for use case

---

PR #541, opened 2026-08-06
fix: refuse an add that would write over content no provenance recorded

### LINKED ISSUE #539: frontx add materializes over undeclared existing content - needs the guard seed already has
## Description

`frontx seed` refuses a target directory that already holds content (#533): the conflict check arbitrates only declared template boundaries, so content that arrived by any other route is declared by nobody and an empty occupied set makes every claim look free. `frontx add` has the same exposure and no guard: against a populated directory with no provenance, `addTemplate` reads `[]`, derives an empty occupied set, passes the conflict check, and materializes - overwriting any existing file at a path the incoming template owns (`writeFileSync` truncates).

Today the only protections are prose: the seed refusal's remedy sentence and the scaffolding skill both warn that `add` arbitrates only declared boundaries. A warning is not a check, and PR #533's review (SysoevAndrey) confirmed the gap by driving `addTemplate` with an empty provenance reader: `ok=true`.

Work:

1. Before materializing, `add` compares the incoming assembly's file set against what already exists in the target outside provenance-recorded ground; an overlap is refused as a handled failure (closed `reason`, message naming the colliding paths), before any write.
2. A directory whose existing content does not collide with the incoming template's ground proceeds - `add` into a populated non-provenance directory stays a supported case (the scaffolding skill offers it as the alternative to seeding a fresh directory).
3. Follow the seed guard's conventions from #533: refusal before the first write, re-probe before materialization, fail closed on unreadable state.

## Estimate

0.5-1 engineer-day.

## Notes

Spun out of PR #533's review threads (the seed guard's remedy currently points at `add` with a prose caveat; this issue makes the pointer safe). Touches `add-template.ts`, which was frozen during #533 while PR #500 was open; #500 merged 2026-08-05, so the surface is free.

itechmeat:
Closes #539.

`frontx seed` refuses a populated target since #533. `frontx add` had the same exposure and no guard: against a populated directory with no provenance it read `[]`, derived an empty occupied set, passed the conflict check, and materialized - `writeFileSync` truncating whatever stood at a path the incoming template owns.

## What the guard does

Before any write, `add` compares the staged assembly's own file set (the same `groupContributionsByPath` grouping materialization composes from) against what the target already holds, skipping every path that falls inside ground an applied template's recorded provenance accounts for. Anything left that exists on disk refuses the whole operation.

- Two probe points through one shared function (`refuseUnlessGroundFree`): once as soon as both inputs exist - the staged assembly and the derived occupied boundaries - and again immediately before `materializeAssembly`, so a directory that gained content while the conflict check ran is refused rather than written over. One function, so the pre-flight refusal and the last-moment refusal cannot drift apart.
- Closed refusal reasons on the existing discriminated union: `target-holds-undeclared-content` (carries the colliding `paths` as evidence, names both remedies: move or delete them, or record the applied provenance of the template that wrote them) and `target-not-directory` (offered with no seed remedy, since seed needs a directory too).
- Fail-closed probing. The adapter answers ENOENT as `absent` and ENOTDIR as `file` (a regular file standing on the way to the path); EACCES, EPERM and anything else propagate, so an unreadable path is never read as free ground. Refusals exit 1 (user error); an unreadable probe exits 2 (internal error).
- The exemption covers only ground a recorded claim and an incoming claim BOTH declare - the same shared-file path, or the same exclusive subtree containing the path - which is exactly what the conflict check compares. A path merely nested inside another template's recorded subtree is arbitrated by nothing (that check compares declared claims for string equality) and is refused like any other occupied path. A recorded subtree is treated as a prefix only at a separator boundary, so `srcx.ts` is not read as being inside `src`.

## What deliberately still passes

- A populated directory whose content stands on no path the assembly would write, including `.git` and stray files. Adding into an unprovenanced project stays supported, which is what makes seed's refusal a redirection to an operation that works.
- A target whose recorded provenance covers ground the incoming claim declares too - most of all the `region-union` shared file an earlier template wrote, whose block materialization carries forward verbatim. A subtree both templates declare is likewise left to the conflict check, which reports it as the contest it is rather than as content nobody recorded.
- A target that does not exist. Materialization creates it, and nothing can be overwritten.

## Surfaces swept

The "add arbitrates declared boundaries only, so it can still overwrite" caveat described a hazard that no longer exists, and it appeared in five places. All of them now state what add does with existing work:

- `packages/cyber-pilot-kit-frontx/skills/project-scaffolding/SKILL.md` (step 2, where the caveat also linked this issue)
- `architecture/features/cli-scaffolding/FEATURE.md` and `architecture/features/ai-project-scaffolding/FEATURE.md`
- seed's own refusal message, which additionally gained a fresh-directory exit clause: add refuses the same directory whenever what it holds stands on the template's own ground, so a refusal naming only add could lead from one refusal to the next
- `README.md` and `QUICK_START.md`, which advertised add against an existing repository while documenting only seed's refusal

The FEATURE spec carries the new instructions (`inst-add-check-ground-free`, `inst-add-if-target-not-directory`, `inst-add-if-ground-occupied`, `inst-add-recheck-ground`), four new state transitions, a new DoD (`cpt-frontx-dod-cli-scaffolding-add-undeclared-content`) with verifiable clauses, and two acceptance criteria. No new ADR.

## Known trade-offs

- A template that declares root files (`template-shell` declares `package.json`, `tsconfig.json` and friends) now refuses over any real repository, because those paths are exactly what such a repository already holds. That is #539's intent - refuse rather than overwrite - and the seed refusal now names the fresh-directory exit so the developer is not sent from one refusal to another. The install-into-an-existing-repository merge story is tracked by #374.
- A dangling symlink is reported as occupied, not absent. `stat` resolves links and finds nothing, but the link is an entry no provenance accounts for and a write through `claimed.txt -> ../escaped.txt` would create the target outside the guarded directory, so the probe consults `lstat` on ENOENT and the guard refuses.
- The re-probe narrows the check-to-write window, it does not close it atomically. Closing it would take an exclusive-create protocol across every write path, out of proportion to what it removes: the guard exists to catch a developer adding into a directory whose content no template recorded, which is not a race. Same reasoning as seed's guard.

## Tests

435 CLI unit tests pass (411 before). New coverage: the refusal itself, the two supported passes above, the recorded-provenance exemption over a `region-union` shared file, the last-moment re-probe, fail-closed probing, target-not-a-directory, the separator-boundary rule, the adapter's own state mapping, and CLI dispatch (which also pins that the real probe is wired at all).

Mutation checks, each confirmed red on the reason the caller receives, then restored:

- neutering both guard call sites: 10 tests fail (nine guard cases plus CLI dispatch)
- deleting only the pre-flight call and keeping the re-probe: 2 tests fail, each reporting `conflict` where the guard refuses first by design (instruction 7 ahead of instruction 10)
- replacing the subtree separator normalization with a bare `startsWith`: 1 test fails - the case where a recorded and an incoming claim both declare `src`, which is what makes the separator decide whether `srcx.ts` reads as ground inside it. The older sibling-path case does not catch this mutation: its incoming template declares no subtree, so the exemption short-circuits before the comparison is reached, and the conflict check backstops the overwrite regardless
- reading only the first contributor's claim (`contributors.slice(0, 1)`): 1 test fails - a path two templates contribute where only the referenced one declares the subtree the recorded claim also declares
- widening the exemption back to any recorded ground: 2 tests fail (nested-subtree, shared-file-under-a-subtree)
- reading a dangling link as absent again: 2 tests fail (the adapter case and the escape case, the latter finding `escaped.txt` written outside the target)

`cfs validate` passes with 0 errors and 171/171 code coverage. `tsc --noEmit`, the test tsconfig, and `eslint --max-warnings 0` are clean.

## Coordination note

The branch for #542 centralizes the segment-aware path compare as `pathWithinSubtree` in `packages/cli/src/paths/relative-path.ts`. `isArbitratedGround` in this PR hand-writes the same rule, and its comment states the conflict check compares subtree strings for equality - which #542 makes untrue. Whichever PR lands second replaces the local compare with the shared predicate and rewrites that comment.

## Summary by CodeRabbit

* **Bug Fixes**
  * `frontx add` now refuses to overwrite existing content outside the template’s declared ownership boundaries.
  * Conflicts report the affected paths and leave existing files unchanged.
  * Added safeguards for files, directories, symlinks, and filesystem access errors.
  * Targets are rechecked immediately before writing to prevent changes during the operation.

* **Documentation**
  * Updated quick-start guides, command references, and scaffolding guidance to explain the safer add behavior and recommend fresh directories when needed.


packages/cli/src/commands/add-template.ts
```diff
@@ -174,3 +244,138 @@ export async function addTemplate(
...
+  claimed: { subtrees: ReadonlySet<string>; sharedFiles: ReadonlySet<string> },
+  occupied: OccupiedBoundaryEntry[],
+): boolean {
+  return occupied.some(
+    ({ boundary }) =>
+      (claimed.sharedFiles.has(path) && boundary.sharedFiles.some((entry) => entry.path === path)) ||
+      boundary.exclusiveSubtrees.some(
+        (subtree) =>
+          claimed.subtrees.has(subtree) &&
+          (path === subtree || path.startsWith(subtree.endsWith('/') ? subtree : `${subtree}/`)),
```
SysoevAndrey, line 389: The description states: *"replacing the subtree separator normalization with a bare `startsWith`: the sibling-path test fails, and the mutated guard lets the overwrite through."*

Neither half reproduces. Applied exactly that mutation in an isolated copy:

```
-  (path === subtree || path.startsWith(subtree.endsWith('/') ? subtree : `${subtree}/`)),
+  (path === subtree || path.startsWith(subtree)),

   Test Files  32 passed (32)
        Tests  432 passed (432)
```

Two independent reasons.

**The named test cannot reach this line.** `entry-flows.test.ts:937` declares `exclusiveSubtrees: []` on the incoming `sibling-template`, so `claimed.subtrees` is empty and the `claimed.subtrees.has(subtree)` conjunct on the preceding line short-circuits before `startsWith` is ever evaluated. That test passes through the `sharedFiles` arm, not the subtree arm its own comment documents.

**The overwrite cannot get through.** For this arm to exempt anything, an incoming and a recorded claim must declare the *identical* subtree string — which `conflict.ts:71` unconditionally reports as a conflict. On a fixture where both declare `src` and the incoming also declares shared file `srcx.ts`, with `srcx.ts` occupied on disk:

| | outcome | file on disk |
|---|---|---|
| as written | `target-holds-undeclared-content` | preserved |
| mutated | `conflict` | preserved |

So the normalization is load-bearing for *which refusal the developer sees*, never for whether a write happens — the conflict check is a silent backstop behind it.

Please (a) correct or drop this claim in the description, and (b) add a test on that fixture asserting `reason === 'target-holds-undeclared-content'`. Nothing currently pins the difference.

packages/cli/src/commands/add-template.ts
```diff
@@ -174,3 +244,138 @@ export async function addTemplate(
...
+// The ground the templates contributing one path declare for themselves — the
+// only ground of theirs a recorded claim can be compared against, since the
+// conflict check compares declared claims and nothing else.
+function claimedGroundOf(
+  contributors: string[],
+  declaredByContributor: Map<string, OwnershipBoundary>,
+): { subtrees: ReadonlySet<string>; sharedFiles: ReadonlySet<string> } {
+  const subtrees = new Set<string>();
+  const sharedFiles = new Set<string>();
+  for (const contributor of contributors) {
```
SysoevAndrey, line 365: Mutating this loop to `contributors.slice(0, 1)` fails **0 tests**. Every guard test has exactly one incoming contributor per grouped path — including the `region-union` case at `entry-flows.test.ts:880`, where the second template sits on the *recorded* side, not among the contributors. The multi-contributor union this function exists to build is unexercised. One test with two incoming templates contributing the same path would close it.

packages/cli/src/commands/add-template.ts
```diff
@@ -127,6 +161,27 @@ export async function addTemplate(
...
+  // @cpt-begin:cpt-frontx-flow-cli-scaffolding-add-template:p1:inst-add-abort-target-not-directory
+  // @cpt-begin:cpt-frontx-flow-cli-scaffolding-add-template:p1:inst-add-if-ground-occupied
+  // @cpt-begin:cpt-frontx-flow-cli-scaffolding-add-template:p1:inst-add-abort-ground-occupied
+  // @cpt-begin:cpt-frontx-state-cli-scaffolding-assembly-op:p2:inst-as-resolved-aborted-target-not-directory
+  // @cpt-begin:cpt-frontx-state-cli-scaffolding-assembly-op:p2:inst-as-resolved-aborted-ground-occupied
+  // Runs as soon as both inputs exist — the staged assembly's own paths and the
+  // ground the target's provenance accounts for — so a repository holding
+  // undeclared content is refused without the pairwise conflict check below
+  // having to run at all.
+  const preflight = await refuseUnlessGroundFree(applyResult.assembly, alreadyOccupied.occupied, targetDir, readTargetPathStateFn);
```
SysoevAndrey, line 169: Deleting only this pre-flight call and leaving the re-check in place fails exactly one test — `refuses on the last-moment re-probe...` — and only because its fake counts probe calls (`probes === 1 ? 'absent' : 'file'`). No test observes an *outcome* difference.

There is no test where the target holds both undeclared content **and** a real boundary conflict. With the pre-flight gone, that case silently reports `conflict` instead of `target-holds-undeclared-content` and the suite stays green. Since running this before `checkAssemblyConflicts` is a deliberate design point the spec states (`FEATURE.md:137` ahead of `:140`), it deserves a test asserting which refusal wins.

architecture/features/cli-scaffolding/FEATURE.md
```diff
@@ -361,8 +409,10 @@ The system **MUST** assemble a repository from one or more independently-applied
...
+- [x] Seeding is refused with no files written and no template resolved when the target directory holds any entry outside the closed non-content set (`.git`, `.DS_Store`, `Thumbs.db`), naming the directory, the content entries found, the add command qualified by what add does with that content — it writes only the ground the template declares and refuses rather than overwriting where content already stands on it — and seeding into a fresh directory as the exit when add would refuse the same directory; a nonexistent target is created, and an existing target that is empty or holds only non-content entries proceeds. (`target`)
 - [x] Seeding is refused, naming the path and offering no add remedy, when the target path exists and is not a directory. (`target`)
+- [x] Adding is refused with no files written when the target directory holds content at a path the staged assembly owns that no already-applied template's recorded provenance accounts for, naming the directory, the occupied paths, and both remedies; a target holding content on no path the assembly owns proceeds untouched, and a path inside an already-applied template's recorded boundaries does not refuse the add. (`target`)
```
SysoevAndrey: > `...and a path inside an already-applied template's recorded boundaries does not refuse the add.`

Contradicted by three things:

- `FEATURE.md:346` — *"A path that merely falls inside another template's recorded exclusive subtree, arbitrated by no check, **is refused** like any other occupied path"*
- `isArbitratedGround` — the exemption requires the **incoming** claim to declare the same subtree; merely falling inside a recorded one is not enough
- `entry-flows.test.ts` — `refuses a path nested inside another template's recorded subtree, which no check arbitrates`

This is `180af47c` narrowing the exemption without updating the criterion. It is also checked `[x]` while the code does the opposite.

Suggested: `...and a path on ground a recorded claim and an incoming claim both declare is left to the conflict check.`

packages/cli/src/commands/add-template.ts
```diff
@@ -127,6 +161,27 @@ export async function addTemplate(
...
+  if (preflight) return preflight;
+  // @cpt-end:cpt-frontx-state-cli-scaffolding-assembly-op:p2:inst-as-resolved-aborted-ground-occupied
+  // @cpt-end:cpt-frontx-state-cli-scaffolding-assembly-op:p2:inst-as-resolved-aborted-target-not-directory
+  // @cpt-end:cpt-frontx-flow-cli-scaffolding-add-template:p1:inst-add-abort-ground-occupied
+  // @cpt-end:cpt-frontx-flow-cli-scaffolding-add-template:p1:inst-add-if-ground-occupied
+  // @cpt-end:cpt-frontx-flow-cli-scaffolding-add-template:p1:inst-add-abort-target-not-directory
+  // @cpt-end:cpt-frontx-flow-cli-scaffolding-add-template:p1:inst-add-if-target-not-directory
+  // @cpt-end:cpt-frontx-flow-cli-scaffolding-add-template:p1:inst-add-check-ground-free
```
SysoevAndrey: Seven begin/end pairs wrap the same two lines, and spec instructions 7/8/9 — siblings in `FEATURE.md` — are nested inside one another here.

`.cf-studio/config/kits/sdlc/codebase/rules.md:65` requires *"wrap the SMALLEST code fragment implementing one CDSL instruction in each begin/end pair"*, and line 92 names the anti-pattern directly: *"a single begin/end pair wrapping the entire multi-instruction function body."*

The distinct fragments live in `refuseUnlessGroundFree` — the target-state read (`:271`), the `'file'` branch and its return, the `occupiedPaths` loop, the final return — and that function carries no markers at all. Moving them there also fixes the sibling/nesting inversion.

Same theme, worth folding into the fix: the re-check at `:214` carries **no** state markers while the identical call at `:175` carries both — and it aborts from `CONFLICT_CHECKED`, a transition `FEATURE.md:236-245` does not declare. Either add the transition, or state in transitions 7/8 that they cover the last-moment re-probe too.

SysoevAndrey (changes requested):
**Request changes.**

The guard is right and it closes #539: neutering both `refuseUnlessGroundFree` call sites kills 9 tests, the fail-closed probing is sound, and the `lstat`-on-ENOENT reasoning about a write escaping through a dangling link is well argued. The e2e fixtures running the real adapter against the real templates are what convinced me the canonical `seed shell -> add mfe` path is intact.

Build claims verified: 432/432 stable over 5 runs, 0 skipped, `tsc --noEmit` and the test tsconfig exit 0, `eslint --max-warnings 0` clean, `cfs validate` `Errors: 0` / `Code coverage: 126/126`. The `411 before` baseline is correct (merge-base with `develop` is `32d9194e`).

What blocks merge is the evidence, not the mechanism. One of the four mutation checks in the description does not reproduce, and re-running the set turned up two more mutations the suite does not catch. The final commit `180af47c` narrowed the exemption without updating an acceptance criterion or the mutation claims, so the artifacts now describe the pre-narrowing rule while the code implements the narrowed one.

packages/cli/src/commands/add-template.ts
itechmeat: Confirmed on both counts - the claim was wrong and nothing pinned the difference. Description corrected; a test on your exact fixture added in 3f1b42c8, asserting `target-holds-undeclared-content`. The bare `startsWith` mutation now reds exactly that test.

packages/cli/src/commands/add-template.ts
itechmeat: Added in 3f1b42c8: two templates contribute the same path and the exemption only holds on their unioned claims. `contributors.slice(0, 1)` now reds exactly that test.

packages/cli/src/commands/add-template.ts
itechmeat: Added in 3f1b42c8: the target holds both undeclared content and a contested subtree, asserting `target-holds-undeclared-content` wins. Deleting the pre-flight now reds it on outcome.

architecture/features/cli-scaffolding/FEATURE.md
itechmeat: Applied your wording in 3f1b42c8.

packages/cli/src/commands/add-template.ts
itechmeat: Restructured in 3f1b42c8: markers moved onto their fragments inside `refuseUnlessGroundFree`, instructions 8/9 un-nested. The re-check now carries state markers via two new `FROM CONFLICT_CHECKED TO ABORTED` transitions - each table row declares exactly one FROM/TO pair, so folding the re-probe into 7/8 would have broken the row shape. cfs validate: Errors 0, 126/126. The same stacked pattern exists at the seed call site - out of this PR's scope, left as is.

itechmeat:
All five taken, one commit 3f1b42c8 on top of 180af47c, no rebase. Ledger re-verified, every mutation now reds on outcome: (a) neuter both call sites - 10 tests; (b) drop lstat-on-ENOENT - 2; (c) widen the exemption - 2; (d) bare `startsWith` - 1, the new fixture test only; (e) `contributors.slice(0, 1)` - 1, its new test only; (f) delete the pre-flight only - 2. 435 tests, 0 skipped; tsc, eslint and cfs validate clean. The description's mutation paragraph is corrected to match.

SysoevAndrey (approved):

packages/cli/src/adapters/fs-target-path.ts
```diff
@@ -0,0 +1,64 @@
...
+import type { ReadTargetPathStateFn, TargetPathState } from '../commands/add-template';
+
+/** Builds the real `stat`-backed target-path probe the `frontx` executable uses. */
+export function createFsReadTargetPathStateFn(): ReadTargetPathStateFn {
+  return async function readTargetPathState(absolutePath: string): Promise<TargetPathState> {
+    try {
+      const stats = await stat(absolutePath);
+      return stats.isDirectory() ? 'directory' : 'file';
+    } catch (error) {
+      if (isErrnoCode(error, 'ENOENT')) return (await standsAsDanglingLink(absolutePath)) ? 'file' : 'absent';
```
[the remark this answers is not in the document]
itechmeat: Valid scenario, but not introduced by this PR - writes could traverse live links long before this guard existed. The spec deliberately scopes this PR's symlink rule to the dangling final component; containment for live ancestor links is deferred to a follow-up hardening pass. No change in 9e13f310.

GeraBart (approved):

Issue #531, opened 2026-08-04
[ARCH] Two configs still point at the pre-split src/ layout: a CI loop that matches nothing and knip patterns that never resolve


gs-layer:
## What happens

Two config files point at directories that do not exist anywhere in this repository. One CI step has therefore never done anything; one linter reports the same three warnings on every run.

## Reproduce

**`.github/workflows/main.yml:41`** — the "Install MFE dependencies" step:

```bash
for mfe in src/mfe_packages/*/; do
```

There is no `src/mfe_packages/` at the repo root. MFE packages live at `template-mfe/src-app/mfe_packages/*`. The glob matches nothing, the loop body never runs, and the step passes — which is why nobody noticed. Its comment explains why each MFE needs its own `npm ci`; that reasoning is still there, doing nothing.

**`knip.json:6-10`**:

```json
"entry": ["src/app/main.tsx", "src/screensets/*/*[Ss]creenset.tsx"],
"project": ["src/**/*.{ts,tsx}"],
"ignore": ["src/screensets/_blank/**"]
```

Neither `src/app/main.tsx` nor `src/screensets/` exists. Run `npm run arch:check` and every run prints:

```
[src/app/main.tsx, …]   knip.json  Refine entry pattern (no matches)
[src/**/*.{ts,tsx}]     knip.json  Refine project pattern (no matches)
/@types\/*/             knip.json  Remove from ignoreDependencies
```

## Why

Both paths describe the pre-split layout, when the template's app lived at the repo root. The split into `template-shell/` + `template-mfe/` (#470) moved everything under `src-app/`, and neither config followed.

## Why it matters

Same class as #483 — a check that exists and runs and passes without checking anything. The knip case is the more corrosive of the two: three warnings on every `arch:check` train people to skim past that section, which is exactly where a real finding would appear.

## What a fix needs

- Point the MFE loop at the real path, or delete the step if `template-validate` (#491) already covers it — worth deciding rather than assuming.
- Repoint knip's entry/project globs at the template trees, or scope knip out of the root workspace if it no longer has one to analyse.

tscbmstubp:
Cross-link: the knip half of this issue overlaps #483 item 2 — PR #547 and PR #554 currently carry byte-identical `knip.json` rewrites. Proposed split (per review on #547): #547 lands `knip.json`, #554 keeps the `main.yml` fix.

PR #547, opened 2026-08-07
fix(arch): point knip at the post-split tree and fail on dead source roots (#483)

tscbmstubp:
## Summary

Item 2 of #483 (items 1 and 3 land via #491): the root knip workspace still pointed `entry`/`project` at `src/`, gone from the repo root since the template split (#470). Every root glob matched nothing, knip demoted that to a "Configuration hints" footnote, and `arch:unused` runs with `--no-exit-code` — a check that looked live while analyzing zero root files.

## Changes

- **`knip.json`** — root workspace now describes what the root actually owns post-split: `scripts/**/*.{ts,mjs}` as the project, `scripts/**/*.test.{mjs,ts}` as entries (mirroring `vitest.scripts.config.mjs`'s include), `scripts/manual-auth-tests/` ignored (imports `@gears-frontx/framework`, a template-shell package unresolvable from this root). Also dropped config referring to things that no longer exist: the `@types/*` `ignoreDependencies` entry (knip itself hinted it suppresses nothing) and the `@gears-frontx/depcruise-config/*` entry (no hint fired for it, but it's equally dead — the dependency resolves through `packages/api`'s own declaration), ignore patterns for `src/screensets`, `eslint-plugin-local/`, `vite.config.ts`, `.eslintrc.cjs`, and the `*Screen.tsx` `ignoreIssues` rule whose subjects all live in `template-*/` trees knip doesn't scan.
- **`package.json`** — both knip scripts gain `--treat-config-hints-as-errors`, which overrides `--no-exit-code` (verified: exit 1). `arch:unused` keeps unused-code findings advisory, but a configured source root matching nothing now fails instead of hinting. This gate runs everywhere `arch:check` does — CI **and** local commits via `.pre-commit-config.yaml`.
- **`QUICK_START.md`** — the `arch:unused` line now notes that configuration hints fail the run.

## Trade-off

`--treat-config-hints-as-errors` gates the whole configuration-hint class, not just dead globs: once unused files exceed 20, knip synthesizes top-level/workspace-unconfigured hints (`reporters/util/configuration-hints.js`), and those would fail `arch:unused` too despite `--no-exit-code`. The repo is at 6 unused files today, so this is a latent tripwire, accepted deliberately — knip 5 offers no per-hint-type selector, and tripping it would mean unused code piled up enough to deserve attention anyway.

## Verification

- Healthy config: `npm run arch:unused` exits 0, zero configuration hints.
- Regression proof: re-pointing `project` at the dead `src/**` glob makes `arch:unused` exit 1 with the two "no matches" hints surfaced as errors.
- Coverage proof: a dead `.mjs` file dropped into `scripts/` is reported under "Unused files".
- Finding set is byte-identical before and after the rewrite — no new noise, nothing previously reported was lost.

## Coordination

#554 currently carries a near-identical `knip.json` rewrite from #531 (identical except manual-auth-tests exclusion style, changed here per CodeRabbit review). Proposed split (per review): this PR carries `knip.json`; #554 drops that half and keeps its `main.yml` change.

Refs #483.

## Summary by CodeRabbit

* **Documentation**
  * Clarified that configuration hints cause the unused-code validation to fail.

* **Chores**
  * Improved unused-code validation coverage for test and source files.
  * Updated validation commands to treat configuration hints as errors.
  * Excluded manual authentication tests from unused-code analysis.


knip.json
```diff
@@ -3,11 +3,10 @@
   "workspaces": {
     ".": {
       "entry": [
-        "src/app/main.tsx",
-        "src/screensets/*/*[Ss]creenset.tsx"
+        "scripts/**/*.test.{mjs,ts}"
       ],
-      "project": ["src/**/*.{ts,tsx}"],
-      "ignore": ["src/screensets/_blank/**"]
+      "project": ["scripts/**/*.{ts,mjs}"],
```
itechmeat: 🟠 Open #554 carries a character-for-character identical rewrite of this same knip.json, from a different issue (#531) - two open PRs, same file content, and neither references the other. develop has not touched the file, so both are live and whichever merges second gets a manual conflict across all these lines. Since this PR is three days older and scoped exactly to #483 item 2, while #554 also carries an unrelated main.yml deletion, the cheapest split looks like: this one lands knip.json and @gs-layer drops that half from #554 - but that needs an explicit decision rather than two PRs assuming. Cross-linking #483 item 2 and #531 would keep the duplication from being re-derived.

package.json
```diff
@@ -53,8 +53,8 @@
     "arch:guards": "npm run build:packages:config && npx tsx scripts/verify-guard-configs.ts",
     "lint:sdk": "eslint packages/api/src --max-warnings 0",
     "arch:deps:core": "npx dependency-cruiser packages/api/src packages/gts-plugin/src packages/mfes/src packages/telemetry/src --config internal/depcruise-config/core.cjs --output-type err-long",
-    "arch:unused": "knip --no-exit-code",
-    "arch:unused:strict": "knip",
+    "arch:unused": "knip --no-exit-code --treat-config-hints-as-errors",
```
itechmeat, line 56: 🟡 [Not a blocker] --treat-config-hints-as-errors gates more than the dead-glob class this PR fixes: once unused files exceed 20, knip synthesizes top-level/workspace-unconfigured hints (reporters/util/configuration-hints.js), and with this flag those become errors - so arch:unused would exit 1 despite --no-exit-code. The repo is at 6 unused files today, so this is a latent tripwire, not a present break, and knip 5 has no per-hint-type selector - worth a sentence in the body as a documented trade-off.

itechmeat, line 56: 🟡 [Not a blocker] The body frames the new gate as CI ("run in CI via arch:check"), but arch:check is also wired into .pre-commit-config.yaml, so knip config rot now fails local commits too - worth saying. And QUICK_START.md:268 still lists arch:unused as "unused-code report (knip)": the findings half is still a report, but the config-hint half is a gate now, so the line undersells what a failure there means.

knip.json
```diff
@@ -20,20 +19,10 @@
...
     "**/template-sources/**",
-    "eslint-plugin-local/**",
-    "**/*.config.{ts,js,cjs,mjs}",
-    "vite.config.ts",
-    ".eslintrc.cjs"
-  ],
-  "ignoreDependencies": [
-    "@types/*",
-    "@gears-frontx/depcruise-config/*"
+    "**/*.config.{ts,js,cjs,mjs}"
```
itechmeat, line 24: 🔵 [Not a blocker] The body says knip hinted that both dropped ignoreDependencies entries suppress nothing, but the captured outputs in #531 and #554 show that hint only for @types/*. The removal still looks safe - @gears-frontx/depcruise-config resolves through packages/api's own declaration, and definitely-typed packages auto-ride their host package - but the justification as written cites a hint that seems not to have fired for the second entry.

itechmeat (commented):
The change itself is verified sound - the flag exists in the pinned knip, genuinely overrides --no-exit-code for the hint class (reproduced: exit 1 on a dead glob, exit 0 with zero hints on this config), and every dropped config entry checks out as dead. What holds approval is coordination, not code: open #554 carries a byte-identical rewrite of this same knip.json from a different issue, and an explicit decision about which PR carries the file should land before either merges. Everything else below is a sentence here or there.

package.json
```diff
@@ -53,8 +53,8 @@
...
+    "arch:unused:strict": "knip --treat-config-hints-as-errors",
```
itechmeat, line 57: 🔵 [Not a blocker] grep finds no caller of arch:unused:strict - not CI, not arch:check, not pre-commit, not the docs. Keeping it flag-synced with arch:unused costs nothing and is right as-is; just checking the intent is "manual strict run" rather than wiring that was meant to happen.

gs-layer (approved):

knip.json
tscbmstubp: Agreed on the split: this PR carries `knip.json` (older, scoped exactly to #483 item 2), and #554 drops that half, keeping its `main.yml` change. Added a Coordination section to the body and I'm cross-linking #483/#531 so the duplication isn't re-derived. @gs-layer — flagging for your explicit ack before either merges.

package.json
tscbmstubp: Documented — the body now has a Trade-off section spelling out the >20-unused-files synthesized-hint class becoming exit-1, why it's accepted (no per-hint-type selector in knip 5), and that we're at 6 today.

tscbmstubp: Both fixed: the body now says the gate runs everywhere `arch:check` does — CI and local commits via `.pre-commit-config.yaml` — and QUICK_START.md:268 now reads "unused-code report (knip); config hints fail the run" (8c978d2c).

knip.json
tscbmstubp: Corrected — the body now attributes the hint to `@types/*` only, and justifies the `@gears-frontx/depcruise-config/*` removal on its own terms (resolves through `packages/api`'s declaration, no hint fired).

package.json
tscbmstubp: Intent is a manual strict run — no wiring was meant to happen. Kept flag-synced so a hand-run gives the same config-hint behavior as the CI path.

knip.json
```diff
@@ -3,11 +3,10 @@
...
     ".": {
       "entry": [
-        "src/app/main.tsx",
-        "src/screensets/*/*[Ss]creenset.tsx"
+        "scripts/**/*.test.{mjs,ts}"
       ],
-      "project": ["src/**/*.{ts,tsx}"],
-      "ignore": ["src/screensets/_blank/**"]
+      "project": ["scripts/**/*.{ts,mjs}"],
+      "ignore": ["scripts/manual-auth-tests/**"]
```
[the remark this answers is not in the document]
tscbmstubp: Applied in bdb27911 — `manual-auth-tests` is now excluded via the negated `project` glob. For the record, the severity was overstated: verified empirically on knip 5.68.0 that both variants produce byte-identical findings and exit codes, including the edge case of a `scripts/` helper imported only from `manual-auth-tests/` (flagged unused under both, since those files aren't entries and are never traversed). Applied as a style alignment with knip's documented preference for project boundaries over `ignore`, not as a functional fix.

GeraBart (approved):

Issue #546, opened 2026-08-07
Boundary declarations no check refuses: manifest self-overlap, shared file inside a foreign exclusive subtree, degenerate spellings


itechmeat:
## Description

The segment-aware boundary work (#542) closes prefix capture and nested claims between two templates' exclusive subtrees. Three adjacent declaration shapes still pass every check, and all three describe ground two parties cannot both hold:

1. **Self-overlap within one manifest.** A template declaring both `src/` and `src/config/` in its own `exclusiveSubtrees` passes: the conflict check compares claims of different owners only, and manifest validation does not compare a manifest's declarations against each other.
2. **Shared file inside another template's exclusive subtree.** Template A declares `src/` exclusively; template B declares a shared file at `src/config.json`. The conflict check compares subtrees against subtrees and shared files against shared files, so the cross-tier pair never meets. By the DoD's own argument this is the same impossibility: an exclusive subtree admits no other writer at any depth.
3. **Degenerate spellings.** `isWellFormedRepoRelativePath` admits `src//` (and `/`-ish variants beyond a single trailing slash). Under the segment rule such a claim addresses nothing at all, so it is dead weight that reads as coverage; before #542 the same spelling captured by bare prefix.

Item 1 and 3 belong to manifest pre-publish validation (`validate-contract.ts`) - refusing at authoring time, before any assembly exists. Item 2 needs a decision first: ADR-0032 frames the conflict check as arbitrating within each tier, and widening it across tiers changes what the check means, so this is an architect call rather than a mechanical extension.

## Notes

Spun out of #542's review round. None of the three is a regression - all pre-date the segment-aware change; #542 makes them the remaining members of the class it closed.

itechmeat:
External review of the #542 branch adds two members to this class:

- **More degenerate spellings pass `isWellFormedRepoRelativePath`** beyond `src//`: a `.` segment (`./src`), a literal backslash (`src\config`), surrounding whitespace (` src`). Same failure shape: the segment predicate matches none of them, so the claim is dead weight that reads as coverage, and a pair like `src//` vs `src/config` escapes conflict detection while both manifests validate. Same fix venue: tighten the well-formedness contract to the discipline `isSafeRelativePath` already applies.
[...]

PR #548, opened 2026-08-07
fix: compare ownership boundaries by path segment, not by string

### LINKED ISSUE #542: Ownership boundary matching is not segment-aware: bare prefix in the assembler filter, nested subtree claims never conflict
## Description

Two sites arbitrate ownership boundaries with string comparisons that ignore path-segment structure. Both are latent today only because the shipped templates' declarations happen to avoid the trap.

1. `packages/cli/src/scaffold/assembler.ts:16` - `isWithinDeclaredBoundaries` filters assembly content with a bare `item.path.startsWith(subtree)`. A subtree claim `src` also captures `src-app/...` and `srcx.ts`. Masked today because both reference templates declare directory subtrees with a trailing slash; a template author who writes `"src"` gets silent capture of sibling trees. PR #541 introduces a segment-aware compare for the add guard's own exemption check, but this assembler filter runs for both seed and add materialization and still uses the bare prefix.

2. `packages/cli/src/scaffold/conflict.ts:69` - `checkAssemblyConflicts` compares two templates' `exclusiveSubtrees` claims by string equality only, so nested claims (`src` vs `src/config`) never register as a conflict. The e2e suite documents this as a KNOWN BLIND SPOT (`template-split.e2e.test.ts`), and PR #541 had to narrow the add guard's provenance exemption specifically because arbitration cannot detect nesting.

Work:

1. Segment-aware comparison in both sites: a path is inside a subtree iff `path === subtree || path.startsWith(subtree + '/')`; nesting between two claims is the same rule applied both ways. `pathsNest` in `packages/cli/src/paths/relative-path.ts:39` already implements the two-way check - reuse a shared helper rather than grow a third copy.
2. Flip the KNOWN BLIND SPOT e2e tests from documenting the hole to proving the detection; sync the test-helper mirror (`__tests__/helpers/template-split-fixtures.ts:118`).
3. The behavior change is refusal-only: assemblies that pass today keep passing unless a claim actually captures a foreign path or nests inside another claim.

## Estimate

0.5 engineer-day.

## Notes

Found during the July template-mechanism review; re-verified against develop @ 32d9194e. The class does not exist on the upgrade path (`ownershipForPath` matches `sharedFiles` exactly), so this issue covers the scaffold layer only. Independent of #494 and of PR #541 by files.

### LINKED ISSUE #549: gitignore and test-fixture exclusions do not cover the .omo agent-runtime directory
## Description

`.omo/` is an agent-session runtime directory (same family as the `.omx`/`.omc`/`.prs` set #532 already ignores), and it is covered neither by `.gitignore` nor by the test fixtures' directory exclusions. Two incidents from one working session:

1. **`git add -A` stages runtime state.** An agent's `git add -A` pulled `.omo/run-continuation/ses_*.json` into the index; the `end-of-file-fixer` pre-commit hook rewrote the live session file and aborted the commit. Until `.omo/` is ignored, `git add -A` is unsafe in a repository agents work in.
2. **Irreproducible test failures.** `template-split.e2e.test.ts` fixtures that walk the real `template-shell/`/`template-mfe/` trees byte-for-byte failed once and never again in nine runs. The fixture helper's `EXCLUDED_DIRS` documents this exact class for `.omc` ("the local-fetch walk would read live session files that this very agent session can be rewriting concurrently") but does not list `.omo`, so a session directory landing under a template tree makes those fixtures fail for a reason that does not exist.

## Fix

Two one-liners: add `.omo/` to `.gitignore` next to the #532 set, and add `'.omo'` to `EXCLUDED_DIRS` in `packages/cli/src/__tests__/helpers/template-split-fixtures.ts`.

itechmeat:
## Description

Ownership boundaries were arbitrated by string comparison in two places, and both ignored path-segment structure. The assembler's content filter used a bare `startsWith`, so a subtree claim `src` also captured `src-app/...` and `srcx.ts`; the conflict check compared two templates' `exclusiveSubtrees` claims by string equality only, so nested claims (`src` vs `src/config`) never registered as a conflict - the e2e suite carried both as documented KNOWN BLIND SPOT tests. Both were latent only because the shipped templates' declarations happen to avoid the trap. Closes #542.

One predicate now carries the rule: `pathWithinSubtree` in `paths/relative-path.ts` (a path is inside a subtree iff it equals it or extends it past a `/`; one trailing slash on either side is normalized away). `pathsNest` is rewritten as the two-way application, the assembler filter and the conflict check consume it, and the reserved-namespace checks in `validate-contract.ts` - two hand-written copies of the same rule in a file that already imported from this module - now call it too. Nested claims between two templates refuse with a ground naming both claims (`ground "src/ overlaps src/config/" contested by: outer, inner`); identical claims keep the old single-string ground. The blind-spot tests are flipped from documenting the hole to proving the detection, and the test helper that mirrored the production rule now calls it instead.

The widening is refusal-only, with one deliberate narrowing found by external review: the conflict check compares only pairs where at least one side is a staged (incoming) claim. Two already-recorded templates whose boundaries nest - recordable under the old equality check - no longer block unrelated adds; arbitration at add time judges the incoming assembly against existing state, not the existing records against each other. Both regression tests pin this, and a dedicated mutation probe reddens exactly them.

Spec-first: FEATURE `cli-scaffolding` instructions, DoD and acceptance criteria updated; ADR-0031, ADR-0032 and DESIGN.md rewritten in place where they still described the equality-only comparison ("the same exclusive subtree" → intersecting, with the definition living in ADR-0032 and the others pointing at it). No new ADR.

## Conscious decisions

- `isArbitratedGround` in PR #541 hand-writes the same segment compare; whichever PR lands second replaces it with the shared predicate and rewrites the comment this PR makes untrue (coordination note also sits in #541's description).
- Three adjacent declaration shapes no check refuses (manifest self-overlap, a shared file inside a foreign exclusive subtree, degenerate spellings like `src//` or `./src`) are tracked in #546 - tightening the well-formedness contract is authoring-time validation, a different layer.
- Windows separators in content paths (platform `path.join` vs `/`-segment claims) pre-date this change on both sides and are noted in #546.

## Verification

- 460 CLI tests (+49 over develop), monorepo suite 8/8, tsc/eslint/`cfs validate` (125/125) clean, `validate:templates` passes both shipped templates - their declarations are disjoint under the segment rule, nothing newly refuses.
- Mutation evidence: reverting the assembler to bare `startsWith` reddens 2 tests; reverting the conflict check to string equality reddens 7; dropping trailing-slash normalization reddens 27; degrading the predicate to a bare prefix reddens 27; removing the staged-pair narrowing reddens exactly the 2 regression tests that pin it.
- Real-binary QA: nested claims across two applied templates refuse naming both templates and both claims with the target byte-identical after refusal; claims sharing only a string prefix (`src/` + `src-app/`) proceed; a regular template-shell seed is unaffected.
- The direct predicate table (33 cases) documents the deliberate non-normalizations, including that a declaration spelled `src//` addresses nothing (#546).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved conflict detection for overlapping exclusive subtrees, including nested paths and trailing-slash variations.
  * Prevented false conflicts for similarly prefixed but unrelated paths.
  * Added clearer conflict reporting and checks against existing repository boundaries.
  * Corrected reserved namespace validation to use whole path segments.

* **Security & Privacy**
  * Excluded `.omo` session-state directories from fetched, generated, and committed project files.

* **Tests**
  * Added comprehensive coverage for path boundaries, conflict scenarios, namespace validation, and local fetching.

Along the way this PR also closes #549: the `.omo/` agent-session runtime directory is now excluded on all three surfaces - `.gitignore`, the test fixtures' directory walk, and the local-fetch adapter's default exclusions (the `.omc` precedent, one directory over) - so `git add -A` no longer stages live session files and a session directory landing under a template tree no longer reads as undeclared template content.

Closes #549


packages/cli/src/scaffold/conflict.ts
```diff
@@ -64,17 +75,39 @@ export function checkAssemblyConflicts(
...
+      // not a relaxation of anything that used to be caught.
+      if (!claimA.staged && !claimB.staged) continue;
+
       // @cpt-begin:cpt-frontx-algo-cli-scaffolding-conflict-check:p1:inst-cc-if-subtree-clash
-      // Detect two templates claiming the same exclusive subtree.
+      // Overlap (`pathsNest`), not string equality: nothing downstream
+      // arbitrates a claim this check waves through, so admitting only
+      // identical spellings would leave a nested pair unarbitrated
+      // (`cpt-frontx-dod-cli-scaffolding-conflict-check`).
       for (const subtreeA of claimA.boundary.exclusiveSubtrees) {
```
SysoevAndrey, line 105: `checkAssemblyConflicts` is documented as the SOLE authority for boundary-collision arbitration, but the widened rule is applied only within like kinds: `exclusiveSubtrees` x `exclusiveSubtrees` here, and `sharedFiles` x `sharedFiles` below. An exclusive subtree that *contains* another template's shared-file path is never compared, so that collision passes pre-flight and reaches materialization.

Verified by probe against this branch:

- Template B is applied with `sharedFiles: [{ path: 'src/main.ts', mergeStrategy: 'region-union', ownedRegions: ['k'] }]`, its region block on disk and in provenance. `frontx add A`, where A declares `exclusiveSubtrees: ['src/']` and ships `src/main.ts`, gets `{ ok: true }` from this function -- A's subtrees are compared only against B's (empty) `exclusiveSubtrees`, never against B's shared files. `composeSharedFiles` then takes the single-exclusive-contributor branch, which never reads the file on disk, and writes A's whole file. B's region block is destroyed, provenance still names B as its owner, exit 0, no warning.
- Mirror case inside one assembly: A declares `exclusiveSubtrees: ['src/']`, B declares `sharedFiles: [{ path: 'src/config.ts', mergeStrategy: 'exclusive' }]`. The check passes; `composeSharedFiles` hits `if (!hasRegionUnionContribution) continue`, so `src/config.ts` is written by nobody -- seed reports success with the file silently missing.

The gap predates this PR, but this PR is the one that centralizes the containment rule and re-states this function's authority in the docblock right above. Either add the cross-kind pair (`pathWithinSubtree(sharedX.path, subtreeY)` in both directions) or record the hole explicitly alongside the shapes already tracked in #546, so the "SOLE authority" claim stays honest.

packages/cli/src/scaffold/conflict.ts
```diff
@@ -64,17 +75,39 @@ export function checkAssemblyConflicts(
...
+      // impossible for a reason no later add introduced.
+      //
+      // The pair reaches here at all because a repository can hold one:
+      // arbitration only became segment-aware after these records were written,
+      // and the equality comparison that admitted a nested occupant is gone
+      // while its provenance record remains. Under that comparison an EQUAL
+      // occupied pair was unreachable — the second template's own add would have
+      // been refused — which is why the narrowing was not needed before and is
+      // not a relaxation of anything that used to be caught.
+      if (!claimA.staged && !claimB.staged) continue;
```
SysoevAndrey, line 98: The justification in this comment does not hold, and the narrowing hides a reachable case.

> Under that comparison an EQUAL occupied pair was unreachable -- the second template's own add would have been refused

That holds only if occupied boundaries were frozen at add time. They are not: `occupiedBoundariesFromProvenance` reads each installed template's *current* manifest, so a claim can widen after its provenance record was written.

Reachable sequence: `shell` occupies `src-app/app/`, `mfe` occupies `src-app/mfe_packages/`, both adds passed. `frontx install` upgrades `shell` to a version whose manifest now claims `src-app/`. The two occupied claims now genuinely overlap, on ground both templates hold files in. Before this change any `frontx add` refused and named `src-app/ ... shell, mfe` -- the only signal the developer gets that the repository is inconsistent and that the shell upgrade must be rolled back or the mfe relocated. After it, every add proceeds silently, and the next bounded update of the shell clears `src-app/` recursively and takes mfe's files with it.

The part that is actually wrong is "the developer has no move that resolves a contest between them". They do have a move -- roll back the shell, or relocate the mfe -- they just no longer get told they need to make it. If the goal is only to stop unrelated adds from being blocked, an occupied x occupied overlap could warn instead of refuse, which keeps the signal without the collateral.

packages/cli/src/paths/relative-path.ts
```diff
@@ -30,11 +34,62 @@ function hasUnsafePathChar(value: string): boolean {
...
 // outer one clears its directory recursively and takes the inner one's files
 // with it, leaving the inner template listed in the index with nothing on disk.
+// The guard cannot change that caller's verdicts — `isSafeRelativePath` rejects
+// an identity with an empty segment, so no identity reaching here carries a
+// slash for it to fire on.
+// Over two manifests' exclusive-subtree claims it is the overlap the pre-flight
+// conflict check refuses (`cpt-frontx-dod-cli-scaffolding-conflict-check`).
 export function pathsNest(a: string, b: string): boolean {
-  return a === b || a.startsWith(`${b}/`) || b.startsWith(`${a}/`);
+  if (addressesNoLocation(a) || addressesNoLocation(b)) return false;
```
SysoevAndrey, line 102: This guard also changes the verdict for two *identical* degenerate claims, which is a regression against the old equality check rather than part of the widening.

`isWellFormedRepoRelativePath` accepts `src//` -- it rejects empty, absolute and `..` segments, but never an empty segment -- so two templates can both publish `exclusiveSubtrees: ['src//']`. Old behaviour: `subtreeA !== subtreeB` was false, the conflict check refused the assembly and named both templates. New behaviour: `pathsNest('src//', 'src//')` is `false`, the assembly passes, and `isWithinDeclaredBoundaries` admits nothing for either side (root `src/` never prefixes `src/main.ts`). `frontx seed a,b ./repo` reports success and writes a provenance record for each template while the repository receives none of their files.

The reasoning above the function ("a declaration addressing no location nests with nothing, because there is no file it and another claim could both own") is right for `src//` vs `src`, but not for `src//` vs `src//`: there the two claims are the same declaration, and the refusal was the developer's only signal that the assembly would produce nothing. An `if (a === b) return true;` ahead of the guard preserves the old verdict without touching anything else in the rewrite.

SysoevAndrey (commented):
Three findings from a detailed review of the boundary-arbitration change. All three are about the pre-flight check's coverage rather than the predicate itself -- `pathWithinSubtree` and its centralization look right, including the trailing-slash normalization and the deliberate non-normalizations documented in the predicate table.

packages/cli/src/paths/relative-path.ts
itechmeat: Implemented in 6380c9a9 as proposed; predicate table, docblock and DoD scoped to distinct claims, plus a refusal test for the identical pair.

packages/cli/src/scaffold/conflict.ts
itechmeat: Justification rewritten in 6380c9a9 (conflict.ts, FEATURE.md, ADR-0032): confirmed occupied boundaries are read from the current manifest, so both old assertions were wrong. The comment now states the real rationale - a pre-existing occupied-occupied inconsistency should not block unrelated adds - and names the cost (a post-add widening is no longer surfaced at add time) and the developer's moves. Warn-instead-of-refuse needs a warning channel ConflictVerdict does not have - a contract change this PR does not make. One correction to the scenario: the next bounded shell update would not clear src-app/ recursively - upgrade removes only files in the shell's own baseline-to-target diff, so the exposure is per-file overwrite of paths both templates ship, not a recursive wipe.

packages/cli/src/scaffold/conflict.ts
itechmeat: Took the second option in 6380c9a9: docblock, FEATURE.md and ADR-0032 now bound the sole-authority claim to the shapes the check judges, with the subtree x foreign-shared-file gap pointed at #546. The cross-kind comparison stays out of this PR - the gap predates the widening and needs its own ground shape and instruction.

SysoevAndrey (approved):

packages/cli/src/commands/add-template.ts
```diff
@@ -384,9 +386,7 @@ function isArbitratedGround(
     ({ boundary }) =>
       (claimed.sharedFiles.has(path) && boundary.sharedFiles.some((entry) => entry.path === path)) ||
       boundary.exclusiveSubtrees.some(
-        (subtree) =>
-          claimed.subtrees.has(subtree) &&
-          (path === subtree || path.startsWith(subtree.endsWith('/') ? subtree : `${subtree}/`)),
+        (subtree) => claimed.subtrees.has(subtree) && pathWithinSubtree(path, subtree),
       ),
```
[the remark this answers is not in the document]
itechmeat: Real one, fixed in 5d403ac6. Took mutual containment rather than the proposed triple: it says "the same directory" and, unlike `pathsNest`, still excludes nesting, which this exemption must keep refusing. Worth noting the outcome was a refusal either way — the conflict check reports the pair it is handed — so what changed is which cause the developer reads, not whether the add is stopped. Regression test pins it; restoring the string match reddens it.

gs-layer (approved):
Approved. Current head fixes the segment-aware ownership boundary issue and the latest add-guard raw-string gap. The previous review findings are addressed or explicitly scoped to #546, and all live checks are green. Merge-order context: this overlaps #545/#556 in materialize.ts, but I do not see a required dependency; normal rebase/conflict resolution is enough if either lands first.

gs-layer (approved):

architecture/ADR/0031-template-ownership-boundary-declaration.md
```diff
@@ -61,7 +61,7 @@ The scope of this decision is the shape and tiers of the ownership-boundary decl
...
+Compliance is confirmed by pre-publish validation and a continuous-integration check on the CLI package: pre-publish validation rejects a template whose boundary declaration is malformed — an exclusive subtree that is not a well-formed path, or a shared-file region without a declared merge (`cpt-frontx-fr-cli-template-validate-prepublish`). A fixture check asserts two templates each declaring disjoint regions of one shared file are both accepted and that their declared merges compose the file, while two templates declaring intersecting exclusive subtrees — the same subtree, one nested inside the other, or one directory written with and without a trailing slash — or the same shared-file region are detected as an intersection by the conflict check (`cpt-frontx-adr-assembly-conflict-prevention` owns what "intersecting" means). Design and code review confirm the two tiers are carried in the manifest and that the concrete schema is owned by the manifest FEATURE.
```
GeraBart, line 64: Actually, I think that this ADR should leave on either CLI package level, not the root project level. Maybe no need to address right now, because it seems that it's some leftover of the recent SDLC artifacts refactoring. So we can create a follow-up issue

GeraBart (approved):

Issue #551, opened 2026-08-07
Build scaffolded screens from @gears-frontx/ui-kit instead of vendored copies (_blank-mfe first)


itechmeat:
## Goal

Screens realized by the project-scaffolding flow should be built from `@gears-frontx/ui-kit` (published 2026-08-07, `0.3.0-alpha.1`) instead of the vendored shadcn-style copies the templates carry today. The kit is a versioned npm dependency with one source of truth - copying components as files into every project is exactly the drift this package exists to end.

## Scope: `_blank-mfe` only

Replace the vendored components in `template-mfe/src-app/mfe_packages/_blank-mfe/src/components/ui/` (`button.tsx`, `card.tsx`, `skeleton.tsx`) with imports from `@gears-frontx/ui-kit`, pinned **exactly** to `0.3.0-alpha.1` (no ranges - the mfes/gts-plugin publish skew taught us what loose pins cost). `_blank-mfe` is the scaffold source the skill copies for every new screen, so this one change makes every future screen kit-based automatically.

Out of scope: the shell chrome (`template-shell/src-app/app/components/ui/` - sidebar, sonner, avatar, skeleton) and `demo-mfe`'s 38-file showcase set. Higher risk, no demo payoff; separate follow-up.

## Verified facts (checked 2026-08-07 in a live seeded project)

- MFEs render inside Shadow DOM: the host attaches a shadow root (`attachShadow` in `@gears-frontx/mfes`), and `ThemeAwareReactLifecycle.mount()` explicitly handles `container instanceof ShadowRoot`.
- A style bridge already exists: `adoptHostStylesIntoShadowRoot()` clones every `<style>` and `<link rel="stylesheet">` from `document.head` into the shadow root at mount time - the architecture already expects component styles to arrive via head.
- `@gears-frontx/ui-kit` ships per-component CSS Modules (`import './chunks/button.css'`, hashed class names) plus `dist/theme.css` with the semantic tokens (shadcn-derived structure, Studio palette values).
- Peer deps `react`/`react-dom` `^19` match the template (19.2.4).

## Risks to verify FIRST (before committing to the approach)

1. **CSS injection timing.** The style bridge copies head styles once, at mount. Verify that in a federated MFE build the kit's CSS chunks land in `document.head` before the MFE mounts (top-level imports should guarantee this; lazy-loaded kit components would not). A half-hour spike - one kit Button inside `_blank-mfe`, built and viewed in the browser - answers this.
[...]

PR #552, opened 2026-08-07
fix: keep template example MFEs out of an applied project's product menu

### LINKED ISSUE #550: Seeded projects surface template example screens in the product menu
## Problem

A freshly seeded project (`frontx seed` template-shell + `frontx add` template-mfe) surfaces the template's example screens in the product menu. Observed in a scaffolding test run on 2026-08-07: the user asked for two screens (Login, Tasks), the running app showed a left menu with **8 items** - the four example MFE packages shipped by template-mfe (`_blank-mfe`, `demo-mfe`, `widgets-fixture-a`, `widgets-fixture-b`) self-register as screen extensions alongside the user's two, and the shell's menu renders every registered screen extension (`presentation.label/icon/route/order` from each `mfe.json`).

The scaffolding agent had to detect this mid-verification, delete all four example packages, regenerate MFE manifests, and restart the dev server to get the product the user asked for. That costs time on every run and would look broken in front of any real user: the product menu is the user's surface, examples are the template's.

## Constraint

`_blank-mfe` cannot simply be deleted from the template: the project-scaffolding skill uses it as the scaffold source it copies for every new screen. Whatever the fix, the blank scaffold must remain present in the seeded tree (or reachable by the skill) while staying out of the product menu.

## Direction (implementer's choice, options in observed-preference order)

1. Example MFE packages don't register in the menu by default: no `presentation` block (or an explicit dev/example flag the shell excludes from menu building) in the four example `mfe.json` files.
2. Or: the project-scaffolding skill removes/deregisters the example packages as an explicit step right after `frontx add`, before realization - so the cleanup is deterministic, not a mid-verification improvisation.

Option 1 fixes the product for every consumer, including people not using the skill; option 2 is skill-only. They can also be combined (1 for fixtures/demo, 2 for nothing).

## Also in scope: scaffolding-run rules confirmed by timing measurements

Two run rules proved out in measured test runs (2026-08-06/07) and belong in the kit, not in user prompts:

- **Sequential screen realization** (project-scaffolding SKILL.md): screens are built one after another in the main loop, not in parallel background subagents. Measured: two screens sequentially 7m36s vs 12m28s in parallel subagents - each subagent re-explores context from scratch, and the second sequential screen reuses the first one's patterns (2m27s vs 5m09s).
- **File search stays inside the project tree** (kit AGENTS.md rule resource): one runaway `find /` from a scaffolding subagent cost 93.6s of pure filesystem scan.

## Out of scope

- `type-check` failure TS2353 in `packages/react/src/mfe/components/__tests__/ExtensionDomainSlot.test.tsx` (fixture references `resolveLifecycleStage*Id` APIs missing from the published `@gears-frontx/mfes@0.3.0-alpha.1`): this is the publish-skew family, resolved by republishing mfes/gts-plugin and PR #535, not by editing the fixture.
- TS2724 `createMfeRegistryFactory` - same skew, same resolution.

## Acceptance

- Fresh `seed` + `add` + realization of N screens yields a product menu with exactly N items, with no cleanup step improvised by the agent mid-run.
- The blank scaffold remains available to the project-scaffolding skill.
- SKILL.md carries the sequential-realization rule; the kit rule resource carries the search-scope rule.

itechmeat:
## What changed

- Each of template-mfe's four shipped packages (`demo-mfe`, `_blank-mfe`, `widgets-fixture-a`, `widgets-fixture-b`) declares `"templateExample": true` in its `mfe.json`, and both discovery scanners - manifest generation (`discoverPackages`) and build/dev orchestration (`getMFEPackages`) - drop flagged packages. A seeded project's menu holds the developer's own screens and nothing else, and the project stops building and serving four dev servers it has no use for. `FRONTX_INCLUDE_TEMPLATE_EXAMPLES=1` (or `true`, any casing) puts the examples back for a run that means to watch them work; the skip is reported by one line in the `generate:mfe-manifests` output.
- Kit rules, in scope per the issue: a sequential unit-realization rule in the project-scaffolding SKILL.md (measured: 7m36s sequential vs 12m28s parallel background agents for two units) and a filesystem search-scope section in the kit AGENTS.md (one unscoped `find /` cost 93.6s and surfaced nothing).
- Wiring tests run both real scanners against a fixture tree with one flagged and one plain package; deleting either filter call fails exactly those tests (mutation-checked). `getMFEPackages` now takes its directory as an argument and `ManifestGenerator` is importable without side effects (CLI entry guarded on being the process entry point).
- Docs: the rule's full explanation lives in one place, `template-mfe/src-app/mfe_packages/README.md`; the other human-facing surfaces (both template READMEs, QUICK_START) carry one line plus a pointer. Agent-facing copies (SKILL, workflow, guideline) stay self-contained because each is read in isolation.

## Decisions worth naming

- A flag rather than the `_` name prefix: one mechanism covers all four packages, it is visible in the file itself, and `_` is not a reserved prefix a project may not legitimately use.
- Exclusion at discovery rather than in `Menu.tsx`: the generated manifest set has more consumers than the menu (build, dev orchestrator), and filtering at the source covers all of them at once.
- The flag stays out of `ExtensionPresentation` in `@gears-frontx/mfes`: extending the published package would reintroduce exactly the publish skew the issue rules out of scope. The optional field is declared on the plugin's local `MfeJson` shape only; both consumers read it from the source `mfe.json`, never from the enriched output.

## Not touched

- `run-mfe-type-checks.mjs` and the root `workspaces` glob deliberately keep covering flagged packages - a scaffold that stops compiling is a scaffold nobody can copy. The boundary is now stated in `mfe-package-contract.md`.
- `_blank-mfe` component sources and its `package.json` - that is issue #551's territory, PR to follow.
- The TS2724/TS2353 publish-skew family - out of scope per the issue, resolved by republication.

## Checks

Root `tsc --noEmit`, `lint`, `test:unit` (8 projects), `validate:templates` - pass. Kit `test:unit` - 13 files / 147 tests pass (includes the forbidden-solution-name scan over both edited kit files). template-shell `type-check` and `test:unit` - pass, all 7 sub-suites; the package suite carries the 4 new wiring tests. Pre-existing and unrelated: 8 `react-hooks/refs` lint errors in `template-shell/packages/react` (fixed on develop by `3598ef98`, which is not in this base), and `dist-lib/` build output is gitignored but not eslint-ignored.

## Deliberately undone

- In the all-examples case a full `dev:all` prints the orchestrator's empty-set notice and then the generator's skip line - two lines about the same fact, neither contradicting the other. Deduping them needs cross-process plumbing between the orchestrator and its spawned child, judged not worth it.
- template-shell tests still do not run in monorepo CI - a pre-existing gap for the whole `template-shell/__tests__/` directory, not widened here.

Closes #550.


itechmeat:
Closing: superseded by #565, which landed the same fix in develop on Aug 17 and closed #550. Nothing left to pick up here.

PR #553, opened 2026-08-07
feat: build _blank-mfe's screen from @gears-frontx/ui-kit instead of vendored shadcn copies

### LINKED ISSUE #551: Build scaffolded screens from @gears-frontx/ui-kit instead of vendored copies (_blank-mfe first)

itechmeat:
## What changed

- `_blank-mfe`'s screen is built from `@gears-frontx/ui-kit@0.3.0-alpha.1` (exact pin, real dependency) instead of vendored shadcn copies. Deleted: `components/ui/{button,card,skeleton}.tsx`, `lib/utils.ts`, their tests; dropped deps: `tailwindcss`, `clsx`, `tailwind-merge`, `class-variance-authority`, `@radix-ui/react-slot`.
- The kit's `theme.css` is imported `?inline` and injected into the MFE's shadow root through the `initializeStyles` hook, with `:root` rewritten to `:host` by `src/shared/anchorKitThemeOnShadowHost.ts` - a `ShadowRoot` is a `DocumentFragment`, so an unmodified `:root` block delivers nothing there. The rewrite strips comments first and anchors on `/:root\b/g`, covering minified single-line CSS and `selector,:root` lists; tests pin both shapes.
- `HomeScreen` uses no Tailwind: inside the shadow root the re-anchored kit tokens shadow the shell's HSL triplets, so Tailwind color utilities would silently resolve to `hsl(<hex>)` and drop - a scaffold copied for every new screen must not carry classes that do nothing. Styling is `HomeScreen.module.css` over kit tokens; the skeleton silhouette literals are the stated exception (the kit ships no skeleton dimensions).
- Host theme bridges to the kit's two scopes via an explicit enumeration: `dark`, `dracula`, `dracula-large` → `dark`, everything else → `light`; the screen root always carries `data-theme` so `prefers-color-scheme` cannot leak through. Verified live on all three shell themes - dracula renders dark-on-dark. Residual limitation, stated in the code: a host theme is a full palette, not a light/dark bit, so `dracula` lands in the kit's dark greys, not Dracula's purples.

## Spike findings (the issue's two risks)

- **CSS injection timing - not a blocker; the premise was wrong.** Kit component CSS never travels through `document.head`: ui-kit is deliberately absent from `sharedDeps`, so it bundles into the MFE and its CSS lands in the MFE's own `exposeAssets.css.sync`, which `MfeHandlerMF` injects into the mount container before `mount()` runs. The one-shot head adoption bridges only the shell's Tailwind - a separate path. This is also why ui-kit must stay out of `sharedDeps`: externalizing it would strand the CSS.
- **Theme token source of truth - real conflict, third option taken.** The shell consumes tokens as HSL triplets via `hsl(var(--x))`; the kit declares the same names as complete colors. Loading `theme.css` at document level would break every Tailwind color utility shell-wide, and loading it unchanged into the shadow root delivers nothing (`:root` matches no shadow tree). Hence the consumer-side `:root`→`:host` re-anchor. The right long-term fix is in the kit itself - `theme.css` declaring `:root, :host { ... }` would delete the rewriter module, its test and the lifecycle override - but that needs another kit alpha publish and a re-pin, so it stays a named follow-up.

## Known collision, measured harmless

`ThemeAwareReactLifecycle.injectBaseResets` paints `:host` from `hsl(var(--background))`; the re-anchored hex tokens make that expression invalid, so the host computes transparent under all themes. Nothing is visibly unpainted: the screen root is the shadow root's only rendered child, sits at the host origin and covers it in both axes, and the kit's `[data-theme]` rule paints it. The `initializeStyles` docblock records the condition under which this stops being harmless.

## Not touched

- `mfe.json`, registration/discovery, kit rule files - PR #552's territory; the branches touch disjoint files.
- `demo-mfe`'s showcase and the shell chrome - explicitly out of scope per the issue.
- The older, unrelated `--uikit none` template flag and its vendored copies - the issue's naming caution is respected.
- `template-mfe`'s harness `overrides` do not redirect ui-kit to local source - deliberate: the scaffold resolves exactly what a seeded project resolves; both harness doc blocks now name the exception.

## Checks

`_blank-mfe`: `tsc --noEmit`, vitest (8 files / 24 tests), `vite build` - pass. Root: `lint`, `type-check`, `policy:template-pin-drift`, `validate:templates` - pass. Browser (shell dev + MFE preview): light, dark and dracula - zero console errors or warnings. Pre-existing, unrelated: one `react-hooks/set-state-in-effect` lint error in `HomeScreen.tsx`, reproduced identically on the base revision.

## Environment issues found on the way (pre-existing, not addressed here)

- `template-mfe`'s lockfile cannot regenerate: it pins `@gears-frontx/frontx-template-shell@0.1.0-alpha.1`, which is unpublished (registry has only `alpha.0`), and `npm install` with the committed lock silently drops all nine `file:`-linked `@gears-frontx` entries. The lock was extended by splicing npm's own generated entries; `npm ci` and `npm ci --omit=dev` both verified green. Worth its own issue.
- `template-shell`'s `dev:all` references `concurrently`, which no package declares.

Closes #551.


itechmeat:
Closing: superseded by #584, which moved _blank-mfe (and demo-mfe) onto @gears-frontx/ui-kit with the same shadow-host theme bridge. Still worth picking up: issue #551 is formally open because auto-close does not fire for PRs into develop.

PR #555, opened 2026-08-10
feat(ui-kit): select dropdown placement, button colours, input slots, badge dot/icon

SysoevAndrey:
Four requested API changes to the kit, plus two follow-ups that came out of reviewing the result. Backward compatibility was explicitly not a goal — the package is in early alpha.

## Select — the list always sits below the input

`SelectContent` no longer overlays the trigger the way a native `<select>` does; the popup opens on its `side` (below by default). With nothing selected the list starts at the top, and a selected option beyond the fold is scrolled into view when the popup opens — Base UI's own `scrollIntoView` on open does this, so the kit carries no scroll code of its own. `alignItemWithTrigger` is gone from the props.

The dropdown's inner padding moved from `SelectGroup` to the list itself, so a select whose items sit directly under `SelectContent` is padded identically to a grouped one — previously a groupless list had its options flush against the popup edges, and the fix means `SelectGroup` is now purely about labelled sections. `SelectSeparator` took over the full gap between sections so the inter-group rhythm survived the move.

## Button — arbitrary colours from a consumer className

Variant colours now resolve through `var(--button-bg, <variant token>)` and friends, so a consumer class can recolour a button without a new variant:

```css
.buy { --button-bg: var(--success); --button-bg-hover: color-mix(in oklab, var(--success) 90%, var(--foreground) 10%); }
```

Five hooks: `--button-bg`, `--button-fg`, `--button-border`, `--button-bg-hover`, `--button-fg-hover`. Setting only `--button-bg` keeps the button that colour in every state — hover falls back to your rest colour rather than snapping back to the variant's token. Before this, a rest-state override was a stylesheet-order race and hover always lost.

## Input — `icon` and `end` slots

`icon` renders a leading decorative icon (aria-hidden, no pointer events); `end` renders trailing interactive content — a clear button, a reveal toggle — after the input in the DOM, so tab order is field then slot. With neither prop the component is still a bare `<input>` with no wrapper.

## Badge — opt-in dot, new icon slot

The status dot is now the `dot` prop instead of an unconditional `::before`, and a new `icon` render prop takes its place when both are given. Default is neither. The `shape` axis frees the name: `pill | plain`.

## Focus ring, reworked

Button's focus ring used to be two-toned — an outer border in `--info` plus an inner inset shadow — because it was drawn on the button's edge, where a same-hue ring cannot clear WCAG 1.4.11's 3:1 against the button's own fill. The result was a blue ring in light mode and a cyan one in dark, next to a violet button.

The ring now sits outside the element (`outline` + `outline-offset: 2px`), so it only ever borders the page background. That allows a single tone per variant in the button's own colour family: `--primary-ring` for default (6.79:1 light / 7.23:1 dark against the page), `--destructive-ring` for destructive (6.01 / 7.89), and the shared `--ring` for the rest (4.05 / 3.78). The previous outer tone measured 3.91:1 in light mode, so this is more headroom, not less. `--button-focus-ring-inner` is retired.

One consequence worth knowing: the ring occupies 4px beyond the button's box, so in a tightly packed group with no gaps it sits closer to neighbours than the old on-edge ring did, and an `overflow: hidden` ancestor can clip it.

## Breaking changes

| Change | Migration |
|---|---|
| `SelectContent` dropped `alignItemWithTrigger` | Remove the prop; the behaviour it selected is now the only behaviour |
| `type="search"` no longer auto-renders a magnifier | Pass one via `icon` |
| Badge renders no dot by default | Add `dot` where the marker is wanted |
| `shape="dot"` → `shape="plain"` | Rename |
| `--button-focus-ring-inner` retired | Set `--button-focus-ring` alone; it is a single tone now |
| Badge no longer sizes a bare `<svg>` child | Move the icon into the `icon` prop |

`llms.txt` and every touched component's `.md` were updated in step, since insight-front's migration reads them.

## Verification

284 tests pass (21 files), including new coverage for the derived `data-dot` precedence, the slot DOM/tab order, a disabled field's `end` slot going inert, dropdown padding belonging to the list rather than the group, the popup's resolved side, and each variant's focus-ring token identity. `type-check`, `type-check:test`, `type-check:demo`, `eslint --max-warnings 0`, `test:consumer` (vite / esbuild barrel + subpath / webpack / types) and `cfs validate` (54 artifacts, 0 errors) are all clean.

Checked in a browser as well: popup below the trigger with the pre-selected item already visible, identical 4px padding across grouped and groupless dropdowns, a `--button-bg` override holding through hover, the search field's slots with correct tab order, and the badge dot/icon states. The focus ring was verified per variant in both themes against its expected token.

Version moves to `0.3.0-alpha.2` — the next alpha of the in-flight 0.3.0 line, which has never shipped, so the breaks ride the alpha counter rather than burning a minor (per Gera's review).

## Summary by CodeRabbit

- **New Features**
  - Badges support optional status dots or icons, with a new plain shape.
  - Inputs support explicit leading icons and trailing content, including clear actions; search icons are no longer added automatically.
  - Select menus provide improved grouping, spacing, scrolling, and positioning.
  - Buttons support customizable colors and improved focus-ring styling.

- **Documentation**
  - Updated component guidance and examples for Badge, Button, Input, and Select.

- **Chores**
  - Updated the UI kit release version.

### DIFF
--- a/packages/ui-kit/demo/components-page.tsx
+++ b/packages/ui-kit/demo/components-page.tsx
@@ -34,6 +34,7 @@ import {
   RadioGroupItem,
   Select,
   SelectContent,
+  SelectGroup,
   SelectItem,
   SelectTrigger,
   SelectValue,
@@ -63,6 +64,11 @@ const REGIONS = [
   { value: 'eu-west', label: 'Dublin' },
   { value: 'us-east', label: 'Virginia' },
 ];
+
+const HOURS = Array.from({ length: 24 }, (_, hour) => ({
+  value: String(hour),
+  label: `${String(hour).padStart(2, '0')}:00`,
+}));
 
 function LoadingDemo() {
   const [busy, setBusy] = useState(false);
@@ -185,6 +191,20 @@ export function ComponentsPage() {
                     {item.label}
                   </SelectItem>
                 ))}
+              </SelectContent>
+            </Select>
+            <Select defaultValue="18" items={HOURS}>
+              <SelectTrigger aria-label="Hour">
+                <SelectValue placeholder="Hour" />
+              </SelectTrigger>
+              <SelectContent>
+                <SelectGroup>
+                  {HOURS.map((item) => (
+                    <SelectItem key={item.value} value={item.value}>
+                      {item.label}
+                    </SelectItem>
+                  ))}
+                </SelectGroup>
               </SelectContent>
             </Select>
           </Row>
--- a/packages/ui-kit/demo/components-page.tsx
+++ b/packages/ui-kit/demo/components-page.tsx
@@ -58,7 +58,7 @@ import {
   TooltipTrigger,
 } from '@gears-frontx/ui-kit';
 
-import { DemoIcon, Row, Section } from './shared';
+import { CloseIcon, DemoIcon, Row, Section } from './shared';
 
 const REGIONS = [
   { value: 'eu-central', label: 'Frankfurt' },
@@ -187,7 +187,7 @@ export function ComponentsPage() {
               onChange={(e) => setQuery(e.currentTarget.value)}
               placeholder="Search projects…"
               icon={<MagnifierIcon />}
-              end={<Button variant="ghost" size="sm" icon={<DemoIcon />} aria-label="Clear search" onClick={() => setQuery('')} />}
+              end={<Button variant="ghost" size="sm" icon={<CloseIcon />} aria-label="Clear search" onClick={() => setQuery('')} />}
             />
           </Field>
           <Field name="notes">
--- a/packages/ui-kit/demo/components-page.tsx
+++ b/packages/ui-kit/demo/components-page.tsx
@@ -140,19 +140,27 @@ export function ComponentsPage() {
           <Badge>muted</Badge>
         </Row>
         <Row>
-          <Badge variant="success" shape="dot">
+          <Badge variant="success" shape="plain" dot>
             success
           </Badge>
-          <Badge variant="warning" shape="dot">
+          <Badge variant="warning" shape="plain" dot>
             warning
           </Badge>
-          <Badge variant="info" shape="dot">
+          <Badge variant="info" shape="plain" dot>
             info
           </Badge>
-          <Badge variant="danger" shape="dot">
+          <Badge variant="danger" shape="plain" dot>
             danger
           </Badge>
-          <Badge shape="dot">muted</Badge>
+          <Badge shape="plain" dot>muted</Badge>
+        </Row>
+        <Row>
+          <Badge variant="success" dot>
+            with dot
+          </Badge>
+          <Badge variant="info" icon={<DemoIcon />}>
+            with icon
+          </Badge>
         </Row>
       </Section>
 
@@ -323,7 +331,7 @@ export function ComponentsPage() {
             <TableRow data-state="restricted">
               <TableCell>gears-vault (restricted)</TableCell>
               <TableCell>
-                <Badge shape="dot">no access</Badge>
+                <Badge shape="plain" dot>no access</Badge>
               </TableCell>
               <TableCell>—</TableCell>
             </TableRow>
--- a/packages/ui-kit/demo/components-page.tsx
+++ b/packages/ui-kit/demo/components-page.tsx
@@ -70,6 +70,15 @@ const HOURS = Array.from({ length: 24 }, (_, hour) => ({
   label: `${String(hour).padStart(2, '0')}:00`,
 }));
 
+function MagnifierIcon() {
+  return (
+    <svg viewBox="0 0 16 16" fill="none" aria-hidden="true">
+      <circle cx="7" cy="7" r="4.5" stroke="currentColor" strokeWidth="1.5" />
+      <path d="m10.5 10.5 3 3" stroke="currentColor" strokeWidth="1.5" strokeLinecap="round" />
+    </svg>
+  );
+}
+
 function LoadingDemo() {
   const [busy, setBusy] = useState(false);
   return (
@@ -162,7 +171,12 @@ export function ComponentsPage() {
           </Field>
           <Field name="query">
             <FieldLabel>Search</FieldLabel>
-            <Input type="search" placeholder="Search projects…" />
+            <Input
+              type="search"
+              placeholder="Search projects…"
+              icon={<MagnifierIcon />}
+              end={<Button variant="ghost" size="sm" icon={<DemoIcon />} aria-label="Clear search" />}
+            />
           </Field>
           <Field name="notes">
             <FieldLabel>Notes</FieldLabel>
--- a/packages/ui-kit/demo/components-page.tsx
+++ b/packages/ui-kit/demo/components-page.tsx
@@ -96,6 +96,7 @@ function LoadingDemo() {
 }
 
 export function ComponentsPage() {
+  const [query, setQuery] = useState('');
   const [region, setRegion] = useState<string | null>(null);
   return (
     <>
@@ -181,9 +182,11 @@ export function ComponentsPage() {
             <FieldLabel>Search</FieldLabel>
             <Input
               type="search"
+              value={query}
+              onChange={(e) => setQuery(e.currentTarget.value)}
               placeholder="Search projects…"
               icon={<MagnifierIcon />}
-              end={<Button variant="ghost" size="sm" icon={<DemoIcon />} aria-label="Clear search" />}
+              end={<Button variant="ghost" size="sm" icon={<DemoIcon />} aria-label="Clear search" onClick={() => setQuery('')} />}
             />
           </Field>
           <Field name="notes">
--- a/packages/ui-kit/demo/components-page.tsx
+++ b/packages/ui-kit/demo/components-page.tsx
@@ -36,6 +36,7 @@ import {
   SelectContent,
   SelectGroup,
   SelectItem,
+  SelectLabel,
   SelectTrigger,
   SelectValue,
   Skeleton,
@@ -224,6 +225,7 @@ export function ComponentsPage() {
               </SelectTrigger>
               <SelectContent>
                 <SelectGroup>
+                  <SelectLabel>Hours</SelectLabel>
                   {HOURS.map((item) => (
                     <SelectItem key={item.value} value={item.value}>
                       {item.label}
--- a/packages/ui-kit/design-notes.md
+++ b/packages/ui-kit/design-notes.md
@@ -259,27 +259,34 @@ Architecture's build bullet).
    instruction that a color failing WCAG gets a new color, not a "kept as
    drawn" footnote:
    - **Button focus rings** (every variant, both themes) now clear the
-     3:1 floor against both the fill and the page background — `default`
-     and `destructive` needed a genuinely two-toned ring (the geometry
-     already supported one: an outer border color plus a separately
-     colorable inset shadow), `outline` gave up on `--border-strong`
-     entirely and now falls back to the kit-wide `--ring`. Ratios as
-     measured (light/dark), with button.test.tsx recomputing them from
-     the live CSS on every run:
+     3:1 floor against the page background. Originally shipped two-toned
+     (an outer border color plus a separately colorable inset shadow,
+     because the ring was drawn ON the button's edge and so had two WCAG
+     neighbors — the page outside, the button's own fill inside) — revised
+     after ship because `--info` (the `default` variant's outer tone) is
+     cyan in dark mode and blue in light, a hue that appears nowhere else
+     near the violet button, and because two custom properties per variant
+     was more machinery than the fix needed. The ring now sits OUTSIDE the
+     button instead (`outline` + `outline-offset`, not a border recolor
+     plus inset shadow), so its only WCAG neighbor is ever the page
+     background and a single tone per variant is enough: `default` and
+     `destructive` get their own family color (new tokens
+     `--primary-ring`/`--destructive-ring`, since violet/red read better
+     against the page than the plain kit-wide ring does), everything else
+     (`outline`, `secondary`, `ghost`, `link`) falls through to the
+     kit-wide `--ring`. `outline` was already on `--ring`, unchanged by
+     this revision. Ratios as measured (light/dark), with button.test.tsx
+     recomputing them from the live CSS on every run:
 
-     | variant | outer tone vs page bg | inner tone vs fill |
-     |---|---|---|
-     | `default` | `--info` 3.91 / 10.89 | `--foreground` 3.79 / 4.86 |
-     | `destructive` | `--primary-hover` 4.98 / 4.18 | `--destructive-foreground` 4.70 / 4.62 |
-     | `secondary` | `--primary-hover` 4.98 / 4.18 | same tone vs `--secondary` 4.76 / 3.10 |
-     | `ghost`, `link` | `--primary-hover` 4.98 / 4.18 | no fill of their own — the page bg again |
-     | `outline` | `--ring` 4.05 / 3.78 | `--ring` vs `--surface-elevated` 4.23 / 3.40 |
+     | variant | ring tone vs page bg |
+     |---|---|
+     | `default` | `--primary-ring` 6.79 / 7.23 |
+     | `destructive` | `--destructive-ring` 6.01 / 7.89 |
+     | `outline`, `secondary`, `ghost`, `link` | `--ring` 4.05 / 3.78 |
 
-     `secondary`, `ghost` and `link` are single-tone: `--primary-hover`
-     alone clears 3:1 on both of their surfaces, so they set no inner
-     override. 3.10 (dark `secondary`, the ring against its own fill) is
-     the tightest number in the table and the one to watch if `--secondary`
-     or `--primary-hover` ever moves.
+     `outline`/`secondary`/`ghost`/`link` share one number because they all
+     resolve to the same `--ring` token; 3.78 (dark) is the tightest in the
+     table and the one to watch if `--ring` ever moves.
    - **`--subtle-foreground`** (the table header label): each mode's drawn
      value failed the 4.5:1 AA floor against the header's own `--surface`
      fill — light `#94a3b8` at 2.56:1, dark `#667085` at 3.74:1. Both
--- a/packages/ui-kit/design-notes.md
+++ b/packages/ui-kit/design-notes.md
@@ -365,7 +365,7 @@ Architecture's build bullet).
      the kit is driven by `variant` (+ `size`, or the occasional real extra
      axis: Badge's `shape`, Table's `density`). `shape` rather than `size`
      because the values are pill vs. plain (dot is a separate opt-in flag,
-   orthogonal to shape) and Badge has no size axis;
+     orthogonal to shape) and Badge has no size axis;
      not `form`, which is a real HTML attribute a styling prop would shadow
      for anyone rendering a form-associated element via `render`.
    - Four new tokens: `--link-foreground` (Button's `link` variant text),
--- a/packages/ui-kit/design-notes.md
+++ b/packages/ui-kit/design-notes.md
@@ -223,11 +223,14 @@ Architecture's build bullet).
    "Studio / shadcn" variable collection. Landed: theme.css carries the
    palette and the new token groups (see Architecture's theme bullet), and
    the existing components follow the mockups' component specs — Badge on
-   semantic intents (pill/dot; the shadcn variant list retired), Button's
+   semantic intents (`pill`/`plain` shapes, with `dot` and `icon` both
+   opt-in; the shadcn variant list retired), Button's
    `icon` slot + auto icon-only + `loading`, Tabs on the trackless Kind=tab
    look (the spec's Kind=segment is the planned `toggle-group`'s styling,
    per the design file's own component description), the unified Field set
-   for Input/Textarea/Select with the `search` and `filter` types, Table
+   for Input/Textarea/Select with the `filter` type (Select's compact
+   toolbar chip) and `search`'s native searchbox role (no automatic icon;
+   pass one via `icon`), Table
    density + selected/stale/restricted row hooks, and component CSS on the
    metric tokens. The drawn-vs-spec control-height discrepancy was ruled
    in favor of the drawn specimens (32/36/40; buttons map sm/default/lg
@@ -361,7 +364,8 @@ Architecture's build bullet).
      than by an axis name only this component used, so every component in
      the kit is driven by `variant` (+ `size`, or the occasional real extra
      axis: Badge's `shape`, Table's `density`). `shape` rather than `size`
-     because the values are pill vs. bare dot and Badge has no size axis;
+     because the values are pill vs. plain (dot is a separate opt-in flag,
+   orthogonal to shape) and Badge has no size axis;
      not `form`, which is a real HTML attribute a styling prop would shadow
      for anyone rendering a form-associated element via `render`.
    - Four new tokens: `--link-foreground` (Button's `link` variant text),
--- a/packages/ui-kit/llms.txt
+++ b/packages/ui-kit/llms.txt
@@ -56,7 +56,7 @@ Rules for generated code:
 ## Components
 
 - [Badge](dist/docs/badge.md): status label, an intent-colored caption with
-  an opt-in status dot (`dot`) or a leading `icon` render prop (icon wins
+  an opt-in status dot (`dot`) or a leading `icon` slot prop (icon wins
   when both are given; default is neither); variants
   success|warning|info|danger|muted (semantic states ONLY — pick by meaning,
   never by color; neutral/count/category = muted), shapes pill|plain (pill
--- a/packages/ui-kit/llms.txt
+++ b/packages/ui-kit/llms.txt
@@ -55,19 +55,23 @@ Rules for generated code:
 
 ## Components
 
-- [Badge](dist/docs/badge.md): status label, a colored dot + matching
-  caption; variants success|warning|info|danger|muted (semantic states ONLY
-  — pick by meaning, never by color; neutral/count/category = muted), shapes
-  pill|dot (pill adds a soft fill; dot is the bare pair for dense/inline
-  spots); no Base UI primitive; `render` (e.g. `<a>`) makes it a clickable
-  tag — hover feedback only appears once it is.
+- [Badge](dist/docs/badge.md): status label, an intent-colored caption with
+  an opt-in status dot (`dot`) or a leading `icon` render prop (icon wins
+  when both are given; default is neither); variants
+  success|warning|info|danger|muted (semantic states ONLY — pick by meaning,
+  never by color; neutral/count/category = muted), shapes pill|plain (pill
+  adds a soft fill; plain is the bare label for dense/inline spots); no
+  Base UI primitive; `render` (e.g. `<a>`) makes it a clickable tag — hover
+  feedback only appears once it is.
 - [Button](dist/docs/button.md): action button; variants
-  default|destructive|outline|secondary|ghost|link, sizes default|sm|lg;
-  icons go in the `icon` slot prop, NEVER in children — `icon` with no
-  children makes a square icon-only button (always pass `aria-label`
-  then); `loading` spins, disables, and keeps the width and accessible
-  name; render prop for button-semantic actions over a real anchor (not
-  a link substitute).
+  default|destructive|outline|secondary|ghost|link, sizes default|sm|lg; a
+  consumer className can recolor a button in every state via
+  `--button-bg`/`--button-fg`/`--button-bg-hover`/`--button-fg-hover`/
+  `--button-border` — no new variant needed; icons go in the `icon` slot
+  prop, NEVER in children — `icon` with no children makes a square
+  icon-only button (always pass `aria-label` then); `loading` spins,
+  disables, and keeps the width and accessible name; render prop for
+  button-semantic actions over a real anchor (not a link substitute).
 - [Card](dist/docs/card.md): bounded content surface (Card/CardHeader/
   CardTitle/CardDescription/CardAction/CardContent/CardFooter); header
   becomes a two-column/two-row grid automatically when it contains a
@@ -91,9 +95,12 @@ Rules for generated code:
   aria-describedby, and validation display automatically — the default
   way to build forms.
 - [Input](dist/docs/input.md): single-line text field; `onValueChange`
-  for the string value, `aria-invalid` for the error state;
-  `type="search"` renders the leading magnifier and announces as a
-  searchbox — no extra prop; wires into Field automatically.
+  for the string value, `aria-invalid` for the error state; `icon` for a
+  leading decorative icon, `end` for trailing interactive content (e.g. a
+  clear button; rendered after the input, so tab order is field → slot) —
+  neither renders automatically; `type="search"` gives the searchbox role
+  but no icon by itself — pass one via `icon`; wires into Field
+  automatically.
 - [Label](dist/docs/label.md): caption for a form control; native
   `<label>` props; associate via `htmlFor` or by nesting the control.
 - [RadioGroup](dist/docs/radio-group.md): mutually exclusive options
@@ -102,8 +109,10 @@ Rules for generated code:
 - [Select](dist/docs/select.md): dropdown picking one value from a
   list (Select/SelectTrigger/SelectValue/SelectContent/SelectGroup/
   SelectItem); trigger sizes default|sm, trigger variants default|filter
-  (filter = compact toolbar filter chip, label stays muted); keyboard
-  and typeahead built in.
+  (filter = compact toolbar filter chip, label stays muted); the popup
+  always opens below the trigger (never overlays it), scrolling a selected
+  item into view when it sits beyond the fold; keyboard and typeahead
+  built in.
 - [Separator](dist/docs/separator.md): thin visual divider between content
   sections; orientation horizontal|vertical (default horizontal); no
   `decorative` prop — pass `aria-hidden="true"` for a purely visual
--- a/packages/ui-kit/package.json
+++ b/packages/ui-kit/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@gears-frontx/ui-kit",
-  "version": "0.3.0-alpha.1",
+  "version": "0.3.0-alpha.2",
   "description": "Standard React component base for FrontX templates - Base UI behavior, CSS Modules styling on semantic tokens, CVA variants",
   "type": "module",
   "types": "./dist/index.d.ts",
--- a/packages/ui-kit/src/components/badge/badge.module.css
+++ b/packages/ui-kit/src/components/badge/badge.module.css
@@ -1,7 +1,8 @@
 /*
  * Badge styles — the F-mockups' Badge/Status design (Figma frame 196:397):
- * a colored status dot plus an intent-colored label, as a soft pill or a
- * bare dot+label. Replaces the shadcn-translated variant set; the focus
+ * an intent-colored label, as a soft pill or a bare label, with an
+ * opt-in status dot (`dot` prop) or leading icon (`icon` slot) in front of
+ * it. Replaces the shadcn-translated variant set; the focus
  * ring/forced-colors/aria-invalid plumbing carries over from that
  * translation. Consumes only theme.css variables — no raw colors.
  *
--- a/packages/ui-kit/src/components/badge/badge.module.css
+++ b/packages/ui-kit/src/components/badge/badge.module.css
@@ -59,13 +59,14 @@
 }
 
 /*
- * The status dot, drawn in both forms (the pill is dot+label on a fill, the
- * dot form is the same pair bare). Pure decoration duplicating the label's
- * intent color, which is why the dot itself is not held to the 3:1 non-text
+ * The status dot, drawn in both shapes when opted in via the `dot` prop
+ * (badge.tsx derives data-dot from dot/icon: the icon slot, when present,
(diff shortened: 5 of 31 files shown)


packages/ui-kit/src/components/badge/badge.tsx
```diff
@@ -45,12 +47,46 @@ const badgeVariants = cva(styles.badge, {
...
+      children: (
+        <>
+          {icon != null && (
+            <span className={styles.icon} aria-hidden="true">
+              {icon}
+            </span>
+          )}
+          {children}
+        </>
+      ),
```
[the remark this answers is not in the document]
SysoevAndrey: Valid, and wider than Badge — fixed in caa3f6d7.

`icon != null` was wrong in three places, not one: Badge suppressed the dot and rendered an empty icon span, Input added its wrapper plus slot padding for nothing (both `icon` and `end`), and Button was worst — `icon={false}` with no children produced an empty square icon-only button.

All three now use the predicate `button.tsx` already had for its label: `Children.toArray(slot).some((child) => child !== "")`, which drops `false`/`null`/`undefined` and the empty string while keeping `0`. Coverage added per component, including the `<Badge dot icon={false}>` case you asked for.

packages/ui-kit/src/components/button/button.md
```diff
@@ -36,6 +36,34 @@ polymorphism, correct disabled/focus behavior, `type="button"` by default
...
+## Custom colors
+
+Variant colors are consumed through CSS custom properties, so a one-off
+brand/status button is a consumer class away — no new variant needed:
+
+| Property | Drives |
+|----------|--------|
+| `--button-bg` / `--button-fg` | fill / text at rest |
+| `--button-bg-hover` / `--button-fg-hover` | fill / text on hover |
+| `--button-border` | border color (visible on `outline`; transparent elsewhere) |
```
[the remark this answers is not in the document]
SysoevAndrey: Correct — the hook lives on the shared `.button` rule, so setting it shows a border on every variant; `outline` only differs in its fallback. Reworded in caa3f6d7 to "border color (transparent by default outside `outline`)".

packages/ui-kit/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@gears-frontx/ui-kit",
-  "version": "0.3.0-alpha.1",
+  "version": "0.4.0-alpha.0",
```
GeraBart: let's keep minor versions aligned for now? I'd suggest `0.3.0-alpha.2`

SysoevAndrey: Agreed, changed to `0.3.0-alpha.2` in 9a0fab17

packages/ui-kit/src/components/badge/badge.tsx
```diff
@@ -18,11 +19,12 @@ import styles from './badge.module.css';
  * itself keeps the kit-wide `variant` name rather than an axis name of its
  * own, so every component in the kit is driven the same way.
  *
- * `shape`, not `size`: the two values are pill vs. bare dot, and Badge has
- * no size axis at all (the pill is a fixed 24px). Not `form` either — that
- * is a real HTML attribute, so a styling prop by that name would shadow it
- * for anyone rendering a form-associated element via `render`. Both shapes
- * carry the colored dot; `pill` adds the fill behind it.
+ * `shape`, not `size`: the two values are pill vs. plain (bare text, no
```
itechmeat, line 22: 🟠 design-notes.md still describes the retired API in three places: line 364 says the shape values are "pill vs. bare dot", line 226 lists "(pill/dot; ...)", and line 230 still credits Input with a `search` type that now does nothing beyond the native attribute. Since the PR's migration story leans on the docs being swept, these lines are worth catching in the same pass.

packages/ui-kit/src/components/select/select.tsx
```diff
@@ -125,19 +123,24 @@ export function SelectContent({
...
+        /*
+         * Never the Base UI default overlay mode (selected item aligned over
+         * the trigger): the popup always opens on `side`, below by default.
+         * With nothing selected the list starts at the top; a selected item
+         * beyond the fold is scrolled into view natively on open — Floating
+         * UI's useListNavigation calls scrollIntoView({block: 'nearest'})
+         * when a selectedIndex exists (verified against @base-ui/react 1.6.0
+         * sources), so no scroll code of ours is needed.
+         */
+        alignItemWithTrigger={false}
```
itechmeat, line 135: 🟡 [Not a blocker] The popup's scroll arrows are absolutely positioned inside the scroll container with an opaque background, and Base UI's scrollIntoView with block:'nearest' puts the selected item flush at the scrollport edge - so a pre-selected item deep in the list can land exactly under the down arrow when the popup opens. Did the 24-hour demo select look right with, say, hour 18 selected? If it does overlap, scroll-padding on the popup would give the item room.

packages/ui-kit/demo/components-page.tsx
```diff
@@ -156,7 +180,14 @@ export function ComponentsPage() {
...
           <Field name="query">
             <FieldLabel>Search</FieldLabel>
-            <Input type="search" placeholder="Search projects…" />
+            <Input
+              type="search"
+              value={query}
+              onChange={(e) => setQuery(e.currentTarget.value)}
+              placeholder="Search projects…"
+              icon={<MagnifierIcon />}
+              end={<Button variant="ghost" size="sm" icon={<DemoIcon />} aria-label="Clear search" onClick={() => setQuery('')} />}
```
itechmeat: 🟡 [Not a blocker] The focus ring now extends 4px outside the element, so the 32px clear button in the search demo focuses as a 40px ring - the same height as the input it sits inside, landing the ring right on the field's own border. Since this is the composition input.md recommends for `end`, did it look right focused in the browser?

packages/ui-kit/src/components/input/input.tsx
```diff
@@ -1,40 +1,67 @@
...
+        styles.input,
+        hasIcon && styles.hasIcon,
+        hasEnd && styles.hasEnd,
+        className,
+      )}
       {...props}
     />
   );
-  if (type !== 'search') {
+  if (!hasIcon && !hasEnd) {
```
itechmeat, line 54: 🟡 [Not a blocker] `type="search"` losing its magnifier and Badge losing the default dot are silent breaks - no type error, nothing at runtime, the visuals just change. Inside the monorepo nothing imports the package, so the whole blast radius is insight-front on alpha.1. Do we know its current call sites for these two, or is the plan that its migration picks them up from the breaking-change table?

packages/ui-kit/src/components/button/button.module.css
```diff
@@ -65,20 +69,21 @@
...
+ * on every focus/blur) no longer applies. The per-variant color arrives via
+ * --button-focus-ring (set on .button above, overridden by the variants
+ * below).
  */
 .button:focus-visible {
-  outline: none;
-  border-color: var(--button-focus-ring);
-  box-shadow: inset 0 0 0 var(--ring-inset)
-    var(--button-focus-ring-inner);
+  outline: var(--border-width-focus) solid var(--button-focus-ring);
```
itechmeat, line 85: 🟡 [Not a blocker] The forced-colors block below still explains itself with "forced-colors mode drops box-shadow, so the ring above leaves the button with no visible focus at all" - there is no box-shadow ring anymore. The color swap to Highlight still does real work against a custom ring color, but the comment's stated reason is gone and the outline/offset lines now duplicate the base rule - could the comment say what the block actually does now?

packages/ui-kit/demo/components-page.tsx
```diff
@@ -187,6 +218,20 @@ export function ComponentsPage() {
                 ))}
               </SelectContent>
             </Select>
+            <Select defaultValue="18" items={HOURS}>
+              <SelectTrigger aria-label="Hour">
+                <SelectValue placeholder="Hour" />
+              </SelectTrigger>
+              <SelectContent>
+                <SelectGroup>
```
itechmeat, line 227: 🟡 [Not a blocker] select.md now says SelectGroup is purely for labelled sections, but the demo's 24-hour select wraps its items in a bare SelectGroup with no SelectLabel - exactly the case the doc says the group is not for. Either a SelectLabel or dropping the group would make the demo model the documented usage; the page's other two selects already show the groupless form.

packages/ui-kit/llms.txt
```diff
@@ -55,19 +55,23 @@ Rules for generated code:
...
 ## Components
 
-- [Badge](dist/docs/badge.md): status label, a colored dot + matching
-  caption; variants success|warning|info|danger|muted (semantic states ONLY
-  — pick by meaning, never by color; neutral/count/category = muted), shapes
-  pill|dot (pill adds a soft fill; dot is the bare pair for dense/inline
-  spots); no Base UI primitive; `render` (e.g. `<a>`) makes it a clickable
-  tag — hover feedback only appears once it is.
+- [Badge](dist/docs/badge.md): status label, an intent-colored caption with
+  an opt-in status dot (`dot`) or a leading `icon` render prop (icon wins
```
itechmeat: 🔵 [Not a blocker] llms.txt calls Badge's `icon` a "render prop", but `render` is a different real prop of the same component five lines down, and the Button entry in the same file correctly says "slot prop". Since this file exists for machine consumption, the one-word fix avoids generated code like `<Badge render={<Icon/>}>`.

packages/ui-kit/src/components/input/input.module.css
```diff
@@ -112,29 +112,48 @@
...
 }
 
-.searchInput {
-  /* Icon inset + the icon + a gap, so the text clears the magnifier. */
+.icon svg {
+  width: var(--icon-size-sm);
+  height: var(--icon-size-sm);
+}
+
+.end {
```
itechmeat, line 143: 🔵 [Not a blocker] A disabled input dims only itself - `.input:disabled` opacity does not reach the `end` slot, which is a sibling inside the wrapper, so a clear button in a disabled field stays fully opaque and clickable. If leaving that to the consumer is the intent, a line in input.md would make it deliberate.

packages/ui-kit/src/components/input/input.tsx
```diff
@@ -1,40 +1,67 @@
...
+ * input either way, so styling targets the same element regardless. Field
+ * wiring is unaffected — the primitive is still the control Field talks to.
+ */
+export function Input({ className, icon, end, ...props }: InputProps) {
+  // `icon != null`/`end != null` are true for `icon={false}`/`end={false}` —
+  // a valid ReactNode that renders nothing — which is exactly what
+  // `icon={cond && <Icon/>}` passes when `cond` is false. Same predicate as
+  // Button's `hasLabel`, applied to a slot instead of children: it answers
+  // "is this renderable" rather than "is this set".
+  const hasIcon = Children.toArray(icon).some((child) => child !== '');
```
itechmeat, line 40: 🔵 [Not a blocker] The renderable-slot predicate `Children.toArray(x).some((child) => child !== '')` now has three copies (button, badge, input) each carrying its own restatement of the same rationale, plus three more retellings in the tests. It has become a kit-wide contract with no single home - a shared helper would give it one before a fourth slot copies it again. Fine as a follow-up.

packages/ui-kit/src/components/select/select.tsx
itechmeat, line 135: 🔵 [Not a blocker] A note from today's standup: the dropdown's arrow was called out as looking off and needing a rework. Is that meant to be part of this PR's select changes or tracked separately? Just making sure the ask does not fall between the two.

packages/ui-kit/src/components/input/input.test.tsx
```diff
@@ -8,25 +8,75 @@ afterEach(cleanup);
...
     expect(input.className).toContain(styles.input);
-    // Every non-search type stays a bare input — no wrapper element.
-    expect(input.parentElement?.className ?? '').not.toContain(styles.searchWrap);
+    // No slots — no wrapper element: the input is the root.
+    expect(input.parentElement).toBe(container);
   });
 
-  it('type="search" adds the decorated wrapper and keeps the searchbox role', () => {
-    render(<Input type="search" placeholder="Find" className="consumer" />);
+  it('type="search" no longer decorates itself', () => {
```
itechmeat: ⚪ [Not a blocker] Two new titles describe the diff rather than the behaviour: `type="search" no longer decorates itself` and select's `renders the popup without the align-with-trigger overlay mode` only mean something next to this PR. Present-tense behaviour sentences (`type="search" stays a bare input with the searchbox role and no icon`) keep them readable a year out.

packages/ui-kit/src/components/button/button.md
```diff
@@ -36,6 +36,34 @@ polymorphism, correct disabled/focus behavior, `type="button"` by default
...
+brand/status button is a consumer class away — no new variant needed:
+
+| Property | Drives |
+|----------|--------|
+| `--button-bg` / `--button-fg` | fill / text at rest |
+| `--button-bg-hover` / `--button-fg-hover` | fill / text on hover |
+| `--button-border` | border color on every variant (transparent by default outside `outline`) |
+
+```css
+.buy { --button-bg: var(--success); --button-bg-hover: color-mix(in oklab, var(--success) 90%, var(--foreground) 10%); --button-fg: #fff; }
```
itechmeat: ⚪ [Not a blocker] The custom-colours example sets `--button-fg: #fff` next to three token values, in a kit whose own modules ban raw colors - `var(--primary-foreground)` already means exactly this. Consumer stylesheets are outside the rule, but the example is what gets copied.

packages/ui-kit/src/styles/theme.css
```diff
@@ -393,6 +417,10 @@
     --primary: #7c4fde;
     --primary-hover: #8257e6;
     --primary-foreground: #ffffff;
+    /* Button default variant's focus-ring tone, measured against
```
itechmeat: ⚪ [Not a blocker] theme.css's own convention keeps notes in the light block only - no other token carries a comment in either dark block. The two new ring tokens got a three-line comment in all three places, six blocks for two tokens. The measured dark ratios could fold into the light block's note and the dark copies go.

packages/ui-kit/src/components/badge/badge.module.css
```diff
@@ -99,11 +101,18 @@
...
  */
-.badge > svg {
+.badge .icon {
+  display: inline-flex;
+  flex-shrink: 0;
+  color: var(--badge-accent);
   pointer-events: none;
+}
+
+.badge .icon svg {
```
itechmeat, line 117: ⚪ [Not a blocker] Two icon slots land in this PR: Input sizes its svg with `--icon-size-sm`, Badge with a raw `0.75rem`. The value is inherited from the old `.badge > svg` and no 12px token exists - so, if 12px is a real size in the system it may deserve a token; if not, is the divergence from the Input slot deliberate?

packages/ui-kit/src/components/button/button.module.css
```diff
@@ -20,29 +20,33 @@
...
-  --button-focus-ring-inner: var(--button-focus-ring);
   /* Anchors the absolutely-centered .spinner. */
   position: relative;
   display: inline-flex;
   align-items: center;
   justify-content: center;
   gap: var(--space-2);
   white-space: nowrap;
-  border: var(--border-width) solid transparent;
+  border: var(--border-width) solid var(--button-border, transparent);
```
itechmeat, line 49: ⚪ [Not a blocker] The `--button-*` hooks are inherited custom properties with no reset on `.button`, so setting `--button-bg` on a container recolors every Button underneath - including a clear button inside an Input `end` slot. If container-level theming is intended, worth a sentence in button.md; if not, worth knowing.

itechmeat (changes requested):
Solid, carefully verified work - the direction of all four API changes is right, the focus-ring rework is a genuine accessibility improvement, and the diff shows its homework: measured contrast ratios that check out, strengthened test guards, an honest breaking-change table. What keeps it one step from approve is doc hygiene rather than code - the PR's own migration story leans on the docs, and one touched file still describes the retired API in a few places. Beyond that, only questions: two visual edge cases worth a screenshot and a consumer-migration check, none of which I expect to move code.

packages/ui-kit/src/components/select/select.tsx
SysoevAndrey: You were right about the mechanism, and chasing it turned up something bigger: **the scroll arrows never rendered at all** — at any list length, at any scroll position. Fixed in 4de033e7.

Base UI applies its scroller styles (`LIST_FUNCTIONAL_STYLES`) to `Select.List` only while `alignItemWithTriggerActive` is true (`select/list/SelectList.mjs`). This PR pinned that mode off, so the List stopped scrolling and our CSS scrolled `.popup` instead — but arrow visibility is computed from `store.state.listElement`, i.e. the List, whose `scrollHeight === clientHeight` forever. Measured on the demo: popup `scrollTop 62`, `scrollHeight 776`, `clientHeight 550`, and zero elements matching the arrow classes. A regression this branch introduced, not a pre-existing gap.

The list is now the scroller (popup is a clipping flex column) and `.list` carries `scroll-padding-block` of exactly one arrow height, so `block: 'nearest'` can no longer park an item under an arrow — your suggested remedy, needed for real once the arrows came back. Verified in the browser: both arrows mount and unmount at the right edges, and `18:00` opens fully clear of the down arrow.

One trap worth recording: the first attempt left `.scrollArrow` content-box, so its own padding pushed it to 32px against 24px of reserved scroll padding and reproduced the exact overlap. There is now a static test tying the arrow's height, `.list`'s scroll padding and that `box-sizing` together — jsdom cannot check the layout, but it can stop those three drifting apart.

packages/ui-kit/src/components/badge/badge.tsx
SysoevAndrey: Right, design-notes.md was missed in the docs sweep — fixed in b8baa583. Corrected the "pill vs. bare dot" line, the "(pill/dot; ...)" listing and the Input `search`-type claim, and swept the rest of the file for other retired-API statements while in there.

packages/ui-kit/demo/components-page.tsx
SysoevAndrey: Measured it: the ring is pixel-exact flush with the field, not overlapping it. Button 32px at top 411.5/bottom 443.5, ring inflated 4px per side to 407.5/447.5, input border-box 407.5/447.5 — identical to sub-pixel. Horizontally it stays well inside. So nothing clips or spills, and it reads clean in the screenshot; the real risk is cosmetic — on a theme where ring and border colours are close they merge into one line instead of two strokes with a gap. Left as is; if it bothers anyone, the lever is the ring offset on small buttons, not the input.

packages/ui-kit/src/components/button/button.module.css
SysoevAndrey: Correct — the box-shadow that comment justified itself with is gone. Rewritten in b8baa583 to say what the block actually does now: forced-colors replaces author colours with system ones, so the block hands the indicator to `Highlight` instead of the variant ring colour.

packages/ui-kit/demo/components-page.tsx
SysoevAndrey: My miss — I changed the doc and left the demo contradicting it. Added a `SelectLabel` to the Hour select in b8baa583, so the page now shows both forms: two groupless selects and one labelled group.

packages/ui-kit/llms.txt
SysoevAndrey: Fixed in b8baa583 — it is a slot prop, same wording as the Button entry. Checked badge.md for the same slip.

packages/ui-kit/src/components/input/input.module.css
SysoevAndrey: Agreed, and it is worth more than a doc line: a disabled field whose clear button is still bright and clickable is just wrong. b8baa583 dims the `end` slot to the kit's 0.42 disabled opacity and makes it inert via `.input:disabled ~ .end`, with the behaviour documented in input.md.

packages/ui-kit/src/components/input/input.test.tsx
SysoevAndrey: Fair — both titles only meant something next to this diff. Renamed in b8baa583 to present-tense behaviour sentences; the assertions are unchanged.

packages/ui-kit/src/components/button/button.md
SysoevAndrey: Fixed in b8baa583 — `var(--primary-foreground)` instead of `#fff`. The example is the thing that gets copied, so it should follow the same rule the kit's own modules do.

packages/ui-kit/src/styles/theme.css
SysoevAndrey: Right, six comment blocks for two tokens. Folded into the light block's note (measured dark ratios included) and deleted the two dark copies in b8baa583. No token values changed.

packages/ui-kit/src/components/badge/badge.module.css
SysoevAndrey: Not deliberate — inherited from the retired `.badge > svg` rule. Added `--icon-size-xs: 0.75rem` to theme.css and Badge now consumes it, so the two icon slots read from the same scale (b8baa583).

packages/ui-kit/src/components/button/button.module.css
SysoevAndrey: Confirmed, and it is intended as container-level theming rather than something to reset — a reset would kill that use. Documented explicitly in button.md (b8baa583): set the hooks on the button's own class for one button, on an ancestor to theme every button inside, including one sitting in an Input `end` slot.

packages/ui-kit/src/components/input/input.tsx
SysoevAndrey: Agreed it is now a kit-wide contract with no home, and I am leaving it as a follow-up rather than doing it here — with a reason worth recording.

The kit currently has zero cross-component imports: every component directory is self-contained, and the only shared module in `src/` is `__test-utils__`, which is test-only. So this is not "move a function into the obvious shared file" — it is introducing the first shared runtime module, which drags in placement, the export surface (`index.ts` and the `./components/*: null` export blocks in package.json), and knip's unused-export detection. That is a structural call better made deliberately than as a rider on a review round.

Worth doing though — three copies plus three test retellings is exactly the point at which a contract wants one home, and a fourth slot will copy it.

packages/ui-kit/src/components/select/select.tsx
SysoevAndrey: Checked with Andrey: the standup remark is about fixes to the current select in this PR, not new work — nothing separate is planned. And it lines up with what came out of your other comment: the scroll arrows were not rendering at all, which is very likely the "looking off" that was called out. Fixed in 4de033e7, details in the reply on that thread.

packages/ui-kit/src/components/input/input.tsx
SysoevAndrey: No, we do not have their call sites — nothing in this monorepo imports the package, so the blast radius is entirely insight-front on alpha.1 and it is invisible from here. The plan is the one you describe: their migration picks these up from the breaking-change table in the description, which lists both silent ones explicitly (type="search" losing the magnifier, pass icon instead; Badge rendering no dot, add dot), and every break also carries a BREAKING CHANGE footer on its commit.

Worth being blunt that this is the weak link: both are visual-only, so nothing fails their build or their tests — the fields and badges just quietly look different. If someone has a line to whoever owns those screens, a heads-up beats the table.

packages/ui-kit/demo/components-page.tsx
```diff
@@ -156,7 +181,14 @@ export function ComponentsPage() {
...
           <Field name="query">
             <FieldLabel>Search</FieldLabel>
-            <Input type="search" placeholder="Search projects…" />
+            <Input
+              type="search"
+              value={query}
+              onChange={(e) => setQuery(e.currentTarget.value)}
+              placeholder="Search projects…"
+              icon={<MagnifierIcon />}
+              end={<Button variant="ghost" size="sm" icon={<DemoIcon />} aria-label="Clear search" onClick={() => setQuery('')} />}
```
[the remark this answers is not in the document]
SysoevAndrey: Caught me — `DemoIcon` is a plus sign, so the clear button was advertising "Clear search" while drawing "+". Added a proper X icon to the demo helpers and wired it up in a4869e78.

packages/ui-kit/src/components/select/select.test.tsx
```diff
@@ -129,4 +134,74 @@ describe('Select', () => {
...
+  });
+
+  it('always opens the popup below the trigger, with no align-with-trigger attribute', () => {
+    renderSelect({ defaultOpen: true });
+    const popup = screen.getByRole('listbox').closest(`.${styles.popup}`);
+    // The popup exists and no longer advertises the retired overlay mode —
+    // the attribute is gone entirely, not set to "false".
+    expect(popup).not.toBeNull();
+    expect(popup?.hasAttribute('data-align-trigger')).toBe(false);
+  });
```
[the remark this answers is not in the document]
SysoevAndrey: Both fair, and the second one is the sharper catch — that assertion could not fail. The attribute was removed from the component entirely, so nothing can bring it back, and the test stayed green no matter where the popup went.

a4869e78:
- The groupless-list test now parses `select.module.css` and asserts `.list` declares `padding: var(--space-1)` while `.group` declares none — the actual contract (the list owns the inset, so grouping cannot change it), not just the class name. Checked it can fail: stripped the padding, watched it go red, restored.
- For placement, jsdom does expose Base UI's resolved placement — `data-side="bottom"` / `data-align="center"` on the popup — so the test now asserts `data-side="bottom"`, which does depend on the popup actually being placed below the trigger. The `data-align-trigger` absence check stays alongside it as a guard against the prop returning.

gs-layer (approved):

packages/ui-kit/src/components/input/input.md
```diff
@@ -23,16 +23,22 @@ error, validation state) when rendered inside one.
...
-`type="search"` upgrades the field visually as well as semantically: a
-leading magnifier icon (decorative, `aria-hidden`) renders inside the
-field via a presentational wrapper element, and the input announces as a
-searchbox. No prop needed — the native type is the switch.
+Icons are explicit: `icon` draws a leading icon inside the field (via a
+presentational wrapper the component adds only when a slot is present),
+`end` overlays trailing content such as a clear button. Semantics still
+come from the native `type` — a search field is `type="search"` (searchbox
+role) plus a magnifier passed to `icon`; nothing renders automatically.
+A disabled input dims and disarms its `end` slot too (0.42 opacity,
```
itechmeat: 🟠 `pointer-events: none` only stops the mouse - a button inside `end` keeps its tab stop and still fires on Enter/Space, so on the exact composition input.md recommends, a keyboard user can tab from a disabled field onto the dimmed clear button and activate it. The doc's "disarms" and the CSS comment's "inert" promise more than the CSS delivers - either soften the wording to pointer-only, or actually propagate disabled into the slot.

packages/ui-kit/src/components/input/input.module.css
```diff
@@ -112,29 +112,60 @@
...
+}
+
+/*
+ * .end is a sibling inside .wrap, not a descendant of .input, so
+ * .input:disabled above never reaches it: a disabled field's clear button
+ * (or whatever else lives in `end`) would otherwise stay fully opaque and
+ * clickable. Match the same 0.42 disabled dim as .input itself and make it
+ * inert.
+ */
+.input:disabled ~ .end {
```
itechmeat: 🟡 [Not a blocker] `~` only matches following siblings and the `icon` slot renders before the input, so the new rule structurally cannot reach it - a disabled search field dims its border, text and placeholder to 0.42 while the magnifier next to them stays at full strength. `.wrap:has(.input:disabled) .icon` would cover both slots; if leaving the icon bright is deliberate, neither the CSS comment nor input.md says so.

packages/ui-kit/src/components/select/select.module.css
```diff
@@ -184,14 +193,50 @@
...
+ * column, .list would otherwise refuse to shrink below its content's
+ * height and grow the popup instead of scrolling within it.
+ *
+ * `scroll-padding-block` keeps a scrolled-into-view item clear of the
+ * arrows: Base UI scrolls the active item with `block: 'nearest'`, which
+ * can park it flush against the scrollport edge, and the arrows are
+ * `position: absolute` against .popup, overlaying that same edge — this
+ * reserves exactly one arrow's box height as scroll padding so `nearest`
+ * never lands an item underneath either arrow. The height is a local
+ * custom property, not a shared one on .popup: the token-guard test
```
itechmeat: 🟡 [Not a blocker] The comment justifies duplicating the `--select-scroll-arrow-height` formula by the token-guard scoping a local declaration to the rule's own leading class - but tokens.test.ts sanctions a second shape explicitly: an ancestor declaration read by descendants (`.declarer .reader`, the Card `--card-spacing` pattern). `leadingClass()` yields `.popup` for `.popup .list`, and both `.list` and `.scrollArrow` are DOM children of `.popup`, so declaring once on `.popup` passes the guard with no duplication, no 30-line rationale and none of the three bespoke tests policing the copies. Two small things ride along: a private internal wearing the public `--select-*` hook prefix, and the arrow-height formula using `--space-2` where the element's own padding is 2 x `--space-1` - agreeing only numerically.

packages/ui-kit/src/styles/theme.css
```diff
@@ -119,21 +119,26 @@
   --control-height-sm: 2rem;
   --control-height-md: 2.25rem;
   --control-height-lg: 2.5rem;
+  /* Badge's icon/dot slot only — no drawn specimen at any other size uses
+   * 12px; every other icon consumer sits on sm/md/lg above. */
+  --icon-size-xs: 0.75rem;
```
itechmeat, line 124: 🔵 [Not a blocker] The new token joined the invariants block next to `--icon-size-sm/md/lg`, but `THEME_INVARIANT_TOKENS` in tokens.test.ts still lists only its three siblings - so both convention guards (declared in the invariants block, never redefined per theme) skip it. A later `--icon-size-xs` in a dark block would sail through; one line in the list closes the hole this change opened.

packages/ui-kit/src/components/button/button.module.css
```diff
@@ -65,31 +69,35 @@
...
-    var(--button-focus-ring-inner);
+  outline: var(--border-width-focus) solid var(--button-focus-ring);
+  outline-offset: 2px;
 }
 
 /*
- * Forced-colors mode drops box-shadow, so the ring above leaves the button with
- * no visible focus at all. Hand the indicator back to the UA outline, which the
- * mode does paint, and keep it off the shadow.
+ * Forced-colors mode replaces author colors with a fixed system palette, so
```
itechmeat, line 90: ⚪ [Not a blocker] The rewritten forced-colors comment here no longer mentions box-shadow, but input.module.css:85 still delegates to it: "See button.module.css: forced-colors drops box-shadow, restore UA outline". Input's own rule stays correct - its ring really is a box-shadow - only the pointer is broken now. Worth giving input (and any sibling carrying the same cross-reference) its own one-line reason instead of delegating.

packages/ui-kit/design-notes.md
```diff
@@ -354,7 +364,8 @@ Architecture's build bullet).
      than by an axis name only this component used, so every component in
      the kit is driven by `variant` (+ `size`, or the occasional real extra
      axis: Badge's `shape`, Table's `density`). `shape` rather than `size`
-     because the values are pill vs. bare dot and Badge has no size axis;
+     because the values are pill vs. plain (dot is a separate opt-in flag,
+   orthogonal to shape) and Badge has no size axis;
```
itechmeat: ⚪ [Not a blocker] Three-space indent on a continuation line whose siblings all use five - it renders via lazy continuation but reads as a stray line in a file this round asks reviewers to trust as swept.

packages/ui-kit/src/components/select/select.module.css
```diff
@@ -145,14 +145,23 @@
...
+  flex-direction: column;
   position: relative;
   isolation: isolate;
   z-index: 50;
   max-height: var(--available-height);
   width: var(--anchor-width);
   min-width: 9rem;
-  overflow-x: hidden;
-  overflow-y: auto;
+  overflow: hidden;
```
itechmeat, line 177: ⚪ [Not a blocker] The `.item` comment at line 255 still names ".popup's own overflow-x: hidden above" - that declaration was folded into the `overflow: hidden` shorthand here, so the grep target it names no longer exists. Behaviour is unchanged; only the wording is stale.

packages/ui-kit/src/components/input/input.test.tsx
```diff
@@ -8,25 +8,75 @@ afterEach(cleanup);
...
     expect(input.className).toContain(styles.input);
-    // Every non-search type stays a bare input — no wrapper element.
-    expect(input.parentElement?.className ?? '').not.toContain(styles.searchWrap);
+    // No slots — no wrapper element: the input is the root.
+    expect(input.parentElement).toBe(container);
   });
 
-  it('type="search" adds the decorated wrapper and keeps the searchbox role', () => {
-    render(<Input type="search" placeholder="Find" className="consumer" />);
+  it('type="search" stays a bare input with the searchbox role and no icon', () => {
```
itechmeat, line 19: ⚪ [Not a blocker] The delta adds four plain `it()` blocks (+1 input, +3 select, no `.each`), so the Verification section's "276 tests pass (21 files)" now undercounts by four - worth updating in step, since that section is what a reviewer is asked to take on trust.

itechmeat (changes requested):
Strong response round - 13 of 15 items closed cleanly, both deferrals honestly reasoned, and the scroll-arrow question turned into a genuine catch: the arrows never rendered on this branch at all, and the fix is correctly root-caused against Base UI's own source rather than patched around. The new scroller architecture checks out, and the mid-round follow-up closing both CodeRabbit threads does too. One cheap thing is all that stands between this and approve: input.md now promises a disabled field "disarms" its end slot, but the CSS only stops pointer interaction - a keyboard user can still tab in and fire it. One sentence (or propagating disabled) closes it; the rest below are non-blocking notes.

packages/ui-kit/src/components/input/input.md
SysoevAndrey: You are right, and the wording was the smaller half of it — `pointer-events: none` leaves the tab stop and Enter/Space intact, so the hole was real on exactly the composition input.md recommends. d8148526 sets `inert` on the end-slot wrapper, which does remove it from the tab order and block activation.

One honest limit, now stated in input.md and the code comment instead of papered over: only the direct `disabled` prop is observable here. A `<Field disabled>` merges Base UI's field state one layer deeper (FieldControl), so it reaches the rendered input but never this component's props — that path still dims (see the `:has()` change from your other comment, which covers it) but cannot lose the tab stop. Consumers disabling through Field should disable their own end content.

packages/ui-kit/src/components/input/input.module.css
SysoevAndrey: Correct — `~` only looks forward and `icon` renders before the input, so the magnifier stayed bright in a dimmed field. Switched to `.wrap:has(.input:disabled)` covering both slots in d8148526, with the reason in the comment. Useful side effect: `:has()` keys off the rendered input, so it dims for a Field-driven disable too, which the prop-based `inert` cannot see.

packages/ui-kit/src/components/select/select.module.css
SysoevAndrey: Good catch — I had the guard wrong. `leadingClass()` reads only the selector's first class, so `.popup .list` buckets as `.popup`, exactly the `--card-spacing` shape. d8148526 declares `--select-scroll-arrow-height` once on `.popup`, reads it from `.popup .list` / `.popup .scrollArrow`, cuts the rationale to a line, and deletes the two drift tests that existed only to police the copies. Kept the `box-sizing: border-box` one — that guards a real regression the browser check caught (without it the arrow grows past the padding reserved for it).

Formula fixed too: it is now the icon size plus twice `--space-1`, matching the arrow's actual padding rather than agreeing with it by coincidence. On the prefix: the kit already uses this shape for private locals (`--card-spacing`, `--badge-accent`) — what marks a property public is the fallback-consumed hook pattern the token guard exempts, not the name, so I left it.

packages/ui-kit/src/styles/theme.css
SysoevAndrey: Right, the guards silently skipped it. Added `--icon-size-xs` to `THEME_INVARIANT_TOKENS` in d8148526, so both conventions — declared in the invariants block, never redefined per theme — now cover it.

packages/ui-kit/src/components/button/button.module.css
SysoevAndrey: Fixed in d181abea, and the sweep was bigger than expected: grepping for that delegated reference turned up 11 occurrences across 10 modules (badge, checkbox, input, radio-group, select, switch, table, tabs ×2, textarea, toast). Each now carries its own reason, which also separates the two mechanisms that had been collapsed into one pointer — a box-shadow ring that forced-colors drops entirely, versus an outline ring that survives but needs recolouring to the system Highlight.

packages/ui-kit/design-notes.md
SysoevAndrey: Fixed in d181abea.

packages/ui-kit/src/components/select/select.module.css
SysoevAndrey: Fixed in d181abea — the shorthand swallowed the declaration that comment was pointing at, so the grep target no longer existed.

packages/ui-kit/src/components/input/input.test.tsx
SysoevAndrey: Right, that section is stale. The count is now 284 across 21 files; updating the description.

itechmeat (approved):

Issue #530, opened 2026-08-04
[BUG] Provenance records always say the template occupies the repository root


gs-layer:
## What happens

Every provenance record ever written says the template occupies `"."` — the repository root — no matter what the template actually declares.

## Reproduce

Seed or add any template, then open `.frontx/provenance.json`:

```json
{ "templateIdentity": "@gears-frontx/frontx-template-mfe", "occupiedOwnershipBoundary": "." }
```

`template-mfe` declares `src-app/mfe_packages/` and `.frontx/ai/@gears-frontx/frontx-template-mfe/`. Neither is the root. The field says otherwise, and it says the same thing for the shell, and for any future template.

## Why

`materializeAssembly` builds each record from three fields — `templateIdentity`, `scaffoldedFromVersion`, `sourceSpec` (`packages/cli/src/scaffold/materialize.ts:289-293`). It never sets `occupiedOwnershipBoundary`. The field is optional, so nothing complains, and `provenance/write.ts:39` fills the blank:

```ts
occupiedOwnershipBoundary: template.occupiedOwnershipBoundary ?? '.',
```

The fallback was meant for the rare record that genuinely has no boundary. In practice it is the only value the field has ever held.

Two things kept this quiet:

- `provenance/types.ts:12` promises "`writeProvenance` always fills a concrete value on the written record" — technically true, which is why reading the type does not raise a flag.
- The one test that touches the field, `composition.test.ts:295`, asserts `toHaveProperty('occupiedOwnershipBoundary')` — presence, not value. It stays green whatever is written.

## Why it matters

`occupiedBoundariesFromProvenance` re-reads each applied template's manifest to reconstruct boundaries, so `add` works today in spite of this. But the recorded value is the durable answer to "what did this template occupy when it was applied" — the question that matters when a template's manifest later changes, or when the template is no longer installed. Right now that answer is wrong for every repository seeded so far.

## What a fix needs

- Populate the field from the boundaries actually staged at materialization time.
- Decide what to do with repositories already carrying `"."` — they are indistinguishable from a legitimate root-owning template, so a migration or a re-derivation step is needed, not just a code change.
- Make the test assert the value, not the key.

Raised during review of #500; deliberately left out of that PR because the migration question is bigger than the write path.

Issue #506, opened 2026-08-03
[BUG] Make upgrade application atomic through the provenance write


gs-layer:
## Problem

`applyChangeSet` restores file mutations only while its project-file loop is protected. The final `deps.writeProvenance(...)` runs after that boundary. If the filesystem write fails, the function throws after approved project files were changed, leaving the repository partially upgraded and bypassing the promised `ApplyResult` failure path.

The gap predates PR #500. That PR removes two earlier throw paths before the first write, but intentionally leaves this final write outside the rollback boundary.

## Required outcome

- Treat the provenance write as part of the same atomic application as project-file writes.
- On a provenance-write failure, restore every snapshotted file and the prior provenance bytes, then return `{ ok: false }` rather than throwing.
- Keep the successful path preserving the full multi-record provenance array added by #500.

## Acceptance criteria

1. A throwing `writeProvenance` leaves every changed project file and `.frontx/provenance.json` byte-for-byte at its pre-upgrade state.
2. `applyChangeSet` returns a typed failure for that case.
3. A targeted regression test covers the failure after at least one project-file mutation.
4. Upgrade FEATURE markers and rollback contract remain accurate.

## Context

Found in review discussion on PR #500, `packages/cli/src/upgrade/apply.ts` around `inst-app-try` and `inst-app-update-prov`.

PR #556, opened 2026-08-10
fix(cli): record each template's occupied boundary, atomicize upgrade's provenance write

gs-layer:
## Summary

Follow-up fixes for the provenance boundary and atomic upgrade review on PR #556.

- **#530 / boundary encoding** — `materializeAssembly` writes each template's resolved occupied ownership boundary through `formatOccupiedBoundary()`. Resolved boundaries are stored as canonical JSON, including the empty owns-nothing boundary `{"exclusiveSubtrees":[],"sharedFiles":[]}`; legacy `.` remains readable only as an omitted-field / pre-schema sentinel.
- **#531 / target boundary refresh** — `computeChangeSet` now scopes the diff to the target version's declared ownership boundary and carries `targetOccupiedOwnershipBoundary` through apply, so the provenance record refreshes version and boundary together.
- **#506 / atomic provenance write** — `applyChangeSet` performs the provenance write inside the rollback-covered apply block. Failure returns `{ ok: false }` after attempting restore from the pre-upgrade snapshot.
- **Kit mirror** — the cyber-pilot kit mirror `ChangeSet` type and tests now include `targetOccupiedOwnershipBoundary`.
- **Mergeability** — the branch is merged up with `develop@9f32349`, preserving the scoped artifact move under `packages/*/architecture`.

## Changes

- `packages/cli/src/provenance/boundary.ts` — canonical JSON for all resolved boundaries; empty boundary no longer serializes to `.`.
- `packages/cli/src/upgrade/compute.ts` — files outside the target ownership boundary are skipped instead of treated as exclusive ownership.
- `packages/cli/src/upgrade/apply.ts` — failure message now reflects best-effort snapshot restore.
- `packages/cli/architecture/features/{composed-provenance,upgrade-changeset}/FEATURE.md` — schema, algorithm, DoD, and AC updated for target boundary refresh and canonical boundary encoding.
- Tests cover deterministic boundary serialization, empty target boundary behavior, provenance-write rollback, and kit mirror type fixtures.

## Test plan

- [x] `npm run test:unit --workspace=@gears-frontx/cli -- src/__tests__/upgrade.test.ts src/__tests__/composition.test.ts src/__tests__/provenance-boundary.test.ts`
- [x] `npm run test:unit --workspace=@gears-frontx/cyber-pilot-kit-frontx -- src/upgrade-orchestration/__tests__/enrich.test.ts src/upgrade-orchestration/__tests__/invoke-upgrade-command.test.ts src/upgrade-orchestration/__tests__/orchestrate.test.ts`
- [x] `npm run type-check --workspace=@gears-frontx/cli`
- [x] `npm run type-check:test --workspace=@gears-frontx/cli`
- [x] `npm run type-check --workspace=@gears-frontx/cyber-pilot-kit-frontx`
- [x] `npm run type-check:test --workspace=@gears-frontx/cyber-pilot-kit-frontx`
- [x] `npm run arch:check`
- [x] `cfs --json validate --artifact packages/cli/architecture/features/upgrade-changeset/FEATURE.md --skip-code`
- [x] `cfs --json validate --artifact packages/cli/architecture/features/composed-provenance/FEATURE.md --skip-code`
- [x] `cfs --json validate-toc packages/cli/architecture/features/upgrade-changeset/FEATURE.md`
- [x] `cfs --json validate-toc packages/cli/architecture/features/composed-provenance/FEATURE.md`
- [x] `cfs --json validate`


packages/cli/src/upgrade/apply.ts
```diff
@@ -176,6 +176,32 @@ export async function applyChangeSet(
       // @cpt-end:cpt-frontx-algo-upgrade-changeset-apply:p1:inst-app-apply-entry
     }
     // @cpt-end:cpt-frontx-algo-upgrade-changeset-apply:p1:inst-app-for-each-entry
+
+    // @cpt-begin:cpt-frontx-algo-upgrade-changeset-apply:p1:inst-app-update-prov
```
itechmeat, line 180: 🟠 The code moves `inst-app-update-prov` inside `inst-app-try`, but upgrade-changeset/FEATURE.md still lists it as step 4 after the CATCH, and step 3.2 still reads "RETURN failure without updating provenance" - which can now be a caught-and-restored write attempt. After the follow-up commit the drift widened: step 4, the DoD and the AC all still say the record is updated "to the newer version" only, while the code now also refreshes the occupied boundary. #506's own acceptance criterion 4 is "Upgrade FEATURE markers and rollback contract remain accurate", and `cfs validate` cannot see this class of drift. Could the FEATURE's numbered algorithm move with the code (step under TRY, 3.2 reworded, an error scenario for the restored write, the boundary refresh named in step 4/DoD/AC) - rewritten in place rather than appended?

packages/cli/src/provenance/boundary.ts
```diff
@@ -0,0 +1,41 @@
...
+ * `ProvenanceRecord.occupiedOwnershipBoundary` holds. Non-empty boundaries use
+ * canonical JSON so shared-file region ownership stays durable while the
+ * provenance schema remains backward-compatible with a string field.
+ *
+ * A boundary with neither exclusive nor shared ownership is legitimately
+ * root-owning (issue #530) and collapses to '.', the same sentinel
+ * `provenance/write.ts` fills in when a caller omits the field entirely.
+ */
+export function formatOccupiedBoundary(boundary: OwnershipBoundary): string {
+  if (boundary.exclusiveSubtrees.length === 0 && boundary.sharedFiles.length === 0) return '.';
```
itechmeat: 🟠 Four linked things. The comment above this line cites #530's "legitimately root-owning" as the licence for the `'.'` sentinel, while the issue uses those words as the problem: '.'-records are "indistinguishable from a legitimate root-owning template, so a migration or a re-derivation step is needed". Second, the semantics invert: per manifest/types.ts and assembler.ts an empty boundary contributes zero files - it owns nothing, while `'.'` reads as "the whole repository". Third, new with the refresh commit: the boundary used to be passthrough on upgrade, so a correct value survived - now it is unconditionally replaced, an empty-boundary manifest is conforming (validate-contract defaults both arrays with no violation), and upgrading to such a version overwrites a previously correct record with the sentinel; the suite already exercises this unasserted (registerVersion's default fixture has an empty boundary) and even hardcodes `targetOccupiedOwnershipBoundary: '.'` as the expectation. Fourth, #530's second required bullet - what happens to repos already carrying `'.'` - is still neither done nor pointed to. A guard on the empty case at refresh (or a distinct encoding), the exclusion recorded in the owning artifact, and the migration pointer would close all four - which way do you want to go?

packages/cli/src/provenance/boundary.ts
```diff
@@ -0,0 +1,41 @@
...
+      .sort((a, b) => {
+        const byPath = compareStrings(a.path, b.path);
+        if (byPath !== 0) return byPath;
+        const byMergeStrategy = compareStrings(a.mergeStrategy, b.mergeStrategy);
+        if (byMergeStrategy !== 0) return byMergeStrategy;
+        return compareStrings(a.ownedRegions.join('\0'), b.ownedRegions.join('\0'));
+      }),
+  };
+
+  return JSON.stringify(canonicalBoundary);
```
itechmeat, line 35: 🟠 composed-provenance/FEATURE.md owns this contract's schema (ADR-0019 delegates "each record's exact field layout" to it) and still describes the field only as "occupied ownership boundary"; the `ProvenanceRecord` doc comment was not touched either, so nothing tells a reader the value is canonical JSON of the whole boundary plus a `'.'` sentinel - nothing parses it back today, and for template-shell it is ~600 chars of escaped JSON in a file the FEATURE asks to keep human-readable. The refresh commit widens this: the same undocumented encoding now crosses the ChangeSet contract too - the new required `targetOccupiedOwnershipBoundary` lands in the `inst-change-set-types` marker span whose instruction id exists nowhere under architecture/ (the dangling id predates this PR), and `inst-cmp-return-changeset` still enumerates the change set as "clean entries + flagged conflicts" and the compute algorithm's Output as added/modified/removed files plus conflicts - neither names the new field. The design itself is defensible - string field stays backward-compatible, lossless beats a lossy path list, writer-before-reader is the right order - but the schema needs stating where the FEATURE says it lives, now in both places.

packages/cli/src/upgrade/types.ts
```diff
@@ -36,6 +36,10 @@ export interface ChangeSet {
   templateIdentity: string;
   baselineVersion: string;
   targetVersion: string;
+  // Canonical `ProvenanceRecord.occupiedOwnershipBoundary` for the target
+  // version's ownership boundary. Apply persists it with the version bump so
+  // the next upgrade starts from the ground this template now occupies.
+  targetOccupiedOwnershipBoundary: string;
```
itechmeat, line 42: 🟡 [Not a blocker] packages/cyber-pilot-kit-frontx/src/upgrade-orchestration/types.ts declares itself a "Structural mirror of the F14 engine's ChangeSet ... carried across the frontx upgrade command surface's JSON contract" - and with this required field added on the engine side it no longer is one. No compiler catches it: the value crosses the process boundary as JSON and is cast (`as unknown as ChangeSet`), so kit-side consumers of the documented contract just cannot see the field in the type. Nothing breaks today, but the mirror's self-description is now checkably false - one line on the kit side keeps it honest.

packages/cli/src/provenance/boundary.ts
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * A boundary with neither exclusive nor shared ownership is legitimately
+ * root-owning (issue #530) and collapses to '.', the same sentinel
+ * `provenance/write.ts` fills in when a caller omits the field entirely.
+ */
+export function formatOccupiedBoundary(boundary: OwnershipBoundary): string {
+  if (boundary.exclusiveSubtrees.length === 0 && boundary.sharedFiles.length === 0) return '.';
+
+  const canonicalBoundary: OwnershipBoundary = {
+    exclusiveSubtrees: [...new Set(boundary.exclusiveSubtrees)].sort(compareStrings),
```
itechmeat, line 19: 🟡 [Not a blocker] The PR body's Changes list names four files and says nothing about `upgrade/compute.ts`, `upgrade/types.ts`, or the boundary-refresh-on-upgrade behaviour - a third of what the PR does after the follow-up commit. Two older phrases also still describe the first implementation: "collapses ... paths into the record's single string field" (it serializes the full structure including mergeStrategy and ownedRegions) and the scope line naming adapters/provenance-io.ts, which is not in the diff. ("Deduplicated and sorted" became accurate again once the Set returned.)

packages/cli/src/provenance/boundary.ts
```diff
@@ -0,0 +1,41 @@
+// @cpt-dod:cpt-frontx-dod-composed-provenance-provenance-at-scaffold:p1
+import type { OwnershipBoundary } from '../manifest/types';
+
+function compareStrings(a: string, b: string): number {
```
itechmeat, line 4: 🔵 [Not a blocker] For string arrays this is byte-for-byte what bare `.sort()` does (the first commit used it), and nothing in the file says why an explicit comparator became necessary. The sharedFiles tiebreak - mergeStrategy, then ownedRegions joined - orders two entries with the same path in one manifest, which is either impossible or a manifest bug that validation should reject rather than silently order. Probably just `.sort()` and sort by path?

packages/cli/src/__tests__/provenance-boundary.test.ts
```diff
@@ -0,0 +1,32 @@
+import { describe, expect, it } from 'vitest';
+import { formatOccupiedBoundary } from '../provenance/boundary';
```
itechmeat, line 3: 🔵 [Not a blocker] boundary.ts carries the provenance-at-scaffold DoD marker, but the test file proving it carries none - the instruction-to-proof link exists in one direction only, while the sibling test files open with their markers. One line - though note the coupling: adding it today would claim coverage of an instruction that does not yet describe the tested encoding, so it lands after the schema paragraph does.

packages/cli/src/upgrade/apply.ts
```diff
@@ -176,6 +176,32 @@ export async function applyChangeSet(
...
+    const updatedRecords = existingRecords.map((record, index) =>
+      index === targetIndex
+        ? {
+            ...record,
+            scaffoldedFromVersion: changeSet.targetVersion,
+            occupiedOwnershipBoundary: changeSet.targetOccupiedOwnershipBoundary,
+          }
+        : record,
+    );
+    await deps.writeProvenance(provPath, JSON.stringify(updatedRecords, null, 2));
```
itechmeat, line 203: ⚪ [Not a blocker] For the exact scenario this fix targets - the disk write failing - the restore write of the same file goes through the same mkdir+write pair (both real implementations match), fails for the same reason, gets swallowed by the best-effort catch, and the caller still reads "Apply failed and was restored". Best-effort predates this PR and is probably deliberate - but should the message promise restoration as a fact?

packages/cli/src/scaffold/materialize.ts
```diff
@@ -290,6 +291,11 @@ export async function materializeAssembly(
       templateIdentity: contribution.templateName,
       scaffoldedFromVersion: manifestResult && manifestResult.ok ? manifestResult.manifest.version : '',
       sourceSpec: entry?.source ?? '',
+      // The boundary this template actually occupies at materialization time
+      // (issue #530) — not the '.' every record fell back to when this field
+      // was left unset, which was indistinguishable from a legitimate
+      // root-owning template for every template, always.
+      occupiedOwnershipBoundary: formatOccupiedBoundary(contribution.ownershipBoundaries),
```
itechmeat, line 298: ⚪ [Not a blocker] The comment at the moved write in apply.ts is now fourteen lines, mostly the bug's history and why the old placement was wrong - PR-body and FEATURE material rather than code. And materialize.ts's "not the '.' every record fell back to when this field was left unset" describes a state that stops existing once this merges. The package's precedent is verbose comments, so up to you - trimming the retrospective phrasing is the minimum.

itechmeat (changes requested):
Both fixes are real and correctly built, and the follow-up commit closes the boundary-refresh gap (self-caught, credit where due) and all three bot threads the right way - the code side of this PR is in good shape. What blocks approval is that the spec keeps not moving with the code: the FEATURE still describes the old apply order and says nothing about the boundary refresh, the new ChangeSet field, or the on-disk encoding; the issue's migration question is still dropped without a pointer; and the refresh now makes the '.' ambiguity writable through upgrade over previously correct records. All of it closes with artifact and text edits plus one guard decision - nothing structural.

packages/cli/src/upgrade/apply.ts
gs-layer: Fixed in `53c4f00a` and carried through the `develop` merge in `36ae661d`. `packages/cli/architecture/features/upgrade-changeset/FEATURE.md` now moves `inst-app-update-prov` under the TRY block, rewords the CATCH return for a restored provenance-write attempt, adds the provenance-write failure error scenario, and names `targetOccupiedOwnershipBoundary` / boundary refresh in the algorithm output, return step, DoD, and AC.

packages/cli/src/provenance/boundary.ts
gs-layer: Fixed in `53c4f00a` / `36ae661d` by taking the distinct-encoding path. `formatOccupiedBoundary()` now writes canonical JSON even for an empty owns-nothing boundary (`{"exclusiveSubtrees":[],"sharedFiles":[]}`), so upgrade refresh no longer overwrites a precise record with the legacy `.` sentinel. The compute path also skips files outside the target boundary, and `packages/cli/architecture/features/composed-provenance/FEATURE.md` records `.` as legacy/omitted-field only with the issue #530 migration/re-derivation pointer.

packages/cli/src/provenance/boundary.ts
gs-layer: Fixed in `53c4f00a` / `36ae661d`. The schema owner now states the concrete record fields and encoding in `packages/cli/architecture/features/composed-provenance/FEATURE.md`, including canonical JSON, empty-boundary encoding, sorted/deduped collections, and the legacy `.` caveat. `packages/cli/architecture/features/upgrade-changeset/FEATURE.md` now also names `targetOccupiedOwnershipBoundary` in compute Output and `inst-cmp-return-changeset`, so the ChangeSet contract is documented where F14 owns it.

packages/cli/src/upgrade/types.ts
gs-layer: Fixed in `53c4f00a` / `36ae661d`. The kit mirror `ChangeSet` in `packages/cyber-pilot-kit-frontx/src/upgrade-orchestration/types.ts` now includes required `targetOccupiedOwnershipBoundary: string`, and the kit orchestration tests were updated so the mirror remains type-checked against its fixtures.

packages/cli/src/provenance/boundary.ts
gs-layer: Fixed. The PR body now names `upgrade/compute.ts`, `upgrade/types.ts`, the boundary-refresh-on-upgrade behavior, and the scoped artifact paths after the `develop` merge. I also removed the stale `adapters/provenance-io.ts` scope and replaced the lossy path-collapse wording with the full canonical boundary-structure wording.

packages/cli/src/__tests__/provenance-boundary.test.ts
gs-layer: Fixed in `53c4f00a` / `36ae661d`. `packages/cli/src/__tests__/provenance-boundary.test.ts` now opens with `// @cpt-dod:cpt-frontx-dod-composed-provenance-provenance-at-scaffold:p1`, after the schema paragraph was added to the composed-provenance FEATURE.

packages/cli/src/upgrade/apply.ts
gs-layer: Fixed in `53c4f00a` / `36ae661d`. The returned message now says `Apply failed; attempted restore from pre-upgrade snapshot: ...`, so it no longer promises restoration as a completed fact when the restore writes are best-effort.

packages/cli/src/upgrade/compute.ts
```diff
@@ -31,12 +32,15 @@ export type ComputeResult =
...
   path: string,
   boundary: OwnershipBoundary,
-): { mergeStrategy: string; ownedRegions: string[] } {
+): { mergeStrategy: string; ownedRegions: string[] } | undefined {
   const sharedEntry = boundary.sharedFiles.find((entry) => entry.path === path);
-  if (!sharedEntry) {
+  if (sharedEntry) {
+    return { mergeStrategy: sharedEntry.mergeStrategy, ownedRegions: sharedEntry.ownedRegions };
+  }
+  if (boundary.exclusiveSubtrees.some((subtree) => path.startsWith(subtree))) {
```
itechmeat, line 40: 🟠 Both existing containment checks normalize the separator and allow an exact match (assembler.ts builds `subtree.endsWith('/') ? subtree : subtree + '/'` and tests `path === subtree || path.startsWith(prefix)`; add-template.ts is the same shape) - this new gate is a bare startsWith. A manifest may declare a subtree without the trailing slash (validation allows it, template-shell's own manifest mixes both styles), so a template declaring `exclusiveSubtrees: ['src']` while shipping `src-app/vite-env.d.ts` gets that file pulled into the change set and written by apply - a path the assembler never contributed and the boundary never claimed, which is exactly what this PR's own new AC forbids. Reusing the existing predicate (extracted to one shared home) instead of a third implementation would close it; the comment above the function also still describes the old default-to-exclusive behaviour. A test with a no-trailing-slash subtree plus a sibling-prefix path would pin it.

packages/cli/architecture/features/composed-provenance/FEATURE.md
```diff
@@ -135,16 +135,18 @@ Recorded debt — occupied ownership boundary: the shipped writer fills each rec
...
 
+Encoding (owned here as the concrete schema owner): `occupiedOwnershipBoundary` is a string field for backward compatibility. Current writers store canonical JSON for every resolved boundary, including an empty owns-nothing boundary: `{"exclusiveSubtrees":[],"sharedFiles":[]}`. Non-empty canonical JSON has `exclusiveSubtrees` deduplicated and lexically sorted, and `sharedFiles` sorted by path, merge strategy, and owned-region list; each `ownedRegions` list is also deduplicated and sorted. The legacy `.` value remains readable only as a pre-schema or omitted-field sentinel and is not written for a resolved empty boundary, because an empty boundary owns no files while `.` is ambiguous with a historical whole-repository placeholder. Repositories that already carry `.` require an explicit migration or re-derivation pass before that value can be interpreted precisely; that migration remains tracked by issue #530 and is outside this feature's current write contract.
```
itechmeat, line 142: 🟠 Four lines above this new paragraph, line 38 of the same file still carries the "Recorded debt - occupied ownership boundary" note: "the shipped writer fills each record's ... field with the whole-repository placeholder `.` because no caller yet threads the assembly's declared boundaries into the records ... This feature's status stays unchecked until records carry each template's actual occupied boundary and upgrade diffs scope to it". Every clause of it is false at head - this PR is exactly the work that debt was gated on - so the owning FEATURE now makes two mutually exclusive statements of the write contract in one section, and the feature-status checkbox at line 24 stays unchecked for a reason that no longer exists. Rewriting the debt paragraph in place (and deciding the checkbox) is the same discipline this round already applied to upgrade-changeset.

packages/cli/src/upgrade/compute.ts
```diff
@@ -126,7 +130,9 @@ export async function computeChangeSet(
 
   // @cpt-begin:cpt-frontx-algo-upgrade-changeset-compute:p1:inst-cmp-for-each-file
   for (const filePath of allPaths) {
-    const { mergeStrategy, ownedRegions } = ownershipForPath(filePath, ownershipBoundary);
+    const ownership = ownershipForPath(filePath, ownershipBoundary);
+    if (!ownership) continue;
```
itechmeat, line 134: 🟡 [Not a blocker] The skip itself is a real fix - compute reads template content unfiltered, unlike the assembler, so the old default-to-exclusive genuinely produced entries for undeclared paths. But one direction used to be correct and changed: a path present in baseline and absent from target used to yield a remove entry regardless of boundary, so a template dropping `scripts/` from its boundary and content got `scripts/*` cleaned up on upgrade - now such paths fail the target-boundary gate and are silently skipped, leaving permanently unowned files behind. Neither the FEATURE nor the AC describes the boundary-shrink case, and the only test of the skip is (j) with a fully empty target boundary (whose fixture ships content under an empty boundary - a combination the self-containment validator would reject in reality); there is no test for an out-of-boundary path under a non-empty boundary. Worth one FEATURE sentence deciding the orphan semantics plus a pinning test.

packages/cli/architecture/features/upgrade-changeset/FEATURE.md
```diff
@@ -101,40 +102,40 @@ Internal system functions and procedures that do not interact with actors direct
...
 3. [x] - `p1` - **CATCH** application error: - `inst-app-catch`
    1. [x] - `p1` - Restore `target` all affected files from the pre-upgrade snapshot, leaving the project byte-for-byte unchanged - `inst-app-restore-on-error`
-   2. [x] - `p1` - Report the error and **RETURN** failure without updating provenance - `inst-app-return-failure`
-4. [x] - `p1` - Update `target` the selected applied template's provenance record to the newer version - `inst-app-update-prov`
-5. [x] - `p1` - Retain `target` the pre-upgrade snapshot for rollback until the developer explicitly releases it or a new upgrade cycle begins - `inst-app-retain-snapshot`
-6. [x] - `p1` - **RETURN** success: applied entries, updated provenance, rollback available - `inst-app-return-success`
+   2. [x] - `p1` - Report the error and **RETURN** failure after restoring the pre-upgrade snapshot; if the provenance write was attempted and then failed, the provenance file is restored from the same snapshot rather than left partially updated - `inst-app-return-failure`
```
itechmeat, line 136: 🟡 [Not a blocker] The restore loop swallows its own failures (best-effort catch in apply.ts) - which is exactly why this same commit softened the message to "attempted restore from pre-upgrade snapshot". But step 3.2 and the apply DoD, rewritten in the same commit, assert the strong form ("the provenance file is restored ... rather than left partially updated"; "restores the pre-upgrade project files and provenance bytes"), while the error scenario added alongside gets it right with "attempts to restore". The happy path is honestly pinned by test (o); only the double-fault wording over-claims - aligning 3.2 and the DoD with the scenario's "attempts" closes it.

packages/cli/src/upgrade/compute.ts
```diff
@@ -247,6 +253,7 @@ export async function computeChangeSet(
       templateIdentity,
       baselineVersion: scaffoldedFromVersion,
       targetVersion,
+      targetOccupiedOwnershipBoundary: formatOccupiedBoundary(ownershipBoundary),
```
itechmeat, line 256: 🟡 [Not a blocker] At head the field is written by materialize and apply, type-validated and asserted in tests - but no production line reads it: compute scopes the diff to the freshly resolved target manifest, not the record. Meanwhile composed-provenance FEATURE line 140 promises "later upgrade reads the occupied boundary", and compute's step 1 (untouched while its neighbours were rewritten) still says the record's occupied ownership boundary is extracted, though the destructuring takes only identity/version/sourceSpec. Either the two sentences say what the code does - the diff scopes to the target's declared boundary, the record is a snapshot for a future reader - or the field gains its reader.

packages/cli/src/provenance/boundary.ts
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Serializes the declared/staged `OwnershipBoundary` into the single string
+ * `ProvenanceRecord.occupiedOwnershipBoundary` holds. Boundaries use canonical
+ * JSON so shared-file region ownership stays durable while the provenance
+ * schema remains backward-compatible with a string field. `.` remains only the
+ * legacy/omitted-field sentinel filled by `provenance/write.ts`.
+ */
+export function formatOccupiedBoundary(boundary: OwnershipBoundary): string {
+  const canonicalBoundary: OwnershipBoundary = {
+    exclusiveSubtrees: [...new Set(boundary.exclusiveSubtrees)].sort(compareStrings),
```
itechmeat, line 19: 🔵 [Not a blocker] Refining the unanswered round-1 thread, including a correction of my own half of it: for the two string arrays `.sort(compareStrings)` is byte-identical to bare `.sort()`, so the argument is redundant there - but the helper itself is genuinely needed by the sharedFiles object comparator, so "delete the helper" was the wrong ask. And the round-1 claim that the tiebreak orders an impossible case was inaccurate: manifest validation never checks sharedFiles path uniqueness, a duplicate passes, so the tiebreak is reachable and reasonable (and `ownershipForPath`'s `.find` silently takes the first duplicate - an argument for a validation rule, not against the sort).

packages/cli/src/__tests__/composition.test.ts
```diff
@@ -22,6 +23,8 @@ const readProjectFileFn: ReadProjectFileFn = async () => null;
 
 const NO_BOUNDARY: OwnershipBoundary = { exclusiveSubtrees: [], sharedFiles: [] };
 
+const occupiedBoundary = (boundary: OwnershipBoundary): string => formatOccupiedBoundary(boundary);
```
itechmeat, line 26: ⚪ [Not a blocker] The round-1 ask was to use the canonical serializer - it is used now, but through a kept one-line passthrough alias. The six call sites can point at formatOccupiedBoundary directly and the line can go.

packages/cli/src/scaffold/materialize.ts
itechmeat, line 298: ⚪ [Not a blocker] Carrying the second unanswered round-1 thread, unchanged at head: the multi-line bug-history comment at the moved write in apply.ts (lines 181-193), and materialize.ts's "not the '.' every record fell back to when this field was left unset" - which after merge describes a state no reader will have seen. The package's precedent is verbose comments, so up to you - the retrospective phrasing is the minimum.

itechmeat (changes requested):
🟠 Round 2 closes out well: all three round-1 blockers were addressed for real - the FEATURE moved with the code, the empty boundary got its distinct encoding pinned by tests at both levels, the schema now lives where the FEATURE says it does, and the kit mirror is typed again. Two new things keep it one step from approve, both cheap: the new boundary gate in compute uses a bare startsWith while both existing containment checks normalize the separator, so a declared 'src' matches 'src-app/...' and an out-of-boundary file can ride into the change set against the PR's own new acceptance criterion; and the FEATURE's old "Recorded debt" paragraph now contradicts the Encoding paragraph added a few lines below it. The rest is wording-level follow-through.

PR #558, opened 2026-08-10
fix: align template-mfe with the shell's react 19.2 types and hooks-lint rules

tscbmstubp:
## Summary

After #557 fixed the lockfile drift, #491's Template Validate advanced to the composed shell+overlay stage and hit two more develop-side drifts in template-mfe (nothing on develop runs the composed checks, so both were silent):

1. **Mixed `@types/react`** — 3ea1ee93 moved template-shell to `@types/react` 19.2.18 (needed for `useEffectEvent`) while the four MFE fixture manifests still pinned 19.0.8. Composed, both copies install and `_blank-mfe`'s type-check dies in `button.tsx` (TS2786/TS2322: Radix Slot typed against one copy, the ref against the other) — [failing run](https://github.com/constructorfabric/gears-frontx/actions/runs/31365633043/job/93383280793). All four manifests plus the template-mfe lockfile move to 19.2.18.

2. **react-hooks 7.1.1 lint violations** — the composed eslint resolves `eslint-plugin-react-hooks` upward from the repo root, and its recommended set flags the MFE screens' bridge-resync pattern (`set-state-in-effect`) — the same violations #491 fixed in the shell's own hooks and deferred for the five MFE screens. Fixed with the shell's own patterns:
   - Five screens (HomeScreen, HelloWorldScreen, CurrentThemeScreen, ProfileScreen, UIKitElementsScreen): initial theme/language via lazy `useState` initializers (first paint now reflects the host's shared properties); effects reduced to subscriptions. UIKitElementsScreen keeps its initial text-direction DOM mutation in the effect.
   - ProfileDetailsCard, date-picker: "adjusting state during render" (react.dev) with a prev-key comparison instead of a sync-with-props effect.
   - data-table: targeted `eslint-disable` for `react-hooks/incompatible-library` (TanStack Table's `useReactTable` is documented as uncompilable; composed lint runs `--max-warnings 0`).

## Verification

Replicated the CI job's composition locally (copy overlay subtrees into the shell, drop nested lockfiles, `npm install`):
- `npm run build`, `type-check`, and lint over `src-app/mfe_packages` are clean
- `test:unit` passes 627 tests across all workspaces including the four MFEs
- `npm ci --dry-run` passes in template-mfe; `policy:version-check` / `policy:template-pin-drift` green

Unblocks the composed half of Template Validate on #491 (together with #557's already-merged standalone half).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved initial theme and language handling across screens, with reliable defaults when settings are unavailable.
  * Ensured theme, language, date selections, and profile details update promptly and consistently.
  * Improved form reset behavior when switching users or leaving edit mode.
  * Synchronized text direction correctly when the selected language changes.

* **Maintenance**
  * Updated React type definitions.
  * Improved compatibility with React performance tooling.


tscbmstubp:
Addressed in 4df73e90. One note on the framing: the four screens never applied the initial direction — the dir write only ever lived in the subscription callbacks, so this was a pre-existing gap rather than something this PR removed. But with `language` now correct from the first render the fix is natural: each screen gets one effect keyed on the `language` state that applies the shadow-host `dir` after commit, covering the initial value and every later change. That also subsumes the per-callback dir duplication (callbacks shrink to `setLanguage`) and collapses UIKitElementsScreen's separate initial-dir bridge read into the same shape. Verified on the composed shell+overlay tree: type-check and lint clean, all unit suites pass.

template-mfe/package-lock.json
```diff
@@ -2970,9 +2970,9 @@
       "license": "MIT"
     },
     "node_modules/@types/react": {
-      "version": "19.0.8",
-      "resolved": "https://registry.npmjs.org/@types/react/-/react-19.0.8.tgz",
-      "integrity": "sha512-9P/o1IGdfmQxrujGbIMDyYaaCykhLKc0NGCtYcECNUr9UAaDe4gwvV9bR6tvd5Br1SG0j+PBpbKr2UYY8CwqSw==",
+      "version": "19.2.18",
```
itechmeat, line 2973: 🟡 [Not a blocker] The `@types/react` entry has its version, resolved URL and integrity swapped, but its `dependencies` block still reads `"csstype": "^3.0.2"` - the range 19.0.8 declared. Published 19.2.18 declares `^3.2.2`, so the entry no longer matches what npm would write, and the next real install will rewrite it as surprise churn (harmless today, since csstype resolves to 3.2.3 and satisfies both). template-shell's lockfile carries the same leftover from 3ea1ee93 while the root lockfile shows the regenerated shape, so regenerating both would close it for good.

template-mfe/src-app/mfe_packages/demo-mfe/src/screens/helloworld/HelloWorldScreen.tsx
```diff
@@ -37,23 +37,22 @@ const languageModules = import.meta.glob('./i18n/*.json') as Record<
  */
 export const HelloWorldScreen: React.FC<HelloWorldScreenProps> = ({ bridge }) => {
   const containerRef = useRef<HTMLDivElement>(null);
-  const [theme, setTheme] = useState<string>('default');
-  const [language, setLanguage] = useState<string>('en');
+  // Initial values are read lazily during the first render so the first paint
+  // already reflects the host's shared properties; the effect below only
+  // subscribes to changes (setState inside an effect body is a lint error).
+  const [theme, setTheme] = useState<string>(() => {
```
itechmeat, line 43: 🟡 [Not a blocker] `@gears-frontx/react`, already imported here, ships `useSharedProperty` for exactly this "read the current value, then subscribe" job, built on `useSyncExternalStore`. Beyond the line count there is a real difference: the read now happens during render while the subscription starts after commit, so a property update arriving in that window is dropped, where the previous code had no window because the read and the subscribe were adjacent statements. It is not a drop-in - that hook needs MfeProvider context, which template-mfe never sets up - but a local hook of the same shape in `src/shared/`, next to useScreenTranslations, would give both.

template-mfe/src-app/mfe_packages/_blank-mfe/src/screens/home/HomeScreen.tsx
```diff
@@ -102,6 +95,17 @@ export const HomeScreen: React.FC<HomeScreenProps> = ({ bridge }) => {
     };
   }, [bridge]);
 
+  // Keep the shadow host's text direction in sync with the language state —
+  // including the initial language read during render. A DOM host mutation has
+  // to wait for commit, so this lives in an effect rather than the render body.
+  useEffect(() => {
+    const rootNode = containerRef.current?.getRootNode();
+    if (rootNode && 'host' in rootNode) {
+      const rtlLanguages = ['ar', 'he', 'fa', 'ur'];
```
itechmeat, line 104: 🟡 [Not a blocker] The RTL list and the `'rtl'`/`'ltr'` literals are now written a fifth time, along with the same six-line and three-line comments, in five screens across two packages. Text direction is knowledge `@gears-frontx/i18n` owns - `SUPPORTED_LANGUAGES` carries a `direction` per language and `TextDirection`/`getLanguageMetadata` are re-exported from `@gears-frontx/react`, which every one of these files already imports. The sets agree today (ar, he, fa, ur), so nothing misbehaves, but these screens' own tests already import `TextDirection` from the package, so the code and its test disagree about where the constant lives.

template-mfe/src-app/mfe_packages/demo-mfe/src/screens/uikit/UIKitElementsScreen.tsx
```diff
@@ -123,6 +110,17 @@ export const UIKitElementsScreen: React.FC<UIKitElementsScreenProps> = ({ bridge
     };
   }, [bridge]);
 
+  // Keep the shadow host's text direction in sync with the language state —
```
itechmeat, line 114: 🔵 [Not a blocker] Line 91, just above the effect this PR rewrites, still says "Handle theme subscription; language + direction come from useScreenTranslations". That effect subscribes to language as well, direction is now the new effect's job right below it, and useScreenTranslations only loads translation JSON - it never touches `dir` or the screen's language state. Worth correcting while the effect is being edited.

template-mfe/src-app/mfe_packages/demo-mfe/src/screens/uikit/UIKitElementsScreen.tsx
```diff
@@ -123,6 +110,17 @@ export const UIKitElementsScreen: React.FC<UIKitElementsScreenProps> = ({ bridge
...
+  // including the initial language read during render. A DOM host mutation has
+  // to wait for commit, so this lives in an effect rather than the render body.
+  useEffect(() => {
```
itechmeat, line 117: 🔵 [Not a blocker] The description says "UIKitElementsScreen keeps its initial text-direction DOM mutation in the effect", which was true at 150c5613 but not after 4df73e90 removed exactly that block and moved the screen onto the shared language-keyed effect like the other four. The comment thread explains it, but the body is what a reviewer reads first and what the merge history keeps.

template-mfe/src-app/mfe_packages/_blank-mfe/src/screens/home/HomeScreen.tsx
```diff
@@ -102,6 +95,17 @@ export const HomeScreen: React.FC<HomeScreenProps> = ({ bridge }) => {
     };
   }, [bridge]);
 
+  // Keep the shadow host's text direction in sync with the language state —
+  // including the initial language read during render. A DOM host mutation has
+  // to wait for commit, so this lives in an effect rather than the render body.
+  useEffect(() => {
```
itechmeat, line 101: 🔵 [Not a blocker] All five screens have suites and none of the thirteen files here touches them, while two new properties landed: the first render already carries the host's values, and direction is applied at mount rather than only on a later language change. Only HomeScreen's test mounts into a real shadow root, so it is the one place the mount-time write is observable, and it does not assert it; the other four install the getRootNode stub after render, so they cannot see that pass at all. One assertion - an initial `ar` yielding `dir="rtl"` before any event - would pin what the review thread asked for.

template-mfe/src-app/mfe_packages/demo-mfe/src/components/ui/data-table/data-table.tsx
```diff
@@ -57,6 +57,10 @@ export function DataTable<TData, TValue>({
   const [columnVisibility, setColumnVisibility] = React.useState<VisibilityState>({})
   const [rowSelection, setRowSelection] = React.useState({})
 
+  // TanStack Table's useReactTable returns functions the React Compiler cannot
+  // memoize safely; it skips this component, which is the documented expectation
+  // for this library.
+  // eslint-disable-next-line react-hooks/incompatible-library
```
itechmeat, line 63: 🔵 [Not a blocker] `react-hooks/incompatible-library` only exists in the plugin's v7 line, and neither template pins it: template-shell/eslint.config.js imports `eslint-plugin-react-hooks` while template-shell/package.json does not list it at all, so the only pin is the repo root's 7.1.1, reached by resolving upward from the composed tree. Where the rule is not defined, ESLint reports "Definition for rule ... was not found" as an error rather than ignoring the directive, so in a tree resolving an older plugin this line becomes a lint failure itself. Where should the plugin version come from for a scaffolded project?

template-mfe/src-app/mfe_packages/_blank-mfe/package.json
```diff
@@ -26,7 +26,7 @@
   "devDependencies": {
     "@vitejs/plugin-react": "4.3.4",
     "@module-federation/vite": "^1.14.2",
-    "@types/react": "19.0.8",
+    "@types/react": "19.2.18",
```
itechmeat, line 29: ⚪ [Not a blocker] This fixes the instance; the class stays open. `check-test-dependency-versions.mjs` skips the `template-shell` and `template-mfe` directories outright and does not cover `@types/*`, and `template-pin-drift-check.mjs` only governs `@gears-frontx`-scope names, so the sole guard for third-party pins is #491's composed type-check - which catches only drift that breaks types. It is live right now: `@types/react-dom` stays 19.0.3 in both templates against a 19.2.4 react-dom runtime, and packages/ui-kit sits at `@types/react` 19.2.17. Worth a follow-up so the next one is not found by a red composed job.

template-mfe/src-app/mfe_packages/demo-mfe/src/components/ui/date-picker.tsx
```diff
@@ -193,10 +193,14 @@ function DatePickerInput({
   const [inputValue, setInputValue] = React.useState(date ? formatDate(date) : "")
   const [month, setMonth] = React.useState<Date | undefined>(date)
 
-  // Sync input value when date changes externally
-  React.useEffect(() => {
+  // Sync input value when the date changes externally, using the "adjusting
+  // state during render" pattern (react.dev) instead of an effect: the synced
+  // value lands in the same render pass instead of one commit later.
+  const [prevSyncKey, setPrevSyncKey] = React.useState({ date, formatDate })
```
itechmeat, line 199: ⚪ [Not a blocker] The pattern terminates: the render-phase write re-renders this component, the key then matches, and the second pass stops - and the default `formatDate` is a module-level function, so identity is stable. One thing worth knowing: if a consumer passes an inline `formatDate` to `DatePicker`, the key differs on every parent render and each one costs an extra pass. The old effect behaved the same way, so this is not a regression, just a sharper sensitivity to identity than before.

itechmeat (approved):
Verified by running rather than reading: the react-hooks 7.1.1 rule set reports 8 problems across the MFE tree at the merge base and zero at head, for both .tsx and .ts, with template-shell clean too, so the composed lint really is unblocked. The type pins line up everywhere now, the lockfile integrity hash matches the registry, and the behaviour swap holds up on inspection - getProperty is a pure map read, and the dropped re-read is not a regression because the MFE lifecycle remounts rather than re-rendering with a new bridge. Nothing here blocks merge. The notes below are about what keeps this from drifting back: five screens now carry the same twenty hand-written lines that a platform primitive already covers, the lockfile entry looks hand-patched rather than regenerated, and the new first-paint behaviour has no test pinning it.

gs-layer (approved):
Approved. I rechecked the current head and CI: the React type alignment and hook-lint changes address the composed template-mfe failures, the CodeRabbit RTL-direction concern is fixed by the second commit, and the remaining notes are non-blocking follow-ups. This PR should merge before #491 because it unblocks #491's current Template Validate failure; #558 itself does not need to wait for another PR.

template-mfe/src-app/mfe_packages/demo-mfe/src/components/ui/data-table/data-table.tsx
tscbmstubp: Good catch — as written, the disable only works where resolution reaches the repo root's 7.1.1. The pin should live in template-shell's own `package.json`: `eslint.config.js` there already imports `eslint-plugin-react-hooks`, so the plugin belongs in its devDependencies (pinned to the v7 line, matching the root), and a scaffolded project then resolves v7 locally instead of by accident of hoisting. That's a template-shell manifest + lockfile change, so I'd rather land it as a follow-up together with the third-party-pin guard gap you noted on `_blank-mfe/package.json` than widen this PR beyond template-mfe.

template-mfe/src-app/mfe_packages/demo-mfe/src/screens/uikit/UIKitElementsScreen.tsx
tscbmstubp: Fixed in 389c42be — the comment now says the effect subscribes to later theme and language changes, with direction applied by the language-keyed effect below.

GeraBart (approved):

PR #562, opened 2026-08-12
fix: drop the concrete Default* exports from the mfes barrel behind the ADR-0003 import-boundary guard (#540)

### LINKED ISSUE #540: [ARCH] mfes public barrel exports DefaultMfeRegistry, contradicting ADR-0003's internal-implementation decision
> [!NOTE]
> **Status update (2026-08-12):** the factory half of this issue has since been fixed on `develop` — `6a50d21d` added `createMfeRegistryFactory()` as the single creation path (concrete factory no longer exported) and migrated template-shell's `registry.ts` to it; `546d137e` followed up on the factory's cache keying. Steps 2 and 3 below remain open: the barrel still exports `DefaultMfeRegistry`, `DefaultLifecycleManager`, and the sibling `Default*` internals (now with zero external consumers), and the ADR-0003 import-boundary CI check still does not exist. The body below is updated accordingly; original text struck through where overtaken.

## Problem

`packages/mfes/src/index.ts` exports the concrete `DefaultMfeRegistry` from the public barrel (annotated "Stage 1 extraction"). ADR-0003 (`mfe-runtime-public-surface`) chose the abstract-facade-with-factory-injection surface: consumers depend on the abstract `MfeRegistry` and build one via `createMfeRegistryFactory().build({ typeSystem })`, and "the concrete factory and registry implementation are internal and are not part of the package's public surface."

The drift is wider than the registry class:

- ~~**The factory side does not honor the decision either.** `createMfeRegistryFactory()` does not exist anywhere in `src/` — it survives only in stale `dist/` typedoc comments. The barrel exports the concrete `DefaultMfeRegistryFactory` directly (`index.ts:146`), and template-shell already couples to it: `template-shell/packages/framework/src/mfe/registry.ts:19-21` imports it from the barrel and does `new DefaultMfeRegistryFactory()` — exactly the coupling ADR-0003's confirmation section says must not happen.~~ **Fixed by `6a50d21d`:** `createMfeRegistryFactory()` now exists, is the only creation path across the barrel, and template-shell's `registry.ts` uses it.
- **The CI check ADR-0003 promised was never built.** The confirmation section calls for "an import-boundary grep" verifying no consumer imports the concrete classes; nothing in `scripts/` or `.github/` implements it, which is why the drift went unnoticed.
- **The sibling `Default*` exports share the disease.** The barrel also ships `DefaultLifecycleManager`, `DefaultMountManager`, `DefaultExtensionManager`, `DefaultRuntimeBridgeFactory`, `DefaultExtensionMounter`, `DefaultDomainLifecycleTrigger`, `DefaultMfeStateContainer`, and `MfeBridgeFactoryDefault` — and ADR-0003 explicitly names mount management, the actions-chains mediator, and bridge wiring as internals that "must not be part of the public surface."

## Why now

None of the concrete `Default*` exports has an external consumer anymore (verified 2026-08-12: template-shell imports only `createMfeRegistryFactory`, `MfeHandlerMF`, and types from the barrel; the only outside `DefaultMfeRegistry` reference is a comment in a template-shell test fixture), and the package is pre-1.0/unstable — removal is nearly free today. ~~But the window is already closing: template-shell's direct `new DefaultMfeRegistryFactory()` is the first real consumer of a concrete class, so the fix is no longer "delete two lines" — the factory function has to be added first.~~ The factory-coupling risk that made this urgent has been defused by `6a50d21d`, but every release that ships the concrete exports invites a new consumer and reopens it.

## Fix

1. ~~**Restore the creation function.** Add `createMfeRegistryFactory(): MfeRegistryFactory` to the mfes package (the shape the dist docs already advertise), export it from the barrel, and migrate the single call site in template-shell's `registry.ts` to it.~~ ✅ Done in `6a50d21d` (+ `546d137e` cache-keying follow-up).
2. **Drop the concrete exports.** Remove `DefaultMfeRegistry` and `DefaultLifecycleManager` from the barrel (no consumers remain). Audit the remaining `Default*` exports: remove the unused ones, and keep anything template-shell legitimately needs (mount strategies, `MfeHandlerMF`) either as sanctioned public surface or as recorded debt with an objective removal criterion (not "temporary").
3. **Add the CI check ADR-0003 promised.** A small script in `scripts/` (following the existing pattern, e.g. `template-pin-drift-check.mjs`) that fails when code outside `packages/mfes` imports the concrete runtime classes from `@gears-frontx/mfes`.

## Acceptance

The barrel and ADR-0003 agree — the concrete registry/lifecycle/coordination exports are gone (consumers use `createMfeRegistryFactory()`), any remaining concrete exports are recorded debt with a removal criterion, and the import-boundary check from ADR-0003's confirmation section exists in CI so the drift cannot recur.

tscbmstubp:
Implements steps 2 and 3 of #540 (step 1 — `createMfeRegistryFactory()` — already landed in `6a50d21d`).

## Step 2 — drop the concrete exports

The barrel no longer ships the concrete coordination machinery ADR-0003 declared internal. Removed (all verified consumer-less repo-wide): `DefaultMfeRegistry`, `DefaultLifecycleManager`, `DefaultActionsChainsMediator` (its `NoHandlerForActionTargetError` error class stays public), `DefaultExtensionManager`, `DefaultMountManager` (+ its `HandlerResolver` type), `DefaultRuntimeBridgeFactory`, `DefaultExtensionMounter`, `DefaultDomainLifecycleTrigger`, `DefaultMfeStateContainer`, and `MfeBridgeFactoryDefault`.

`MfeHandlerMF` stays as sanctioned public surface, with a barrel comment saying why: hosts construct it and pass it to the registry, and it wires its own bridge factory internally — nothing forces `MfeBridgeFactoryDefault` across the boundary.

## Step 3 — the CI check ADR-0003 promised

`scripts/mfes-import-boundary-check.mjs` (+ 61 unit tests under the `repo-scripts` vitest project), following the `template-pin-drift-check.mjs` house pattern. The forbidden set is a **naming rule** (`Default`-prefixed/suffixed, plus an explicit denylist seeded for the debt register), not a class list, so the next `DefaultFooManager` is born guarded. Both halves of the boundary are checked:

- the barrel must not export a concrete-implementation name (value or `type`, re-exported or declared directly in the barrel);
- no file outside `packages/mfes` may import a concrete name from `@gears-frontx/mfes`, nor deep-import any subpath past the barrel. Aliases don't launder names, and `import type` counts.

Comments are stripped before matching in both halves, so a comment can neither launder a concrete name past the rule nor be mistaken for the statement's own `import` keyword. The strip is string-, template-, and regex-literal aware (round 4): a `/*` inside an ordinary string — a glob like `'src/**'`, which exists in real files in this repo — cannot open a strip that silently deletes the code between it and the next `*/`. `parseImportedNames` routes through the same shared `stripComments` rather than carrying a second copy. The barrel's direct-declaration sweep also covers `async`/generator/`default`/mid-line declarations now — `export default class DefaultX {}` matters most, since a default import has no brace block for the consumer half to see.

Zero scanned files fails closed rather than passing vacuously. Wired as `npm run policy:mfes-import-boundary` and a step in `main.yml` next to the other policy checks.

## Docs kept in agreement

ADR-0003's confirmation section now names the actual script and the `createMfeRegistryFactory().build({ typeSystem })` path; the mfes `DESIGN.md` public-surface table no longer lists the concrete classes.

## Verification

- guard passes against the repo (689 consumer files scanned — down from the originally reported 817 exactly by the dot-directory exclusion), its 61 tests pass
- mfes: type-check (src + test configs), 68 unit tests, and the tsup build with DTS all pass
- pre-existing on clean develop, unrelated (confirmed by stash-and-rerun): `validate-templates.test.mjs` failures (stale local `packages/cli/dist`) and gts-plugin type-check (missing local workspace links)

**Deliberately out of scope** (now recorded, not just flagged): the barrel still exports nine concrete-but-not-`Default`-named internals — `ChildMfeBridgeImpl`/`ParentMfeBridgeImpl`, `ChildDomainForwardingHandler`, `OperationSerializer`, `LoadExtHandler`, `WeakMapRuntimeCoordinator`, `InvalidatableDomainContext`, `LruCache`, `RetryHandler`. The naming rule wouldn't catch them; each is carried as debt in the mfes DESIGN public-surface debt register (owner + removal criterion), and the guard seeds `CONCRETE_EXPORT_DENYLIST` so retiring one is a one-line append.

**Also out of scope**: the version bump. mfes stays `0.3.0-alpha.2` here; the `alpha.3` bump rides the separate publish step per house practice.

Closes #540

## Summary by CodeRabbit

* **New Features**
  * MFES now provides a streamlined factory-based registry creation API.
  * The supported public API emphasizes stable contracts and approved runtime integrations.

* **Documentation**
  * Updated architecture and registration guidance to reflect the current MFES API and implementation boundaries.

* **Tests**
  * Added automated checks covering supported imports, exports, file scanning, exclusions, and command-line validation.


architecture/ADR/0003-mfe-runtime-public-surface.md
```diff
@@ -57,7 +57,7 @@ Chosen option: **abstract registry facade with factory injection**, because it i
 
 ### Confirmation
 
-Architecture review confirms the package's public surface exports only the abstract `MfeRegistry` contract and the `mfeRegistryFactory`, and that the concrete implementation and factory remain internal. A continuous-integration check (an import-boundary grep) confirms that no consumer outside the runtime package imports the concrete registry or factory classes, and that every instance is obtained through `mfeRegistryFactory.build({ typeSystem })`.
+Architecture review confirms the package's public surface exports only the abstract `MfeRegistry` contract and the `createMfeRegistryFactory` creation function, and that the concrete implementation and factory remain internal. A continuous-integration check (`scripts/mfes-import-boundary-check.mjs`, wired as `npm run policy:mfes-import-boundary` in `.github/workflows/main.yml`) confirms both halves of the boundary: the barrel exports no concrete-implementation name (the `Default*` naming rule), and no consumer outside the runtime package imports a concrete class or deep-imports past the barrel — every instance is obtained through `createMfeRegistryFactory().build({ typeSystem })`.
```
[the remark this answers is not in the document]
tscbmstubp: Fixed in eee4054c — the confirmation sentence now scopes the 'only' claim to registry construction ("registry construction crosses the package's public surface only as…") instead of the package's whole surface.

packages/mfes/architecture/DESIGN.md
```diff
@@ -223,17 +223,17 @@ Applications need to gain user-facing functionality from independently developed
...
 | `createShadowRoot`, `injectCssVariables`, `injectStylesheet` | Shadow-DOM mount utilities used by handlers to isolate rendered output. |
-| `ExtensionManager`, `DefaultExtensionManager`, `MountManager`, `DefaultMountManager`, `LifecycleManager`, `DefaultLifecycleManager`, `RuntimeBridgeFactory`, `DefaultRuntimeBridgeFactory`, `OperationSerializer`, `WeakMapRuntimeCoordinator`, `MfeStateContainer`, `DefaultMfeStateContainer` | The internal coordination machinery behind the facade — governance, mounting, lifecycle orchestration, bridge construction, and state, exported for handler and extension authors. |
+| `ExtensionManager`, `MountManager`, `LifecycleManager`, `RuntimeBridgeFactory`, `OperationSerializer`, `WeakMapRuntimeCoordinator`, `MfeStateContainer` | The coordination contracts behind the facade — governance, mounting, lifecycle orchestration, bridge construction, and state, exported for handler and extension authors. Their `Default*` implementations stay internal (ADR-0003): the registry wires them itself and nothing outside the composition root may construct a rival set. |
```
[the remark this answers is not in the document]
tscbmstubp: Fixed in eee4054c — added a public-surface debt register below the table covering `ChildMfeBridgeImpl`, `ParentMfeBridgeImpl`, `OperationSerializer`, and `WeakMapRuntimeCoordinator`: each classified as debt (not sanctioned), owner recorded (mfes maintainers, tracked on the #540 thread), with an objective removal criterion — drop from the barrel in the next pre-1.0 minor once a repo-wide scan shows zero external imports (already true for all four), adding the name to an explicit denylist in the guard in the same change. The two rows naming them now point at the register, and `MfeHandlerMF` is explicitly marked sanctioned.

scripts/mfes-import-boundary-check.mjs
```diff
@@ -0,0 +1,302 @@
...
+    const head = content.slice(0, match.index);
+    const statementStart = Math.max(head.lastIndexOf('import'), head.lastIndexOf('export'));
+    if (statementStart === -1) continue;
+    const statement = content.slice(statementStart, match.index + match[0].length);
+    sites.push({
+      file: relativeFilePath,
+      specifier: match[1],
+      statement,
+      names: parseImportedNames(statement),
+    });
```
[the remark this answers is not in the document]
tscbmstubp: Fixed in eee4054c — a second pattern now matches from-less forms: side-effect `import '@gears-frontx/mfes/…'` AND dynamic `import('@gears-frontx/mfes/…')`, both yielding sites that hit the deep-import rule. Regression tests added for both shapes (parser-level and end-to-end violation).

scripts/mfes-import-boundary-check.mjs
```diff
@@ -0,0 +1,302 @@
...
+      if (isConcreteImplementationName(exportedName)) {
+        violations.push({
+          file: MFES_BARREL_RELATIVE_PATH,
+          kind: 'barrel-export',
+          detail: exportedName,
+        });
+      }
+    }
+  }
+  return violations;
```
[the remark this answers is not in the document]
tscbmstubp: Fixed in eee4054c — you were right, the alias exemption was a real hole. The barrel rule now flags a concrete name on EITHER side of an `as` alias (`export { DefaultMfeRegistry as MfeRegistryImpl }` is a violation), and wildcard re-exports (`export * [as ns] from`) are prohibited outright as a new `wildcard-export` violation kind, since the guard can't resolve them without becoming a module-graph walker. Tests updated: the permitted-alias expectation flipped to a violation, plus wildcard coverage for both forms.

scripts/mfes-import-boundary-check.mjs
```diff
@@ -0,0 +1,327 @@
...
+import fs from 'node:fs';
+import path from 'node:path';
+import process from 'node:process';
+import { pathToFileURL } from 'node:url';
+
+export const MFES_PACKAGE_NAME = '@gears-frontx/mfes';
+export const MFES_BARREL_RELATIVE_PATH = path.join('packages', 'mfes', 'src', 'index.ts');
+
+/** Directories never walked: generated output and third-party trees. */
+const SKIPPED_DIR_NAMES = new Set(['node_modules', 'dist', '.git', 'coverage']);
```
itechmeat: 🟠 The walk recurses from the repo root skipping only four directory names, so it scans `.claude/worktrees/` - gitignored disposable checkouts of this same repo. Reproduced on a live tree: 34,705 files instead of ~700 and exit 1 on two `DefaultMfeRegistryFactory` imports from other branches (one inside a `template-standard/` tree deleted by #470), while CI on a fresh checkout stays green - red only for developers is how a guard gets trained out of people. `.dependency-cruiser.cjs` already excludes `.claude` for exactly this reason, and once this branch lands in a worktree the guard's own test fixtures become "violations" too, since the two exclusions compare full paths against the root only. Cheapest fix: add `.claude` to the skip set, or drive discovery from `git ls-files` and get `.gitignore` for free.

packages/mfes/architecture/DESIGN.md
```diff
@@ -223,19 +223,21 @@ Applications need to gain user-facing functionality from independently developed
 
 | Public surface | Purpose |
 |----------------|---------|
-| `MfeRegistry`, `MfeRegistryFactory`, `mfeRegistryFactory`-produced `DefaultMfeRegistry` / `DefaultMfeRegistryFactory` | The abstract registry facade and its factory-with-cache builder — the sole way a consumer obtains a registry instance bound to an injected `TypeSystemPlugin`. |
+| `MfeRegistry`, `MfeRegistryFactory`, `createMfeRegistryFactory` | The abstract registry facade, its abstract factory contract, and the creation function — the sole way a consumer obtains a registry instance bound to an injected `TypeSystemPlugin`. The concrete `DefaultMfeRegistry` / `DefaultMfeRegistryFactory` stay internal (ADR-0003, `scripts/mfes-import-boundary-check.mjs`). |
```
itechmeat, line 226: 🟠 The edited row makes `createMfeRegistryFactory` the named creation path, but the same DESIGN.md still says `mfeRegistryFactory` in four places the barrel does not export - lines 49, 83 (the mermaid edge), 199 and 243, the last one three lines below this row inside the same hunk. ADR-0003 has the same half-sweep: the Confirmation now says `createMfeRegistryFactory` while lines 41, 47 and 92 still describe `mfeRegistryFactory.build({ typeSystem })`. One grep for `mfeRegistryFactory` across both docs and they stop arguing with themselves.

packages/mfes/src/index.ts
```diff
@@ -141,26 +138,15 @@ export {
   importBlobModule,
 } from './handler/mf-dynamic-module-ops';
 
-// Default registry — concrete facade implementation (Stage 1 extraction)
-export { DefaultMfeRegistry } from './runtime/DefaultMfeRegistry';
-// Only the creation function crosses the barrel: the concrete factory stays
-// internal so no consumer can build a rival registry past the composition root
+// Only the creation function crosses the barrel: the concrete registry and
```
itechmeat, line 141: 🟠 A repo-wide grep for the eleven removed names outside `packages/mfes` leaves exactly one live instruction surface: `packages/cyber-pilot-kit-frontx/AGENTS.md:34` still says "MFEs register through `@gears-frontx/mfes` DefaultMfeRegistry". An agent following it writes precisely the import the new guard fails on, and the guard cannot catch the doc since it only scans code. One line.

packages/mfes/architecture/DESIGN.md
```diff
@@ -223,19 +223,21 @@ Applications need to gain user-facing functionality from independently developed
...
 | `extractGtsPackage` | A string-parsing utility over a GTS-shaped entity identifier. It imports no type-definition specification, but its segment rules encode the GTS identifier grammar — a known tension with the format-agnosticism MFES-1/MFES-4 intend, and a candidate for relocation to the type-system provider. |
 
+**Public-surface debt register** (#540 follow-up). The barrel still exports four concrete implementations that ADR-0003's encapsulation driver would keep internal but that the `Default*` naming rule enforced by `scripts/mfes-import-boundary-check.mjs` does not detect: `ChildMfeBridgeImpl`, `ParentMfeBridgeImpl` (bridge wiring), `OperationSerializer`, and `WeakMapRuntimeCoordinator` (registry coordination). Each is debt, not sanctioned surface. Owner: the mfes package maintainers, tracked on the #540 thread. Removal criterion (objective, not "temporary"): drop an item from the barrel in the next pre-1.0 minor once a repo-wide consumer scan shows zero imports of it outside `packages/mfes` — already true for all four as of 2026-08-12 — and, in the same change, add its name to an explicit denylist in `mfes-import-boundary-check.mjs` so it cannot return.
```
itechmeat: 🟡 The PR body itself names `LoadExtHandler` among the deliberately kept concrete exports, but the register does not list it - the register and the PR description disagree within one change. `ChildDomainForwardingHandler` looks like the same class of debt (`src/bridge/index.ts`'s docblock literally calls these "internal implementations"). An undercounting register is worse than none - either add both names, or record why they are sanctioned the way `MfeHandlerMF`'s row does.

scripts/mfes-import-boundary-check.mjs
```diff
@@ -0,0 +1,327 @@
...
+ * @param {string} relativeFilePath repo-root-relative, for reporting
+ * @returns {ParsedImportSite[]}
+ */
+export function findMfesImportSites(content, relativeFilePath) {
+  /** @type {ParsedImportSite[]} */
+  const sites = [];
+  const specifierPattern = /\bfrom\s*['"](@gears-frontx\/mfes(?:\/[^'"]*)?)['"]/g;
+  for (const match of content.matchAll(specifierPattern)) {
+    const head = content.slice(0, match.index);
+    const statementStart = Math.max(head.lastIndexOf('import'), head.lastIndexOf('export'));
```
itechmeat: 🟡 [Not a blocker] `lastIndexOf('import')` matches any substring, so a comment like `// imported for typing only` inside a multi-line import block resets the statement start and the names list comes back empty - `DefaultLifecycleManager` a line below walks through (reproduced against the head functions). `import { DefaultMfeRegistry, importManifest }` slips the same way, the backtrack landing on `importManifest`. Anchoring on `import\s*(type\s*)?{` instead of backtracking fixes both - worth a regression test each, since this is the half that catches real consumers.

scripts/mfes-import-boundary-check.mjs
```diff
@@ -0,0 +1,327 @@
...
+      if (offendingName) {
+        violations.push({
+          file: MFES_BARREL_RELATIVE_PATH,
+          kind: 'barrel-export',
+          detail: offendingName,
+        });
+      }
+    }
+  }
+  for (const match of barrelContent.matchAll(/export\s*\*(?:\s*as\s+\w+)?\s*from\s*['"]([^'"]+)['"]/g)) {
```
itechmeat: 🟡 [Not a blocker] `export type * from './runtime/internals'` is not matched - the `type` sits between `export` and `*`, the same hole one level below the alias one this PR just closed. A `(?:type\s+)?` in the pattern closes it while the file is open. Same family: a `Default*` class declared directly in the barrel (`export class DefaultX {}`) is invisible too, if that ever becomes real.

scripts/mfes-import-boundary-check.mjs
```diff
@@ -0,0 +1,327 @@
...
+ * shape and is not matched.
+ *
+ * @param {string} content
+ * @param {string} relativeFilePath repo-root-relative, for reporting
+ * @returns {ParsedImportSite[]}
+ */
+export function findMfesImportSites(content, relativeFilePath) {
+  /** @type {ParsedImportSite[]} */
+  const sites = [];
+  const specifierPattern = /\bfrom\s*['"](@gears-frontx\/mfes(?:\/[^'"]*)?)['"]/g;
```
itechmeat, line 382: 🟡 [Not a blocker] eslint 10 has both halves of this rule - `no-restricted-imports` with `importNamePattern` and `no-restricted-exports` with `restrictedNamedExportsPattern` - and the shared configs already use `no-restricted-imports`. The script has a strong answer (the consumer half must cover the templates, which ecosystem CI does not lint - `scripts/verify-guard-configs.ts` says so), but the docblock currently argues rule-vs-list, not script-vs-lint-rule. One sentence there settles this permanently.

packages/mfes/architecture/DESIGN.md
itechmeat: 🔵 [Not a blocker] "Already true for all four as of 2026-08-12" and "tracked on the #540 thread" are a point-in-time observation and an external status pointer - the date starts rotting the day this lands. The removal criterion is the durable part; the issue can hold the status. Also worth considering: exclusions the implementation relies on conventionally live in the owning ADR's Consequences - a reader of ADR-0003 alone currently sees a clean boundary with no mention of the four.

scripts/mfes-import-boundary-check.mjs
```diff
@@ -0,0 +1,327 @@
...
+ * `.github/workflows/main.yml`. Core logic is exported for unit tests in
+ * `scripts/mfes-import-boundary-check.test.mjs`.
+ */
+import fs from 'node:fs';
+import path from 'node:path';
+import process from 'node:process';
+import { pathToFileURL } from 'node:url';
+
+export const MFES_PACKAGE_NAME = '@gears-frontx/mfes';
+export const MFES_BARREL_RELATIVE_PATH = path.join('packages', 'mfes', 'src', 'index.ts');
```
itechmeat, line 57: 🔵 [Not a blocker] The register's removal criterion promises to "add its name to an explicit denylist in `mfes-import-boundary-check.mjs`", but the script has no denylist mechanism today - the next person reads the doc, greps, and finds nothing. Either seed the (empty) denylist now so the follow-up is a one-line append, or reword the criterion to a mechanism that exists.

scripts/mfes-import-boundary-check.mjs
```diff
@@ -0,0 +1,327 @@
...
+
+/**
+ * Finds every `import`/`export ... from` statement in `content` whose module
+ * specifier is `@gears-frontx/mfes` or a subpath of it — including the
+ * from-less side-effect form (`import '@gears-frontx/mfes/…'`), which
+ * executes the module's top level and is a deep import like any other.
+ *
+ * Anchored on the quoted specifier and backtracked to the nearest preceding
+ * `import`/`export` keyword, so multi-line named-import blocks are captured
+ * whole. Anything else that merely mentions the package name — an eslint
```
itechmeat: 🔵 [Not a blocker] Verified against the real functions: a JSDoc example line `* import { DefaultLifecycleManager } from '@gears-frontx/mfes';` yields a concrete-name violation, and prose like `// never import from '@gears-frontx/mfes/internal/wiring'` yields a deep-import one - while the docblock guarantees prose "is not matched". Harmless today (zero such comments in consumers), but the comment states a guarantee the code does not hold. Same family: `require('@gears-frontx/mfes')` is invisible though `.cjs`/`.js` are scanned, and a commented-out `export { DefaultX }` in the barrel fails the check. Either align the docblock with what is really matched, or strip comments before matching.

packages/mfes/src/index.ts
```diff
@@ -126,13 +126,10 @@ export type { ActionChainExecutor, LifecycleTrigger } from './runtime/mount-mana
 // Runtime bridge factory (Phase 7)
 export { RuntimeBridgeFactory } from './runtime/runtime-bridge-factory';
 
-// Governance concrete classes (Phase 7)
-export { DefaultExtensionManager } from './runtime/default-extension-manager';
-export { DefaultMountManager, type HandlerResolver } from './runtime/default-mount-manager';
-
 // MFE Isolation — handler, trust kernel, types (Phase 8)
+// MfeHandlerMF is sanctioned public surface: hosts construct it and pass it to
```
itechmeat: ⚪ [Not a blocker] This is the same sentence, near word for word, as the DESIGN public-surface row and the guard's docblock - three copies of one rationale, and only one of them is the source of truth. The counterargument is fair (without a comment the next barrel cleanup deletes `MfeHandlerMF` along the way), so shrinking this to the constraint plus a pointer at DESIGN would keep the protection without the drift.

packages/mfes/src/index.ts
```diff
@@ -172,8 +158,8 @@ export type { LifecycleActionPayload } from './runtime/extension-lifecycle-actio
 // Runtime coordination — default WeakMap-based implementation
 export { WeakMapRuntimeCoordinator } from './runtime/coordination/weak-map-runtime-coordinator';
 
-// MFE state container — abstract contract and default implementation
-export { MfeStateContainer, DefaultMfeStateContainer } from './state';
+// MFE state container — abstract contract; the default implementation stays internal (ADR-0003)
+export { MfeStateContainer } from './state';
```
itechmeat, line 162: ⚪ [Not a blocker] Ten exports leave the published surface but mfes stays 0.3.0-alpha.2, and publish is version-gated - the registry copy the templates pin keeps shipping the old exports. House practice does bumps as separate commits, so this is probably deliberate - worth one line in the body though: alpha.3 here, or a separate step?

scripts/mfes-import-boundary-check.mjs
itechmeat, line 57: ⚪ [Not a blocker] The sibling policy guards make a point of discovering their scope structurally; this one names one package and one barrel path. Fair for a rule owned by one package's ADR, and the missing-barrel path fails closed if the file moves - but gts-plugin and api have barrels too, so: generalize now, or when a second owner shows up? Genuinely either way.

scripts/mfes-import-boundary-check.test.mjs
```diff
@@ -0,0 +1,267 @@
...
+  });
+
+  it('leaves the sanctioned surface alone', () => {
+    for (const name of ['MfeRegistry', 'createMfeRegistryFactory', 'MfeHandlerMF', 'ConcurrentMountStrategy', 'InvalidatableDomainContext']) {
+      expect(isConcreteImplementationName(name)).toBe(false);
+    }
+  });
+});
+
+describe('findMfesImportSites / parseImportedNames', () => {
```
itechmeat: ⚪ [Not a blocker] Two small ones in an otherwise well-titled suite: this `describe` bundles two units (its last case only exercises `parseImportedNames`), and `leaves the sanctioned surface alone` at line 62 states no condition - "leaves a contract or creation-function name alone" would read as a behaviour sentence like the rest.

itechmeat (changes requested):
The direction is right and most of the execution is exemplary - a naming rule instead of a class list, both halves of the boundary guarded, fail-closed on zero files, the sibling policy-script pattern followed to the letter, tests driving the real functions. What blocks approval is cheap but real: the guard currently exits red on any dev machine with agent worktrees while staying green in CI, and the docs this PR makes authoritative still disagree with themselves after the spot edits - a half-swept rename and an undercounted debt register. All mechanical fixes; the design itself needs no rework.

scripts/mfes-import-boundary-check.mjs
tscbmstubp: Fixed in b886da77 (previous push) — `findConsumerFiles` now skips every dot-directory, so `.claude/worktrees/*` copies (and their fixture-quoting `scripts/` trees) are never walked. Regression tests cover both the walker exclusion and a stale worktree copy carrying the pre-#540 `DefaultMfeRegistryFactory` coupling; a live-tree run is back to ~700 files and green.

packages/mfes/architecture/DESIGN.md
tscbmstubp: Fixed in b886da77 — swept both docs. `mfeRegistryFactory` now appears nowhere in ADR-0003 or the mfes DESIGN except as `createMfeRegistryFactory` or the abstract `MfeRegistryFactory` type name (verified by grep in the change).

packages/mfes/src/index.ts
tscbmstubp: Fixed in b886da77 — the kit AGENTS.md line now reads “MFEs register through the abstract `MfeRegistry` from `@gears-frontx/mfes`, obtained via `createMfeRegistryFactory().build({ typeSystem })`”. No instruction surface names a concrete class anymore.

packages/mfes/architecture/DESIGN.md
tscbmstubp: Fixed in b886da77 — the register now carries six items: `ChildMfeBridgeImpl`, `ParentMfeBridgeImpl`, `ChildDomainForwardingHandler` (bridge wiring), `OperationSerializer`, `WeakMapRuntimeCoordinator` (registry coordination), and `LoadExtHandler` (lifecycle action wiring), each verified against the barrel before listing. The PR body and the register no longer disagree.

scripts/mfes-import-boundary-check.mjs
tscbmstubp: Fixed in b886da77 — both reproduced escapes and one more you were a hair away from: the backtrack now anchors on whole-word `import`/`export` (so `importManifest` and `// imported for typing only` no longer reset it), and comments are stripped inside the brace block before the comma-split — a `//` line directly above a concrete name was gluing onto it in the split and laundering it past the naming rule. All three arrangements are exact-equality regression tests.

scripts/mfes-import-boundary-check.mjs
tscbmstubp: Fixed in b886da77 — the wildcard pattern now takes `(?:type\s+)?`, so `export type * from` is flagged; test added. A `Default*` class declared directly in the barrel stays out of scope for now: the barrel is a pure re-export file and the wildcard ban keeps it that way, so the declaration form has no way to sneak in unreviewed.

scripts/mfes-import-boundary-check.mjs
tscbmstubp: Fixed in b886da77 — the docblock now settles it in one sentence: the consumer half must cover template territory that ecosystem CI does not lint, which an eslint rule in the shared configs can never reach.

packages/mfes/architecture/DESIGN.md
tscbmstubp: Fixed in b886da77 — the point-in-time “already true as of…” claim is gone (status lives on the #540 thread; the removal criterion is what the doc keeps), and ADR-0003's Consequences now records the exclusions with a pointer at the DESIGN debt register, so a reader of the ADR alone no longer sees a clean boundary.

scripts/mfes-import-boundary-check.mjs
tscbmstubp: Fixed in b886da77 — the guard now seeds `CONCRETE_EXPORT_DENYLIST` (empty, honored by both the barrel and consumer halves, with a test that exercises it), and the register names that exact mechanism, so the follow-up really is a one-line append.

scripts/mfes-import-boundary-check.mjs
tscbmstubp: Fixed in b886da77 — the docblock now states the real contract: text with no from-clause/side-effect shape is ignored, and statement-shaped comments or strings ARE matched — a deliberate conservative fail-closed trade, documented by a test asserting a commented-out import is still reported. Comment stripping was added only inside brace blocks, where the token grammar makes it safe. The `require()` form remains unmatched and is acknowledged as out of scope here — worth its own line if a `.cjs` consumer ever appears.

packages/mfes/src/index.ts
tscbmstubp: Fixed in b886da77 — the barrel comment shrank to the constraint plus a pointer at the DESIGN public-surface table, which is now the single home of the rationale. The protection against an over-eager barrel cleanup stays.

packages/mfes/src/index.ts
tscbmstubp: Deliberate — version bumps ride their own commits in the publish flow per house practice, so mfes stays 0.3.0-alpha.2 in this change and the alpha.3 bump lands as the separate publish step. Noting it in the PR body as you suggest.

scripts/mfes-import-boundary-check.mjs
tscbmstubp: Staying single-owner for now: the rule is owned by one package's ADR, and generalizing before gts-plugin or api adopt the same boundary would be speculative structure. The missing-barrel fail-closed path protects a move; when a second owner shows up the walk generalizes then.

scripts/mfes-import-boundary-check.test.mjs
tscbmstubp: Fixed in b886da77 — the bundled describe split in two (`findMfesImportSites` and `parseImportedNames` each own their cases), and the title now reads as a behaviour sentence: “leaves contract and creation-function names alone”.

architecture/ADR/0003-mfe-runtime-public-surface.md
```diff
@@ -57,7 +57,7 @@ Chosen option: **abstract registry facade with factory injection**, because it i
 
 ### Confirmation
 
-Architecture review confirms the package's public surface exports only the abstract `MfeRegistry` contract and the `mfeRegistryFactory`, and that the concrete implementation and factory remain internal. A continuous-integration check (an import-boundary grep) confirms that no consumer outside the runtime package imports the concrete registry or factory classes, and that every instance is obtained through `mfeRegistryFactory.build({ typeSystem })`.
+Architecture review confirms that registry construction crosses the package's public surface only as the abstract `MfeRegistry` contract and the `createMfeRegistryFactory` creation function, and that the concrete implementation and factory remain internal. A continuous-integration check (`scripts/mfes-import-boundary-check.mjs`, wired as `npm run policy:mfes-import-boundary` in `.github/workflows/main.yml`) confirms both halves of the boundary: the barrel exports no concrete-implementation name (the `Default*` naming rule), and no consumer outside the runtime package imports a concrete class or deep-imports past the barrel — every instance is obtained through `createMfeRegistryFactory().build({ typeSystem })`.
```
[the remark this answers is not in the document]
tscbmstubp: Fixed in b886da77 — both remaining `mfeRegistryFactory.build({ typeSystem })` references (Considered Options and More Information) now read `createMfeRegistryFactory().build({ typeSystem })`; the ADR documents one construction path.

scripts/mfes-import-boundary-check.test.mjs
```diff
@@ -0,0 +1,295 @@
...
+  });
+
+  it('ignores mentions with no from-clause: eslint groups, bundler externals, prose', () => {
+    const content = [
+      "const groups = ['@gears-frontx/mfes', '@gears-frontx/mfes/*'];",
+      "// exported from @gears-frontx/mfes for convenience",
+      "export default { external: ['@gears-frontx/mfes'] };",
+    ].join('\n');
+    expect(findMfesImportSites(content, 'a.ts')).toEqual([]);
+  });
```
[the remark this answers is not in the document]
tscbmstubp: Fixed in b886da77 — comments are now stripped inside import/export brace blocks before name parsing (the arrangement that mattered: a `//` line directly above a concrete name laundered it past the rule; exact-equality regression test added). Outside brace blocks the guard deliberately keeps matching statement-shaped text in comments/strings — a conservative fail-closed trade now stated in the docblock and documented by a test, rather than growing a JS parser.

scripts/mfes-import-boundary-check.mjs
```diff
@@ -0,0 +1,374 @@
...
+ *
+ * @param {string} barrelContent
+ * @returns {Violation[]}
+ */
+export function findBarrelViolations(barrelContent) {
+  /** @type {Violation[]} */
+  const violations = [];
+  for (const match of barrelContent.matchAll(/export\s+(?:type\s+)?{([\s\S]*?)}/g)) {
+    for (const rawEntry of match[1].split(',')) {
+      const entry = rawEntry.replace(/\btype\b/g, '').trim();
```
itechmeat, line 535: 🟠 The comment-laundering fixes closed the named repro cases but not the class - both halves still fall to a comment, verified end-to-end at exit 0. Barrel half: `findBarrelViolations` splits the braced list on commas with no comment removal (the exact strip `parseImportedNames` got at 168-170), so `export { // kept for the shell until #999 <newline> DefaultMfeRegistry }` glues the comment onto the name and the check passes with `DefaultMfeRegistry` back in the barrel - the half the docblock calls "structurally impossible to reintroduce". Consumer half: the whole-word backtrack at line 125 takes the LAST `import`/`export` before the specifier, so `// do not import DefaultLifecycleManager elsewhere` inside the block becomes the statement start and the names list comes back empty. One comment strip before matching, applied in both halves plus mirror tests, closes the class rather than the next repro.

packages/mfes/architecture/DESIGN.md
```diff
@@ -223,22 +223,24 @@ Applications need to gain user-facing functionality from independently developed
...
 
+**Public-surface debt register** (#540 follow-up). The barrel still exports six concrete implementations that ADR-0003's encapsulation driver would keep internal but that the `Default*` naming rule enforced by `scripts/mfes-import-boundary-check.mjs` does not detect: `ChildMfeBridgeImpl`, `ParentMfeBridgeImpl`, `ChildDomainForwardingHandler` (bridge wiring), `OperationSerializer`, `WeakMapRuntimeCoordinator` (registry coordination), and `LoadExtHandler` (lifecycle action wiring). Each is debt, not sanctioned surface. Owner: the mfes package maintainers; current status lives on the #540 thread, not in this doc. Removal criterion (objective, not "temporary"): drop an item from the barrel in the next pre-1.0 minor once a repo-wide consumer scan shows zero imports of it outside `packages/mfes`, and, in the same change, add its name to the `CONCRETE_EXPORT_DENYLIST` in `mfes-import-boundary-check.mjs` so it cannot return.
```
itechmeat: 🟡 [Not a blocker] Re-scanning the barrel at head: `InvalidatableDomainContext` (index.ts:62 - the concrete implementation of the `DomainContext` interface exported one line above, the exact pattern the register catalogs; the new test at line 73 meanwhile pins it as a "contract" name), `LruCache` (index.ts:132) and `RetryHandler` (index.ts:133) are concrete classes covered by neither the register, nor the surface table, nor the naming rule. Maybe all three are deliberately sanctioned utilities - then that is worth recording the way `MfeHandlerMF`'s row does, since ADR-0003 leans on the register as an exhaustive inventory and "six" is again a countable claim the barrel disagrees with.

scripts/mfes-import-boundary-check.mjs
```diff
@@ -0,0 +1,374 @@
...
+ *    not the local alias, and `import type` counts — a type-level dependency
+ *    on a concrete class couples the consumer to its shape just the same.
+ *
+ * Zero scanned files is a hard failure, never a vacuous pass (same rule as
+ * `template-pin-drift-check.mjs`): a walk that stops matching means the guard
+ * is broken, and a human needs to see that as red.
+ *
+ * Why a script and not eslint `no-restricted-imports`/`no-restricted-exports`:
+ * the consumer half of this boundary must cover template territory
+ * (`frontx-template.json` trees) that ecosystem CI does not lint at all — an
```
itechmeat: 🟡 [Not a blocker] Round 1 asked for the reason to be written down - it is, but with a wrong fact: main.yml (lines 252-269) does run `npm run lint` inside every self-contained template, and the composition step covers the overlay, so "ecosystem CI does not lint at all" is false. The true reason sits next door and is stronger: root eslint.config.js (131-132) ignores both template trees, so a root-owned rule cannot reach them, and per-template configs are files the templates freely edit - not a place to own a repo-wide boundary. One sentence swap.

packages/mfes/architecture/DESIGN.md
itechmeat: 🔵 [Not a blocker] Table row 229 lists the bridge exports without `ChildDomainForwardingHandler`, row 236 the coordination ones without `LoadExtHandler` - both are real barrel exports (index.ts:72 and :155) and both are named as debt by the register three lines below. The table titled "Public surface" and the register in the same section currently disagree about what crosses the barrel; two cell edits align them.

scripts/mfes-import-boundary-check.test.mjs
```diff
@@ -0,0 +1,364 @@
+// @cpt-dod:cpt-frontx-dod-unit-test-generation-and-agent-verification-standard-test-convention:p1
```
itechmeat, line 1: 🔵 [Not a blocker] The PR's own record went stale behind the fixes: the body still says "20 unit tests" and "817 consumer files scanned" while head has 30 tests and 689 files - and the 817 -> 689 delta is precisely the dot-directory fix, worth stating as such. The squashed commit message meanwhile still lists four register items where DESIGN and the body now carry six, so history lands disagreeing with the doc it ships. A body refresh plus a message amend.

scripts/mfes-import-boundary-check.mjs
```diff
@@ -0,0 +1,374 @@
...
+ * nobody can review what crosses a barrel that doesn't name its exports, and
+ * this guard cannot resolve them without becoming a module-graph walker.
+ *
+ * @param {string} barrelContent
+ * @returns {Violation[]}
+ */
+export function findBarrelViolations(barrelContent) {
+  /** @type {Violation[]} */
+  const violations = [];
+  for (const match of barrelContent.matchAll(/export\s+(?:type\s+)?{([\s\S]*?)}/g)) {
```
itechmeat: ⚪ [Not a blocker] Both barrel patterns look only at braced lists and wildcards, so `export class DefaultFooManager {}` declared directly in the barrel passes at exit 0 (verified through runCli). Latent today - the barrel is re-export-only by convention - but a one-line extra sweep (`/^\s*export\s+(?:abstract\s+)?(?:class|const|function)\s+(\w+)/m`) closes the family, conveniently in the same commit as the comment strip.

itechmeat (changes requested):
🟠 Round 1 was closed properly - all four blockers verified fixed by content, not by reply: the walk skips dot-directories (live run: 689 files, exit 0), both docs swept clean, the kit AGENTS.md rewritten against the real factory signature, the register grown to six names and the denylist seeded with a test. What keeps this one step from approve is a single new hole the comment-laundering fixes did not close as a class: a comment inside an import or export block still defeats both halves of the guard, verified end-to-end at exit 0 with `DefaultMfeRegistry` back in the barrel. The fix is a two-line port of the comment strip `parseImportedNames` already got. One mechanical change plus mirror tests, and this is an approve.

scripts/mfes-import-boundary-check.mjs
tscbmstubp: Fixed in 53e2879c — closed as a class this time: `stripComments()` (block + line comments to a single space, newlines kept) now runs before all matching in BOTH halves, so a comment can neither glue onto a name in the comma-split nor become the backtrack's statement start. That flips the round-2 stance, deliberately: a commented-out import is no longer reported at all — the false negative comments could produce is the one failure a guard must not have, and the narrower fail-closed case (statement-shaped STRING literals) still matches. Both of your repros are exact-equality regression tests, plus a block-comment variant, plus the flip itself documented as a test.

packages/mfes/architecture/DESIGN.md
tscbmstubp: Fixed in 53e2879c — all three verified at zero consumers outside `packages/mfes` (`InvalidatableDomainContext` constructed only by `DefaultMfeRegistry`; `LruCache`/`RetryHandler` used only inside `MfeHandlerMF`), so they're debt, not sanctioned surface: the register now carries nine items and the table rows list all three with debt pointers. The test that pinned `InvalidatableDomainContext` in the contract-names list now uses `ExtensionManager` instead.

scripts/mfes-import-boundary-check.mjs
tscbmstubp: Fixed in 53e2879c — sentence swapped for the true reason: root `eslint.config.js` deliberately ignores both template trees (each ships its own config), so a root-owned rule can never reach them, and a template's own config is a file the template freely edits — not a place a repo-wide boundary can durably live. The false “ecosystem CI does not lint templates at all” claim is gone.

packages/mfes/architecture/DESIGN.md
tscbmstubp: Fixed in 53e2879c — both cells updated: `ChildDomainForwardingHandler` joins the bridge/handler row and `LoadExtHandler` the coordination row (plus the three newly-registered names from the re-scan), so the table and the register no longer disagree about what crosses the barrel.

scripts/mfes-import-boundary-check.test.mjs
tscbmstubp: Fixed in 53e2879c — body refreshed (39 tests now; 689 consumer files, with the 817→689 delta attributed to the dot-directory exclusion; nine register items) and the squashed commit message amended to match: the guard bullet describes the comment strip and declaration sweep, and the docs bullet lists all nine register items.

scripts/mfes-import-boundary-check.mjs
tscbmstubp: Fixed in 53e2879c — a third barrel pattern sweeps direct declarations (`export (declare)? (abstract)? class|function|const|let|var|enum|type|interface Name`) through the same naming rule, in the same commit as the comment strip as you suggested. Tests: `export class DefaultFooManager {}` flagged, a directly-declared abstract contract left alone, `export const DefaultConfig` flagged; `export type Ellipsis` re-exports can't false-positive (needs a bare identifier after the keyword).

scripts/mfes-import-boundary-check.mjs
```diff
@@ -0,0 +1,435 @@
...
+ * LAUNDER a concrete name past the boundary rule (a false negative), even at
+ * the cost of no longer flagging statement-shaped text that only exists
+ * inside a comment (previously a deliberate false positive — see the
+ * `findMfesImportSites` docblock for the round-3 rationale).
+ *
+ * @param {string} source
+ * @returns {string}
+ */
+export function stripComments(source) {
+  return source.replace(/\/\*[\s\S]*?\*\//g, ' ').replace(/\/\/[^\n]*/g, ' ');
```
itechmeat: 🟠 The block-comment pass runs first, over raw text, with `[\s\S]*?` - so a `/*` inside an ordinary string, template literal or regex opens a strip that runs across lines to the next `*/` anywhere in the file, deleting the real code in between. This inverts the round-2 class: real code is now discarded as a comment - a false negative, the thing the guard's own docblock says must never happen. Executed proof on plain code: a glob string `['src/**']` above an import and a normal JSDoc below it swallow the `DefaultMfeRegistryFactory` import whole, exit 0 in both halves - and `stripComments` run over all 689 consumer files shows `template-shell/vitest.shared.ts` already losing real statements today to its `'**/*.config.*'` globs. The docblock also understates the trade ("pathological one-liner" - the cross-line `/*` half is wider and undocumented), and no stripComments test touches a string containing `//` or `/*`. A string-aware single-scan pass plus regression tests for these shapes closes it.

scripts/mfes-import-boundary-check.mjs
```diff
@@ -0,0 +1,435 @@
...
+  }
+  for (const match of barrelContent.matchAll(/export\s+(?:type\s+)?\*(?:\s*as\s+\w+)?\s*from\s*['"]([^'"]+)['"]/g)) {
+    violations.push({
+      file: MFES_BARREL_RELATIVE_PATH,
+      kind: 'wildcard-export',
+      detail: match[1],
+    });
+  }
+  for (const match of barrelContent.matchAll(
+    /^\s*export\s+(?:declare\s+)?(?:abstract\s+)?(?:class|function|const|let|var|enum|type|interface)\s+(\w+)/gm,
```
itechmeat: 🟡 [Not a blocker] Executed: `export async function DefaultFoo() {}`, `export default class DefaultX {}` and `const x = 1;export class DefaultX {}` all pass at exit 0 (`async`/`default` are not in the modifier group, `^\s*` wants line start; `function*` misses like `async`). The default case is the sharpest: a default import has no brace block, `parseImportedNames` returns `[]` by contract, so `export default class DefaultMfeRegistry {}` plus `import Registry from '@gears-frontx/mfes'` is invisible to both halves. `(?:default\s+)?`, `(?:async\s+)?` and `\*?` after `function` close all four shapes.

scripts/mfes-import-boundary-check.mjs
```diff
@@ -0,0 +1,435 @@
...
+ * `type`, and commas, so removing block comments and line-comment runs here
+ * cannot corrupt a real binding name.
+ *
+ * @param {string} statement
+ * @returns {string[]}
+ */
+export function parseImportedNames(statement) {
+  const braced = statement.match(/{([\s\S]*?)}/);
+  if (!braced) return [];
+  const withoutComments = braced[1]
```
itechmeat: 🔵 [Not a blocker] Every in-repo caller reaches `parseImportedNames` through `findMfesImportSites`, which already stripped the whole file (line 150) - this second pass can never see a comment, yet its docblock still argues it as the load-bearing fix. Worse, it duplicates the exact regex pair with the string-awareness defect, so fixing `stripComments` alone leaves a second buggy copy for any direct caller. Either drop the pass and record the comment-free-input guarantee as the caller's, or route it through `stripComments`.

itechmeat (changes requested):
🟠 Round 2 is fully closed and verified by execution - both exploits now fail the guard, the register's inventory finally matches the barrel completely, the docs and the PR's own record all agree with head. What blocks approval is one new hole the fix itself opened: stripComments is not string-aware, so a `/*` inside an ordinary string or regex literal starts a strip that runs to the next `*/` anywhere in the file and silently deletes the real code in between - the inverse of the round-2 class, a false negative instead of a false positive, and it already fires on real files in this repo (glob literals in template-shell/vitest.shared.ts). One string-aware pass plus regression tests for the probed shapes, and this is an approve - the rest of the delta held up under attack.

tscbmstubp:
Round 4 pushed as `f21e86dd` (amended squash, message updated to match).

**The blocker — `stripComments` string-awareness**: replaced the raw-text regex pair with a character-level scanner that recognizes `'…'`/`"…"` strings, template literals, and regex literals (division vs regex decided by the standard lexer heuristic, char classes honored) before comment detection. A `/*` inside `'src/**'` is now literal text; verified directly against `template-shell/vitest.shared.ts` — all 7 imports survive the strip, where round 3 deleted them. Block comments now keep their newlines so line-anchored patterns downstream keep their line structure. Regression tests pin all three probed literal kinds, the end-to-end glob-string-above-import repro, and the unterminated-string cap.

**Same-file non-blockers folded in:**
- `findBarrelViolations`' declaration sweep now covers `async`/generator/`default` modifiers and mid-line (`;`/`}` -prefixed) statements, plus bare `export default DefaultX;` binding re-exports — `export default class DefaultX {}` was the sharp one since the consumer half is structurally blind to default imports. All four probed shapes now exit 1, each with a test.
- `parseImportedNames` routes through the shared `stripComments` instead of carrying a duplicated (and now defective-by-copy) regex pair; its docblock records that in-repo callers arrive pre-stripped.

**Record refresh:** PR body updated (48 guard tests now; the 689 consumer-file count is the CI number — a local tree with materialized `packages/cli/templates/` links scans 788, all clean), commit message updated in the same amend.

Deliberate residue, documented in the scanner's docblock: string *content* is preserved, so statement-shaped text inside a string still matches (fail closed, unchanged from round 3), and a comment nested inside a template `${…}` expression is kept for the same reason.

Verification: guard suite 48/48, full repo-scripts suite 356/356, `tsc -p tsconfig.scripts.json` clean, live `npm run policy:mfes-import-boundary` exit 0.

scripts/mfes-import-boundary-check.mjs
tscbmstubp: Fixed in `f21e86dd`. `stripComments` is now a character-level scanner: `'…'`/`"…"` strings, template literals, and regex literals (division vs regex by the lexer heuristic, char classes honored) are recognized before comment detection, so a `/*` inside `'src/**'` is literal text. Verified against your repro file — all 7 imports in `template-shell/vitest.shared.ts` survive the strip, where round 3 deleted them. Block comments now also preserve their newlines, keeping the line-anchored barrel patterns honest. Regression tests pin the three probed literal kinds, the end-to-end glob-above-import shape, and the unterminated-string cap.

Deliberate residue, written into the docblock: string content is preserved (statement-shaped text in a string still matches, fail closed, unchanged from round 3), and a comment nested inside a template `${…}` expression is kept for the same reason.

scripts/mfes-import-boundary-check.mjs
tscbmstubp: Fixed in `f21e86dd`. The declaration sweep now takes `(?:declare\s+)?(?:default\s+)?(?:abstract\s+)?(?:async\s+)?` plus an optional `*` after the keyword, and anchors on `(?:^|[;})])\s*` instead of line start — so `export async function DefaultFoo()`, `export function* DefaultGen()`, `export default class DefaultX {}`, and `const x = 1;export class DefaultX {}` all exit 1 now, each pinned by a test. A second small sweep catches the bare `export default DefaultX;` binding re-export. The one that stays invisible is an anonymous `export default class {}` — no name for a naming rule to judge; noted in the docblock.

scripts/mfes-import-boundary-check.mjs
tscbmstubp: Fixed in `f21e86dd` — took the routing option: `parseImportedNames` now calls the shared `stripComments` on the braced content, so there is exactly one strip implementation and the string-awareness fix covers direct callers too. Its docblock records that every in-repo caller arrives pre-stripped via `findMfesImportSites`, and that the pass exists to keep the function's contract self-contained.

scripts/mfes-import-boundary-check.mjs
```diff
@@ -0,0 +1,560 @@
...
+ * @param {string} source
+ * @param {number} start index of the opening quote
+ * @returns {number}
+ */
+function scanStringLiteral(source, start) {
+  const quote = source[start];
+  for (let i = start + 1; i < source.length; i++) {
+    const ch = source[i];
+    if (ch === '\\') { i++; continue; }
+    if (ch === quote) return i + 1;
```
itechmeat, line 201: 🟠 `scanStringLiteral` opened on a backtick returns at the first backtick it meets - for `` `a ${`b`} c` `` that is the inner template's OPENING backtick, so everything after scans as code and a `//` or `/*` inside the inner template starts a comment strip. A URL is enough: `` const u = `${`https://a.b`}`; import { DefaultMfeRegistry } from '@gears-frontx/mfes'; `` passes at exit 0 (control without the nesting exits 1), and the barrel half equally passes `DefaultMfeRegistry` back in. The docblock meanwhile claims the opposite in two places (lines 133-135 and 192-194: `${...}` content "copied verbatim", a nested comment "kept... errs toward fail-closed") - both inverted, it is a silent false negative by deletion. Latent today (repo-wide sweep: two files with nested templates, no comment markers inside, zero head files lose code), but it is exactly the class this round was raised to close. Tracking `${}` nesting in scanStringLiteral plus regression tests closes it; the two docblock sentences go with the fix.

scripts/mfes-import-boundary-check.mjs
```diff
@@ -0,0 +1,560 @@
...
+ */
+function isRegexContext(code) {
+  const trimmed = code.trimEnd();
+  if (trimmed === '') return true;
+  const last = trimmed[trimmed.length - 1];
+  if (/[\w$]/.test(last)) {
+    const word = /[\w$]+$/.exec(trimmed)?.[0] ?? '';
+    return REGEX_PRECEDING_KEYWORDS.has(word);
+  }
+  return !/[)\]'"`]/.test(last);
```
itechmeat: 🟡 [Not a blocker] When the slash is read as division, the regex BODY scans as code - and a legal `[/*]` character class then enters the block-comment branch: `if (x) /[/*]/.test(y);` plus an import plus a later `*/` gives exit 0 (control exits 1); same after `]` and after a closing quote. Contrived trigger, so not a blocker on its own - but it is the same root as the nested-template hole (any literal-boundary mis-scan re-opens the deletion class), worth fixing in the same pass. Either way the docblock's closing claim at 110-112, "no comment stripping goes wrong either way", is false and needs the correction.

scripts/mfes-import-boundary-check.mjs
```diff
@@ -0,0 +1,560 @@
...
+  }
+  for (const match of barrelContent.matchAll(/export\s+(?:type\s+)?\*(?:\s*as\s+\w+)?\s*from\s*['"]([^'"]+)['"]/g)) {
+    violations.push({
+      file: MFES_BARREL_RELATIVE_PATH,
+      kind: 'wildcard-export',
+      detail: match[1],
+    });
+  }
+  for (const match of barrelContent.matchAll(
+    /(?:^|[;})])\s*export\s+(?:declare\s+)?(?:default\s+)?(?:abstract\s+)?(?:async\s+)?(?:class|function|const|let|var|enum|type|interface)\b\s*\*?\s*(\w+)/gm,
```
itechmeat: 🟡 [Not a blocker] Executed, all three MISSED: `export const enum DefaultKind { A }` (the alternation eats `const`, `(\w+)` captures `enum`), `export default new DefaultMfeRegistry();` (`(\w+)` captures `new`, the required `;|$` fails), `export default { DefaultMfeRegistry };` (the braced loop wants `export {`). The docblock names the anonymous `export default class {}` as the only remaining gap, which now reads as exhaustive when it is not. Cheap: `(?:const\s+)?enum` closes the first; the gap sentence widens for the rest.

itechmeat: 🔵 [Not a blocker] Executed: `const s = "a; export class DefaultX";` and a template literal containing `export class DefaultX {}` on its own line are both FLAGGED as barrel-export violations. Strings rightly survive comment stripping, but the declaration sweep then matches inside them - a barrel that ever carries such a string fails CI on a violation that does not exist. Unrealistic for a re-export-only barrel, hence not blocking; shielding literal spans from the sweep (the scanner already knows them) removes the class.

scripts/mfes-import-boundary-check.mjs
```diff
@@ -0,0 +1,560 @@
...
+ * number, a closing quote, `)` or `]` it is division. This is the standard
+ * lexer heuristic, not a parser — `if (x) /re/.test(y)` misreads as division
+ * and leaves the slash alone, which merely means the regex body is not
+ * shielded there; no comment stripping goes wrong either way.
+ *
+ * @param {string} code everything emitted so far
+ * @returns {boolean}
+ */
+function isRegexContext(code) {
+  const trimmed = code.trimEnd();
```
itechmeat: 🔵 [Not a blocker] `code` is the whole accumulated output, and both `trimEnd()` and the end-anchored `/[\w$]+$/` rescan it in full on every `/`. Measured on division-heavy input: 46 KB - 121 ms, 184 KB - 2.0 s, 736 KB - 34 s (~quadratic); 1.5 MB missed a two-minute timeout. Live exposure is negligible today (35 of 689 files reach the strip, largest 34 KB, 2.6 ms total), so not blocking - but one large committed file mentioning the package could hang the CI job, and carrying the last significant character in a variable removes the class for free.

scripts/mfes-import-boundary-check.mjs
```diff
@@ -0,0 +1,560 @@
...
+ * import, `import * as ns`, bare re-export) yields no names — the barrel-side
+ * check is what guards those, since a namespace object only contains what the
+ * barrel actually exports.
+ *
+ * Comments are stripped from the braced content before the comma-split (via
+ * the shared `stripComments` — one implementation, not a second copy with its
+ * own defects): a `//` line comment directly above a name otherwise glues
+ * onto that name in the split (`// imported for typing only\n
+ * DefaultLifecycleManager` becomes one entry, not two), laundering a concrete
+ * name past the naming rule. Every in-repo caller reaches this through
```
itechmeat, line 422: ⚪ [Not a blocker] The docblock says this pass "only matters for a direct caller" - but through the scanner's blind spots a comment can survive the whole-file strip inside an import's braces, and then it is this SECOND strip that blocks the round-2 launder (reproduced: an input where `stripComments(input) === input`, yet parseImportedNames still cleans `// DefaultLifecycleManager,`). An argument for keeping the pass, giving the barrel half a symmetric one (it has no such backstop), and correcting the sentence.

itechmeat (changes requested):
🟠 Round 3 is fully closed and the new character-level scanner is genuinely good - it survived ~45 adversarial probes (division-vs-regex in every shape, escapes, unterminated strings, shebang), a repo-wide sweep shows zero files losing code, and the declaration sweep now catches every named shape. One hole keeps this from approve: the scanner does not track `${...}` nesting, so a nested template literal's inner content lands in code position and a `//` or `/*` inside it - a URL is enough - re-opens the delete-real-code class, verified at exit 0 on both halves against exit-1 controls, while the docblock asserts the opposite in two places. Latent today (no head file trips it), but it is the same class round 4 was raised to close, and the fix is scoped: track nesting in scanStringLiteral plus regression tests. The rest of the delta held up under attack; after this one, approve.

tscbmstubp:
Round 4 addressed in 5dfaccdc (since squashed into the PR's single commit):

- **Blocker (nested `${}` templates)**: `scanStringLiteral` now finds the outer literal's real closing backtick via a nesting-aware `scanTemplateExpression` (brace depth, nested strings/templates, comments, regexes — boundary scan only, content still copied verbatim). Both probed exploits now exit 1; the two inverted docblock claims are corrected.
- **`[/*]`-in-regex-after-condition** (same deletion class): slash disambiguation moved to incremental lexer state with a paren stack marking `if`/`for`/`while`/`switch`/`catch` condition parens, so `if (x) /[/*]/.test(y)` scans as a regex. The false "no comment stripping goes wrong either way" claim is gone. The incremental state also removes the quadratic rescan (~34s at 736KB → ~0.3s measured).
- **Declaration sweep**: `export const enum Default…` is caught; the gap sentence now names the expression-bodied default-export shapes (`export default new X()`, `export default { X }`) as outside the sweep.
- **Backstop**: the barrel's braced sweep got the symmetric second strip, and the `parseImportedNames` docblock now describes the pass as the backstop it is.
- **Skipped**: shielding literal spans from the declaration sweep (statement-shaped text in strings staying matched is the guard's documented fail-closed trade; flagged text there is a false positive that fails loud, not a false negative).

Verified: 58 unit tests with exit-1 controls for every probed shape, live guard at exit 0 over 788 consumer files, and a repo-wide 887-file sweep asserting the strip removes nothing outside genuine comments.

scripts/mfes-import-boundary-check.mjs
tscbmstubp: Fixed in 5dfaccdc. `scanStringLiteral` hands `${` off to a new `scanTemplateExpression` that tracks brace depth and skips nested strings, templates (recursively), comments, and regexes — so the outer literal ends at its real closing backtick. Boundary scan only: the expression content is still copied verbatim, and both docblock sentences now state that instead of the inverse. Your probe (`` `${`https://a.b`}` `` before a concrete import) exits 1 on both halves, with controls; regression tests also pin a comment inside `${}` kept verbatim, brace depth via an object literal, a backtick inside a regex within `${}`, and an unterminated `${` shielding (never deleting) the rest of the file.

scripts/mfes-import-boundary-check.mjs
tscbmstubp: Fixed in the same pass (5dfaccdc), as suggested. Slash disambiguation now runs on incremental lexer state with a paren stack that marks `if`/`for`/`while`/`switch`/`catch` condition parens, so the slash after `if (x)` starts a regex and its `[/*]` character class never reaches the comment branch. Regression test covers the probe plus a division control (`f(x) / 2`, `(a + b) / c`). The false "no comment stripping goes wrong either way" claim at 110–112 is gone; the replacement docblock states what a residual misread does and which classes are closed.

scripts/mfes-import-boundary-check.mjs
tscbmstubp: Fixed per the prescription in 5dfaccdc: `const\s+enum` joined the alternation (before `const`, so it cannot be eaten), with a regression test on `export const enum DefaultKind { A }`. The gap sentence no longer reads as exhaustive — it now names the expression-bodied default-export shapes (`export default new DefaultX();`, `export default { DefaultX };`) alongside the anonymous default class as what the sweep cannot see without a parser.

tscbmstubp: Skipped, deliberately. Statement-shaped text inside string literals staying matched is the guard's documented fail-closed trade (same as the consumer half's string-literal rule), and this false positive fails LOUD in CI on a barrel that is re-export-only by convention — unlike the silent delete-real-code false negatives these rounds closed. If it ever fires on a real barrel, shielding literal spans (the scanner already knows them) is the ready fix.

scripts/mfes-import-boundary-check.mjs
tscbmstubp: Fixed in 5dfaccdc, exactly as suggested: the strip no longer rescans everything emitted — slash disambiguation is incremental state (`createSlashContext`) carrying the last word and paren stack forward, so each `/` decides in O(1). Re-measured your division-heavy shape: 736 KB strips in ~0.3 s (was ~34 s).

scripts/mfes-import-boundary-check.mjs
tscbmstubp: Fixed in 5dfaccdc: the docblock now describes the pass as the backstop your repro shows it to be (a comment surviving the whole-file strip through a residual blind spot lands inside the braces, and this second strip keeps it from gluing onto a name), and the barrel half's braced sweep got the symmetric second strip with a comment saying why.

scripts/mfes-import-boundary-check.test.mjs
```diff
@@ -0,0 +1,614 @@
...
```
itechmeat, line 1: 🔵 [Not a blocker] Two record fixes, nothing more: the body still says "48 unit tests" twice - head has 58 (counted and run). The file count is the other way around: the body's 689 is the reproducible number (live run on a clean archive of head prints 689, exit 0; round-4 head prints 689 too, the discovery logic is byte-identical), while the round-5 comment's "788" is not what the committed code prints - 788 is exactly 689 + 63 files under packages/mfes + 36 under scripts, i.e. a count that includes the two trees the guard deliberately excludes, or a local tree with materialized generated dirs. Worth one body edit and one line in the thread so the canonical number is 689.

scripts/mfes-import-boundary-check.mjs
```diff
@@ -0,0 +1,686 @@
...
+ *    sits inside an arbitrary expression only a parser could enumerate.
+ *
+ * @param {string} barrelContent
+ * @returns {Violation[]}
+ */
+export function findBarrelViolations(barrelContent) {
+  barrelContent = stripComments(barrelContent);
+  /** @type {Violation[]} */
+  const violations = [];
+  for (const match of barrelContent.matchAll(/export\s+(?:type\s+)?{([\s\S]*?)}/g)) {
```
itechmeat: 🔵 [Not a blocker] `export\s+` demands at least one whitespace, but `export{DefaultMfeRegistry}from'./r';` and `export*from'./r';` are valid TypeScript with none - both return empty at head (executed; round-4 head behaves the same, so this predates the reviewed delta, from round 1). The consumer half is unaffected (its anchor is `\bfrom\s*['"]`), and so is the declaration sweep (JS syntax forces the space there). Zero instances in the repo and no Prettier to ever introduce the shape - a one-character `export\s*` fix in both patterns whenever the file is next open.

scripts/mfes-import-boundary-check.mjs
```diff
@@ -0,0 +1,686 @@
...
+ * The rules are the standard lexer heuristic, not a parser: a regex may start
+ * after nothing, after most punctuation, or after a keyword like `return`;
+ * after an identifier, a number, a closing quote/literal, or `]` the slash is
+ * division. A `)` is division UNLESS its matching `(` followed a condition
+ * keyword (tracked with a paren stack) — without that distinction, `if (x)
+ * /[/*]/.test(y)` read the regex body as code and its `[/*]` character class
+ * as a block-comment opener, deleting real code up to the next `*\/` in the
+ * file. The heuristic can still misread a slash in shapes this repo does not
+ * contain (e.g. a method NAMED `if` called before a division), which mis-scans
+ * literal boundaries from that point; every such residue class found by review
```
itechmeat: ⚪ [Not a blocker] The docblock's residue example says "shapes this repo does not contain (e.g. a method NAMED `if` called before a division)" - but CONDITION_KEYWORDS is matched against the bare lastWord and a preceding dot does not disqualify it: `promise.catch(fn)` and `Symbol.for('x')` push the same `true` onto the paren stack as a real `catch`/`for`, and the repo has dozens of those call sites. The consequence is confined to the fail-loud direction (a misread division copies text verbatim, can only over-report), the corpus census shows zero slashes after a condition paren, and the r4-vs-r5 strip diff over every file is empty - so zero live footprint, only the example sentence is wrong. One sentence swap when convenient.

scripts/mfes-import-boundary-check.mjs
```diff
@@ -0,0 +1,686 @@
...
+ * @param {{ rootDir: string; log: (line: string) => void; logError: (line: string) => void }} context
+ * @returns {number}
+ */
+function check({ rootDir, log, logError }) {
+  const barrelPath = path.join(rootDir, MFES_BARREL_RELATIVE_PATH);
+  if (!fs.existsSync(barrelPath)) {
+    logError(`[mfes-import-boundary-check] FAIL: barrel not found at ${MFES_BARREL_RELATIVE_PATH} — the boundary this guard protects has moved, update the guard.`);
+    return 1;
+  }
+  const violations = findBarrelViolations(fs.readFileSync(barrelPath, 'utf8'));
```
itechmeat, line 633: ⚪ [Not a blocker] Barrel violations are computed here and then discarded by the zero-files early return a few lines down: a run with a broken walk reports only "the walk is broken" and never names the concrete exports it already found. Exit code is still 1, so no false negative - purely diagnostic completeness for an out-of-CI branch; printing the found violations before returning would cost two lines.

itechmeat (approved):
🟢 Round 5 closes it. All five fixes verify by execution, the deliberate skip is exactly as declared (fail-loud beats silent deletion), and the rewritten scanner now holds under real attack: every earlier exploit exits 1, 240,000 structured-fuzz cases across both halves missed nothing, the strip is behavior-preserving over the whole repo, and the round-4 quadratic case dropped from 34 s to 30 ms. Approving. What remains is non-blocking housekeeping - the body's test count went stale (58 now), and for the record the guard's reproducible count is 689 files; the 788 in the round-5 comment is not what the committed code prints.

scripts/mfes-import-boundary-check.test.mjs
tscbmstubp: Fixed in 420d7399 (body edited in place) — the body now says 61 unit tests: your counted 58 at head, plus the three regression tests this round added (two no-space export shapes for the `export\s*` fix, one broken-walk barrel-violation-output test). 689 is the canonical checked-file count — agreed the 788 in the round-5 comment included the two deliberately excluded trees; the body's 689 stands unchanged.

scripts/mfes-import-boundary-check.mjs
tscbmstubp: Fixed in 420d7399 — `export\s+` → `export\s*` in both barrel patterns (braced re-export and wildcard), consumer half and declaration sweep untouched as you noted. Regression tests added for both no-space shapes: `export{DefaultMfeRegistry}from'./r';` → barrel-export violation, `export*from'./r';` → wildcard-export violation.

scripts/mfes-import-boundary-check.mjs
tscbmstubp: Fixed in 420d7399 — the residue example now names the shape the repo DOES contain: dotted calls like `promise.catch(fn)` / `Symbol.for('x')` set `lastWord` to the bare keyword and raise the condition flag, but the over-report only fires if a `/` immediately follows the closing paren, and no such site exists. Comment-only change.

scripts/mfes-import-boundary-check.mjs
tscbmstubp: Fixed in 420d7399 — the broken-walk early return now prints every already-found barrel violation (via a shared `explainViolation` helper, so both report paths use identical wording) before returning 1. Regression test added: concrete re-export in the barrel + zero consumer files asserts both the broken-walk message and the named violation appear.

scripts/mfes-import-boundary-check.test.mjs
tscbmstubp: Branch squashed for handoff — the round-6 fix (previously 420d7399) now lives in 894108cb, the branch's single commit. Content unchanged (byte-identical tree).

scripts/mfes-import-boundary-check.mjs
tscbmstubp: Branch squashed for handoff — the round-6 fix (previously 420d7399) now lives in 894108cb, the branch's single commit. Content unchanged (byte-identical tree).

scripts/mfes-import-boundary-check.mjs
tscbmstubp: Branch squashed for handoff — the round-6 fix (previously 420d7399) now lives in 894108cb, the branch's single commit. Content unchanged (byte-identical tree).

scripts/mfes-import-boundary-check.mjs
tscbmstubp: Branch squashed for handoff — the round-6 fix (previously 420d7399) now lives in 894108cb, the branch's single commit. Content unchanged (byte-identical tree).

GeraBart (approved):

PR #563, opened 2026-08-12
fix: harden the DCO commit-msg hook and point agents at CONTRIBUTING.md

tscbmstubp:
> [!NOTE]
> Reworked after `49bc1f0c` landed the `require-signoff` hook on develop as an inline bash grep (this PR and that commit were written independently for the same reason — PR #562's DCO failure). Rebased; this now **hardens that hook** rather than adding a competing one: same hook id, same wiring, the check moves into a tested script.

## What the inline grep gets wrong (and this fixes)

- **Merge commits are rejected.** The grep runs on every commit-msg invocation, so any unsigned `git merge` fails locally — but the CI DCO check exempts merge commits, and nobody rewrites a merge to sign it. The script mirrors the CI exemption.
- **A `Signed-off-by` inside git's `#` comment lines counts as a trailer.** E.g. amending with the old message shown in comments can false-pass the grep. The script strips comment lines before matching.
- **No tests.** The check is now `scripts/check-dco-signoff.mjs` with 10 unit tests in the house `runCli` pattern, like every other guard in `scripts/`.

## The discovery gap that caused #562

- **`AGENTS.md` / `CLAUDE.md`**: a one-line pointer to CONTRIBUTING.md's commit requirements. The agent entry points route to `architecture/` and studio rules, so CONTRIBUTING.md never entered an agent's reading path — that's how #562 shipped unsigned commits.
- **`CONTRIBUTING.md`**: the enforcement paragraph now names the backing script and contrasts the local hook (fix = one re-run with `-s`) with the CI check (fix = history rewrite).

## Verified

- 10/10 unit tests pass.
- End-to-end in a fresh clone after `prek install`: unsigned commit rejected with the `git commit -s` pointer, signed commit lands, merge-shaped message exempt.

## Heads-up for existing checkouts

A checkout with a stale `core.hooksPath` (e.g. leftover `.husky/_` from an earlier husky setup — my clone had one) silently bypasses **all** prek hooks, this one included. `git config --unset core.hooksPath && npx prek install` fixes it. Possibly worth a follow-up doctor check, but out of scope here.

## Summary by CodeRabbit

* **New Features**
  * Added automated commit sign-off validation with author and committer identity matching.
  * Added support for merge-commit exemptions and clear feedback for invalid or mismatched sign-offs.

* **Documentation**
  * Updated contribution guidance with sign-off requirements, hook limitations, CI enforcement, and manual verification steps.
  * Added reminders to review contribution guidelines before committing.

* **Tests**
  * Expanded coverage for sign-off validation, identity resolution, repository discovery, merge commits, and error scenarios.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -65,8 +65,9 @@ repos:
         entry: npm run arch:guards
         language: system
         pass_filenames: false
+      # See check-dco-signoff.mjs's own docblock for what this hook checks.
       - id: require-signoff
         name: Require Signed-off-by trailer (commit with `git commit -s`)
-        entry: 'bash -c ''grep -qE "^Signed-off-by: .+ <.+>$" "$1" || { echo "Commit rejected: missing Signed-off-by trailer. Commit with: git commit -s"; exit 1; }'' --'
+        entry: node scripts/check-dco-signoff.mjs
         language: system
         stages: [commit-msg]
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -10,4 +10,6 @@ AI tooling and FrontX development guidelines are provided by the Constructor Stu
 
 Treat `architecture/` as the authority on what the system is and why: PRD for intent, DESIGN for structure, ADRs for decisions, and each FEATURE for the behaviour its numbered instructions specify. Ecosystem code carries `@cpt-` markers back to those instructions; `cfs validate` checks that the chain holds. Template territory — any top-level directory carrying a `frontx-template.json` manifest (how `scripts/template-discovery.mjs` finds templates) — sits outside this chain: markers found there are non-authoritative residue being retired as files change (`cpt-frontx-adr-template-territory-traceability`).
 
+Commit requirements — DCO sign-off (`git commit -s`), branch targets, PR flow — live in `CONTRIBUTING.md`; read it before your first commit, including how it's enforced.
+
 ALL user requests MUST be handled by the Orchestrator agent.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -10,4 +10,6 @@ AI tooling and FrontX development guidelines are provided by the Constructor Stu
 
 Treat `architecture/` as the authority on what the system is and why: PRD for intent, DESIGN for structure, ADRs for decisions, and each FEATURE for the behaviour its numbered instructions specify. Ecosystem code carries `@cpt-` markers back to those instructions; `cfs validate` checks that the chain holds. Template territory — any top-level directory carrying a `frontx-template.json` manifest (how `scripts/template-discovery.mjs` finds templates) — sits outside this chain: markers found there are non-authoritative residue being retired as files change (`cpt-frontx-adr-template-territory-traceability`).
 
+Commit requirements — DCO sign-off (`git commit -s`), branch targets, PR flow — live in `CONTRIBUTING.md`; read it before your first commit, including how it's enforced.
+
 ALL user requests MUST be handled by the Orchestrator agent.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -63,7 +63,7 @@ git commit -s -m "feat: describe the change"
 
 To add a missing sign-off to the latest commit, run `git commit --amend -s --no-edit`. To sign off a range of commits, run `git rebase --signoff <base>` (e.g. `git rebase --signoff HEAD~3` to cover the last three), then push with `--force-with-lease`. Note that `git rebase --signoff` appends the trailer even when a `Signed-off-by` already exists earlier in the message (it only skips when an identical sign-off is last), so rebasing over already-signed commits that end in `Co-Authored-By` adds a duplicate line - pick a `<base>` that spans only the commits missing the trailer.
 
-This is enforced locally by a `commit-msg` hook (`require-signoff` in [`.pre-commit-config.yaml`](.pre-commit-config.yaml)); commits without the trailer are rejected before they're made. The hook is installed automatically by `npm install` (via the `prepare` script) — if commits aren't being checked, run `npx prek install`.
+This is enforced locally by a `commit-msg` hook (`require-signoff` in [`.pre-commit-config.yaml`](.pre-commit-config.yaml), backed by `scripts/check-dco-signoff.mjs`); commits without the trailer, or whose trailer matches neither the author nor the committer, are rejected before they're made; merge commits are exempt, as they are in CI. The hook only runs for `git commit` — `git cherry-pick`, `git rebase` and `git rebase --signoff` bypass `commit-msg` hooks entirely, so check those results yourself. CI's DCO check enforces the same rules after push, where fixing it means rewriting history. The hook is installed automatically by `npm install` (via the `prepare` script) — if commits aren't being checked, run `npx prek install`.
 
 ## Versioning
 
--- a/scripts/check-dco-signoff.mjs
+++ b/scripts/check-dco-signoff.mjs
@@ -1,0 +1,239 @@
+/**
+ * DCO Sign-off Commit Guard.
+ *
+ * CONTRIBUTING.md requires a `Signed-off-by` trailer on every commit
+ * (Developer Certificate of Origin), and CI enforces it with the DCO check —
+ * but CI is the most expensive possible place to learn about a missing
+ * trailer: the PR is already open, the push already made, and the fix is a
+ * history rewrite (`git rebase --signoff` + `--force-with-lease`). PR #562
+ * shipped exactly that way. This guard moves the failure to commit time,
+ * where the fix is re-running one command.
+ *
+ * Wired as a `commit-msg`-stage hook in `.pre-commit-config.yaml`
+ * (installed for every contributor by `npm install` → `prek install`; see
+ * `default_install_hook_types` there). prek passes the commit-message file
+ * path as the only argument.
+ *
+ * What counts as signed: a `Signed-off-by: Name <email>` trailer line, the
+ * exact shape `git commit -s` writes. Merges are exempt, matching the DCO CI
+ * check's own exemption — but that exemption is keyed off actual merge state
+ * (`MERGE_HEAD` present, the same signal `dco2` reads), never off the
+ * commit message's first line: a hand-written "Merge the two config loaders"
+ * subject is not a merge, and a real merge with a custom subject still is.
+ *
+ * When a `Signed-off-by` trailer is present, its name/email is also compared
+ * (case-insensitively) against the commit's author and committer idents —
+ * the same `SignOffMismatch` comparison the CI DCO check makes, which accepts
+ * a match against either. The author comes from `GIT_AUTHOR_NAME`/`_EMAIL`,
+ * which git exports into the hook's environment (so `--author=`, `git am`
+ * and `cherry-pick -x` relays keep their original sign-off); the committer
+ * from `git var GIT_COMMITTER_IDENT`, which git resolves even when
+ * `user.name` is unset. Only if neither ident can be resolved does the check
+ * fall back to trailer-presence only.
+ *
+ * CLI entry: `node scripts/check-dco-signoff.mjs <commit-msg-file>`
+ * (exit 0 when signed or exempt). Core logic is exported for unit tests in
+ * `scripts/check-dco-signoff.test.mjs`.
+ */
+import { execFileSync } from 'node:child_process';
+import fs from 'node:fs';
+import path from 'node:path';
+import process from 'node:process';
+import { pathToFileURL } from 'node:url';
+
+export const SIGNOFF_PATTERN = /^Signed-off-by: (.+) <(.+@.+)>[ \t]*$/gim;
+
+/**
+ * @param {string} message full commit message, comment lines included
+ * @returns {{ name: string; email: string }[]}
+ */
+function trailers(message) {
+  // Assumes the default git comment char `#`; does not read `core.commentChar`
+  // — if that's overridden, git's comment block isn't stripped here.
+  const effective = message
+    .split('\n')
+    .filter((line) => !line.startsWith('#'))
+    .join('\n');
+  return [...effective.matchAll(SIGNOFF_PATTERN)].map(([, name, email]) => ({ name, email }));
+}
+
+/** @typedef {{ name: string; email: string }} Identity */
+
+/**
+ * @param {string} message full commit message, comment lines included
+ * @param {{ isMerge?: boolean; identities?: Identity[] }} [options]
+ *   `identities` — the idents a sign-off may match (author and committer);
+ *   empty or absent means presence-only checking.
+ * @returns {{ ok: true } | { ok: false, reason: 'missing-signoff' | 'sign-off-mismatch', message: string }}
+ */
+export function checkMessage(message, options = {}) {
+  const { isMerge = false, identities = [] } = options;
+
+  if (isMerge) return { ok: true };
+
+  const found = trailers(message);
+  if (found.length === 0) {
+    return {
+      ok: false,
+      reason: 'missing-signoff',
+      message:
+        'Commit message carries no `Signed-off-by` trailer.\n\n' +
+        'CONTRIBUTING.md (Commit Requirements) requires a DCO sign-off on every commit:\n' +
+        '  git commit -s ...\n' +
+        'To sign the commit you were just writing, re-run the same command with -s added.\n' +
+        'The CI DCO check enforces the same rule after push, where fixing it means rewriting history.',
+    };
+  }
+
+  if (identities.length > 0) {
+    const matches = found.some((trailer) =>
+      identities.some(
+        (id) => trailer.name.toLowerCase() === id.name.toLowerCase() && trailer.email.toLowerCase() === id.email.toLowerCase(),
+      ),
+    );
+    if (!matches) {
+      const expected = identities.map((id) => `${id.name} <${id.email}>`).join(' or ');
+      return {
+        ok: false,
+        reason: 'sign-off-mismatch',
+        message:
+          `Commit message's \`Signed-off-by\` trailer matches neither the author nor the committer ` +
+          `(SignOffMismatch): expected ${expected}.\n\n` +
+          'The CI DCO check makes the same comparison. If this is your own commit, re-run:\n' +
+          '  git commit -s ...\n' +
+          "so the trailer is written from your ident. If you are relaying someone else's commit, keep their\n" +
+          'trailer and preserve them as the author (`--author=`, `git am`, `cherry-pick -x`) instead.',
+      };
+    }
+  }
+
+  return { ok: true };
+}
+
+/**
+ * @param {string} gitDir
+ * @returns {boolean}
+ */
+export function detectIsMerge(gitDir) {
+  try {
+    return fs.existsSync(path.join(gitDir, 'MERGE_HEAD'));
+  } catch {
+    return false;
+  }
+}
+
+/**
+ * @param {string} [cwd]
+ * @returns {string | undefined} the `$GIT_DIR` of the repository at `cwd`
+ */
+export function resolveGitDir(cwd = process.cwd()) {
+  try {
+    const dir = execFileSync('git', ['rev-parse', '--git-dir'], { cwd, encoding: 'utf8', stdio: ['ignore', 'pipe', 'pipe'] }).trim();
+    return path.resolve(cwd, dir);
+  } catch {
+    return undefined;
+  }
+}
+
+/**
+ * Parses `git var` output — `Name <email> <timestamp> <tz>`.
+ * @param {string} ident
+ * @returns {Identity | undefined}
+ */
+function parseIdent(ident) {
+  const match = /^(.+?) <(.+@.+?)>(?: |$)/.exec(ident.trim());
+  return match ? { name: match[1], email: match[2] } : undefined;
+}
+
+/**
+ * @param {'GIT_AUTHOR_IDENT' | 'GIT_COMMITTER_IDENT'} name
+ * @param {string} cwd
+ * @param {NodeJS.ProcessEnv} env
+ * @returns {Identity | undefined}
+ */
+function gitVar(name, cwd, env) {
+  try {
+    return parseIdent(execFileSync('git', ['var', name], { cwd, env, encoding: 'utf8', stdio: ['ignore', 'pipe', 'pipe'] }));
+  } catch {
+    return undefined;
+  }
+}
+
+/**
+ * The idents a sign-off may match, as CI sees them: the author (from the
+ * `GIT_AUTHOR_*` environment git hands every hook, falling back to
+ * `git var GIT_AUTHOR_IDENT`) and the committer (`git var GIT_COMMITTER_IDENT`).
+ * @param {{ cwd?: string; env?: NodeJS.ProcessEnv }} [options]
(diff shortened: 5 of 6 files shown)


scripts/check-dco-signoff.mjs
```diff
@@ -0,0 +1,91 @@
...
+ */
+export function checkMessage(message) {
+  // Strip the comment lines git appends below the message — a Signed-off-by
+  // quoted inside the commented diff/status must not count as a trailer.
+  const effective = message
+    .split('\n')
+    .filter((line) => !line.startsWith('#'))
+    .join('\n');
+
+  if (/^Merge /.test(effective)) return { ok: true };
```
itechmeat: 🟠 The exemption keys on the message text, but CI's DCO check exempts a merge by parent count (`parents.len() > 1` in cncf/dco2), not by subject. That cuts both ways: `git commit -m "Merge the two config loaders into one"` now passes this hook unsigned and fails after push - the exact late failure this guard exists to prevent, and one the inline grep it replaces could not produce - while a real merge with a custom `-m` message is rejected locally though CI would exempt it. At commit-msg time `MERGE_HEAD` is present in `$GIT_DIR`, which is the same signal CI reads; passing that in as a flag from `runCli` would keep `checkMessage` pure.

scripts/check-dco-signoff.mjs
```diff
@@ -0,0 +1,91 @@
...
+ *
+ * CLI entry: `node scripts/check-dco-signoff.mjs <commit-msg-file>`
+ * (exit 0 when signed or exempt). Core logic is exported for unit tests in
+ * `scripts/check-dco-signoff.test.mjs`.
+ */
+import fs from 'node:fs';
+import process from 'node:process';
+import { pathToFileURL } from 'node:url';
+
+export const SIGNOFF_PATTERN = /^Signed-off-by: .+ <.+@.+>$/m;
```
itechmeat: 🟡 [Not a blocker] CI does not just want a trailer, it wants one whose name and email match the author or committer - cncf/dco2 reports `SignOffMismatch` otherwise. A copy-pasted `Signed-off-by` from someone else's commit passes this hook and still goes red after push, which is the case the PR's promise covers. `user.name`/`user.email` are available at commit-msg time, though declining the identity check is defensible (an environment with no configured identity would then fail every commit). If it is declined, is it worth narrowing the "matching the CI DCO check" wording, which currently appears in four places?

CONTRIBUTING.md
```diff
@@ -63,7 +63,7 @@ git commit -s -m "feat: describe the change"
...
 
-This is enforced locally by a `commit-msg` hook (`require-signoff` in [`.pre-commit-config.yaml`](.pre-commit-config.yaml)); commits without the trailer are rejected before they're made. The hook is installed automatically by `npm install` (via the `prepare` script) — if commits aren't being checked, run `npx prek install`.
+This is enforced locally by a `commit-msg` hook (`require-signoff` in [`.pre-commit-config.yaml`](.pre-commit-config.yaml), backed by `scripts/check-dco-signoff.mjs`); commits without the trailer are rejected before they're made — CI's DCO check enforces the same rule after push, where fixing it means rewriting history. The hook is installed automatically by `npm install` (via the `prepare` script) — if commits aren't being checked, run `npx prek install`.
```
itechmeat: 🟡 [Not a blocker] The rewritten sentence says commits without the trailer are rejected before they are made, which stopped being true for merge commits in this same PR, and the carve-out currently lives only in a YAML comment and the script's docblock - neither of which a contributor reads. Line 58 above carries the same absolute wording. Since this paragraph is already open in the diff, one clause naming the merge exemption would keep it accurate.

scripts/check-dco-signoff.mjs
```diff
@@ -0,0 +1,91 @@
...
+
+/**
+ * @param {string} message full commit message, comment lines included
+ * @returns {{ ok: true } | { ok: false, reason: string }}
```
itechmeat: 🟡 [Not a blocker] House style wants a handled failure's `reason` to come from a closed literal set that callers switch on, with the human wording in a separate `message` field. Here `reason` is the prose itself, and the test at check-dco-signoff.test.mjs:41 asserts `toContain('git commit -s')` - so rewording the help text breaks a unit test. Splitting it into `reason: 'missing-signoff'` plus `message` would let the test assert the closed value instead.

CLAUDE.md
```diff
@@ -10,4 +10,6 @@ AI tooling and FrontX development guidelines are provided by the Constructor Stu
 
 Treat `architecture/` as the authority on what the system is and why: PRD for intent, DESIGN for structure, ADRs for decisions, and each FEATURE for the behaviour its numbered instructions specify. Ecosystem code carries `@cpt-` markers back to those instructions; `cfs validate` checks that the chain holds. Template territory — any top-level directory carrying a `frontx-template.json` manifest (how `scripts/template-discovery.mjs` finds templates) — sits outside this chain: markers found there are non-authoritative residue being retired as files change (`cpt-frontx-adr-template-territory-traceability`).
 
+Commit requirements — DCO sign-off (`git commit -s`), branch targets, PR flow — live in `CONTRIBUTING.md`; read it before your first commit. A `commit-msg` hook (`prek`, installed by `npm install`) rejects unsigned commits locally, and CI rejects them after push.
```
itechmeat: 🔵 [Not a blocker] The first sentence does the job - point at CONTRIBUTING.md. The second copies its enforcement fact into two more files, so the same statement now lives in four places (CONTRIBUTING.md, AGENTS.md, CLAUDE.md, the YAML comment) and three of them go stale the first time the hook changes. If the point is that a bare pointer gets skimmed past, would a short clause do instead of a second sentence?

.pre-commit-config.yaml
```diff
@@ -65,8 +65,12 @@ repos:
         entry: npm run arch:guards
         language: system
         pass_filenames: false
+      # The script, not an inline grep: merge commits are exempt (matching
```
itechmeat: 🔵 [Not a blocker] These four lines restate what the script's own header says, and the `entry` right below already names the file. The `arch-*` comment above earns its place by explaining wiring that lives nowhere else; this one duplicates a docblock one file away. Trimming it to a single line would keep the hook list readable.

scripts/check-dco-signoff.test.mjs
```diff
@@ -0,0 +1,85 @@
...
+function run(argv) {
+  const lines = [];
+  const record = (line) => lines.push(line);
+  const exitCode = runCli({ argv, log: record, logError: record });
+  return { exitCode, output: lines.join('\n') };
+}
+
+describe('checkMessage', () => {
+  it('accepts the exact trailer shape `git commit -s` writes', () => {
+    expect(checkMessage('fix: thing\n\nSigned-off-by: Vuk D. <davydowiktor@gmail.com>\n').ok).toBe(true);
```
itechmeat: 🔵 [Not a blocker] Every other case in this file uses the neutral `A B <a@b.c>`, but this one hardcodes a real personal address into a permanent repo file. The case tests trailer shape, which is identical with a neutral fixture value.

scripts/check-dco-signoff.test.mjs
```diff
@@ -0,0 +1,85 @@
...
+  const file = path.join(workDir, 'COMMIT_EDITMSG');
+  await writeFile(file, content);
+  return file;
+}
+
+/** Runs the guard with its output captured, so a case can assert what it named. */
+function run(argv) {
+  const lines = [];
+  const record = (line) => lines.push(line);
+  const exitCode = runCli({ argv, log: record, logError: record });
```
itechmeat: 🔵 [Not a blocker] Both injected sinks write into the same array, so no test can tell stderr from stdout. A regression that printed the failure to stdout - where a hook runner may not surface it - would pass all four runCli tests unchanged. Two separate arrays would pin the stream contract.

scripts/check-dco-signoff.mjs
```diff
@@ -0,0 +1,91 @@
...
+ */
+export function checkMessage(message) {
+  // Strip the comment lines git appends below the message — a Signed-off-by
+  // quoted inside the commented diff/status must not count as a trailer.
+  const effective = message
+    .split('\n')
+    .filter((line) => !line.startsWith('#'))
```
itechmeat, line 55: ⚪ [Not a blocker] `core.commentChar` is configurable - with it set to anything else git's comment block stops being stripped and a quoted `Signed-off-by` false-passes again, which is the bug this line fixes. It also drops a legitimate body line like `#562 fixed the thing`, harmless since a sign-off cannot start with `#`. Probably fine to ignore, but worth one line in the header stating the assumption.

scripts/check-dco-signoff.mjs
itechmeat: ⚪ [Not a blocker] The CI regex is case-insensitive and tolerates trailing whitespace (`(?mi)…\s*$`); this one is neither, and git's default `--cleanup=whitespace` strips that trailing space only after the commit-msg hook has run - so the hook can reject what CI would accept. Not a regression (the old grep behaved the same) and it errs safe, just another spot where "the same rule as CI" is not literal.

scripts/check-dco-signoff.mjs
```diff
@@ -0,0 +1,91 @@
...
+  const logError = options.logError ?? console.error;
+
+  const messageFile = argv[0];
+  if (!messageFile) {
+    logError('[check-dco-signoff] FAIL: no commit-message file argument — the hook is miswired, not the commit unsigned.');
+    return 1;
+  }
+  let message;
+  try {
+    message = fs.readFileSync(messageFile, 'utf8');
```
itechmeat: ⚪ [Not a blocker] `log`/`logError` are injected but the read is not, so all four runCli tests go through real mkdtemp/writeFile/rm and the unreadable-file case is only reachable via a hardcoded `/nonexistent/COMMIT_EDITMSG`. `check-test-dependency-versions.mjs` takes its file access as a parameter, though `validate-templates.mjs` and `template-pin-drift-check.mjs` use real tmpdirs too, so the convention here is genuinely mixed.

AGENTS.md
```diff
@@ -10,4 +10,6 @@ AI tooling and FrontX development guidelines are provided by the Constructor Stu
 
 Treat `architecture/` as the authority on what the system is and why: PRD for intent, DESIGN for structure, ADRs for decisions, and each FEATURE for the behaviour its numbered instructions specify. Ecosystem code carries `@cpt-` markers back to those instructions; `cfs validate` checks that the chain holds. Template territory — any top-level directory carrying a `frontx-template.json` manifest (how `scripts/template-discovery.mjs` finds templates) — sits outside this chain: markers found there are non-authoritative residue being retired as files change (`cpt-frontx-adr-template-territory-traceability`).
 
+Commit requirements — DCO sign-off (`git commit -s`), branch targets, PR flow — live in `CONTRIBUTING.md`; read it before your first commit. A `commit-msg` hook (`prek`, installed by `npm install`) rejects unsigned commits locally, and CI rejects them after push.
```
itechmeat: ⚪ [Not a blocker] The two files are byte-identical and this PR correctly edited both, but nothing goes red if the next edit lands in only one - every doc change has to be made twice on trust. A symlink, or a small guard next to the other `scripts/` ones, would close it. Follow-up territory, not this PR's debt.

itechmeat (changes requested):
The direction is right and the work is solid: the inline grep really did reject local merges and really did see git's comment lines, the replacement is tested, and the tests are genuinely wired into the runner CI executes. One thing keeps it from approve, and it sits in the feature the PR is named for: the merge exemption matches the message text rather than the fact of a merge, so a hand-written "Merge ..." subject now ships unsigned and fails at CI - the late, history-rewrite failure this guard exists to prevent, and a hole the grep it replaces did not have. MERGE_HEAD is on disk at commit-msg time and is the same signal CI reads. Everything else is questions and polish.

scripts/check-dco-signoff.mjs
tscbmstubp: Fixed in b2fda725: `checkMessage()` now takes an explicit `isMerge` flag instead of pattern-matching the subject line. `runCli()` resolves it from whether `MERGE_HEAD` exists in `$GIT_DIR` — the same signal `dco2` reads — overridable via `options.isMerge` for tests. Added regression tests for both directions: a real merge with a non-"Merge" subject now passes, and a hand-written "Merge ..." subject with no trailer now correctly fails.

scripts/check-dco-signoff.mjs
tscbmstubp: Fixed in b2fda725: after confirming a `Signed-off-by` trailer is present, `checkMessage()` now compares its name/email (case-insensitively) against `git config user.name`/`user.email` when both are configured, and fails with a `SignOffMismatch`-style reason on a mismatch — same comparison `dco2` makes. When identity isn't fully configured locally, it falls back to the prior presence-only check rather than failing every commit, per the tradeoff you flagged.

CONTRIBUTING.md
tscbmstubp: Fixed in 8c9a69ff: added an "except merge commits, which are exempt" clause so this sentence stays accurate now that the merge exemption exists.

scripts/check-dco-signoff.mjs
tscbmstubp: Fixed in 8c9a69ff: `checkMessage()`'s failure return is now `{ ok: false, reason: 'missing-signoff' | 'sign-off-mismatch', message: <prose> }` — `reason` is a closed literal callers can switch on, `message` carries the human-readable text. `runCli` now logs `result.message`, and the tests asserting `git commit -s` in the help text were updated to assert against `message` (and the literal `reason` code) instead.

CLAUDE.md
tscbmstubp: Fixed in 8c9a69ff: trimmed the second sentence to a short clause ("...read it before your first commit, including how it's enforced.") instead of restating the enforcement fact — same treatment applied to AGENTS.md so the two stay identical.

.pre-commit-config.yaml
tscbmstubp: Fixed in 8c9a69ff: trimmed the 4-line comment to a single line pointing at the script's own docblock, matching the style of the other hooks in this file.

scripts/check-dco-signoff.test.mjs
tscbmstubp: Fixed in 8c9a69ff: replaced the real email with the neutral `A B <a@b.c>` fixture already used elsewhere in the file.

scripts/check-dco-signoff.test.mjs
tscbmstubp: Fixed in 8c9a69ff: the `run()` test helper now captures stdout/stderr into two separate arrays instead of one shared sink, and one test now also asserts stdout stayed empty on a failure — pinning the stream contract you flagged.

scripts/check-dco-signoff.mjs
tscbmstubp: Fixed in 8c9a69ff: added a line to the docblock above the comment-stripping filter stating it assumes the default `#` `core.commentChar` and doesn't read the config — doc-only, no behavior change.

scripts/check-dco-signoff.mjs
tscbmstubp: Fixed in 8c9a69ff: `SIGNOFF_PATTERN` is now case-insensitive on `Signed-off-by` and tolerant of trailing whitespace (`/^Signed-off-by: (.+) <(.+@.+)>[ \t]*$/gim`), matching CI's regex tolerance. Capture-group indices are unchanged; added a test for a mixed-case, trailing-whitespace trailer.

scripts/check-dco-signoff.mjs
tscbmstubp: Fixed in 8c9a69ff: the message-file read is now injectable via `options.readFile` (defaulting to `fs.readFileSync`), matching the existing `log`/`logError` DI pattern. Added a test exercising the unreadable-file path via injection, alongside the existing hardcoded-path case.

AGENTS.md
tscbmstubp: Fixed in 8c9a69ff: applied the same trim as CLAUDE.md here so both files stay byte-identical (verified with a diff after editing).

scripts/check-dco-signoff.mjs
```diff
@@ -0,0 +1,188 @@
...
+    (() => {
+      try {
+        const gitDir = execFileSync('git', ['rev-parse', '--git-dir'], { encoding: 'utf8', stdio: ['ignore', 'pipe', 'pipe'] }).trim();
+        return detectIsMerge(gitDir);
+      } catch {
+        return false;
+      }
+    })();
+
+  const identity = options.identity ?? {
```
itechmeat: 🟡 [Not a blocker] `git commit -s` writes the trailer from the effective committer ident, and CI (cncf/dco2) accepts a sign-off matching the author **or** the committer - but this reads `git config user.*`, which is neither. Two consequences, both reproducible: a relayed commit (author Bob, committer Alice, trailer Bob - `git am`, `cherry-pick -x`, `--author=`) is rejected here though CI accepts it, and the failure text then advises `git commit -s`, which would rewrite Bob's sign-off to Alice and change who certified the contribution; and when `user.name` is unset while git still resolves an ident from system data, the check silently falls back to presence-only at line 83 and accepts a stranger's trailer - the case the check was added for. Git already puts `GIT_AUTHOR_NAME`/`GIT_AUTHOR_EMAIL` in the hook's environment and `git var GIT_COMMITTER_IDENT` is always resolvable, so accepting a match against either would mirror CI exactly.

scripts/check-dco-signoff.test.mjs
```diff
@@ -0,0 +1,174 @@
...
+function run(argv, overrides = {}) {
+  const stdout = [];
+  const stderr = [];
+  // isMerge/identity default to values that don't depend on the real git
+  // state of whatever checkout runs this suite.
+  const exitCode = runCli({
+    argv,
+    log: (line) => stdout.push(line),
+    logError: (line) => stderr.push(line),
+    isMerge: false,
```
itechmeat, line 41: 🔵 [Not a blocker] `run()` is the only `runCli` call site in the suite and it hardcodes `isMerge: false, identity: {}` ahead of `...overrides`, with no case passing `undefined` - so the `??` fallbacks never execute and `detectIsMerge`, the `rev-parse --git-dir` call and `readGitConfigValue` have no coverage at all. Those are the three functions doing real IO, and the mechanism the docblock and CONTRIBUTING.md now promise. Neither helper is exported, so a test cannot reach them even deliberately; exporting them would make a temp-repo case possible.

CONTRIBUTING.md
```diff
@@ -63,7 +63,7 @@ git commit -s -m "feat: describe the change"
...
 
-This is enforced locally by a `commit-msg` hook (`require-signoff` in [`.pre-commit-config.yaml`](.pre-commit-config.yaml)); commits without the trailer are rejected before they're made. The hook is installed automatically by `npm install` (via the `prepare` script) — if commits aren't being checked, run `npx prek install`.
+This is enforced locally by a `commit-msg` hook (`require-signoff` in [`.pre-commit-config.yaml`](.pre-commit-config.yaml), backed by `scripts/check-dco-signoff.mjs`); commits without the trailer are rejected before they're made, except merge commits, which are exempt — CI's DCO check enforces the same rule after push, where fixing it means rewriting history. The hook is installed automatically by `npm install` (via the `prepare` script) — if commits aren't being checked, run `npx prek install`.
```
itechmeat: 🔵 [Not a blocker] The hook now also rejects a *present* trailer whose identity does not match, and that reason (`sign-off-mismatch`) appears on no contributor-facing surface - AGENTS.md and CLAUDE.md were trimmed to a pointer last round, so this sentence is the only place left for it. The merge clause is also narrower than reality: `commit-msg` does not run for `git cherry-pick`, `git rebase` or `git rebase --signoff` at all (verified), so merges are not the only exception - and `git rebase --signoff` is the remedy the paragraph above recommends.

itechmeat (approved):
🟢 All twelve round-1 findings are genuinely closed, and the merge exemption is now correct - I verified the MERGE_HEAD path against real git in five repo layouts (root, subdirectory, linked worktree, submodule, conflicted merge), and `--squash` correctly stays subject to sign-off. Tests went from 10 to 22 and pass. Approving. Three non-blocking notes remain, all in the identity check this round asked for: it compares against `git config user.*` rather than the ident git actually stamps, which rejects a relayed commit CI would accept and advises a fix that would overwrite someone else's sign-off. Cheap to close - `GIT_AUTHOR_*` is already in the hook's environment - and none of it needs to hold up the merge.

scripts/check-dco-signoff.mjs
tscbmstubp: Fixed in 1373eee0: the check now resolves the author from `GIT_AUTHOR_NAME`/`GIT_AUTHOR_EMAIL` (falling back to `git var GIT_AUTHOR_IDENT`) and the committer from `git var GIT_COMMITTER_IDENT`, and accepts a trailer matching either — the same either/or dco2 applies. Relayed commits (`--author=`, `git am`, `cherry-pick -x`) keep their original sign-off, and the mismatch text names both idents and tells a relayer to preserve the author rather than re-sign as themselves. Presence-only fallback now happens only if neither ident resolves, which `git var` makes practically unreachable; a temp-repo test covers the unset-`user.name` case resolving the committer anyway.

scripts/check-dco-signoff.test.mjs
tscbmstubp: Fixed in 1373eee0: `detectIsMerge`, `resolveGitDir` and `resolveIdentities` are exported (the latter two take `cwd`/`env` so they don't depend on the suite's own checkout), and a new describe block runs them against a temp `git init` repo: MERGE_HEAD false→true across a real `merge --no-commit`, `rev-parse --git-dir` from a subdirectory and outside a repo, and identity resolution with/without `GIT_AUTHOR_*` and with `user.name` unset. 30 tests now.

CONTRIBUTING.md
tscbmstubp: Fixed in 1373eee0: the sentence now says a trailer matching neither author nor committer is also rejected, keeps the merge exemption, and states that `git cherry-pick`, `git rebase` and `git rebase --signoff` bypass `commit-msg` hooks entirely so those results need checking by hand.

gs-layer (approved):
Looks good to merge. The new script is a solid replacement for the inline grep: merge/squash exemption, comment-line stripping, and author/committer identity matching are all correctly handled and covered by tests (verified this directly against real git repos, not just by reading the diff).

Two minor, non-blocking things worth a look, either now or as a fast follow-up:

1. SIGNOFF_PATTERN's email group is greedy ((.+@.+)), so a malformed trailer with a doubled closing bracket (e.g. Signed-off-by: A B <a@b.c>>) still passes. Doesn't happen from a normal git commit -s, but [^>]+ instead of .+ would close it for free.
[scripts/check-dco-signoff.mjs#L43-L45](https://github.com/constructorfabric/gears-frontx/blob/fb76875d16654392436e4f15402fe57b77dd03e0/scripts/check-dco-signoff.mjs#L43-L45)

2. The PR description flags that a stale core.hooksPath (leftover husky config) can silently skip all pre-commit hooks, including this new DCO check, and leaves it out of scope. Given this repo already hit hook-config fragility once (issue 
[#239](https://github.com/gs-layer/gears-frontx/issues/239)
), a small follow-up — e.g. a doctor check for stale core.hooksPath — seems worth opening so the gap doesn't sit open indefinitely.

Nothing here blocks the merge.

GeraBart (approved):

Issue #507, opened 2026-08-03
[ARCH] Define region-union marker grammar and host-document semantics


gs-layer:
## Problem

The `region-union` implementation needs one durable contract for both its marker grammar and the surrounding file model. Today the reverse scanner can only resolve whitespace-delimited `identity:key` tokens, while the manifest contract does not restrict region-key characters or require a delimiter before a comment closer. Separately, composition emits only marker blocks, so unmarked content is discarded even at seed, while exclusive paths overwrite their file whole.

These are one contract question owned by the template-manifest feature, not independent scanner fixes.

## Decisions required

1. Specify the legal marker grammar: identities, region keys, whitespace, comment closers, delimiters, and escaping if needed.
2. Specify how malformed, unterminated, nested, and otherwise unresolvable markers are handled.
3. Decide whether a `region-union` file is strictly a concatenation of owned blocks or a host document with preserved user content and anchors.
4. Define the policy for developer edits in template-owned files, including the intentional asymmetry between `region-union` and `exclusive`.
5. Update the owning FEATURE and any dependent CLI-scaffolding instructions, then add contract-level tests.

## Scope boundary

PR #500 should still make its local scanner safe for normal valid markers: exact end-token matching and refusal rather than silent loss when it recognizes an incomplete block. This issue owns the wider schema and document-model decision.

## Context

Raised during review of PR #500; its current docstring documents the parser limitation but cannot serve as the schema authority.

PR #564, opened 2026-08-12
docs(arch): redesign template registration around one project state file (ADR 0035-0042)

### LINKED ISSUE #587: Redesign template registration around one project state file and a thin manifest
## Description

The template mechanism the CLI carries today was assembled one decision at a time, and the pieces no longer fit each other. Overlapping state is spread across several records that can disagree, and the model cannot express the composition its own templates need.

**The manifest declares five categories** — `schemaVersion`, `ownershipBoundaries.exclusiveSubtrees`, `ownershipBoundaries.sharedFiles` with merge strategies and region markers, `referencedTemplates`, and an optional `description`. The ownership categories restate what an ownership computation can derive on its own; `referencedTemplates` names a preset-composition path; and `description` being optional means a template can be published with nothing a caller holding no reference can select it by.

**Provenance is one record per applied template instance**, kept separately from the project-level registration state and from the ownership exceptions a project reserves for itself. All three key off the same template name, origin, version and target, and each is written by a different command — so a partial or interrupted write leaves the set internally inconsistent, with no single document authoritative.

**Composition is inferred rather than requested**: a template's manifest names the other templates to apply alongside it and a transitive resolver expands that graph, so a caller cannot see what will be written until the resolver has run.

**Conflict detection compares targets for equality only**, so a template targeting `src/config/` is accepted beside one targeting `src/`. Containment is not treated as a conflict, and there is correspondingly no way for a host template to deliberately host a guest either — the mechanism has neither the refusal nor the permission.

Replace this with one coherent model, decided and recorded at the architecture layer before any code moves.

## Scope

Fix, at decision altitude, a single template registration model covering all of the following. Each item is a question an existing accepted ADR already owns; the answer is rewritten into that record rather than added beside it (see Constraints).

1. **Thin manifest.** Exactly four declared fields: `name`, `version`, a required non-empty `description`, and `ownership.excludedSubtrees`. `description` is the sole carrier of both selection semantics (what a caller with no reference chooses the template by) and post-instantiation usage semantics (apply-once-then-import vs apply-per-unit). Retire `schemaVersion`, `ownershipBoundaries` with both children, and `referencedTemplates`. Fix the identity rules `name` carries — uniqueness among registrable templates, the character prohibition, and the segment geometry that permits an npm-scoped `/`.

2. **One project state file.** A single git-tracked `.frontx/project.json` holding `formatVersion`, a `templates` map keyed by manifest name (each entry carrying `origin`, `version`, and a `targets` array), and `projectOwnedRoots`. It absorbs the per-instance provenance record, the project registry, and the anticipated ownership-exceptions file. An applied instance's identity is its unique normalized target — no synthetic instance identifier and no registry path. Every mutation is one atomic temp-file-plus-rename write.

3. **Whole-target ownership.** A template owns its entire applied target by default. Effective ownership is the target minus its declared `excludedSubtrees`, minus `projectOwnedRoots`, minus the template's own local origin folder, minus `.frontx`, minus environment-owned entries (`.git`, `.DS_Store`, `Thumbs.db`). This is one definition with one owner; every other artifact points at it rather than restating it. `.frontx/ai/<manifest-name>/` is a CLI-owned write, never claimable by a template.

4. **Explicit batch application.** Composition is the caller's explicit target-keyed batch, `{"templates": {"<manifestName>": ["<target>", ...]}}` — never a manifest naming its companions. `assemble` previews statelessly and writes nothing; `apply` independently re-derives and re-validates the same batch before materializing, trusting no prior preview and persisting no execution plan. Re-applying a recorded target is an idempotent no-op decided by record, not by content.

5. **Nesting-aware conflict prevention.** Containment between two targets is a conflict unless the inner target lies within the outer template's declared `excludedSubtrees`. Ancestor/descendant is decided over whole canonicalized path segments, never a string prefix. Path resolution is fail-closed against symlink and `..` escape. There is no `--force`.

6. **Registration and immutable origin pinning.** `register` resolves a template's origin and pins it — an exact commit SHA or package version, never the typed moving ref — and records it under the manifest name. A local `path:` origin is recorded as given and is not pinned. `register` creates the state file on its own first mutation, so no prior `seed` is required. `seed` is narrowed to auto-registering official defaults.

7. **Atomic all-targets upgrade.** The unit of an upgrade is a template name — every target it has been applied to, atomically, as one reviewable transition against the state file's single `origin`/`version` baseline. A partial upgrade is not a representable state.

8. **Uniform machine-readable output.** One discriminated-union JSON envelope, `{ok: true, data}` / `{ok: false, error}`, across every command, over one stable error-code vocabulary. No command invents a code of its own.

The command surface these decisions govern: `install`, `register`, `unregister`, `list`, `assemble`, `apply`, `seed`, `delete`, `upgrade`, `validate`, `ownership add|remove|list`.

## Constraints

- **Decisions land in the ADR that already owns the question.** A question an accepted ADR already answers is re-answered by rewriting that record in place. Do not add a superseding ADR beside it, and do not leave a `superseded` status behind: ADR names are outcome-agnostic precisely so the outcome can change without a new file, a new ID, and a repointing wave across every citation. A genuinely new question with no existing owner is the only case that warrants a new record.
- **No historical footprint.** Artifacts describe the current project state. No changelog prose, no dated "post-redesign" notes appended under a decision the note contradicts, no "previously"/"no longer"/"earlier revisions" narration. Where a decision changes, the document reads as though the current decision had always been the decision.
- **One definition, many pointers.** Each load-bearing definition — effective ownership, the batch shape, the error vocabulary, the manifest schema — has exactly one owning artifact. Everything else cites it. Independently drifting restatements of the same rule are the defect this redesign exists to remove, not a style preference.
- **No migration path for a legacy manifest.** A manifest carrying any retired field is refused outright by every command that reads one, `install` included, mirroring how a legacy `.frontx/provenance.json` is already refused rather than translated. The org owns both published templates and converts them by hand; a generic translation path would be permanent compatibility surface for a third-party consumer that does not exist.
- **Docs-first, gate-clean.** The deterministic gate (`cfs validate --skip-code`, `validate-toc`, `check-language`) passes over the full artifact set. Code-side traceability lag is expected while `packages/cli/src` still implements the previous model, and needs an explicit, recorded mechanism rather than a silently red required check.

## Acceptance criteria

- Every one of the eight decisions above is recorded in the ADR that owns its question, with Context, Decision Outcome, Consequences arguing honestly against the decision, and a Confirmation naming concrete CI fixtures.
- No artifact in the tree carries a `superseded` status or a `superseded_by` pointer as a result of this work, and no live artifact cites a retired record as a current authority.
- The retired vocabulary (`schemaVersion`, `ownershipBoundaries`, `exclusiveSubtrees`, `sharedFiles`, `referencedTemplates`, per-instance provenance record, `instanceId`, `registryPath`) appears nowhere as a live contract term.
- Every capability the previous model offered and this one does not is declared as a Consequence of the decision that drops it — specifically the loss of per-instance independent upgrade, and the loss of a specified rollback carrier for `upgrade`.
- Effective ownership, the batch shape, and the error-code vocabulary each read identically wherever they are cited, because each is cited rather than restated.
- The CLI PRD, the CLI DESIGN, the root DESIGN and DECOMPOSITION, and the affected FEATURE files agree with the decisions and with each other; `cfs validate --skip-code` and `validate-toc` pass over the full set.
- The conversion of `template-shell` and `template-mfe` onto the new contract is tracked as its own work item, with the behavioural change in `template-shell`'s ownership flagged for its own review rather than folded in as a mechanical edit.

## Out of scope

- **File-level changeset mechanics for upgrade** — diff representation, three-way merge, per-file conflict detection, and the concrete carrier that makes an applied upgrade reversible. This redesign fixes the unit and the baseline of an upgrade and leaves the mechanism to a dedicated decision; `upgrade`'s reversibility requirement stays in force in the PRD meanwhile.
- **The TypeScript rework of `packages/cli/src`** — a separate phase that follows these decisions.
- **The conversion of the two shipped templates** — tracked separately; this work only fixes the contract they convert onto.
- **Any change to the trust policy for template-declared AI extensions.**

gs-layer:
Closes #587.

## What

Docs-first redesign of the template registration system. Eight new ADRs (0035–0042) supersede the composed-preset model (ADR 0017, 0018, 0019, 0020, 0021, 0031, 0032); PRD, DESIGN, DECOMPOSITION and nine FEATURE files are reworked to match, and the kit package's PRD/DESIGN are migrated off the per-instance provenance model.

### Accepted decisions

- **Thin four-field manifest** — `name`, `version`, required non-empty `description` (carries both selection and usage semantics), `ownership.excludedSubtrees` (ADR 0035). `referencedTemplates`, `sharedFiles`/merge regions, `exclusiveSubtrees`, `schemaVersion` retired.
- **One project state document** — Git-tracked `.frontx/project.json` (`templates[name] = {origin, version, targets[]}` + `projectOwnedRoots`) replaces the registry file, per-instance provenance records, and any ownership file; applied-instance identity is the unique **target** (ADR 0036, 0038).
- **Whole-target ownership** — a template owns its entire target minus declared `excludedSubtrees`, `projectOwnedRoots`, environment-owned entries (`.git`, `.DS_Store`, `Thumbs.db`), local origin folders, and `.frontx`; AI-extension bundles under `.frontx/ai/<name>/` are materialized by the CLI itself as a CLI-owned step (ADR 0037).
- **Explicit target-keyed batch** — `assemble` is a stateless preview, `apply` re-derives everything at call time; re-apply is a record-based idempotent no-op (ADR 0038).
- **Nesting-aware, fail-closed conflict prevention** over canonicalized, segment-compared targets; no `--force` (ADR 0039).
- **Manifest-keyed registration with immutable origin pinning**; `path:` local origins; `unregister` refused while targets exist (ADR 0040).
- **Atomic all-targets upgrade** with the project-state entry as baseline; file-level changeset/restore mechanics deliberately deferred to a dedicated future ADR — the first item of the implementation phase (ADR 0041).
- **Uniform JSON envelope** `{ok,data}|{ok,error:{code,message,details}}` with a 16-code stable vocabulary; `CONFIRMATION_REQUIRED` protocol for destructive ops (ADR 0042).

### Review

The wave went through an external adversarial architecture review (1 blocker + 10 major + ~14 minor findings). Every finding was resolved by an explicit decision, incorporated, and independently verified against the documents. Known intentionally-open question: upgrade file-level diff/merge/restore mechanics (tracked in CLI DESIGN §4; upgrade FRs in the PRD are marked blocked-on that future decision).

## CI expectations

- The `validate` matrix (build, tests, `arch:guards`, `validate:templates`, lockfile policies) should stay green — the diff is docs-only and the shipped code/templates remain internally consistent on the old model.
- **`studio-validate` will be red by design**: it runs full code traceability, and `packages/cli/src` still implements the superseded model (~370 expected `code-orphan-ref`/`code-inst-*` findings). This is the docs-first gap; the implementation phase (starting with the deferred changeset ADR, then the CLI rework and the `template-shell`/`template-mfe` restructuring recorded in CLI DESIGN §4) closes it.

## Relation to #560

#560 refines the region-union `sharedFiles` semantics — a mechanism this redesign retires entirely (ADR 0035/0037/0039 supersede the two-tier boundary model). The two cannot both land as-is; per its author, #560 will likely be closed in favor of this wave.

## Summary by CodeRabbit

* **New Features**
  * Added documented support for template registration, local and pinned remote origins, explicit batch application, whole-target ownership, and atomic all-target upgrades.
  * Added a unified machine-readable JSON response format across CLI commands.
  * Added centralized project state tracking and ownership management.
  * Added nesting-aware conflict detection and registration-gated AI extension handling.

* **Documentation**
  * Updated product requirements, architecture decisions, workflows, and feature specifications to reflect the new template lifecycle and CLI behavior.
  * Marked superseded design decisions and documented their replacements.


architecture/ADR/0017-source-spec-syntax.md
```diff
@@ -1,5 +1,6 @@
 ---
-status: accepted
+status: superseded
```
itechmeat: 🔴 This is the first PR to keep superseded ADRs in this tree, and the repo has been here before: 5718cdb3 superseded ADR 0017 with a new record, and 569029ad folded it back with "SDLC artifacts describe the current project state with no historical footprint", repairing the pointers the supersession had broken. This wave redoes that at 7x scale - untouched live ADRs 0033 (the one CLAUDE.md leans on), 0025 and 0028 now cite superseded records as live authorities. The studio kit's own rules genuinely mandate what you did (ALWAYS supersede, NEVER edit an accepted decision - ARCH-ADR-008), so the kit and the house practice are in direct conflict, and 6 of the 8 new ADRs answer the same question as the record they replace. Could we settle the convention once, in writing, before this lands - fold-back per 569029ad, or an explicit local override of the kit rule? @GeraBart wdyt - whichever way, this PR should implement a decision rather than set the precedent silently.

architecture/ADR/0028-cli-internal-decomposition.md
```diff
@@ -93,7 +93,20 @@ Promote each concern to its own npm package.
...
...nent, `cpt-frontx-component-cli-registration`, is added to own the `register`/`unregister`/`install` behavior fixed by `cpt-frontx-adr-template-registration-and-origin-pinning`, a concern the original six-component roster had no place for because registration as a distinct step postdates it. `packages/cli/architecture/DESIGN.md`, where the CLI's internal component roster is detailed, reflects this seventh component. Reconciling the rest of the roster with the redesigned command surface (assemble/apply, delete, ownership) remains deferred to the DESIGN rework that accompanies this redesign wave. Two specific phrases in the Decision Outcome above are dated by that same redesign rather than merely by the roster's growth: the conflict checker's "pre-flight ownership-boundary intersection check" describes the exact-equality boundary model that `cpt-frontx-adr-nesting-aware-conflict-prevention` replaces, and the provenance recorder's "per-applied-template provenance write-at-apply / read-and-update-at-upgrade" describes the per-applied-template record that `cpt-frontx-adr-single-project-state-file` replaces — both are historical descriptions of superseded mechanics, not current behavior.
```
itechmeat: 🟠 Six live ADRs gain dated "Post-redesign note (2026-08-12)" blocks whose content is "the Decision Outcome above is no longer true" - while the Decision Outcome itself stays as written, so a reader of the live section gets the retired model as current (0028 and 0030 are the sharpest cases). The house rule wants the opposite: rewrite the outcome in place so the document reads as if the current decision had always been the decision - and this PR already does exactly that in architecture/DESIGN.md for the component roster. This is the same convention question as the supersession thread one level down: under the kit's "NEVER edit an accepted ADR" an appended note is the only legal move, so the append-vs-rewrite policy needs the same one-time decision rather than per-file patches.

packages/cli/architecture/features/cli-scaffolding/FEATURE.md
```diff
@@ -9,427 +8,486 @@
...
+
+Composition is no longer resolved from a manifest-declared reference graph: a batch names exactly the templates and targets a caller wants applied, and no template may pull in another through its own manifest. Ownership is no longer declared as a two-tier exclusive-subtree-plus-shared-file-region structure: a template owns its entire target by default, narrowed only by its manifest's `ownership.excludedSubtrees` and the project's `projectOwnedRoots`. There is no `sharedFiles` construct, no merge strategy, and no region marker of any kind — no two templates ever write into the same target, so there is nothing to compose at the file level.
```
itechmeat: 🟠 The house rule names these phrasings verbatim - no history, no changelog prose, no "previously" or "no longer" - and the added lines carry 136 of them (38 "no longer", 20 "previously", 3 "Earlier revisions", 5 "has since", 2 "predates"), plus DESIGN.md's ADR roster turned into a seven-entry changelog and 0030's explicit "Historical - retained as a traceability record" block. Whatever the ADR supersession decision turns out to be, the FEATURE/PRD/DESIGN files are not protected by the kit's immutability rule - these can be rewritten to present tense in place.

packages/cli/architecture/features/cli-scaffolding/FEATURE.md
```diff
@@ -9,427 +8,486 @@
...
+
+Earlier revisions of this feature resolved a template's manifest-declared preset (`referencedTemplates`) transitively and composed shared files from declared, merge-strategy-tagged regions; both are retired. Composition is now driven only by the caller's explicit batch, and a target is owned wholly by one template — there is no region-level co-ownership, no `compose-shared-files` algorithm, and no marker-delimited block to locate, carry forward, or refuse. Delete is owned here: the Assembler computes a deletion plan through the Conflict Checker's geometry.
```
itechmeat: 🟠 The instruction rewrite (inst-* ids 363 -> 503, only 186 shared) leaves 373 markers in packages/cli/src dangling - resolution drops from 791/797 to 418/797 - and the kit rename adds 4 more; the body's "~370 expected" is 393 in fact. Docs-first is legitimate and the FEATURE skeleton survived, so the repair is mechanical, but until the cli rework lands cfs validate cannot tell these 373 from real defects and the required gate just sits red with no waiver, threshold or annotation - the next PR cannot see its own findings under the wave (compare studio-spec-coverage in the same workflow, which rebases its bar with measured rationale when expectations change). Could this either stack behind the code MR, or carry an expected-orphan mechanism, or record the explicit team decision that develop tolerates a broken chain for the phase?

architecture/ADR/0035-thin-template-manifest.md
```diff
@@ -0,0 +1,134 @@
+---
+status: accepted
+date: 2026-08-12
+---
+
+# A Thin Manifest Where Description Carries All Selection and Usage Semantics
```
itechmeat: 🟡 [Not a blocker] All eight new ADRs are named for the outcome that won (thin-, single-, whole-target-, explicit-batch-, nesting-aware-, origin-pinning, atomic-all-targets-, uniform-) where the records they replace were mostly problem-named - and the house rule's own worked example of a problem name, 0017-source-spec-syntax, is the very file this PR marks superseded. An outcome name structurally guarantees the next model change cannot be rewritten in place either, i.e. it locks in a future supersession wave. Cheap to rename now, expensive later; free if the fold-back path is taken.

architecture/ADR/0039-nesting-aware-conflict-prevention.md
```diff
@@ -0,0 +1,137 @@
...
+* An ancestor/descendant relationship between two targets is a conflict, **unless** the inner (descendant) target lies strictly inside the outer (ancestor) template's declared `ownership.excludedSubtrees`, in which case the outer template has deliberately carved out that ground and the nested target is permitted. Ancestor/descendant is determined over **whole path segments of the canonicalized path, never a string prefix**: one target is an ancestor of another only when every one of its path segments matches the other's corresponding leading segments in order, so `packages/app` and `packages/app-shell` share no ancestor/descendant relationship despite one string being a character-prefix of the other — they are disjoint, sibling ground. An `excludedSubtree` must itself be a strict descendant of its own template's target (by the same segment-wise rule) and contain no `..` segment, and the template's own payload must not place files inside a subtree it has excluded from its own ownership — a template cannot declare a hole and then fill it.
```
itechmeat: 🟡 [Not a blocker] The rule says "strictly inside", but this ADR's own Confirmation fixture 4 accepts the case where the guest target coincides with the exclusion (src/config/ declared in excludedSubtrees) - which only passes under the looser reading, and three phrasings now circulate ("strictly inside", "inside", "falls entirely inside"). It is load-bearing: DECOMPOSITION has template-shell declare src-app/mfe_packages/ precisely so guests target it - under the strict reading that apply is refused, i.e. the planned conversion's main case does not work. One word decides it; worth picking the intended reading and aligning all three sites.

packages/cli/architecture/PRD.md
```diff
@@ -38,30 +41,33 @@
...
-| assembly | One CLI-driven apply operation that resolves the selected template set, checks ownership claims, and writes accepted project content. |
-| ownership boundary | The declared part of a repository that one template may create or modify. |
+| assembly | One CLI-driven operation — previewed statelessly by `assemble` and materialized by `apply` — that resolves an explicit batch of templates and targets, checks ownership claims, and writes accepted project content. |
+| origin | The immutable source a registered template resolves to: a remote source-spec pinned to the exact commit or package version it resolved to at registration time, or a local `path:` reference to a folder inside the project. |
+| target | The unique repository path a template is applied to. A target's identity is the identity of the applied instance; no separate instance identifier exists. |
+| ownership boundary | The part of a repository one template arbitrates for upgrade, delete, and conflict purposes: its entire target, minus the manifest's declared `excludedSubtrees`, minus the project's `projectOwnedRoots`, minus the template's own local origin folder, minus `.frontx`. |
```
itechmeat: 🟡 [Not a blocker] The owning definition (ADR 0037:57) subtracts six terms; this glossary subtracts five - dropping the environment-owned entries whose subtraction 0037:118 explicitly calls a restoration of a rule "silently dropped" last time - and DESIGN.md:55 drops it too while :190 calls the same subtraction unconditional. Downstream, the delete plan (cli-scaffolding FEATURE :312) and the DoD (:436) omit the local-origin-folder term the apply path (:241) includes - a path:-installed template whose origin folder sits under its own target lands in toDelete - and :436 subtracts excludedSubtrees from a quantity that already excludes it. Could this become one definition with pointers instead of four independently drifting spellings?

architecture/ADR/0037-whole-target-ownership.md
```diff
@@ -0,0 +1,145 @@
...
+
+Effective ownership of a target is therefore: **target − manifest `ownership.excludedSubtrees` − `projectOwnedRoots` − the template's own local origin folder (when installed by local path) − `.frontx` − environment-owned entries (`.git`, `.DS_Store`, `Thumbs.db`)**. This is the whole surface `delete` may remove for that template (itself excluding the same subtractions, plus any nested target and its own exclusions), and the whole surface a second template instance must stay out of. Nothing in this model tracks file content, hashes, or per-file state: once a file is written by `apply`, it is the project's, and the only thing the CLI still knows about it is which subtraction rule, if any, keeps it out of the next `delete` or the next competing template's ownership.
```
itechmeat: 🟡 [Not a blocker] template-shell applies at the repo root, so post-conversion its effective ownership is the entire repository minus exclusions - and delete on it removes everything a developer ever added at root unless they knew about ownership add beforehand, an opt-in learned after the loss. The CONFIRMATION_REQUIRED protocol with delete/preserve lists probably covers it in practice, but is the root-applied case part of the behavioural review DECOMPOSITION 2.22 already flags, or does 0037 need a named rule for it?

architecture/ADR/0041-atomic-all-targets-upgrade.md
```diff
@@ -0,0 +1,131 @@
...
+* Good, because atomicity across all targets of a name rules out a registry that claims one version while its targets sit at a mix of versions — a state the previous per-instance model could produce and that a validator would otherwise need to detect and explain.
+* Bad, because a developer cannot move one target of a multi-target template forward while leaving a sibling target behind (login on v2, registration still on v1); adopting a new version for any one target requires accepting it for all targets of that name, or not upgrading that name yet.
+* Bad, because an upgrade that touches several targets at once must succeed or fail as a whole; a conflict or ownership violation confined to one target blocks the version change for every other target of the same name until that target's issue is resolved.
+* Bad, because this mechanism does not itself implement the reviewable, non-destructive, reversible discipline ADR-0021 established: those properties carry forward as standing requirements on the future changeset decision (see More Information), not as consequences this ADR's atomic unit-and-baseline choice already delivers.
```
itechmeat: 🟡 [Not a blocker] 0021 goes superseded and its replacement honestly says it does not carry 0021's reviewable/reversible guarantees - handing them to a changeset ADR that does not exist yet. They do live on as requirements in the cli PRD (status notes at :268/:280 and the milestoned open item at :587), so nothing is lost - but a reader of 0041 cannot find that out from 0041. One pointer sentence there would keep "an upgrade must be reversible" discoverable as a requirement in force.

architecture/ADR/0039-nesting-aware-conflict-prevention.md
```diff
@@ -0,0 +1,137 @@
...
+
+Reliability treatment (REL): fail-closed is the mechanism's reliability design. **Failure modes** — an unresolvable or escaping path, a cross-template target collision, a containment collision without a matching exclusion, or a target landing on reserved project ground all surface before any file is written, through the same `TARGET_CONFLICT` code in both `assemble` and `apply`. **Non-destructive** — nothing is written until the full batch, checked against already-applied state, passes; a refused batch writes zero files. **Recovery** — a refused batch leaves the repository and `.frontx/project.json` exactly as they were; there is no partial state to roll back. **No override (REL/COMPL)** — an ownership conflict has no `--force`; the only path past a refusal is correcting the manifest (declaring an `excludedSubtree`) or the requested targets, never a flag.
```
itechmeat: 🔵 [Not a blocker] 0039 routes the canonicalization refusal through TARGET_CONFLICT while 0042, the cli DESIGN and the very algorithm this ADR governs (cli-scaffolding FEATURE :257/:399) all say INVALID_PATH - an implementer reading the owning ADR emits the wrong code. And VERSION_MISMATCH is fixed to validate --project and upgrade by 0042/DESIGN, but the upgrade-changeset FEATURE never names it - its validation routes to ORIGIN_UNAVAILABLE/REGISTRATION_CONFLICT/CONTENT_CONFLICT, so half of a stable code's meaning has no realization. Two one-line alignments.

architecture/ADR/0036-single-project-state-file.md
```diff
@@ -0,0 +1,139 @@
...
...mutation of `.frontx/project.json` is serialized through an atomic write — every command that mutates the document writes its full new content to a temporary file and renames it into place, so a reader never observes a partially-written document and two concurrent writers never interleave their bytes into a corrupt merge. This is also why an interrupted write cannot leave "no document" as an outcome distinct from "the prior document": a rename either lands, producing the new document whole, or does not, leaving the file the write started from untouched — there is no window in which the target path exists but is incomplete. A crash between a command materializing its target files and this document recording that write is recovered by re-running the same command: because re-applying a target already recorded in `project.json` is an idempotent no-op on the state write (`cpt-frontx-adr-explicit-batch-application`), and a target not yet recorded is applied as if for the first time, a re-applied command converges the document and the filesystem to the same consistent state regardless of which side of the crash the interruption fell on, with no separate crash-recovery procedure to invoke.
```
itechmeat: 🔵 [Not a blocker] The crash case is precisely "files written, record not written" - on retry that target goes down 0038's unrecorded-target existing-content path, where a half-written file reports as contentConflicts and anything extra blocks as additionalPaths, i.e. the interesting crash is the one that does not converge by re-running. The common case (bytes identical, lands in identicalFiles) does converge - but the claim is written unconditionally. Worth narrowing it, or naming what repairs a torn target.

architecture/ADR/0035-thin-template-manifest.md
```diff
@@ -0,0 +1,134 @@
...
...algorithmically as its entire target minus its declared `excludedSubtrees` (and minus the small set of project-level exclusions the ownership FEATURE also subtracts), so the manifest need declare only the one list that carries information the algorithm cannot derive on its own. `referencedTemplates`, `sharedFiles` with its merge strategies and region markers, `exclusiveSubtrees`, and `schemaVersion` are removed from the contract: composition is now performed by explicit application of each template rather than by one template naming the others it composes with, and total ownership by exclusion removes any need for a category distinguishing exclusive from shared ground. `description` is the sole carrier of semantics for both an AI agent and a human — what the template is, when to choose it, and how to use it once applied, including any apply-once-then-import discipline a template requires. A second structured field for the latter would duplicate the same audience's concern in two places that could disagree. The middle option (a dedicated `usage`/`aiCapabilities` field) was rejected on exactly this ground, and because it re-inflates the manifest the simplification is meant to shrink.
```
itechmeat: 🔵 [Not a blocker] This retires exactly what #560's ADR 0034 decides, the numbering here even skips 0034, but it is not in the supersession list - the body says #560 "will likely be closed in favor of this wave", so could that closure land first (or the supersession be recorded), and could the disposition of #507/#487 - which #560 claims to close - be stated explicitly so they do not fall between the two branches?

architecture/ADR/0018-template-manifest-contract.md
```diff
@@ -1,5 +1,6 @@
 ---
-status: accepted
+status: superseded
```
itechmeat: 🔵 [Not a blocker] The kit's ai-project-scaffolding FEATURE still grounds template selection in ADR 0018 (untouched line :44), under which description was optional - 0035 makes it required and non-empty, so the "entries that declare a description" partition at :192 is unreachable for any conforming manifest. Worth re-grounding on 0035 and dropping the dead partition while the wave is open.

packages/cli/architecture/DESIGN.md
```diff
@@ -510,40 +623,93 @@ sequenceDiagram
...
+    participant GH as Source registry
+    participant Chk as Conflict Checker
+    participant State as Project State Store
     participant Repo as Repository
-    Dev->>CLI: install + apply template(s) / preset (versioned source-spec)
-    CLI->>GH: resolve template(s) + a preset's referenced templates by reference
-    alt registry reachable and references resolvable
+    Dev->>CLI: list --json (defaults + registered + installed, with descriptions)
+    CLI-->>Dev: catalog
+    Dev->>Dev: compose explicit batch {"templates": {"<name>": ["<target>", ...]}}
```
itechmeat, line 611: 🔵 [Not a blocker] The two new sequence diagrams are the fastest-reading part of the wave - could we also get a one-page architect view of the whole model (manifest -> project.json -> ownership -> conflict -> upgrade, and which ADR fixes which edge)? And given this PR redefines the template mechanism end to end, it feels like one worth a synchronous call before it lands, even with the text looking right.

architecture/DECOMPOSITION.md
```diff
@@ -293,6 +302,47 @@ The installed SDLC kit currently defines feature-entry identifiers only in DECOM
...
+
+**Depends On**: `cpt-frontx-feature-template-manifest` (F11 — the four-field manifest contract both templates convert onto)
+
+**Scope** — five concrete changes, across both templates:
+
+1. Replace each legacy manifest (`schemaVersion` + `ownershipBoundaries`) with the current four-field shape: `name`, `version`, a required non-empty `description`, and `ownership.excludedSubtrees`.
+2. Widen `template-shell`'s ownership from its current file-subset claim to whole-target ownership (its target minus its declared `excludedSubtrees`) — extending what it owns, or moving previously unclaimed root files into `projectOwnedRoots`. This is a **behavioural change requiring its own review**, not a mechanical manifest edit, and is tracked here rather than folded into item 1.
+3. Remove both templates' own claim on `.frontx/ai/<manifest-name>/`; that bundle is a CLI-owned write (`cpt-frontx-adr-whole-target-ownership`), never template-owned.
+4. Add `src-app/mfe_packages/` to `template-shell`'s `ownership.excludedSubtrees` (a well-formed target-relative directory path with the trailing `/` `cpt-frontx-adr-thin-template-manifest` requires) and stop carrying payload inside it.
```
itechmeat: ⚪ [Not a blocker] DECOMPOSITION credits the trailing-slash requirement to the thin-manifest ADR, but 0035 explicitly defers field layout (:116) and never states it - the rule lives in template-manifest/FEATURE.md:140. One-line reattribution.

itechmeat (changes requested):
The direction here is genuinely good - a thin manifest, one project state file, no --force, no merge regions is a simpler model, and the internal consistency of the new spec holds up under close reading: the load-bearing definitions agree across their many restatements, the new ADRs argue honestly against themselves, and the retired vocabulary is swept. What blocks approval is the wave's form, not its content: it lands superseded ADR statuses and history prose that this repo's own artifact discipline forbids and has already reverted once before, and it leaves the code-to-spec traceability chain knowingly broken for the whole implementation phase with no mechanism around it. The studio kit's rules genuinely mandate the opposite of the house practice here, so this is a convention decision for the team to make explicitly - not a precedent to set silently by merging. With that settled (and the mechanical cleanups that follow from it), this is strong work I would expect to approve.

Out-of-diff notes:

- ⚪ [Not a blocker] The body's external adversarial-review claim ("1 blocker + 10 major + ~14 minor, independently verified") has no reachable artifact - nothing in the tree, no review on the PR beyond CodeRabbit, and the R1-R3/S1-S4/T1-T2/U1-U2 labels in commit subjects decode to nothing a reader can inspect. Worth attaching the review or dropping the claim.
- ⚪ [Not a blocker] Six of the 17 commit subjects describe the review process rather than the change ("address semantic review findings for PR #564", "close T1-T2 from the latest re-review") - the other half shows the intended form ("retire the generic legacy-manifest migration path").
- ⚪ [Not a blocker] The branch head is a merge commit ("Merge remote-tracking branch 'upstream/develop'...") where the convention for work branches is rebase-only; `git rebase --signoff` onto develop with a force-with-lease push fixes it.

architecture/ADR/0035-thin-template-manifest.md
GeraBart, line 1: I would suggest to rename the new ADRs to match the existing rule, which is ADR's name is related to the covered problem, not the outcome. So that we have flexibility to change the outcome without renaming files and changing traces id-s.

GeraBart:
All project's ADRs are mixed together now. I would suggest to move them to the corresponding package.

architecture/ADR/0017-source-spec-syntax.md
GeraBart: I agree, the project should not have any superseded ADRs at any time. The ADRs have outcome-agnostic names, so instead of superseding it's better to change the outcome. In case the ADR becomes completely inapplicable, it's better to remove it. Superseded ADR has no value to me.

architecture/ADR/0017-source-spec-syntax.md
```diff
@@ -172,6 +181,6 @@ This decision directly addresses the following requirements and design elements:
...
 * `cpt-frontx-fr-cli-template-update-local` — Bounded local update accepts a versioned reference as input, so its input surface carries this shape, including the substitution risk recorded in the consequences.
 * `cpt-frontx-fr-cli-template-list` — The inventory a developer lists may hold several templates acquired from one repository. The subtree distinguishes the references that acquired them, not the identities they are listed under: identity remains what each manifest declares (`cpt-frontx-adr-template-manifest-contract`), so two references differing only in subtree yield two entries when their manifests declare different identities, and a collision when they declare the same one.
-* `cpt-frontx-fr-cli-composed-template-resolution` — A preset's declared sibling references are the motivating case for addressing templates that share a repository, and they carry the shape this decision fixes.
+* The retired composed-template-resolution requirement (superseded by `cpt-frontx-adr-explicit-batch-application`) — a preset's declared sibling references were the motivating case for addressing templates that share a repository, and they carry the shape this decision fixes.
```
GeraBart: SDLC artifacts should not have any historical references, should never mention any expired past decisions.

architecture/ADR/0018-template-manifest-contract.md
```diff
@@ -26,81 +26,98 @@ date: 2026-06-05
...
...algorithmically as its entire target minus its declared `excludedSubtrees` (and minus the small set of project-level exclusions the ownership FEATURE also subtracts), so the manifest need declare only the one list that carries information the algorithm cannot derive on its own. `referencedTemplates`, `sharedFiles` with its merge strategies and region markers, `exclusiveSubtrees`, and `schemaVersion` are removed from the contract: composition is now performed by explicit application of each template rather than by one template naming the others it composes with, and total ownership by exclusion removes any need for a category distinguishing exclusive from shared ground. `description` is the sole carrier of semantics for both an AI agent and a human — what the template is, when to choose it, and how to use it once applied, including any apply-once-then-import discipline a template requires. A second structured field for the latter would duplicate the same audience's concern in two places that could disagree. The middle option (a dedicated `usage`/`aiCapabilities` field) was rejected on exactly this ground, and because it re-inflates the manifest the simplification is meant to shrink.
```
GeraBart: Why is `excludedSubtrees` nested under an `ownership` object instead of being a flat 4th field (`name`, `version`, `description`, `excludedSubtrees`)? Three of the four fields are flat scalars — worth stating explicitly why this one earns its own namespace object (room for future ownership-related keys? self-documentation?) rather than staying flat, especially given the whole point of this decision is minimalism.

architecture/DESIGN.md
```diff
@@ -103,15 +103,23 @@ Published libraries:
...
+* `cpt-frontx-adr-template-ownership-boundary-declaration` — Superseded by `cpt-frontx-adr-template-ownership-boundary-declaration`; previously defined the two-tier ownership-boundary declaration (exclusive subtrees plus shared-file region ownership with a declared merge).
+* `cpt-frontx-adr-assembly-conflict-prevention` — Superseded by `cpt-frontx-adr-assembly-conflict-prevention`; previously detected and refused conflicting assembly via exact-path-equality comparison of declared boundaries plus a post-materialization boundary-honesty guard.
+* `cpt-frontx-adr-composed-template-resolution` — Superseded by `cpt-frontx-adr-composed-template-resolution`; previously assembled a repository by resolving a preset's manifest-declared referenced templates transitively in one operation.
+* `cpt-frontx-adr-project-provenance-record` — Superseded by `cpt-frontx-adr-project-provenance-record`; previously recorded provenance as one record per applied template, with no single whole-repository origin.
```
GeraBart: Six of these read as a decision superseding itself — e.g. `cpt-frontx-adr-template-manifest-contract` — Superseded by `cpt-frontx-adr-template-manifest-contract` — same cpt-ID on both sides. Looks like a leftover from substituting the abandoned intermediate ADR-number citations (0035-0041) with the current cpt-IDs without noticing the fold-back means there's no distinct successor anymore. These decisions were rewritten in place, not superseded — these bullets should be deleted rather than fixed.

GeraBart:
Several places in the current diff still cite ADR numbers 0035–0041 by number ("ADR 0036", "ADR 0040", etc.) — e.g. `architecture/ADR/0018-template-manifest-contract.md` and `architecture/ADR/0028-cli-internal-decomposition.md`, plus the decision index in `architecture/DESIGN.md`. None of those files exist anywhere in this PR or the repo — they were dropped from an earlier revision when the 8 new ADRs got folded back into the original problem-named ADRs (e.g. the content now cited as "ADR 0036" actually lives in ADR 0019).

A reader following one of these citations today finds nothing. Since these are bare numbers in prose rather than `@cpt-` markers, `cfs validate` won't catch the dangling references — worth a repo-wide sweep for `ADR 003[5-9]|ADR 004[01]` before merge.

architecture/ADR/0017-source-spec-syntax.md
GeraBart: Re-surfacing this — it's still unresolved and the current diff still carries 88 occurrences of "previously"/"no longer"/"Historical"/"Superseded"/etc. across living ADR/DESIGN docs, concentrated in the "Post-redesign note" appendices added to ADR 0016/0026/0028 and the decision index in DESIGN.md. Could we settle this as an explicit decision before merge — rewrite those notes into the Decision Outcome in place, the way this PR already did for DESIGN.md's component roster?

GeraBart (approved):

GeraBart:
@gs-layer I have reviewed your proposal, and I support it. Currently this PR can't be merged due to failing CI jobs, but this is happening due to implementation lacking. IMO you can proceed with implementation based on this proposal.

PR #567, opened 2026-08-13
refactor(demo-mfe): extract bridge-property and RTL direction into shared hooks

tscbmstubp:
Implements #502 — follow-up from @itechmeat's review of #491.

## What

Extracts the pieces #491 pasted verbatim into the demo-mfe screens (`RTL_LANGUAGES`, `readBridgeProperty`, the property subscription, and the host `dir` effect) into two shared hooks in `demo-mfe/src/shared/`:

- **`useBridgeProperty<T>(bridge, propertyId, fallback)`** — lazy initial read on first render and the change subscription with cleanup. The value is passed through as the caller's `T`; schema validation stays with the type-system plugin.
- **`useHostDirection(containerRef, language)`** — Shadow DOM host `dir` sync keyed by language; no-op outside a shadow root.

All four demo-mfe screens (UIKitElements, Profile, CurrentTheme, HelloWorld) are rewired to the hooks; each duplicated ~60-line block becomes three lines.

## Tests

`useBridgeProperty` is pinned on initial read, fallback when the property is not yet published, non-string pass-through, update propagation, and unsubscribe on unmount. `useHostDirection` gets rtl/ltr transition and outside-shadow-DOM no-op coverage.

The bridge-swap cases (the `prevBridge` render-time re-read in the screens and the four screen tests pinning it) are removed rather than extracted: the runtime retains one bridge per extension across every mount/unmount cycle and `ThemeAwareReactLifecycle` renders a fresh root per mount, so a mounted screen never sees a different bridge instance (review round 3).

## What stays

Per the issue, `_blank-mfe/HomeScreen` keeps its inline copy and its test: `no-cross-mfe-imports` is error-severity, and per-MFE `src/shared/` duplication is the existing convention (`useScreenTranslations` precedent). No `src-app/mfe_packages/shared/` package is introduced.

## Verification

Reproduced the `template-validate` composition locally: overlaid `template-mfe`'s `exclusiveSubtrees` onto `template-shell` and ran `build`, `type-check`, `arch:deps`, `lint --max-warnings 0`, and `test:unit` over the union — all green (demo-mfe: 13 test files / 41 tests, including the new hook tests).

Behaviour change limited to the two round-3 removals above; otherwise a refactor.

## Summary by CodeRabbit

* **New Features**
  * Improved theme and language synchronization across demo screens.
  * Added automatic right-to-left and left-to-right layout direction based on the selected language.
  * Added consistent fallback handling when shared properties are unavailable.

* **Bug Fixes**
  * Improved responsiveness to theme and language property updates.

* **Tests**
  * Added coverage for property updates, fallback behavior, cleanup, and text-direction handling.

### DIFF
--- a/template-mfe/src-app/mfe_packages/demo-mfe/src/screens/helloworld/HelloWorldScreen.tsx
+++ b/template-mfe/src-app/mfe_packages/demo-mfe/src/screens/helloworld/HelloWorldScreen.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useRef, useState, useCallback } from 'react';
+import React, { useRef, useCallback } from 'react';
 import type { ChildMfeBridge } from '@gears-frontx/react';
 import { FRONTX_ACTION_MOUNT_EXT, FRONTX_SCREEN_DOMAIN, FRONTX_SHARED_PROPERTY_THEME, FRONTX_SHARED_PROPERTY_LANGUAGE } from '@gears-frontx/react';
 import {
@@ -11,6 +11,8 @@ import {
   Skeleton,
 } from '@gears-frontx/ui-kit';
 import { useScreenTranslations } from '../../shared/useScreenTranslations';
+import { useBridgeProperty } from '../../shared/useBridgeProperty';
+import { useHostDirection } from '../../shared/useHostDirection';
 import {
   THEME_EXTENSION_ID,
   PROFILE_EXTENSION_ID,
@@ -36,13 +38,6 @@ const languageModules = import.meta.glob('./i18n/*.json') as Record<
   () => Promise<{ default: Record<string, string> }>
 >;
 
-const RTL_LANGUAGES = ['ar', 'he', 'fa', 'ur'];
-
-function readBridgeProperty(bridge: ChildMfeBridge, property: string, fallback: string): string {
-  const current = bridge.getProperty(property);
-  return current && typeof current.value === 'string' ? current.value : fallback;
-}
-
 /**
  * Hello World Screen for the MFE remote.
  *
@@ -59,59 +54,12 @@ function readBridgeProperty(bridge: ChildMfeBridge, property: string, fallback: 
  */
 export const HelloWorldScreen: React.FC<HelloWorldScreenProps> = ({ bridge }) => {
   const containerRef = useRef<HTMLDivElement>(null);
-  // Initial value read directly from the bridge's lazy useState initializer (runs once,
-  // synchronously, during the first render) instead of via setState in a mount effect —
-  // this avoids an extra render and the set-state-in-effect anti-pattern. The effect
-  // below only subscribes for subsequent property changes.
-  const [theme, setTheme] = useState<string>(() =>
-    readBridgeProperty(bridge, FRONTX_SHARED_PROPERTY_THEME, 'default')
-  );
-  const [language, setLanguage] = useState<string>(() =>
-    readBridgeProperty(bridge, FRONTX_SHARED_PROPERTY_LANGUAGE, 'en')
-  );
-  // The lazy initializers above run only on mount; if the host swaps the bridge
-  // instance, re-read its current properties during render ("adjusting state
-  // during render") — the subscription effect only delivers future changes.
-  const [prevBridge, setPrevBridge] = useState(bridge);
-  if (prevBridge !== bridge) {
-    setPrevBridge(bridge);
-    setTheme(readBridgeProperty(bridge, FRONTX_SHARED_PROPERTY_THEME, 'default'));
-    setLanguage(readBridgeProperty(bridge, FRONTX_SHARED_PROPERTY_LANGUAGE, 'en'));
-  }
+  const theme = useBridgeProperty(bridge, FRONTX_SHARED_PROPERTY_THEME, 'default');
+  const language = useBridgeProperty(bridge, FRONTX_SHARED_PROPERTY_LANGUAGE, 'en');
+  useHostDirection(containerRef, language);
 
   // Load translations using the shared hook
   const { t, loading } = useScreenTranslations(languageModules, bridge);
-
-  useEffect(() => {
-    // Subscribe to theme domain property
-    const themeUnsubscribe = bridge.subscribeToProperty(FRONTX_SHARED_PROPERTY_THEME, (property) => {
-      if (typeof property.value === 'string') {
-        setTheme(property.value);
-      }
-    });
-
-    // Subscribe to language domain property
-    const languageUnsubscribe = bridge.subscribeToProperty(FRONTX_SHARED_PROPERTY_LANGUAGE, (property) => {
-      if (typeof property.value === 'string') {
-        setLanguage(property.value);
-      }
-    });
-
-    return () => {
-      themeUnsubscribe();
-      languageUnsubscribe();
-    };
-  }, [bridge]);
-
-  // Keep the Shadow DOM host's text direction in sync with the active language.
-  // An effect keyed by `language` (rather than logic inside the subscription
-  // callback) also covers the initial language, which never fires a callback.
-  useEffect(() => {
-    const rootNode = containerRef.current?.getRootNode();
-    if (rootNode && 'host' in rootNode) {
-      (rootNode.host as HTMLElement).dir = RTL_LANGUAGES.includes(language) ? 'rtl' : 'ltr';
-    }
-  }, [language]);
 
   // Navigate to Theme Screen
   const handleGoToTheme = useCallback(async () => {
--- a/template-mfe/src-app/mfe_packages/demo-mfe/src/screens/profile/ProfileScreen.tsx
+++ b/template-mfe/src-app/mfe_packages/demo-mfe/src/screens/profile/ProfileScreen.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useEffect, useRef, useState } from 'react';
+import React, { useCallback, useRef } from 'react';
 import type { ChildMfeBridge } from '@gears-frontx/react';
 import {
   FRONTX_SHARED_PROPERTY_THEME,
@@ -9,6 +9,8 @@ import {
 } from '@gears-frontx/react';
 import { Button, Card, CardContent, CardFooter, CardHeader, CardTitle, Skeleton } from '@gears-frontx/ui-kit';
 import { useScreenTranslations } from '../../shared/useScreenTranslations';
+import { useBridgeProperty } from '../../shared/useBridgeProperty';
+import { useHostDirection } from '../../shared/useHostDirection';
 import { kitThemeScopeFor } from '../../shared/kitThemeScope';
 import { AccountsApiService, type UpdateProfileVariables } from '../../api/AccountsApiService';
 import type { GetCurrentUserResponse } from '../../api/types';
@@ -33,13 +35,6 @@ const languageModules = import.meta.glob('./i18n/*.json') as Record<
   () => Promise<{ default: Record<string, string> }>
 >;
 
-const RTL_LANGUAGES = ['ar', 'he', 'fa', 'ur'];
-
-function readBridgeProperty(bridge: ChildMfeBridge, property: string, fallback: string): string {
-  const current = bridge.getProperty(property);
-  return current && typeof current.value === 'string' ? current.value : fallback;
-}
-
 /**
  * Profile Screen for the MFE remote.
  *
@@ -55,25 +50,9 @@ function readBridgeProperty(bridge: ChildMfeBridge, property: string, fallback: 
  */
 export const ProfileScreen: React.FC<ProfileScreenProps> = ({ bridge }) => {
   const containerRef = useRef<HTMLDivElement>(null);
-  // Initial value read directly from the bridge's lazy useState initializer (runs once,
-  // synchronously, during the first render) instead of via setState in a mount effect —
-  // this avoids an extra render and the set-state-in-effect anti-pattern. The effect
-  // below only subscribes for subsequent property changes.
-  const [theme, setTheme] = useState<string>(() =>
-    readBridgeProperty(bridge, FRONTX_SHARED_PROPERTY_THEME, 'default')
-  );
-  const [language, setLanguage] = useState<string>(() =>
-    readBridgeProperty(bridge, FRONTX_SHARED_PROPERTY_LANGUAGE, 'en')
-  );
-  // The lazy initializers above run only on mount; if the host swaps the bridge
-  // instance, re-read its current properties during render ("adjusting state
-  // during render") — the subscription effect only delivers future changes.
-  const [prevBridge, setPrevBridge] = useState(bridge);
-  if (prevBridge !== bridge) {
-    setPrevBridge(bridge);
-    setTheme(readBridgeProperty(bridge, FRONTX_SHARED_PROPERTY_THEME, 'default'));
-    setLanguage(readBridgeProperty(bridge, FRONTX_SHARED_PROPERTY_LANGUAGE, 'en'));
-  }
+  const theme = useBridgeProperty(bridge, FRONTX_SHARED_PROPERTY_THEME, 'default');
+  const language = useBridgeProperty(bridge, FRONTX_SHARED_PROPERTY_LANGUAGE, 'en');
+  useHostDirection(containerRef, language);
 
   const service = apiRegistry.getService(AccountsApiService);
 
@@ -123,36 +102,6 @@ export const ProfileScreen: React.FC<ProfileScreenProps> = ({ bridge }) => {
     },
     [updateProfile]
   );
-
-  // Subscribe to theme and language domain properties
-  useEffect(() => {
-    const themeUnsubscribe = bridge.subscribeToProperty(FRONTX_SHARED_PROPERTY_THEME, (property) => {
-      if (typeof property.value === 'string') {
-        setTheme(property.value);
-      }
-    });
-
-    const languageUnsubscribe = bridge.subscribeToProperty(FRONTX_SHARED_PROPERTY_LANGUAGE, (property) => {
-      if (typeof property.value === 'string') {
-        setLanguage(property.value);
-      }
-    });
-
-    return () => {
-      themeUnsubscribe();
-      languageUnsubscribe();
-    };
-  }, [bridge]);
-
-  // Keep the Shadow DOM host's text direction in sync with the active language.
-  // An effect keyed by `language` (rather than logic inside the subscription
-  // callback) also covers the initial language, which never fires a callback.
-  useEffect(() => {
-    const rootNode = containerRef.current?.getRootNode();
-    if (rootNode && 'host' in rootNode) {
-      (rootNode.host as HTMLElement).dir = RTL_LANGUAGES.includes(language) ? 'rtl' : 'ltr';
-    }
-  }, [language]);
 
   const kitThemeScope = kitThemeScopeFor(theme);
 
--- a/template-mfe/src-app/mfe_packages/demo-mfe/src/shared/useBridgeProperty.test.tsx
+++ b/template-mfe/src-app/mfe_packages/demo-mfe/src/shared/useBridgeProperty.test.tsx
@@ -1,0 +1,94 @@
+import { act, renderHook } from '@testing-library/react';
+import { FRONTX_SHARED_PROPERTY_THEME } from '@gears-frontx/react';
+import { createMfeBridgeFixture } from '@frontx-test-utils/createMfeBridgeFixture';
+import { describe, expect, it } from 'vitest';
+import { useBridgeProperty } from './useBridgeProperty';
+
+const TEST_DOMAIN_ID = 'test-domain';
+const TEST_INSTANCE_ID = 'test-instance';
+
+function themeBridgeFixture(initialProperties: Record<string, unknown> = {}) {
+  return createMfeBridgeFixture({
+    extDomainId: TEST_DOMAIN_ID,
+    extensionId: TEST_INSTANCE_ID,
+    initialProperties,
+  });
+}
+
+describe('useBridgeProperty', () => {
+  it('reads the initial value from the bridge on first render', () => {
+    const { bridge } = themeBridgeFixture({
+      [FRONTX_SHARED_PROPERTY_THEME]: 'initial-theme',
+    });
+
+    const { result } = renderHook(() =>
+      useBridgeProperty(bridge, FRONTX_SHARED_PROPERTY_THEME, 'fallback-theme')
+    );
+
+    expect(result.current).toBe('initial-theme');
+  });
+
+  it('returns the fallback when the property is not published', () => {
+    // No property registered at all: getProperty returns undefined.
+    const { bridge } = themeBridgeFixture();
+
+    const { result } = renderHook(() =>
+      useBridgeProperty(bridge, FRONTX_SHARED_PROPERTY_THEME, 'fallback-theme')
+    );
+
+    expect(result.current).toBe('fallback-theme');
+  });
+
+  it('passes non-string values through untouched', () => {
+    // Schema validation is the type-system plugin's job; the hook does not
+    // filter by runtime type.
+    const fixture = themeBridgeFixture({
+      [FRONTX_SHARED_PROPERTY_THEME]: { mode: 'dark' },
+    });
+
+    const { result } = renderHook(() =>
+      useBridgeProperty<{ mode: string }>(fixture.bridge, FRONTX_SHARED_PROPERTY_THEME, { mode: 'light' })
+    );
+
+    expect(result.current).toEqual({ mode: 'dark' });
+
+    act(() => {
+      fixture.setProperty(FRONTX_SHARED_PROPERTY_THEME, { mode: 'high-contrast' });
+    });
+
+    expect(result.current).toEqual({ mode: 'high-contrast' });
+  });
+
+  it('reacts to property updates', () => {
+    const fixture = themeBridgeFixture({
+      [FRONTX_SHARED_PROPERTY_THEME]: 'initial-theme',
+    });
+
+    const { result } = renderHook(() =>
+      useBridgeProperty(fixture.bridge, FRONTX_SHARED_PROPERTY_THEME, 'fallback-theme')
+    );
+
+    act(() => {
+      fixture.setProperty(FRONTX_SHARED_PROPERTY_THEME, 'updated-theme');
+    });
+
+    expect(result.current).toBe('updated-theme');
+  });
+
+  it('unsubscribes on unmount', () => {
+    const fixture = themeBridgeFixture({
+      [FRONTX_SHARED_PROPERTY_THEME]: 'initial-theme',
+    });
+
+    const { unmount } = renderHook(() =>
+      useBridgeProperty(fixture.bridge, FRONTX_SHARED_PROPERTY_THEME, 'fallback-theme')
+    );
+
+    unmount();
+
+    expect(fixture.unsubscriptions).toHaveLength(1);
+    for (const { unsubscribe } of fixture.unsubscriptions) {
+      expect(unsubscribe).toHaveBeenCalledTimes(1);
+    }
+  });
+});
--- a/template-mfe/src-app/mfe_packages/demo-mfe/src/shared/useBridgeProperty.ts
+++ b/template-mfe/src-app/mfe_packages/demo-mfe/src/shared/useBridgeProperty.ts
@@ -1,0 +1,62 @@
+/**
+ * useBridgeProperty Hook
+ *
+ * Subscribes a screen to one of the host's shared bridge properties
+ * (e.g. theme or language) and keeps the returned value in sync:
+ *
+ * 1. Reads the initial value lazily during the first render, so the first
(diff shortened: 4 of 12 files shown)


template-mfe/src-app/mfe_packages/demo-mfe/src/screens/profile/ProfileScreen.tsx
```diff
@@ -1,7 +1,7 @@
 // @cpt-FEATURE:request-cancellation:p2
 // @cpt-FEATURE:implement-endpoint-descriptors:p4
 
-import React, { useCallback, useEffect, useRef, useState } from 'react';
+import React, { useCallback, useRef } from 'react';
```
itechmeat, line 1: 🟠 ADR-0033 makes marker cleanup a review gate: when a template file is otherwise modified, every residue @cpt- marker it carries is removed, kept distinguishable from the substantive change (e.g. as its own commit). This PR rewires ProfileScreen (6 residue markers; the request-cancellation region also shrank, so its claim got narrower than at base) and HelloWorldScreen (2 markers) but keeps them all; none of the ids binds a live instruction under architecture/. The other two screens carry none, so the strip is 8 comment lines across two files.

template-mfe/src-app/mfe_packages/demo-mfe/src/shared/useBridgeProperty.ts
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * Hook returning the live value of a shared bridge property.
+ *
+ * @param bridge - ChildMfeBridge instance
+ * @param propertyId - Shared property id (e.g. FRONTX_SHARED_PROPERTY_THEME)
+ * @param fallback - Value used while the property is unset or non-string
+ * @returns The current string value of the property
+ */
+export function useBridgeProperty(
```
itechmeat: 🟡 [Not a blocker] @gears-frontx/react already exports useSharedProperty - the same getProperty + subscribeToProperty concept on useSyncExternalStore. It does not drop in here (it pulls the bridge from MfeContext, which demo-mfe never mounts, and has no fallback or non-string filtering), so a local hook is defensible - but two differently-named hooks for the same concept is the kind of pair that drifts silently. Worth either a line in this hook's doc naming the difference, or a follow-up about a bridge-argument form of useSharedProperty.

template-mfe/src-app/mfe_packages/demo-mfe/src/shared/useHostDirection.ts
```diff
@@ -0,0 +1,38 @@
...
+ * ```tsx
+ * const containerRef = useRef<HTMLDivElement>(null);
+ * const language = useBridgeProperty(bridge, FRONTX_SHARED_PROPERTY_LANGUAGE, 'en');
+ * useHostDirection(containerRef, language);
+ * ```
+ */
+
+import { useEffect, type RefObject } from 'react';
+
+const RTL_LANGUAGES = ['ar', 'he', 'fa', 'ur'];
```
itechmeat: 🟡 [Not a blocker] The convention this PR cites actually points the other way: useScreenTranslations.ts is byte-identical in demo-mfe/src/shared/ and _blank-mfe/src/shared/ - the precedent is "copy the hook file per MFE", not "keep the inline block". _blank-mfe is the starter people copy from, and it now teaches the pattern this PR just deleted. No cross-MFE import is needed to align it - copying the two hook files would follow the same precedent. #502 scoped this out deliberately, so a follow-up issue is fine - just worth filing.

template-mfe/src-app/mfe_packages/demo-mfe/src/shared/useBridgeProperty.ts
```diff
@@ -0,0 +1,69 @@
...
+  // synchronously, during the first render) instead of via setState in a mount effect —
+  // this avoids an extra render and the set-state-in-effect anti-pattern. The effect
+  // below only subscribes for subsequent property changes.
+  const [value, setValue] = useState<string>(() =>
+    readBridgeProperty(bridge, propertyId, fallback)
+  );
+  // The lazy initializer above runs only on mount; if the host swaps the bridge
+  // instance, re-read its current property during render ("adjusting state
+  // during render") — the subscription effect only delivers future changes.
+  const [prevBridge, setPrevBridge] = useState(bridge);
```
itechmeat: 🔵 [Not a blocker] The prevBridge dance is what useSyncExternalStore normally replaces - the shell's own hooks use it - but swapping it here would change behaviour: this hook deliberately keeps the last string value when a property updates to a non-string (test-pinned), which a re-reading getSnapshot cannot express. That reasoning is the one constraint worth a line in the hook's doc, so the next reader does not "simplify" it into a regression.

template-mfe/src-app/mfe_packages/demo-mfe/src/shared/useHostDirection.ts
```diff
@@ -0,0 +1,38 @@
...
+ * @param language - Active language code (e.g. 'en', 'ar')
+ */
+export function useHostDirection(
+  containerRef: RefObject<HTMLElement | null>,
+  language: string
+): void {
+  useEffect(() => {
+    const rootNode = containerRef.current?.getRootNode();
+    if (rootNode && 'host' in rootNode) {
+      (rootNode.host as HTMLElement).dir = RTL_LANGUAGES.includes(language) ? 'rtl' : 'ltr';
```
itechmeat, line 41: 🔵 [Not a blocker] RTL_LANGUAGES.includes(language) silently answers ltr for region-tagged codes - ar-EG, he-IL - and region tags are already in the tree (_blank-mfe ships zh-TW.json; LTR, so no live bug). Pre-existing behaviour, so rightly untouched by a behaviour-neutral refactor - but now that the predicate lives in one place, a follow-up splitting on "-" (or Intl.Locale textInfo) is cheap.

template-mfe/src-app/mfe_packages/demo-mfe/src/shared/useBridgeProperty.ts
```diff
@@ -0,0 +1,69 @@
...
+  if (prevBridge !== bridge) {
```
itechmeat: 🔵 [Not a blocker] The subscription effect declares [bridge, propertyId], so the hook advertises a changeable propertyId - but the render-time resync compares only bridges, so a changed propertyId would re-subscribe yet keep returning the old property's value until the new one happens to publish. Both call sites pass constants, so nothing breaks today; one JSDoc sentence pinning propertyId as constant for the component's lifetime would close the gap (or the resync compares it too).

template-mfe/src-app/mfe_packages/demo-mfe/src/screens/helloworld/HelloWorldScreen.tsx
```diff
@@ -44,60 +39,15 @@ function readBridgeProperty(bridge: ChildMfeBridge, property: string, fallback:
...
-  // The lazy initializers above run only on mount; if the host swaps the bridge
-  // instance, re-read its current properties during render ("adjusting state
-  // during render") — the subscription effect only delivers future changes.
-  const [prevBridge, setPrevBridge] = useState(bridge);
-  if (prevBridge !== bridge) {
-    setPrevBridge(bridge);
-    setTheme(readBridgeProperty(bridge, FRONTX_SHARED_PROPERTY_THEME, 'default'));
-    setLanguage(readBridgeProperty(bridge, FRONTX_SHARED_PROPERTY_LANGUAGE, 'en'));
-  }
+  // Theme and language mirror the host's shared bridge properties; the hook
```
itechmeat: 🔵 [Not a blocker] This two-liner is now pasted into all four screens and restates the hook's own JSDoc - the same one-text-in-N-places shape this PR exists to remove. The hook name already carries the message; the screens can drop it.

template-mfe/src-app/mfe_packages/demo-mfe/src/shared/useBridgeProperty.ts
```diff
@@ -0,0 +1,69 @@
...
+ * @param propertyId - Shared property id (e.g. FRONTX_SHARED_PROPERTY_THEME)
+ * @param fallback - Value used while the property is unset or non-string
+ * @returns The current string value of the property
+ */
+export function useBridgeProperty(
+  bridge: ChildMfeBridge,
+  propertyId: string,
+  fallback: string
+): string {
+  // Initial value read directly from the bridge's lazy useState initializer (runs once,
```
itechmeat: ⚪ [Not a blocker] Points 1-2 of the module doc block are restated nearly verbatim by the inline comments at lines 44-53 of the same 69-line file, and the surviving text is the generic React-docs retelling rather than a bridge-specific constraint. One copy - probably the header - seems enough.

template-mfe/src-app/mfe_packages/demo-mfe/src/shared/useHostDirection.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { describe, expect, it } from 'vitest';
+import { useHostDirection } from './useHostDirection';
+
+function DirectionProbe({ language }: { language: string }) {
+  const containerRef = useRef<HTMLDivElement>(null);
+  useHostDirection(containerRef, language);
+  return <div ref={containerRef} />;
+}
+
+function shadowMountNode() {
```
itechmeat, line 17: ⚪ [Not a blocker] The tree already has @frontx-test-utils/mockShadowHost - the three screen tests next door import it. The difference is real (this test wants an actual shadow root as the render container, not a spied getRootNode, which is the more faithful setup for this hook) - but if that is the reason, a one-line comment saying so would stop the next reader from "deduplicating" in either direction.

template-mfe/src-app/mfe_packages/demo-mfe/src/shared/useBridgeProperty.test.tsx
```diff
@@ -0,0 +1,101 @@
...
+    const { bridge } = themeBridgeFixture('initial-theme');
+
+    const { result } = renderHook(() =>
+      useBridgeProperty(bridge, FRONTX_SHARED_PROPERTY_THEME, 'fallback-theme')
+    );
+
+    expect(result.current).toBe('initial-theme');
+  });
+
+  it('returns the fallback when the property is unset or non-string', () => {
```
itechmeat: ⚪ [Not a blocker] The title claims "unset or non-string", but the fixture registers the key with value undefined - so neither the truly-unset half (getProperty returning undefined, the `current &&` branch) nor a non-string value is exercised by this case; the non-string branch is only touched inside the line-41 case, which itself bundles update propagation, non-string rejection and unmount teardown. Splitting per behaviour and adding the two missing inputs would make the titles honest.

template-mfe/src-app/mfe_packages/demo-mfe/src/shared/useHostDirection.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+    expect(host.dir).toBe('ltr');
+
+    host.remove();
+  });
+
+  it('is a no-op outside a Shadow DOM', () => {
+    const { container } = render(<DirectionProbe language="ar" />);
+
+    // The document root has no `host`, so no direction is written anywhere.
+    expect(container.closest('[dir]')).toBeNull();
```
itechmeat: ⚪ [Not a blocker] expect(container.closest('[dir]')).toBeNull() is true no matter what the hook does - the real signal is the documentElement.dir assertion below it. And host.remove() in the sibling case only runs when the assertions above pass, so a failure leaks the node into the next test; afterEach would be safer.

template-mfe/src-app/mfe_packages/demo-mfe/src/shared/useHostDirection.ts
itechmeat, line 41: ⚪ [Not a blocker] The `as HTMLElement` cast is genuinely forced ('host' in rootNode narrows Node only to Node & Record<'host', unknown>), but nothing in the file says so, and the house rule is that a cast carries its reason. One clause on the comment above covers it - the omission was copy-pasted from the screens, and the new module is the place to fix it.

template-mfe/src-app/mfe_packages/demo-mfe/src/shared/useBridgeProperty.ts
```diff
@@ -0,0 +1,69 @@
...
+ * ```tsx
+ * const theme = useBridgeProperty(bridge, FRONTX_SHARED_PROPERTY_THEME, 'default');
+ * const language = useBridgeProperty(bridge, FRONTX_SHARED_PROPERTY_LANGUAGE, 'en');
+ * ```
+ */
+
+import { useEffect, useState } from 'react';
+import type { ChildMfeBridge } from '@gears-frontx/react';
+
+function readBridgeProperty(bridge: ChildMfeBridge, propertyId: string, fallback: string): string {
```
itechmeat: ⚪ [Not a blocker] Three small consistency gaps against useScreenTranslations in the same directory: the private helper sits above the exported hook (house order is exported symbol first, helpers after), useHostDirection's module doc is a prose paragraph where its siblings use the numbered-list form, and RTL_LANGUAGES is a mutable string[] where a contract constant would be `as const`.

itechmeat (changes requested):
Clean, faithful extraction - exactly what #502 asked for. The behaviour-neutral claim survives a token-by-token comparison of all four screens against the hooks, with no drift against the deliberately kept _blank-mfe copy, and the hook test adds bridge-swap coverage demo-mfe did not have. One mechanical thing stands between this and approve: ADR-0033 requires a modified template file to shed its residue @cpt- markers, and two of the rewired screens keep theirs. The rest is questions and follow-up material - mostly about reusing what already exists and trimming restated comments - none of which should move the core code.

template-mfe/src-app/mfe_packages/demo-mfe/src/screens/profile/ProfileScreen.tsx
tscbmstubp: Done in ca636842 (separate commit): all 8 residue markers stripped from ProfileScreen and HelloWorldScreen.

template-mfe/src-app/mfe_packages/demo-mfe/src/shared/useBridgeProperty.ts
tscbmstubp: Added a "Related: useSharedProperty…" paragraph to the module doc naming the difference, and filed #570 for a bridge-argument form.

template-mfe/src-app/mfe_packages/demo-mfe/src/shared/useHostDirection.ts
tscbmstubp: Filed #571 to copy the two hook files (and tests) into _blank-mfe per the useScreenTranslations precedent.

template-mfe/src-app/mfe_packages/demo-mfe/src/shared/useBridgeProperty.ts
tscbmstubp: Documented in the module doc: "Deliberately not useSyncExternalStore: when a property updates to a non-string value the hook keeps the last string value (test-pinned), which a re-reading getSnapshot cannot express."

template-mfe/src-app/mfe_packages/demo-mfe/src/shared/useHostDirection.ts
tscbmstubp: Filed #572 for region-tagged codes (split on "-" or Intl.Locale textInfo), keeping this PR behaviour-neutral.

template-mfe/src-app/mfe_packages/demo-mfe/src/shared/useBridgeProperty.ts
tscbmstubp: Pinned in the @param propertyId JSDoc: must stay constant for the component's lifetime, with the reason (resync compares only bridge instances).

template-mfe/src-app/mfe_packages/demo-mfe/src/screens/helloworld/HelloWorldScreen.tsx
tscbmstubp: Dropped from all four screens — the hook name and its JSDoc carry it.

template-mfe/src-app/mfe_packages/demo-mfe/src/shared/useBridgeProperty.ts
tscbmstubp: Kept the header as the single copy; the inline restatements at the initializer and prevBridge blocks are gone.

template-mfe/src-app/mfe_packages/demo-mfe/src/shared/useHostDirection.test.tsx
tscbmstubp: Added the rationale comment on shadowMountNode: a real shadow root as the render container rather than mockShadowHost's getRootNode spy, so the hook's actual getRootNode() walk is exercised.

template-mfe/src-app/mfe_packages/demo-mfe/src/shared/useBridgeProperty.test.tsx
tscbmstubp: Split per behaviour: separate cases for truly-unset (no property registered), non-string value, update propagation, keeping the last string on a non-string update, and unmount teardown.

template-mfe/src-app/mfe_packages/demo-mfe/src/shared/useHostDirection.test.tsx
tscbmstubp: Dropped the always-true closest('[dir]') assertion (the documentElement.dir check remains) and moved host cleanup to afterEach so a failing assertion no longer leaks the node.

template-mfe/src-app/mfe_packages/demo-mfe/src/shared/useHostDirection.ts
tscbmstubp: Added the reason on the cast: the `in` check narrows Node only to Node & Record<'host', unknown>.

template-mfe/src-app/mfe_packages/demo-mfe/src/shared/useBridgeProperty.ts
tscbmstubp: Helper moved below the exported hook, module doc switched to the numbered-list form. For RTL_LANGUAGES I used `readonly string[]` instead of `as const`: a literal `as const` tuple makes `.includes(language: string)` a type error, and the readonly annotation gives the same immutability.

tscbmstubp:
Review addressed in two commits:

- **ca636842** — the blocker: all 8 residue `@cpt-` markers stripped from ProfileScreen and HelloWorldScreen (own commit per ADR-0033).
- **e3bc7427** — the polish: hook doc lines (useSharedProperty difference, the not-useSyncExternalStore constraint, propertyId pinned constant), inline/module-doc dedupe, cast rationale, house-order/doc-form/readonly consistency, the pasted screen comment dropped, and the test suites split per behaviour with the missing unset/non-string inputs, afterEach cleanup, and the always-true assertion removed.

Follow-ups filed: #570 (bridge-argument useSharedProperty), #571 (_blank-mfe hook copies), #572 (region-tagged RTL codes).

Verified locally: demo-mfe `test:unit` (44 passed) and `type-check` clean.

template-mfe/src-app/mfe_packages/demo-mfe/src/shared/useBridgeProperty.ts
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+ * Hook returning the live value of a shared bridge property.
+ *
+ * @param bridge - ChildMfeBridge instance
+ * @param propertyId - Shared property id (e.g. FRONTX_SHARED_PROPERTY_THEME).
+ *   Must stay constant for the component's lifetime: the render-time resync
+ *   compares only bridge instances, so a changed id would keep returning the
+ *   old property's value until the new one publishes.
+ * @param fallback - Value used while the property is unset or non-string
```
itechmeat: 🔵 [Not a blocker] Read alone, this @param says any non-string yields the fallback - but the module doc this round just sharpened (lines 16-18) states the opposite for updates: after a string has been seen, a non-string update keeps the last string value. One clause ("...on read; a later non-string update keeps the last string") would align the two.

template-mfe/src-app/mfe_packages/demo-mfe/src/shared/useBridgeProperty.test.tsx
```diff
@@ -0,0 +1,145 @@
...
+      useBridgeProperty(bridge, FRONTX_SHARED_PROPERTY_THEME, 'fallback-theme')
+    );
+
+    expect(result.current).toBe('fallback-theme');
+  });
+
+  it('returns the fallback when the property value is not a string', () => {
+    // Property registered, but its value is undefined rather than a string.
+    const { bridge } = themeBridgeFixture({
+      [FRONTX_SHARED_PROPERTY_THEME]: undefined,
```
itechmeat: ⚪ [Not a blocker] This case can only ever supply undefined: the fixture types input as Record<string, string | undefined>, while the production SharedProperty.value is unknown - so a number or object, the inputs the typeof filter actually exists for, stay untested here and in "keeps the last string value". The in-test comment honestly discloses it; widening means a fixture-type change shared with template-shell, so follow-up territory.

template-mfe/src-app/mfe_packages/demo-mfe/src/shared/useBridgeProperty.ts
```diff
@@ -0,0 +1,74 @@
...
+ *    paint already reflects the host's current property (no extra render
+ *    from a mount effect).
+ * 2. Re-reads the current value during render when the host swaps the
+ *    bridge instance ("adjusting state during render") — the subscription
+ *    only delivers future changes, never the new bridge's current state.
+ * 3. Subscribes to subsequent property changes and unsubscribes on
+ *    unmount or bridge/property change.
+ *
+ * Deliberately not useSyncExternalStore: when a property updates to a
+ * non-string value the hook keeps the last string value (test-pinned),
```
itechmeat: ⚪ [Not a blocker] "(test-pinned)" justifies the decision by the test that exists because of the decision - circular. The load-bearing reason is behavioural: a host publishing a non-string should not flash the UI back to 'default'/'en'. The mechanical explanation that follows is correct, so this is just a wording swap.

template-mfe/src-app/mfe_packages/demo-mfe/src/shared/useHostDirection.test.tsx
```diff
@@ -0,0 +1,56 @@
...
+    rerender(<DirectionProbe language="en" />);
+
+    expect(host.dir).toBe('ltr');
+  });
+
+  it('is a no-op outside a Shadow DOM', () => {
+    render(<DirectionProbe language="ar" />);
+
+    // The document root has no `host`, so no direction is written anywhere.
+    expect(document.documentElement.dir).toBe('');
```
itechmeat, line 54: ⚪ [Not a blocker] With the vacuous closest('[dir]') gone, this is the case's only assertion - and it also passes if the hook body never runs (jsdom default). It does still catch a hook that wrongly writes to documentElement, so it is weak rather than vacuous; noting it only because the round-1 fix traded one soft assertion for another rather than for a positive one.

itechmeat (approved):
All 13 threads from the previous round are addressed and verified by content, not by reply: the ADR-0033 marker strip is a clean deletions-only commit, the three follow-ups (#570, #571, #572) exist and say what they need to say, the test split genuinely separates the behaviours (the unset case now really exercises the absent-key branch), and the readonly string[] choice over as const is technically right - checked against the repo's TS 5.4. Four cosmetic notes below; none needs to gate the merge. Approving.

template-mfe/src-app/mfe_packages/demo-mfe/src/shared/useBridgeProperty.ts
tscbmstubp: Aligned in 7b58faa7 (amended into the polish commit): the @param now reads "…on read; a later non-string update keeps the last string value instead."

template-mfe/src-app/mfe_packages/demo-mfe/src/shared/useBridgeProperty.ts
tscbmstubp: Reworded in 7b58faa7: the doc now leads with the behavioural reason (a host publishing a non-string must not flash the UI back to the fallback) instead of "(test-pinned)".

template-mfe/src-app/mfe_packages/demo-mfe/src/shared/useBridgeProperty.test.tsx
tscbmstubp: Agreed — filed #573 for widening the fixture value type to `unknown` (shared with template-shell, so follow-up territory as you say).

template-mfe/src-app/mfe_packages/demo-mfe/src/shared/useHostDirection.test.tsx
tscbmstubp: Acknowledged — for a no-op case the strongest available check is "nothing gained a dir", and asserting documentElement stayed clean is the sharpest form of that jsdom offers. Leaving it as is.

template-mfe/src-app/mfe_packages/demo-mfe/src/shared/useBridgeProperty.ts
```diff
@@ -0,0 +1,76 @@
...
+export function useBridgeProperty(
+  bridge: ChildMfeBridge,
+  propertyId: string,
+  fallback: string
+): string {
+  const [value, setValue] = useState<string>(() =>
+    readBridgeProperty(bridge, propertyId, fallback)
+  );
+  const [prevBridge, setPrevBridge] = useState(bridge);
+  if (prevBridge !== bridge) {
```
GeraBart: this can never happen. bridge is preserved during consecutive re-mounts

template-mfe/src-app/mfe_packages/demo-mfe/src/shared/useBridgeProperty.ts
```diff
@@ -0,0 +1,76 @@
...
+      }
+    });
+  }, [bridge, propertyId]);
+
+  return value;
+}
+
+function readBridgeProperty(bridge: ChildMfeBridge, propertyId: string, fallback: string): string {
+  const current = bridge.getProperty(propertyId);
+  return current && typeof current.value === 'string' ? current.value : fallback;
```
GeraBart: why is that? shared property value is not restricted to string? also, shared property value validation against its schema is reponsibility of type system plugin

GeraBart (changes requested):

template-mfe/src-app/mfe_packages/demo-mfe/src/shared/useBridgeProperty.ts
tscbmstubp: Agreed — removed in ba7844fc (squashed into the PR commit). Confirmed against the runtime: `DefaultMountManager.mountExtension` hands the extension's retained bridge pair back into `acquireBridge` on every remount (the `MfeBridge` contract in handler/types.ts pins the bridge as stable across mounts, disposed only on unregister), and `ThemeAwareReactLifecycle` renders a fresh root per mount, so a mounted screen never sees a different bridge instance. Dropped the `prevBridge` branch, the bridge-swap hook test, and the four screen tests that pinned the same swap (those pre-dated this PR). The hook is now lazy init + one subscribe effect.

template-mfe/src-app/mfe_packages/demo-mfe/src/shared/useBridgeProperty.ts
tscbmstubp: Fair — no reason for it. In ba7844fc (squashed into the PR commit) the hook is generic (`useBridgeProperty<T>(bridge, id, fallback: T): T`) and passes `SharedProperty.value` through untouched; the only remaining fallback case is "property not yet published" (`getProperty` returns undefined). Schema validation stays with the type-system plugin, as you say. The non-string filter, the "keeps the last string" behaviour, their tests and the doc paragraph justifying them are gone; a pass-through test with an object value replaces them.

tscbmstubp:
Round 3 addressed and squashed into the single PR commit **ba7844fc** (both of @GeraBart's threads):

- **Bridge resync dropped.** The `prevBridge !== bridge` branch is unreachable: `DefaultMountManager` hands the retained bridge pair back into `acquireBridge` on every remount and `ThemeAwareReactLifecycle` renders a fresh root per mount. Removed the branch, its hook test, and the four pre-existing screen tests pinning the same swap.
- **String filtering dropped.** `useBridgeProperty` is now generic and passes `SharedProperty.value` through; the only fallback case is "not yet published". Schema validation stays with the type-system plugin.

PR title and description updated to match.

Verified locally: demo-mfe `type-check` clean; hook tests and the theme/uikit screen tests pass. The helloworld/profile screen tests fail identically on the untouched branch in my checkout (ui-kit and base-ui resolving to the root workspace's React copy), so that is an environment issue, not this change.

GeraBart (approved):

PR #569, opened 2026-08-13
fix(ui-kit): ship 'use client' per dist chunk, only where a hook is used

### LINKED ISSUE #568: ui-kit dist ships no 'use client' directives - hook-using components crash in React Server Components
## Problem

`@gears-frontx/ui-kit@0.3.0-alpha.2`'s dist chunks carry no `'use client'` banner — e.g. `dist/badge.js` starts straight with imports:

```js
import { jsxs as d, Fragment as v, jsx as p } from "react/jsx-runtime";
import { mergeProps as g } from "@base-ui/react/merge-props";
import { useRender as m } from "@base-ui/react/use-render";
```

In an RSC framework (Next.js App Router is the dominant consumer environment), a server component importing `Badge`, `Button`, `Select`, etc. straight from the package crashes at render — the component function executes on the server and calls a hook (`useRender`, Base UI internals). The failure is a runtime error in the consumer's page, not a build-time signal, so each consumer discovers it the hard way.

## What consumers do today

Every Next.js consumer has to hand-roll a client-boundary re-export module and route all kit imports through it, e.g. fabric-pass's `src/app/ui-kit.ts` (constructorfabric/fabric-pass#72):

```ts
'use client'
export { Badge, Button /* … */ } from '@gears-frontx/ui-kit'
```

That works, but it is per-app boilerplate for something only the library knows precisely — which modules actually need a client boundary — and a blanket wrapper forces even hook-free components client-side (see below).

## Proposal

Ship `'use client'` **per dist chunk, only where hooks are used** — the established practice across the ecosystem (Radix, MUI, Mantine, Chakra, react-aria-components all added dist banners once the App Router became mainstream):

- The build already emits one chunk per component (`dist/badge.js`, `dist/card.js`, …), which is exactly the right granularity. Vite/Rollup strips module-level directives from library output by default; `rollup-plugin-preserve-directives` (or writing the banner per entry) fixes that.
- Hook-free components — Card, Table, Skeleton — should **stay directive-free**: without a banner they render inside a server component with zero client-side JS, and marking them `'use client'` "just in case" takes that away from RSC consumers.
- The directive is inert for non-RSC bundlers (ignored/stripped, at worst a silenceable warning), so nothing regresses for Vite SPA consumers.

Worth a line in the README/llms.txt either way, stating which components are server-renderable.

Note: Badge is client-bound by design regardless — it unconditionally calls Base UI's `useRender` for its `render` prop. If a server-renderable Badge is ever wanted, that's an API-design question, not a directive one.

SysoevAndrey:
## Summary
- `@gears-frontx/ui-kit` dist chunks shipped with no `'use client'` directive, so a React Server Component (Next.js App Router) importing a hook-using component like Badge crashes at render.
- Adds a small Rollup plugin (`buildPlugin.ts`) that re-attaches `'use client'` to a chunk when its source module declared the directive — Rollup strips module-level directives during bundling, and `rollup-plugin-preserve-directives` requires `output.preserveModules`, which would break this build's one-chunk-per-component output shape.
- Only **Badge** (`useRender`), **DropdownMenu** (`useContext`), and **Toast** (`useToastManager`) call a hook directly in their own render body and get the directive. The other 16 components stay directive-free — they only compose Base UI primitives as JSX, and Base UI's own dist already carries `'use client'` on its hook-using modules, so they remain server-renderable for RSC consumers.
- `verify-consumer.sh` gets a regression guard that derives the full component list from the same glob the build uses and asserts every component is classified in exactly one of the client/server lists — a future component landing in neither (or both) fails loudly instead of shipping unclassified.
- `README.md`/`llms.txt` document which components are server-renderable vs. client-bound.

Fixes #568

## Test plan
- [x] `npm run build --workspace=@gears-frontx/ui-kit`
- [x] `npm run type-check --workspace=@gears-frontx/ui-kit`
- [x] `npm run test:unit --workspace=@gears-frontx/ui-kit` (284/284)
- [x] `bash packages/ui-kit/scripts/verify-consumer.sh` (all 5 legs, including new directive-banner guard)
- [x] `npm run lint` / `npm run arch:check` (12/12)
- [x] Byte-level inspection of all 20 dist chunks: directive present only on `badge.js`, `dropdown-menu.js`, `toast.js`
- [x] Vite SPA demo build unaffected, no directive-related warnings
- [ ] Not verified against a real Next.js App Router harness (none exists in this repo) — server-renderability of the other 16 components is verified structurally (no hooks + Base UI leaves self-mark), not by rendering

## Summary by CodeRabbit

* **New Features**
  * Added React Server Components guidance, clarifying which UI-kit components can render on the server and which require client-side execution.
  * Updated Badge, Dropdown Menu, and Toast for reliable client-side behavior.

* **Bug Fixes**
  * Improved build output so client-only component directives are preserved correctly.

* **Documentation**
  * Documented Server Components compatibility and usage guidance.

* **Tests**
  * Added checks to verify correct client and server component classifications in built bundles.

### DIFF
--- a/packages/ui-kit/README.md
+++ b/packages/ui-kit/README.md
@@ -39,6 +39,20 @@ unused JS bindings the way it prunes JS. If your build uses esbuild directly
 (not through a framework that wraps it, like Vite does), prefer the subpath
 import for CSS you actually want dropped.
 
+In an RSC framework (Next.js App Router being the dominant case), most kit
+components render straight from a Server Component with zero client-side
+JS — they only compose Base UI primitives as JSX, and Base UI's own dist
+already carries `'use client'` on the modules that call hooks. Three
+components call a hook directly in their own render body and ship a
+`'use client'` banner on their own chunk instead: **Badge** (`useRender`,
+for its `render` prop), **DropdownMenu** (`useContext`, for the portal
+container it shares with nested submenus), and **Toast** (`useToastManager`,
+for the live toast list `Toaster` renders). Everything else — including
+interactive primitives like Button, Checkbox, Dialog, Select, Switch,
+RadioGroup, and Tabs — stays server-renderable; each one's own client
+interactivity comes from a Base UI primitive it renders as a child, which
+establishes its own client boundary as needed.
+
 The theme file paints the page (background, text, and UA-owned surfaces like
 the scrollbar) as well as defining the tokens, so dark mode works out of the
 box with no attribute at all: it follows `prefers-color-scheme` by default.
--- a/packages/ui-kit/README.md
+++ b/packages/ui-kit/README.md
@@ -40,21 +40,19 @@ unused JS bindings the way it prunes JS. If your build uses esbuild directly
 import for CSS you actually want dropped.
 
 In an RSC framework (Next.js App Router being the dominant case), most kit
-components render straight from a Server Component with no kit-side client
-JS — they only compose Base UI primitives as JSX, and Base UI's own dist
-already carries `'use client'` on the modules that call hooks (an
-interactive primitive like Select or Dialog still ships Base UI's own
-client bundle; only the kit wrapper around it needs no directive of its
-own). Three
-components call a hook directly in their own render body and ship a
-`'use client'` banner on their own chunk instead: **Badge** (`useRender`,
-for its `render` prop), **DropdownMenu** (`useContext`, for the portal
-container it shares with nested submenus), and **Toast** (`useToastManager`,
-for the live toast list `Toaster` renders). Everything else — including
-interactive primitives like Button, Checkbox, Dialog, Select, Switch,
-RadioGroup, and Tabs — stays server-renderable; each one's own client
-interactivity comes from a Base UI primitive it renders as a child, which
-establishes its own client boundary as needed.
+components render straight from a Server Component with no kit-side client JS —
+they only compose Base UI primitives as JSX, and Base UI's own dist already
+carries `'use client'` on the modules that call hooks (an interactive primitive
+like Select or Dialog still ships Base UI's own client bundle; only the kit
+wrapper around it needs no directive of its own). Three components call a hook
+directly in their own render body and ship a `'use client'` banner on their own
+chunk instead: **Badge** (`useRender`, for its `render` prop), **DropdownMenu**
+(`useContext`, for the portal container it shares with nested submenus), and
+**Toast** (`useToastManager`, for the live toast list `Toaster` renders).
+Everything else — including interactive primitives like Button, Checkbox,
+Dialog, Select, Switch, RadioGroup, and Tabs — stays server-renderable; each
+one's own client interactivity comes from a Base UI primitive it renders as a
+child, which establishes its own client boundary as needed.
 
 The theme file paints the page (background, text, and UA-owned surfaces like
 the scrollbar) as well as defining the tokens, so dark mode works out of the
--- a/packages/ui-kit/README.md
+++ b/packages/ui-kit/README.md
@@ -40,9 +40,12 @@ unused JS bindings the way it prunes JS. If your build uses esbuild directly
 import for CSS you actually want dropped.
 
 In an RSC framework (Next.js App Router being the dominant case), most kit
-components render straight from a Server Component with zero client-side
+components render straight from a Server Component with no kit-side client
 JS — they only compose Base UI primitives as JSX, and Base UI's own dist
-already carries `'use client'` on the modules that call hooks. Three
+already carries `'use client'` on the modules that call hooks (an
+interactive primitive like Select or Dialog still ships Base UI's own
+client bundle; only the kit wrapper around it needs no directive of its
+own). Three
 components call a hook directly in their own render body and ship a
 `'use client'` banner on their own chunk instead: **Badge** (`useRender`,
 for its `render` prop), **DropdownMenu** (`useContext`, for the portal
--- a/packages/ui-kit/llms.txt
+++ b/packages/ui-kit/llms.txt
@@ -19,6 +19,12 @@ Rules for generated code:
   tree-shakes unused components' JS from a root import already, but only
   drops their CSS from a subpath import — Vite and webpack drop both either
   way, so this only matters when esbuild is the actual bundler running.
+- In an RSC framework (Next.js App Router), every component EXCEPT Badge,
+  DropdownMenu, and Toast is safe to import and render directly from a
+  Server Component — no wrapping client-boundary module needed. Those three
+  ship their own `'use client'` banner (each calls a hook — `useRender`,
+  `useContext`, `useToastManager` — directly in its own render body) and so
+  always render on the client, wherever they're imported from.
 - The kit paints the page itself from its tokens and sets `color-scheme`: with
   no `data-theme` attribute anywhere, `prefers-color-scheme: dark` alone
   yields a correct, legible dark page (background, text, and UA-owned
--- a/packages/ui-kit/package.json
+++ b/packages/ui-kit/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@gears-frontx/ui-kit",
-  "version": "0.3.0-alpha.2",
+  "version": "0.3.0-alpha.3",
   "description": "Standard React component base for FrontX templates - Base UI behavior, CSS Modules styling on semantic tokens, CVA variants",
   "type": "module",
   "types": "./dist/index.d.ts",
--- a/packages/ui-kit/package.json
+++ b/packages/ui-kit/package.json
@@ -56,6 +56,7 @@
     "@vitest/coverage-v8": "4.1.4",
     "esbuild": "0.25.12",
     "jsdom": "26.1.0",
+    "magic-string": "0.30.21",
     "react": "19.2.8",
     "react-dom": "19.2.8",
     "rollup-plugin-node-externals": "^9.0.1",
--- a/packages/ui-kit/scripts/buildPlugin.ts
+++ b/packages/ui-kit/scripts/buildPlugin.ts
@@ -153,10 +153,12 @@ export function buildPlugin(): PluginOption[] {
       renderChunk(code, chunk) {
         // Rollup has always already stripped the source directive by this
         // point (that's the whole problem this plugin exists to fix) — the
-        // `startsWith` check is a no-op today, kept only as a cheap guard
+        // `directiveRe` check is a no-op today, kept only as a cheap guard
         // against double-prepending if a future Rollup/Vite version stops
-        // stripping it.
-        if (code.startsWith(USE_CLIENT) || !chunk.moduleIds.some((id) => clientModuleIds.has(id))) {
+        // stripping it. Reuses `directiveRe` (rather than a plain
+        // `startsWith(USE_CLIENT)`) so this guard tolerates the same quote
+        // styles the detection above does.
+        if (directiveRe.test(code) || !chunk.moduleIds.some((id) => clientModuleIds.has(id))) {
           return null;
         }
 
--- a/packages/ui-kit/scripts/buildPlugin.ts
+++ b/packages/ui-kit/scripts/buildPlugin.ts
@@ -1,5 +1,6 @@
 import fs from 'node:fs';
 import path from 'node:path';
+import MagicString from 'magic-string';
 import nodeExternals from 'rollup-plugin-node-externals';
 import type { Plugin, PluginOption, UserConfig } from 'vite';
 import dts from 'vite-plugin-dts';
@@ -30,6 +31,7 @@ export function buildPlugin(): PluginOption[] {
     // possible at all.
     libInjectCss(),
     buildEntries(),
+    preserveUseClientPlugin(),
     dts({
       outDir: 'dist',
       entryRoot: 'src',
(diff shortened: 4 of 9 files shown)


SysoevAndrey:
Retargeting base branch to develop — reopening to refresh checks.

packages/ui-kit/package.json
```diff
@@ -56,6 +56,7 @@
     "@vitest/coverage-v8": "4.1.4",
     "esbuild": "0.25.12",
     "jsdom": "26.1.0",
+    "magic-string": "0.30.21",
```
itechmeat, line 59: 🟠 This is the only package.json edit - `version` stays `0.3.0-alpha.2`, which `8fd80600` already published, and publish-packages.yml only publishes when the version field changes. As merged, this PR ships nothing: the registry keeps the directive-less dist that #568 reports, while "Fixes #568" closes the issue. Is a bump to alpha.3 (or a stated separate publish step) the plan? Once it ships, fabric-pass#72's hand-rolled client boundary can be pointed at it.

packages/ui-kit/scripts/verify-consumer.sh
```diff
@@ -141,6 +141,84 @@ for probe_name in BUTTON_CLASS TABLE_CLASS DIALOG_CLASS; do
...
+# dist/, not a consumer bundle — an RSC framework like Next.js reads the
+# directive straight off the installed package's chunks, before any of this
+# script's bundler legs run, so none of them would catch it drifting. Split
+# by hook usage, not by any other property: badge/dropdown-menu/toast call a
+# hook directly in their own render body (useRender/useContext/
+# useToastManager — see buildPlugin.ts's preserveUseClientPlugin comment);
+# everything else only composes Base UI primitives as JSX, which already
+# carry their own directive in Base UI's dist, so marking them here would
+# take server-renderability away for no reason.
+CLIENT_COMPONENTS=(badge dropdown-menu toast)
```
itechmeat, line 154: 🟡 [Not a blocker] The guard ties dist banners to the lists and the lists to src/components/, but nothing ties the lists to what the sources actually do - add useState to select.tsx without the directive and every check here stays green, which is exactly the regression class #568 was. A cheap partial fix: grep each source for the directive and assert presence matches CLIENT_COMPONENTS. The forgotten-directive direction likely needs a lint rule and is fine as a follow-up - but the guard's comment is worth trimming to claim only what it covers.

packages/ui-kit/scripts/verify-consumer.sh
```diff
@@ -141,6 +141,84 @@ for probe_name in BUTTON_CLASS TABLE_CLASS DIALOG_CLASS; do
...
+done
+
+for name in "${CLIENT_COMPONENTS[@]}"; do
+  if ! head -c 20 "$UIKIT_DIR/dist/$name.js" | grep -qF "use client"; then
+    echo "FAIL: dist/$name.js is missing its 'use client' banner"
+    exit 1
+  fi
+done
+for name in "${SERVER_COMPONENTS[@]}" index; do
+  if head -c 20 "$UIKIT_DIR/dist/$name.js" | grep -qF "use client"; then
```
itechmeat, line 283: 🟡 [Not a blocker] Under set -euo pipefail a missing dist/<name>.js makes head exit non-zero and the if quietly false - so "no file at all" is reported as "checked, banner absent, good" and the success line still prints. assert_absent_from_glob a couple dozen lines below establishes the opposite convention in this same file (emptiness checked first, as a louder failure), and the client-side loop does not have the hole. A [ -f ] check before the head keeps this loop honest the same way.

packages/ui-kit/README.md
```diff
@@ -39,6 +39,20 @@ unused JS bindings the way it prunes JS. If your build uses esbuild directly
 (not through a framework that wraps it, like Vite does), prefer the subpath
 import for CSS you actually want dropped.
 
+In an RSC framework (Next.js App Router being the dominant case), most kit
```
itechmeat, line 42: 🟡 [Not a blocker] The RSC rationale is now written out four times - README, llms.txt, buildPlugin.ts's doc block and the verify-consumer.sh comment - while design-notes.md, which those same build comments point at for architecture, says nothing about RSC. One owner in design-notes.md's Architecture section with the other surfaces pointing at it would keep the copies from drifting; README keeping the consumer-facing half (which components are client-bound) makes sense.

packages/ui-kit/llms.txt
```diff
@@ -19,6 +19,12 @@ Rules for generated code:
   tree-shakes unused components' JS from a root import already, but only
   drops their CSS from a subpath import — Vite and webpack drop both either
   way, so this only matters when esbuild is the actual bundler running.
+- In an RSC framework (Next.js App Router), every component EXCEPT Badge,
```
itechmeat, line 22: 🔵 [Not a blocker] The shell arrays are guarded against src/components/ drift, but this sentence and README's paragraph hand-name the same three components with nothing checking them - a fourth client-bound component leaves both quietly wrong, and llms.txt is what agents read to generate consumer code. Deriving or spot-checking the prose list in the same guard would close it.

packages/ui-kit/scripts/buildPlugin.ts
```diff
@@ -105,6 +107,66 @@ export function buildPlugin(): PluginOption[] {
...
+        }
+        return null;
+      },
+      renderChunk(code, chunk) {
+        // Rollup has always already stripped the source directive by this
+        // point (that's the whole problem this plugin exists to fix) — the
+        // `startsWith` check is a no-op today, kept only as a cheap guard
+        // against double-prepending if a future Rollup/Vite version stops
+        // stripping it.
+        if (code.startsWith(USE_CLIENT) || !chunk.moduleIds.some((id) => clientModuleIds.has(id))) {
```
itechmeat: 🔵 [Not a blocker] directiveRe above deliberately tolerates both quote styles, but this guard checks only the single-quoted form - and esbuild rewrites the source directive to double quotes on transpile, so in the future-rollup scenario this check exists for it would not match and the plugin would prepend a duplicate anyway (harmless - two directives are still a valid prologue). Reusing directiveRe here, or dropping the check, makes it honest.

packages/ui-kit/scripts/buildPlugin.ts
```diff
@@ -105,6 +107,66 @@ export function buildPlugin(): PluginOption[] {
     };
   }
 
+  /**
+   * Rollup strips a source module's `'use client'` directive prologue when
+   * it renders that module's code into a bundled chunk — the directive
+   * survives esbuild's per-file TS transpile (see the component sources'
```
itechmeat: 🔵 [Not a blocker] The doc block points the reader at "the component sources' own comments" for the esbuild claim, but the three sources carry a bare directive and nothing else - the cross-reference resolves to nothing, and an editor of badge.tsx gets no local signal that line 1 is load-bearing and coupled to CLIENT_COMPONENTS in verify-consumer.sh. Either a one-line note in the three sources or dropping the parenthetical.

packages/ui-kit/README.md
```diff
@@ -39,6 +39,20 @@ unused JS bindings the way it prunes JS. If your build uses esbuild directly
...
+components render straight from a Server Component with zero client-side
```
itechmeat: ⚪ [Not a blocker] "Zero client-side JS" holds for Card, Table, Label, Separator, Skeleton but not for Dialog, Select, Tabs, Checkbox or Switch, which the same paragraph vouches for - those pull in Base UI client bundles. The closing clause partly walks it back and llms.txt avoids the claim entirely. A wording like "no kit-side client JS" keeps the sentence true as read.

packages/ui-kit/scripts/buildPlugin.ts
```diff
@@ -105,6 +107,66 @@ export function buildPlugin(): PluginOption[] {
...
+   * for no reason.
+   */
+  function preserveUseClientPlugin(): Plugin {
+    const USE_CLIENT = "'use client';";
+    // Matches the directive as the file's first statement, tolerating
+    // leading whitespace (which also covers a byte-order mark — `\s`
+    // matches U+FEFF), license/comment headers, and either quote
+    // style/semicolon — esbuild only preserves a directive prologue found
+    // in exactly that position.
+    const directiveRe = /^(?:\s|\/\/[^\n]*\n|\/\*[\s\S]*?\*\/)*['"]use client['"];?/;
```
itechmeat, line 143: ⚪ [Not a blocker] directiveRe's documented BOM/comment-header/quote-style tolerance is exercised by nothing - the repo has precedent for unit-testing build scripts (scripts/*.test.mjs) while packages/ui-kit/scripts has none. The regex does behave as documented (checked by hand), so this is only about pinning it against future edits.

itechmeat (commented):
The fix itself is the right mechanism, honestly scoped and carefully verified - the hook classification is exactly right, the plugin ordering is deliberate, and the no-Next.js-harness limitation is declared rather than hidden. The one thing I want settled before approving is delivery: the version stays on the already-published alpha.2 and the publish workflow only fires on a version change, so as merged this fixes #568 in the repo but ships nothing to the consumers who reported it. Everything else is non-blocking - two small holes in the new guard's coverage and some doc-scoping housekeeping.

packages/ui-kit/scripts/verify-consumer.sh
```diff
@@ -141,6 +141,132 @@ for probe_name in BUTTON_CLASS TABLE_CLASS DIALOG_CLASS; do
...
+# lists and of whatever the build produced, as a second, build-independent
+# signal tying the lists to source intent.
+for name in "${CLIENT_COMPONENTS[@]}"; do
+  if ! has_client_directive "$UIKIT_DIR/src/components/$name/$name.tsx"; then
+    echo "FAIL: src/components/$name/$name.tsx is classified as CLIENT_COMPONENTS but its source has no 'use client' directive as its first statement"
+    exit 1
+  fi
+done
+for name in "${SERVER_COMPONENTS[@]}"; do
+  if has_client_directive "$UIKIT_DIR/src/components/$name/$name.tsx"; then
```
itechmeat, line 262: 🟡 [Not a blocker] The dist loops gained a missing-file guard, but the source loops assume src/components/<name>/<name>.tsx exists - if the entry file is ever renamed or split, awk inside has_client_directive exits non-zero printing nothing, grep on empty input returns 1, and the SERVER absence check passes silently: the same "no file reads as clean" defect the dist loops below just fixed. A [ -f ] here too keeps both halves consistent (the CLIENT loop fails loudly in that case, but with a misleading message).

packages/ui-kit/scripts/verify-consumer.sh
```diff
@@ -141,6 +141,132 @@ for probe_name in BUTTON_CLASS TABLE_CLASS DIALOG_CLASS; do
...
+  awk '
+    /^[[:space:]]*$/ { next }
+    /^[[:space:]]*\/\// { next }
+    /^[[:space:]]*\/\*.*\*\/[[:space:]]*$/ { next }
+    { print; exit }
+  ' "$1" | grep -Eq "^[[:space:]]*('use client'|\"use client\");?[[:space:]]*\$"
+}
+
+# The lists are also only checked above against WHICH components exist, not
+# against what each one's source actually does — a hook added to (say)
```
itechmeat: 🔵 [Not a blocker] The comment motivates the new check with the scenario it cannot catch - useState added to select.tsx without the directive leaves select in SERVER_COMPONENTS, the dist bannerless and every check green. What the new loops actually pin is source-directive <-> list agreement, which is worth having - but worth saying precisely, so the paragraph does not read as covering the hook-without-directive direction (that one needs a lint rule).

packages/ui-kit/scripts/buildPlugin.ts
```diff
@@ -105,6 +107,68 @@ export function buildPlugin(): PluginOption[] {
     };
   }
 
+  /**
+   * Rollup strips a source module's `'use client'` directive prologue when
+   * it renders that module's code into a bundled chunk — the directive
+   * survives esbuild's per-file TS transpile (see the component sources'
```
itechmeat: 🔵 [Not a blocker] The new source comments explain why the directive is load-bearing and what it is coupled to - but the claim this line cites them for (surviving esbuild's per-file transpile) still appears in none of them. Either carry the esbuild claim here or add it to those notes; as is, the citation still resolves to nothing.

packages/ui-kit/README.md
```diff
@@ -39,6 +39,23 @@ unused JS bindings the way it prunes JS. If your build uses esbuild directly
 (not through a framework that wraps it, like Vite does), prefer the subpath
 import for CSS you actually want dropped.
 
+In an RSC framework (Next.js App Router being the dominant case), most kit
+components render straight from a Server Component with no kit-side client
+JS — they only compose Base UI primitives as JSX, and Base UI's own dist
+already carries `'use client'` on the modules that call hooks (an
+interactive primitive like Select or Dialog still ships Base UI's own
+client bundle; only the kit wrapper around it needs no directive of its
+own). Three
```
itechmeat: ⚪ [Not a blocker] Leftover reflow: the inserted parenthetical left a two-word line mid-paragraph in a file otherwise wrapped at ~76 columns. Renders fine - source-only nit.

itechmeat (approved):
🟢 Approving - the delta answers everything that gated round 1: the version moves to 0.3.0-alpha.3 so the publish gate trips, the guard is now tied to the sources in both directions (prologue-aware as of the latest commit), the dist loops fail loudly on a missing chunk, and the README claim is true as read. The threads on package.json and the dist absence check, plus the two now-outdated ones (quote guard, README wording), can be resolved. What's left below is small - one latent silent-pass in the new source loops, two wording points and a reflow nit, none gating. The untouched round-1 non-blockers (design-notes ownership, unguarded prose lists, directiveRe tests) stay as open follow-up threads.

GeraBart (approved):

PR #574, opened 2026-08-17
feat(react): add explicit-bridge options form to useSharedProperty (#570)

### LINKED ISSUE #570: Consider a bridge-argument form of useSharedProperty to subsume demo-mfe's useBridgeProperty
Follow-up from #567 review.

`@gears-frontx/react` exports `useSharedProperty`, and demo-mfe now carries a local `useBridgeProperty` — the same getProperty + subscribeToProperty concept. The local hook exists because `useSharedProperty` pulls the bridge from `MfeContext` (which demo-mfe never mounts) and has no fallback or non-string filtering. Two differently-named hooks for one concept can drift silently.

Proposal: add a bridge-argument form (and optional fallback/non-string filtering) to `useSharedProperty`, then retire the template-local hook. The difference is documented in `useBridgeProperty`'s module doc in the meantime.

tscbmstubp:
Closes #570.

## Why

The React package already ships `useSharedProperty` for reading a host-published property, but it only works inside `MfeProvider` because it takes the bridge from React context. Our MFEs receive the bridge as a plain argument at their entry point and never mount that provider, so they cannot use the shipped hook.

The result is that the template MFEs each carry their own copy of the same logic:

- demo-mfe has `useBridgeProperty` (extracted in #567).
- Both demo-mfe and `_blank-mfe` have a `useScreenTranslations` that subscribes to the bridge by hand.
- `_blank-mfe/HomeScreen` and demo-mfe's `CurrentThemeScreen` do the same subscription inline.

That is one concept written five times, each slightly different, and every copy is stamped into every new project generated from the template. The #567 review flagged this and opened #570.

This PR lets the shipped hook take the bridge directly, plus a fallback for before the host has published anything:

```ts
// before: only this form, needs MfeProvider above it
const theme = useSharedProperty('theme');

// after: also this form, works anywhere you hold a bridge
const theme = useSharedProperty('theme', { bridge, fallback: 'light' });
```

Existing callers are unaffected. Once this lands, the five local copies can be deleted (see Follow-up), leaving one hook, one set of tests, and one place to fix bugs.

## What

`useSharedProperty` in `@gears-frontx/react` (template-shell) gains an options form while keeping the zero-options context form intact:

```ts
useSharedProperty(propertyTypeId)                          // unchanged: bridge from MfeContext, throws outside MfeProvider
useSharedProperty(propertyTypeId, { bridge?, fallback? })
```

- `bridge`: an explicit `ChildMfeBridge` for MFEs that never mount `MfeProvider` (demo-mfe's case). Falls back to context when omitted. `useContext(MfeContext)` is still called unconditionally; only the resolved bridge is used conditionally.
- `fallback`: returned while the host has not published the property yet. The `{ fallback: T }` overload makes the return type non-optional. Only the `undefined` sentinel is substituted; a published `null` is a real value.

There is deliberately no consumer-side `select`/validation option. The host validates every published value against the property's schema before storing or delivering it (`updateSharedProperty` -> `TypeSystemPlugin.register` throws on a mismatch), so a malformed value never reaches an MFE. `T` is the caller's declaration of that schema, as with `useState<T>`. The TSDoc, CLAUDE.md and llms.txt say so.

## Notes for review

- **Implementation stays on `useSyncExternalStore`** (bridge as the store, `bridge.getProperty()` as the snapshot). The first render shows the host's current value, a publish between render and commit is caught by React's post-subscribe snapshot re-read, the value is tearing-protected, and a bridge or `propertyTypeId` change is reflected on the same render. `fallback` is only a `getSnapshot` dependency, so passing it inline does not resubscribe.
- **Error text changed**: the no-bridge throw now also names the explicit-`bridge` option.
- **ADR-0033**: template territory. Residual `@cpt-` markers were stripped from the two modified source files (`useSharedProperty.ts`, and `src/index.ts`, touched only to re-export `UseSharedPropertyOptions<T>`). No new markers were added.
- **Version bump**: `@gears-frontx/react` 0.2.0-alpha.6 -> 0.2.0-alpha.7 (new public API), with every exact pin moved in the same commit: template-shell's lockfile self-entry, the four template-mfe fixture manifests (`_blank-mfe`, `demo-mfe`, `widgets-fixture-a`, `widgets-fixture-b`) and their lockfile mirror entries. That is why template-mfe appears in the diff.

## Tests

Eight new cases in `__tests__/mfe/hooks.test.tsx` (package suite 158/158):

- throws outside `MfeProvider` with no explicit bridge
- reads from an explicit bridge without `MfeProvider`
- returns `fallback` while unpublished, then the published value
- resubscribes on bridge-instance swap
- resubscribes on `propertyTypeId` change on the same bridge
- picks up a publish landing between render and subscription
- treats a published `null` as a value, not as unpublished
- does not resubscribe when an inline `fallback` changes identity

Also green: `tsc --noEmit`, repo-root eslint, `policy:template-pin-drift`, `policy:version-check`, `policy:template-lockfile-selflink`, `policy:version-bump-on-change`.

## Follow-up

Once this ships, migrate the hand-rolled copies of the pattern:

- demo-mfe `useBridgeProperty` -> `useSharedProperty(id, { bridge, fallback })`.
- `_blank-mfe/HomeScreen` and both `useScreenTranslations` copies call `getProperty`/`subscribeToProperty` directly; `CurrentThemeScreen` carries the same lazy-init plus bridge-resync logic.

One assumption to verify there: the hook re-reads via `bridge.getProperty()` on notification while `useBridgeProperty` uses the callback payload. These are equivalent because each bridge capability delegates to the single runtime authority (ADR-0008).

### DIFF
--- a/template-mfe/src-app/mfe_packages/_blank-mfe/package.json
+++ b/template-mfe/src-app/mfe_packages/_blank-mfe/package.json
@@ -16,7 +16,7 @@
     "react": "19.2.4",
     "react-dom": "19.2.4",
     "@gears-frontx/frontx-template-shell": "0.1.0-alpha.4",
-    "@gears-frontx/react": "0.2.0-alpha.6",
+    "@gears-frontx/react": "0.2.0-alpha.7",
     "@gears-frontx/ui-kit": "0.4.0-alpha.2",
     "@tanstack/react-query": "5.101.4"
   },
--- a/template-shell/packages/react/__tests__/mfe/hooks.test.tsx
+++ b/template-shell/packages/react/__tests__/mfe/hooks.test.tsx
@@ -6,7 +6,7 @@
 
 import { describe, it, expect, vi } from 'vitest';
 import React from 'react';
-import { renderHook } from '@testing-library/react';
+import { act, renderHook } from '@testing-library/react';
 import { Provider as ReduxProvider } from 'react-redux';
 import { configureStore } from '@reduxjs/toolkit';
 import {
@@ -17,20 +17,32 @@ import {
   useHostAction,
   type MfeContextValue,
 } from '../../src/mfe';
-import type { ChildMfeBridge } from '@gears-frontx/framework';
+import type { ChildMfeBridge, SharedProperty } from '@gears-frontx/framework';
 
 // ============================================================================
 // Mock Data
 // ============================================================================
 
-const mockBridge: ChildMfeBridge = {
-  extDomainId: 'gts.frontx.mfes.ext.domain.v1~frontx.screensets.layout.sidebar.v1',
-  extensionId: 'test-instance-123',
-  executeActionsChain: vi.fn().mockResolvedValue(undefined),
-  subscribeToProperty: vi.fn().mockReturnValue(() => {}),
-  getProperty: vi.fn().mockReturnValue(undefined),
-  registerActionHandler: vi.fn(),
-};
+const TEST_DOMAIN_ID =
+  'gts.frontx.mfes.ext.domain.v1~frontx.screensets.layout.sidebar.v1';
+
+/**
+ * Build a ChildMfeBridge test double. Override only what the test cares
+ * about; everything else is an inert mock.
+ */
+function makeBridge(overrides: Partial<ChildMfeBridge> = {}): ChildMfeBridge {
+  return {
+    extDomainId: TEST_DOMAIN_ID,
+    extensionId: 'test-instance',
+    executeActionsChain: vi.fn().mockResolvedValue(undefined),
+    subscribeToProperty: vi.fn().mockReturnValue(() => {}),
+    getProperty: vi.fn().mockReturnValue(undefined),
+    registerActionHandler: vi.fn(),
+    ...overrides,
+  };
+}
+
+const mockBridge = makeBridge({ extensionId: 'test-instance-123' });
 
 const mockMfeContextValue: MfeContextValue = {
   bridge: mockBridge,
@@ -137,6 +149,181 @@ describe('MfeContext', () => {
 
       unmount();
       expect(unsubscribe).toHaveBeenCalled();
+    });
+
+    it('should throw when used outside MfeProvider and no explicit bridge is given', () => {
+      expect(() => {
+        renderHook(() =>
+          useSharedProperty('gts.frontx.mfes.comm.shared_property.v1~test.user_data.v1')
+        );
+      }).toThrow(/MfeProvider/);
+    });
+
+    it('should read from an explicit bridge without requiring MfeProvider', () => {
+      const propertyId = 'gts.frontx.mfes.comm.shared_property.v1~test.explicit_bridge.v1';
+      const explicitBridge = makeBridge({
+        extensionId: 'explicit-bridge-instance',
+        getProperty: vi.fn().mockReturnValue({ id: propertyId, value: 'hello' }),
+      });
+
+      // No MfeProvider in the tree -- only the option-supplied bridge is used.
+      const { result } = renderHook(() =>
+        useSharedProperty(propertyId, { bridge: explicitBridge })
+      );
+
+      expect(result.current).toBe('hello');
+      expect(explicitBridge.getProperty).toHaveBeenCalledWith(propertyId);
+    });
+
+    it('should return fallback while the property is unpublished, then the published value', () => {
+      const propertyId = 'gts.frontx.mfes.comm.shared_property.v1~test.fallback_unset.v1';
+      let subscribedCallback: ((property: SharedProperty) => void) | undefined;
+      let currentValue: unknown = undefined;
+
+      const explicitBridge = makeBridge({
+        subscribeToProperty: vi.fn((_id, callback) => {
+          subscribedCallback = callback;
+          return () => {};
+        }),
+        getProperty: vi.fn(() =>
+          currentValue === undefined ? undefined : { id: propertyId, value: currentValue }
+        ),
+      });
+
+      const { result } = renderHook(() =>
+        useSharedProperty<string>(propertyId, { bridge: explicitBridge, fallback: 'Untitled' })
+      );
+
+      expect(result.current).toBe('Untitled');
+
+      currentValue = 'Published';
+      act(() => subscribedCallback?.({ id: propertyId, value: 'Published' }));
+
+      expect(result.current).toBe('Published');
+    });
+
+    it('should resubscribe and reflect the new value when the bridge instance is swapped', () => {
+      const propertyId = 'gts.frontx.mfes.comm.shared_property.v1~test.bridge_swap.v1';
+      const unsubscribeA = vi.fn();
+      const unsubscribeB = vi.fn();
+
+      const bridgeA = makeBridge({
+        extensionId: 'instance-a',
+        subscribeToProperty: vi.fn().mockReturnValue(unsubscribeA),
+        getProperty: vi.fn().mockReturnValue({ id: propertyId, value: 'value-a' }),
+      });
+      const bridgeB = makeBridge({
+        extensionId: 'instance-b',
+        subscribeToProperty: vi.fn().mockReturnValue(unsubscribeB),
+        getProperty: vi.fn().mockReturnValue({ id: propertyId, value: 'value-b' }),
+      });
+
+      const { result, rerender } = renderHook(
+        ({ bridge }: { bridge: ChildMfeBridge }) =>
+          useSharedProperty(propertyId, { bridge, fallback: 'Untitled' }),
+        { initialProps: { bridge: bridgeA } }
+      );
+
+      expect(result.current).toBe('value-a');
+      expect(bridgeA.subscribeToProperty).toHaveBeenCalledTimes(1);
+
+      rerender({ bridge: bridgeB });
+
+      expect(unsubscribeA).toHaveBeenCalled();
+      expect(bridgeB.subscribeToProperty).toHaveBeenCalledTimes(1);
+      expect(result.current).toBe('value-b');
+    });
+
+    it('should resubscribe and reflect the new value when propertyTypeId changes on the same bridge', () => {
+      const propertyIdA = 'gts.frontx.mfes.comm.shared_property.v1~test.resync_prop_a.v1';
+      const propertyIdB = 'gts.frontx.mfes.comm.shared_property.v1~test.resync_prop_b.v1';
+      const unsubscribe = vi.fn();
+
+      const bridge = makeBridge({
+        subscribeToProperty: vi.fn().mockReturnValue(unsubscribe),
+        getProperty: vi.fn((id: string) => ({
+          id,
+          value: id === propertyIdA ? 'value-a' : 'value-b',
+        })),
+      });
+
+      const { result, rerender } = renderHook(
+        ({ propertyTypeId }: { propertyTypeId: string }) =>
+          useSharedProperty(propertyTypeId, { bridge }),
+        { initialProps: { propertyTypeId: propertyIdA } }
+      );
+
+      expect(result.current).toBe('value-a');
+
+      rerender({ propertyTypeId: propertyIdB });
+
+      expect(unsubscribe).toHaveBeenCalled();
+      expect(bridge.subscribeToProperty).toHaveBeenLastCalledWith(
+        propertyIdB,
+        expect.any(Function)
+      );
+      expect(result.current).toBe('value-b');
+    });
+
+    it('should pick up a publish landing between render and subscription', () => {
+      const propertyId = 'gts.frontx.mfes.comm.shared_property.v1~test.commit_window.v1';
+      let currentValue = 'published-before-render';
+
+      // The value changes at the moment the hook subscribes, simulating a
+      // publish that landed inside the render->commit window. Nothing ever
+      // fires the subscription callback, so the only way to observe the new
+      // value is useSyncExternalStore's post-subscribe snapshot re-read.
+      const bridge = makeBridge({
+        subscribeToProperty: vi.fn(() => {
+          currentValue = 'published-before-subscribe';
+          return () => {};
+        }),
+        getProperty: vi.fn(() => ({ id: propertyId, value: currentValue })),
+      });
+
+      const { result } = renderHook(() =>
+        useSharedProperty(propertyId, { bridge })
+      );
+
+      expect(result.current).toBe('published-before-subscribe');
+    });
+
+    it('should treat a published null as a value, not as unpublished', () => {
+      const propertyId = 'gts.frontx.mfes.comm.shared_property.v1~test.null_value.v1';
+      const explicitBridge = makeBridge({
+        getProperty: vi.fn().mockReturnValue({ id: propertyId, value: null }),
+      });
+
+      // `null` is a real value, distinct from "unpublished" (getProperty
+      // returning undefined), so it must NOT collapse to `fallback`.
+      const { result } = renderHook(() =>
+        useSharedProperty<string | null>(propertyId, { bridge: explicitBridge, fallback: 'Untitled' })
+      );
+
+      expect(result.current).toBeNull();
+    });
+
+    it('should not resubscribe when an inline fallback changes identity across renders', () => {
+      const propertyId = 'gts.frontx.mfes.comm.shared_property.v1~test.no_churn.v1';
+      const unsubscribe = vi.fn();
+      const explicitBridge = makeBridge({
+        subscribeToProperty: vi.fn().mockReturnValue(unsubscribe),
+        getProperty: vi.fn().mockReturnValue({ id: propertyId, value: 'value' }),
+      });
+
+      // A fresh fallback object each render must not tear down and recreate
+      // the host subscription.
+      const { rerender } = renderHook(() =>
+        useSharedProperty(propertyId, { bridge: explicitBridge, fallback: { label: 'Untitled' } })
+      );
+
+      expect(explicitBridge.subscribeToProperty).toHaveBeenCalledTimes(1);
+
+      rerender();
+      rerender();
+
+      expect(explicitBridge.subscribeToProperty).toHaveBeenCalledTimes(1);
+      expect(unsubscribe).not.toHaveBeenCalled();
     });
   });
 
--- a/template-shell/packages/react/llms.txt
+++ b/template-shell/packages/react/llms.txt
@@ -26,7 +26,7 @@ Part of the FrontX React Layer (L3) - depends on @gears-frontx/framework. Peer d
 | Hook | Purpose |
 |------|---------|
 | `useMfeBridge()` | Access MFE bridge (child MFEs) |
-| `useSharedProperty()` | Subscribe to shared property changes |
+| `useSharedProperty()` | Subscribe to shared property changes, via `MfeContext` or an explicit `bridge` option |
 | `useHostAction()` | Invoke actions on host application |
 | `useDomainExtensions()` | Subscribe to extensions in a domain |
 
@@ -96,6 +96,26 @@ function MyExtension() {
   return <div style={{ backgroundColor: theme?.primaryColor }}>...</div>;
 }
 ```
+
+### Shared Properties Without MfeProvider
+
+`useSharedProperty(propertyTypeId, options)` takes an explicit `bridge` for components that never mount `MfeProvider`, plus a `fallback` returned until the host publishes the property (only `undefined` is substituted; a published `null` is a value). With `fallback` set the result type is `T`. There is no consumer-side validation option: the host validates published values against the property's schema before delivery, so `T` is the caller's declaration of that schema.
+
+```tsx
+import { useSharedProperty } from '@gears-frontx/react';
+import type { ChildMfeBridge } from '@gears-frontx/framework';
+
+function StandaloneComponent({ bridge }: { bridge: ChildMfeBridge }) {
+  const label = useSharedProperty<string>(LABEL_PROPERTY_ID, {
+    bridge,
+    fallback: 'Untitled',
+  });
+
+  return <span>{label}</span>;
+}
+```
+
+Omitting `bridge` falls back to `MfeContext`; the hook still throws when neither is available. Changing `bridge` or `propertyTypeId` moves the subscription and the value follows on the same render.
 
 ### Subscribing to Domain Extensions
 
--- a/template-shell/packages/react/src/index.ts
+++ b/template-shell/packages/react/src/index.ts
@@ -76,6 +76,7 @@ export type {
   MfeContextValue,
   MfeProviderProps,
   ExtensionDomainSlotProps,
+  UseSharedPropertyOptions,
 } from './mfe';
 
 // ============================================================================
@@ -539,8 +540,6 @@ export type {
 import type { EventPayloadMap as FrameworkEventPayloadMap, EventBus } from '@gears-frontx/framework';
 import { eventBus as frameworkEventBus } from '@gears-frontx/framework';
 
-// @cpt-dod:cpt-frontx-dod-react-bindings-event-payload-map:p2
-// @cpt-begin:cpt-frontx-dod-react-bindings-event-payload-map:p2:inst-event-payload-map
 export interface EventPayloadMap extends FrameworkEventPayloadMap { }
 
 /**
@@ -549,4 +548,3 @@ export interface EventPayloadMap extends FrameworkEventPayloadMap { }
  * type-safe access to both framework events and app-layer augmented events.
  */
 export const eventBus: EventBus<EventPayloadMap> = frameworkEventBus as EventBus<EventPayloadMap>;
-// @cpt-end:cpt-frontx-dod-react-bindings-event-payload-map:p2:inst-event-payload-map
--- a/template-shell/packages/react/src/mfe/hooks/useSharedProperty.ts
+++ b/template-shell/packages/react/src/mfe/hooks/useSharedProperty.ts
@@ -1,76 +1,145 @@
 /**
  * useSharedProperty Hook - Shared property subscription
  *
- * Subscribes to shared property updates from the host.
+ * Subscribes to shared property updates from the host, either through the
+ * ambient MfeProvider context or through an explicitly supplied bridge.
  *
  * React Layer: L3
  */
-// @cpt-flow:cpt-frontx-flow-react-bindings-use-shared-property:p1
-// @cpt-algo:cpt-frontx-algo-react-bindings-mfe-context-guard:p1
-// @cpt-dod:cpt-frontx-dod-react-bindings-mfe-hooks:p1
 
-import { useSyncExternalStore, useCallback } from 'react';
-import { useMfeContext } from '../MfeContext';
+import { useCallback, useContext, useSyncExternalStore } from 'react';
+import type { ChildMfeBridge } from '@gears-frontx/framework';
+import { MfeContext } from '../MfeContext';
+
+// ============================================================================
+// Option Types
+// ============================================================================
+
+/**
+ * Options for {@link useSharedProperty}.
+ */
+export interface UseSharedPropertyOptions<T> {
+  /**
+   * Bridge to read the property from, bypassing `MfeContext`.
+   *
+   * Pass this when the calling component never mounts `MfeProvider` (for
+   * example, a component that receives its bridge directly as a prop or from
+   * a non-React entry point). When omitted, the hook falls back to the
+   * bridge from `MfeContext` and throws if no `MfeProvider` is mounted,
+   * matching the zero-argument form.
+   *
+   * `bridge` and `propertyTypeId` identify the subscription: changing either
+   * one unsubscribes from the old source and subscribes to the new one, and
+   * the returned value reflects the new source on the same render.
+   */
+  bridge?: ChildMfeBridge;
+
+  /**
+   * Value returned while the host has not published the property yet
+   * (`bridge.getProperty()` returns `undefined`).
+   *
+   * Only the `undefined` sentinel is substituted: a published `null` is a
+   * real value and is returned as-is. The host validates every published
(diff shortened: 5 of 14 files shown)


template-shell/packages/react/src/mfe/hooks/useSharedProperty.ts
```diff
@@ -1,76 +1,245 @@
...
+  // below only fires on the *new* bridge's future publishes, not its current value.
+  // Uses this render's own `select`/`fallback` closures (not the latest-ref
+  // above), since both are already current at this point in render.
+  const [prevSource, setPrevSource] = useState({ bridge, propertyTypeId });
+  if (prevSource.bridge !== bridge || prevSource.propertyTypeId !== propertyTypeId) {
+    setPrevSource({ bridge, propertyTypeId });
+    setValue(resolveValueOrFallback(bridge, propertyTypeId, select, fallback));
+  }
+
+  useEffect(() => {
```
itechmeat: 🟠 The subscription effect subscribes but never reads - the read happens earlier, during render, so a publish landing between render and effect commit is not picked up until the next publish. The previous `useSyncExternalStore` version had no such window since `getSnapshot` re-read on every render, and the window reopens on every resubscribe (bridge swap, StrictMode's dev remount). Re-reading `bridge.getProperty()` inside the effect right before subscribing, and setting state if it differs, closes it in about three lines - would that work instead of documenting the gap?

template-shell/packages/react/src/mfe/hooks/useSharedProperty.ts
```diff
@@ -1,76 +1,245 @@
...
+   * bridge from `MfeContext` and throws if no `MfeProvider` is mounted,
+   * matching the zero-argument form.
+   */
+  bridge?: ChildMfeBridge;
+
+  /**
+   * Value returned before any value has been accepted, and reused whenever
+   * `select` rejects a subsequent raw value (see keep-last-valid below).
+   *
+   * Not reactive once a real value has been accepted: changing `fallback`
```
itechmeat: 🟠 The `fallback` doc lists "a rejecting publish" as a moment the new value takes effect, but a rejecting publish never substitutes `fallback` - that branch is a deliberate no-op, which is the keep-last-valid contract stated ten lines below. The path that does substitute a fresh `fallback` (an unfiltered publish of a cleared property) is missing from the list, and CLAUDE.md carries the same contradiction: line 314 says `fallback` is reused whenever `select` rejects, line 317 says the opposite. Worth fixing in both, and `select`'s own non-reactivity is currently undocumented while `fallback`'s gets a paragraph.

template-shell/packages/react/src/mfe/hooks/useSharedProperty.ts
```diff
@@ -1,76 +1,245 @@
...
-  // @cpt-begin:cpt-frontx-flow-react-bindings-use-shared-property:p1:inst-read-bridge
-  // @cpt-begin:cpt-frontx-algo-react-bindings-mfe-context-guard:p1:inst-throw-no-mfe-context
-  // Enforce MfeProvider context requirement
-  const { bridge } = useMfeContext(); // Throws if not in MfeProvider
-  // @cpt-end:cpt-frontx-flow-react-bindings-use-shared-property:p1:inst-read-bridge
-  // @cpt-end:cpt-frontx-algo-react-bindings-mfe-context-guard:p1:inst-throw-no-mfe-context
+export function useSharedProperty<T = unknown>(propertyTypeId: string): T | undefined;
+export function useSharedProperty<T>(
+  propertyTypeId: string,
+  options: UseSharedPropertyOptions<T> & { fallback: T }
```
itechmeat, line 104: 🟡 [Not a blocker] The `fallback` overload types the result as non-optional `T`, but substitution only happens for `undefined` - without `select`, a published `null` (or any non-`T` raw value) comes straight through the `as T | undefined` cast. The PR's own null test does exactly this: it calls the hook with `fallback: 'Untitled'`, where the signature says `string`, and asserts `toBeNull()`. Should `fallback` only be accepted together with `select`, or should the doc stop calling the result guaranteed?

template-shell/packages/react/src/mfe/hooks/useSharedProperty.ts
```diff
@@ -1,76 +1,245 @@
...
  *
  * React Layer: L3
  */
-// @cpt-flow:cpt-frontx-flow-react-bindings-use-shared-property:p1
-// @cpt-algo:cpt-frontx-algo-react-bindings-mfe-context-guard:p1
-// @cpt-dod:cpt-frontx-dod-react-bindings-mfe-hooks:p1
 
-import { useSyncExternalStore, useCallback } from 'react';
-import { useMfeContext } from '../MfeContext';
+import { useContext, useEffect, useRef, useState } from 'react';
```
itechmeat: 🟡 [Not a blocker] The note says the ref-cached snapshot pattern is rejected by the React Compiler lint rules, but `useDomainExtensions` in this same folder writes `cacheRef.current` inside its `getSnapshot` and feeds it to `useSyncExternalStore` under the same config, and it is green - same for `useMountedExtensions` and `useRegisteredPackages`. Was it the pattern that got rejected, or the particular shape? Asking because every existing caller of the context form pays for the move with tearing protection, and if there is a real difference (the ref write here depends on the previous snapshot, theirs does not) that distinction is worth a sentence in the note.

template-shell/packages/react/src/mfe/hooks/useSharedProperty.ts
```diff
@@ -1,76 +1,245 @@
...
+    );
+  }
+
+  // Latest `select`/`fallback`, refreshed after every render (in an effect,
+  // never read/written during render -- this repo's react-hooks/refs lint
+  // rule forbids ref access during render). The subscription effect below
+  // reads through this ref instead of listing select/fallback as its own
+  // dependencies, so passing a fresh inline `select`/`fallback` each render
+  // does not tear down and recreate the host subscription.
+  const latestRef = useRef({ select, fallback });
```
itechmeat: 🟡 [Not a blocker] `useLatest` already lives in this package (`src/hooks/useLatest.ts`) with the same deps-less effect and the same render-time-ref-write rationale in its own doc block, and `useApiStream` uses it. Here the mechanism is rebuilt by hand along with a second copy of the explanation - two `useLatest` calls, or one over the packed object, would reuse it.

template-shell/packages/react/src/mfe/hooks/useSharedProperty.ts
```diff
@@ -1,76 +1,245 @@
...
+
+  useEffect(() => {
     return bridge.subscribeToProperty(propertyTypeId, () => {
-      // When property changes, trigger React re-render
-      callback();
+      const { select: currentSelect, fallback: currentFallback } = latestRef.current;
+      const next = resolveValue(bridge, propertyTypeId, currentSelect);
+      if (next !== undefined) {
+        setValue(next);
+      } else if (!currentSelect) {
```
itechmeat: 🟡 [Not a blocker] With `select` set, a host clearing the property is indistinguishable from a malformed publish - both surface as `undefined` from `resolveValue`, and the only path back to `fallback` is the `select`-less branch, so the screen keeps the stale value indefinitely. Without `select` the same clear falls back. Is that asymmetry deliberate? If so, the keep-last-valid paragraph currently frames rejection only as "a host publishing a malformed update" and could name the clear case too.

template-shell/packages/react/__tests__/mfe/hooks.test.tsx
```diff
@@ -138,6 +138,221 @@ describe('MfeContext', () => {
...
+          bridge: explicitBridge,
+          fallback: 'Untitled',
+          select: (value) => (typeof value === 'string' ? value : undefined),
+        })
+      );
+
+      expect(result.current).toBe('Untitled');
+    });
+
+    it('should keep the last valid string value when the host later publishes a non-string', () => {
```
itechmeat: 🟡 [Not a blocker] Since `select` rejects `12345`, the callback takes the no-op branch and no state update happens at all - this test passes identically if the subscription were never wired or if `setValue` were removed. No test drives a valid value through `select` via a publish, so "last accepted value" is only ever the initial render read. Publishing a valid second string first, asserting it lands, then publishing the non-string would make it discriminating.

template-shell/packages/react/src/mfe/hooks/useSharedProperty.ts
```diff
@@ -1,76 +1,245 @@
...
+  );
+
+  // Render-time resync: React's documented pattern for resetting derived state
+  // when an input changes without waiting an extra commit for an effect. Needed
+  // here because a bridge swap must be reflected immediately -- the subscription
+  // below only fires on the *new* bridge's future publishes, not its current value.
+  // Uses this render's own `select`/`fallback` closures (not the latest-ref
+  // above), since both are already current at this point in render.
+  const [prevSource, setPrevSource] = useState({ bridge, propertyTypeId });
+  if (prevSource.bridge !== bridge || prevSource.propertyTypeId !== propertyTypeId) {
```
itechmeat: 🔵 [Not a blocker] Resync on either input changing is promised in the TSDoc, in CLAUDE.md and by the code, but the only resync test swaps the bridge instance with a fixed id. A one-line variant rerendering with a changed `propertyTypeId` on the same bridge would cover the other half.

itechmeat: 🔵 [Not a blocker] The resync branch is a set-state-during-render guarded only by identity, so a `bridge` whose identity is not stable across renders now fires it every render and React bails with "too many re-renders" - the previous `useSyncExternalStore` version merely resubscribed. Since the option docs explicitly invite passing `select`/`fallback` inline, a line saying `bridge` must be referentially stable would keep that from reading as "everything here can be fresh".

template-shell/packages/react/src/mfe/hooks/useSharedProperty.ts
```diff
@@ -1,76 +1,245 @@
...
+ *   host to publish anything.
+ * - The hook resyncs at render time whenever the resolved bridge instance or
+ *   `propertyTypeId` changes (e.g. a host swapping the bridge given to an
+ *   MFE) -- the property subscription only delivers *future* changes, so a
+ *   bridge swap needs an explicit resync to reflect the new bridge's current
+ *   value immediately rather than waiting for its first publish.
  *
- * NOTE: This hook provides the interface and uses useSyncExternalStore with a stub subscription.
- * Full bridge subscription should be implemented when bridge methods are available.
+ * Implementation note (behavior delta vs. the pre-#570 context-form hook):
```
itechmeat: 🔵 [Not a blocker] The implementation note is framed against code that no longer exists ("pre-#570", "the implementation this replaced") and closes by justifying its own presence - a reader of the file cannot see the deleted version and needs the constraint (a publish between render and commit is picked up on the next publish; the value is not tearing-protected), not the comparison. The same `select`/`fallback` text is then repeated near-verbatim across the TSDoc, CLAUDE.md and llms.txt, and CLAUDE.md itself states the convention four lines above an analogous hook: the TSDoc is authoritative, the guide summarizes consumer usage.

template-shell/packages/react/src/mfe/hooks/useSharedProperty.ts
```diff
@@ -1,76 +1,245 @@
...
+ * `=== undefined`, not `??`/truthiness, or `null` will incorrectly collapse
+ * to `fallback`.
+ */
+function resolveValue<T>(
+  bridge: ChildMfeBridge,
+  propertyTypeId: string,
+  select: ((value: unknown) => T | undefined) | undefined
+): T | undefined {
+  const raw = bridge.getProperty(propertyTypeId)?.value;
+  return select ? select(raw) : (raw as T | undefined);
```
itechmeat: 🔵 [Not a blocker] This is the only cast in the file and it is now bare - the deleted implementation carried the justification verbatim (the caller specifies the expected `T`, as with `useState<T>`/`useContext<T>`, and type safety is the caller's responsibility). The doc block above explains the `undefined` sentinel, not why an unvalidated `unknown` may be asserted as `T`.

template-shell/packages/react/src/index.ts
```diff
@@ -76,6 +76,7 @@ export type {
   MfeContextValue,
   MfeProviderProps,
   ExtensionDomainSlotProps,
+  UseSharedPropertyOptions,
```
itechmeat, line 79: ⚪ [Not a blocker] The range from merge-base to head holds exactly one commit - the branch was squashed - and `d1ce6a6d`, named in the body as the separate marker-only commit, is not a valid object in the repo. So the ADR-0033 ask to keep the mechanical marker strip distinguishable from substantive change is not actually met (the strip rides with this export line, and in the hook it is inseparable from the rewrite). Also a small one: the diff adds eight new tests, not seven. Worth correcting the body rather than re-splitting the branch.

template-shell/packages/react/__tests__/mfe/hooks.test.tsx
```diff
@@ -138,6 +138,221 @@ describe('MfeContext', () => {
...
+          select: (value) => (typeof value === 'string' ? value : undefined),
+        })
+      );
+
+      expect(result.current).toBe('first-valid');
+
+      // Host publishes a non-string value; the hook must keep the last valid
+      // string instead of flashing back to `fallback`.
+      currentValue = 12345;
+      subscribedCallback?.({ id: propertyId, value: currentValue });
```
itechmeat: ⚪ [Not a blocker] The subscription callback is a state change originating outside React and is invoked bare here, relying on the following `rerender()` to flush - that produces "not wrapped in act" noise and ties the test to RTL's flush timing. The same double is also typed two incompatible ways across these tests (with a payload at :216, argument-less at :292), neither matching `ChildMfeBridge['subscribeToProperty']`.

template-shell/packages/react/src/mfe/index.ts
```diff
@@ -14,5 +14,6 @@ export {
   useMountedExtensions,
   useRegisteredPackages,
 } from './hooks';
+export type { UseSharedPropertyOptions } from './hooks';
```
itechmeat: ⚪ [Not a blocker] Every other export in this barrel carries its type inline (`export { MfeProvider, type MfeProviderProps }`, and so on), and the sibling barrel this PR also touches uses the inline form - this type could join the existing `from './hooks'` block a few lines above. `src/index.ts` does it the consistent way.

template-shell/packages/react/__tests__/mfe/hooks.test.tsx
```diff
@@ -138,6 +138,221 @@ describe('MfeContext', () => {
...
+      expect(() => {
+        renderHook(() =>
+          useSharedProperty('gts.frontx.mfes.comm.shared_property.v1~test.user_data.v1')
+        );
+      }).toThrow(/MfeProvider/);
+    });
+
+    it('should read from an explicit bridge without requiring MfeProvider', () => {
+      const propertyId = 'gts.frontx.mfes.comm.shared_property.v1~test.explicit_bridge.v1';
+      const explicitBridge: ChildMfeBridge = {
```
itechmeat: ⚪ [Not a blocker] There is already a `mockBridge` at the top of this file, and the new tests add six more full copies of the same literal - a small `makeBridge(overrides)` would cover them. Half of the new ones also use `domainId: 'domain'` while the rest of the file uses real gts ids.

template-mfe/src-app/mfe_packages/_blank-mfe/package.json
```diff
@@ -15,7 +15,7 @@
   "dependencies": {
     "react": "19.2.4",
     "react-dom": "19.2.4",
-    "@gears-frontx/react": "0.2.0-alpha.0",
+    "@gears-frontx/react": "0.2.0-alpha.1",
```
itechmeat: ⚪ [Not a blocker] `_blank-mfe/HomeScreen` and both `useScreenTranslations` copies still call `getProperty`/`subscribeToProperty` directly, and `CurrentThemeScreen` carries the same lazy-init plus previous-bridge resync plus string filter this hook generalizes. The cross-MFE import reason for keeping the blank copy goes away once the hook ships in `@gears-frontx/react`, which `_blank-mfe` already depends on - worth widening the follow-up beyond demo-mfe's `useBridgeProperty`.

itechmeat (changes requested):
This does exactly what #570 asked for, and it is refreshingly honest about its trade-offs - the behaviour delta from the useSyncExternalStore rewrite is written down rather than hidden, the marker cleanup is textbook, and the version pins are complete. Two things keep it from approve, both cheap and both about code disagreeing with its own description: the subscription effect never re-reads the value it subscribes to, so a publish landing between render and commit is lost until the next one, and the `fallback` doc names a substitution the code does not make while omitting the one it does. The rest is questions - a couple of them about whether the rewrite was as forced as the note says.

tscbmstubp:
All findings addressed in 9bd7649b (amended single commit, per the squash convention). Point-by-point:

**Blockers**

- **Subscription effect never re-reads** → fixed as suggested: the effect's callback logic is now a `syncFromBridge()` that runs once immediately before subscribing, closing the render→commit window on mount and on every resubscribe (bridge swap, StrictMode remount). Covered by a new test that changes the bridge value between the render-time read and the effect's re-read.
- **`fallback` doc names the wrong substitution** → fixed in the TSDoc and CLAUDE.md (both contradicting lines): a rejecting publish keeps the last accepted value and never re-substitutes `fallback`; the unfiltered clear is now the documented substitution path. `select`'s non-reactivity is documented alongside `fallback`'s.

**Questions / 🟡**

- **`fallback` overload types non-optional `T` but only `undefined` is substituted** → kept the overload (it matches the common `select`-paired usage and `useState<T>`-style caller responsibility), but the docs no longer call the result guaranteed: TSDoc, `@returns`, and CLAUDE.md now state the substitution is `undefined`-only and that the `T` guarantee holds at runtime only when paired with `select`.
- **Was the `useSyncExternalStore` rewrite forced?** → you were right that the distinction was missing: the sibling hooks' ref-cache is an idempotent memo whose key is recomputable from the store on every call; keep-last-valid is history-dependent ("last *accepted* value"), so `getSnapshot` can't compute it idempotently — a snapshot call from a discarded/replayed render would advance the cache for a commit that never happened. The implementation note now says exactly this instead of blaming the lint config.
- **Hand-rolled `useLatest`** → replaced with the package's `useLatest`, one call over the packed `{ select, fallback }` object.
- **Clear-vs-malformed asymmetry with `select`** → deliberate (the hook can't distinguish them once `select` maps both to `undefined`); now documented in the keep-last-valid paragraph, naming the clear case explicitly in both the "with select" and "without select" directions.
- **Non-discriminating keep-last-valid test** → now publishes a valid second string through the subscription, asserts it lands, then publishes the non-string.

**🔵 / ⚪**

- `propertyTypeId` resync test added (same bridge, changed id).
- `bridge` referential-stability requirement documented on the option (and contrasted with `select`/`fallback` in the no-memoization paragraph).
- Implementation note rewritten to state the constraints (no per-render tearing protection; pre-subscribe re-read) instead of comparing against deleted code; CLAUDE.md now defers to the TSDoc as authoritative per the package convention, dropping the duplicated paragraph.
- Cast justification restored on the `as T | undefined` line.
- PR body corrected: no phantom marker-only commit (the branch is one squashed commit; the commit message itemizes which deletions are mechanical), and the test count fixed (now 10 new tests, 150/150).
- Test doubles unified behind `makeBridge(overrides)` with real gts ids, `SharedProperty`-typed callbacks (matching `ChildMfeBridge['subscribeToProperty']`), and `act()`-wrapped publishes — no more RTL act noise.
- Barrel export made inline (`type UseSharedPropertyOptions` inside the `from './hooks'` block).
- Follow-up widened in the PR body to `_blank-mfe/HomeScreen`, both `useScreenTranslations` copies, and `CurrentThemeScreen`, not just demo-mfe's `useBridgeProperty`.

Checks after the round: `tsc --noEmit`, repo-root eslint, vitest 150/150, `tsup` build, dependency-cruiser — all green.

template-shell/packages/react/src/mfe/hooks/useSharedProperty.ts
```diff
@@ -1,76 +1,275 @@
...
+  fallback?: T;
+
+  /**
+   * Narrows/validates the raw `SharedProperty.value` (typed `unknown`).
+   * Return the narrowed value to accept it, or `undefined` to reject it.
+   *
+   * When a value is rejected, the hook keeps returning the last *accepted*
+   * value (or `fallback` if nothing has been accepted yet) instead of
+   * flashing back to `fallback` -- a host publishing a malformed update
+   * should not visibly regress the UI. A host *clearing* the property is
```
itechmeat: 🟡 [Not a blocker] This says that with `select` set a cleared property is indistinguishable from a rejected one and the hook keeps the last accepted value - true on the publish path, but the initial read and the resync both go through `resolveValueOrFallback`, which substitutes `fallback` for `undefined` even with `select` set. So at mount, or on a bridge/`propertyTypeId` swap, a cleared property does fall back, and the `fallback` paragraph ten lines above says so correctly. Adding "on a publish" here (and in CLAUDE.md's closing paragraph, which carries the same unqualified sentence - its `fallback` bullet gets it right) would settle it.

template-shell/packages/react/src/mfe/hooks/useSharedProperty.ts
```diff
@@ -1,76 +1,275 @@
...
+    const syncFromBridge = () => {
+      const { select: currentSelect, fallback: currentFallback } = latestRef.current;
+      const next = resolveValue(bridge, propertyTypeId, currentSelect);
+      if (next !== undefined) {
+        setValue(next);
+      } else if (!currentSelect) {
+        // No filter configured: `undefined` is itself the correct new value
+        // (property was unset/cleared), matching the hook's original
+        // unfiltered behavior -- substitute `fallback` if one was given.
+        setValue(currentFallback);
```
itechmeat: 🟡 [Not a blocker] Three mutations at head leave 150/150 green: commenting out this `setValue` (so the unfiltered-clear-to-`fallback` path is unguarded), swapping `resolveValueOrFallback` for `resolveValue` in the resync branch on line 245, and - for the resubscribe half of the pre-subscribe read - deleting `syncFromBridge()` fails only the mount test, since the bridge-swap and `propertyTypeId` tests use a constant `getProperty`. This is the same critique this round applied to the keep-last-valid test, one level down: the docs got precise, the tests did not follow.

template-shell/packages/react/src/mfe/hooks/useSharedProperty.ts
```diff
@@ -1,76 +1,275 @@
...
+      }
+      // else: `select` rejected the new value (or the property is unset) --
+      // keep the last accepted one.
+    };
+    // Sync once before subscribing: a publish landing between the render
+    // that seeded/resynced the state and this effect's commit would
+    // otherwise be missed until the next publish. The same window reopens on
+    // every resubscribe (bridge swap, StrictMode's dev remount), so the
+    // pre-subscribe sync runs on each of them too.
+    syncFromBridge();
```
itechmeat: 🟡 [Not a blocker] `syncFromBridge` calls `setValue` unconditionally, so React's `Object.is` bailout is the only guard - with a non-allocating `select` that is one render and two `getProperty` calls, but with an allocating one (a spread, a `zod.parse()`, any narrower returning a fresh object) it is two renders on mount and five instead of two on a bridge swap. Since the docs explicitly invite passing `select` inline without memoizing, would it be worth either noting the cost on `select` or gating the pre-subscribe `setValue` on a real difference from what render already seeded?

template-shell/packages/react/__tests__/mfe/hooks.test.tsx
```diff
@@ -138,6 +150,250 @@ describe('MfeContext', () => {
...
+        ({ propertyTypeId }: { propertyTypeId: string }) =>
+          useSharedProperty(propertyTypeId, { bridge }),
+        { initialProps: { propertyTypeId: propertyIdA } }
+      );
+
+      expect(result.current).toBe('value-a');
+
+      rerender({ propertyTypeId: propertyIdB });
+
+      // Same bridge instance, new property id: the old subscription is torn
```
itechmeat: 🟡 [Not a blocker] Both comments attribute the result to the render-time resync, but `rerender` is act-wrapped, so effects flush before the assertion - delete the `prevSource` resync block and both tests stay green, since the value arrives through the effect anyway. The resubscribe-and-eventually-resync part is well covered; the reason the extra `prevSource` state exists (no stale-value paint in the same commit) is not, and the comments say it is. Either softening the comments to what is asserted or adding a render-order assertion would keep someone from simplifying the block away on a green suite.

template-shell/packages/react/llms.txt
```diff
@@ -97,6 +97,27 @@ function MyExtension() {
 }
 ```
 
+### Shared Properties Without MfeProvider
+
+`useSharedProperty(propertyTypeId, options)` takes an explicit `bridge` for components that never mount `MfeProvider`, plus `fallback` and `select`. `fallback` is returned until a value is accepted and makes the result non-optional; `select` narrows the raw `unknown` value and keeps the last accepted value when it rejects a later publish, instead of flashing back to `fallback`.
```
itechmeat: 🟡 [Not a blocker] The three clarifications this round added to the TSDoc and CLAUDE.md did not reach `llms.txt`: `bridge` referential stability, the caveat that `fallback`'s `T` is only as strong as the published values, and `select`/`fallback` non-reactivity. Nothing here is false, it is just the one surface still missing exactly what round 1 asked to correct - and since this file ships in the package's `files` and is written for an agent reading it alone, "makes the result non-optional" reads as an unconditional guarantee.

itechmeat (approved):
🟢 Both blockers are genuinely closed, and closed in a way that holds up under mutation: deleting the pre-subscribe read fails exactly the one new test written for it, and the keep-last-valid test now fails if the publish path stops setting state. The `useSyncExternalStore` answer is substantive and technically right about the idempotent-memo distinction, the `useLatest` reuse and the test factory landed, and `tsc`/eslint/150 tests reproduce green at the head. Approving. Five things are worth a follow-up pass rather than another round: one new TSDoc sentence is unqualified where the code is not, three newly documented branches survive mutation untested, two resync test comments claim more than the tests isolate, `llms.txt` did not receive the corrections its siblings did, and the pre-subscribe read costs an extra render when `select` allocates - which the docs actively encourage.

template-shell/packages/react/src/mfe/hooks/useSharedProperty.ts
```diff
@@ -1,76 +1,275 @@
...
+ * `undefined` is the one and only "nothing here" sentinel throughout this
+ * hook -- an accepted `null` (unfiltered, or returned by `select`) is a real
+ * value and is never treated as "reject"/"unset". Callers must test with
+ * `=== undefined`, not `??`/truthiness, or `null` will incorrectly collapse
+ * to `fallback`.
+ */
+function resolveValue<T>(
+  bridge: ChildMfeBridge,
+  propertyTypeId: string,
+  select: ((value: unknown) => T | undefined) | undefined
```
GeraBart: Why is it needed? TypeSystemPlugin is responsible for validating shared property value against its schema, which happens at setting shared property value state. So a value, which doesn't comply with the property shouldn't be even set. We shouldn't have custom validators.

GeraBart (changes requested):

template-shell/packages/react/src/mfe/hooks/useSharedProperty.ts
tscbmstubp: You're right — `updateSharedProperty` registers the value with the TypeSystemPlugin before storing or notifying, and `register()` throws on schema mismatch, so a bad value never reaches an MFE. Dropped `select` and the keep-last-valid machinery; the hook is back on `useSyncExternalStore` with just `bridge` + `fallback`. See the "Round 3" section in the body.

template-shell/packages/react/__tests__/mfe/hooks.test.tsx
```diff
@@ -138,6 +150,181 @@ describe('MfeContext', () => {
...
       expect(unsubscribe).toHaveBeenCalled();
     });
+
+    it('should throw when used outside MfeProvider and no explicit bridge is given', () => {
+      expect(() => {
+        renderHook(() =>
+          useSharedProperty('gts.frontx.mfes.comm.shared_property.v1~test.user_data.v1')
+        );
+      }).toThrow(/MfeProvider/);
+    });
```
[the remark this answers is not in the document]
tscbmstubp: Declining. `useSharedProperty` reads only `MfeContext` (plus the optional `bridge` option); it has no dependency on `FrontXProvider`, and the pre-existing `14.5.3` tests in this file already mount it without the root provider (Redux + `MfeProvider` wrapper only). The CLAUDE.md line is consumer guidance for apps, not a unit-test requirement. Wrapping these tests would also hide the one thing the first test asserts: that the hook throws when neither `MfeProvider` nor an explicit `bridge` is present.

template-shell/packages/react/src/mfe/hooks/useSharedProperty.ts
```diff
@@ -1,76 +1,145 @@
...
+   * real value and is returned as-is. The host validates every published
+   * value against the property's schema before it reaches any MFE
+   * (`TypeSystemPlugin.register()` throws on a schema mismatch and the
+   * update is never stored or delivered), so no consumer-side filtering is
+   * needed -- `T` is simply the caller's declaration of that schema.
+   *
+   * `fallback` may be passed inline on every render; it does not affect the
+   * host subscription.
+   */
+  fallback?: T;
```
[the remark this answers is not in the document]
tscbmstubp: Declining — this is intentional, see the "Round 3" section of the PR body and https://github.com/constructorfabric/gears-frontx/pull/574#discussion_r3801218211. `select` was removed because the host already validates every published value against the property's schema at set time (`updateSharedProperty` → `TypeSystemPlugin.register()`, which throws on mismatch), so a malformed value never reaches an MFE and a consumer-side validator has nothing to catch. The #570 wording predates that finding.

gs-layer (approved):

GeraBart (approved):

PR #576, opened 2026-08-17
fix(mfes): add webpackIgnore to the blob-module dynamic import (#504)

### LINKED ISSUE #504: [Bug] MFE blob loader throws "Cannot find module 'blob:…'" under webpack/rsbuild hosts (importBlobModule missing webpackIgnore)
## Description
`MfeHandlerMF` cannot load any MFE when the host application is bundled with
webpack/rsbuild. The blob module import in `importBlobModule` is annotated only
with `/* @vite-ignore */`; webpack does not honor that comment, compiles
`import(<variable>)` into a dynamic context module, and throws
`Cannot find module 'blob:…'` at mount time. Vite/Rolldown hosts are unaffected.

## Environment
- **HAI3 Version**: @gears-frontx/screensets 0.2.0-alpha.0
- **Node.js Version**: v22.x
- **OS**: Windows 11 (bundler-agnostic; reproduces on any OS)
- **Package(s) Affected**: @gears-frontx/screensets (MFE loader `mfe/handler`), when consumed by a webpack/rsbuild host 

## Steps to Reproduce
1. Build a host with webpack/rsbuild that consumes `@gears-frontx/screensets` and mounts an MFE via `mfeRegistryFactory` + `MfeHandlerMF` (e.g. the frontx-frontend-app-adapter, `rsbuild dev`).
2. Register an MFE entry/manifest whose expose chunk is served as a federation remote.
3. Drive the lifecycle through `executeActionsChain` (`load_ext` then `mount_ext`) so the loader mints a blob URL and calls `importBlobModule(blobUrl)`.

## Expected Behavior
The loader imports the blob module natively and the MFE loads and renders, regardless of which bundler built the host.

## Actual Behavior
The blob import is turned into a webpack dynamic context module. The `blob:` URL is not in the generated context map, so webpack's `throwMissingModuleError` runtime fires. The actions chain fails and no MFE content renders (the shadow root mounts but stays empty).

## Error Message / Stack Trace
```
[MfeRegistry] Actions chain failed: Cannot find module 'blob:https://localhost/<uuid>'
  | path: [gts.frontx.mfes.comm.action.v1~frontx.mfes.ext.load_ext.v1~ ...]
[MfeRegistry] Actions chain failed: Cannot find module 'blob:https://localhost/<uuid>'
  | path: [gts.frontx.mfes.comm.action.v1~frontx.mfes.ext.mount_ext.v1~ ...]

# Thrown by webpack/lib/RuntimeTemplate.js -> throwMissingModuleErrorBlock:
#   const err = `Cannot find module '${request}'`;
```

## Screenshots
N/A (console errors above; shadow root mounts with adopted styles but no rendered content).

## Additional Context
- Offending code: `packages/screensets/src/mfe/handler/mf-dynamic-module-ops.ts`, `importBlobModule`:
  ```ts
  return await import(/* @vite-ignore */ blobUrl)
  ```
- The error string is emitted by webpack itself, not by gears-frontx, which is why it is not found anywhere in this repo.
- This is the only webpack-compiled dynamic import in the loader. The `__frontx_lazy` stub in `mf-handler.ts` is emitted as a blob source string and executed as its own blob, so webpack never processes it and it needs no annotation.
- Found while injecting an MFE into the legacy console via the rsbuild-built frontx-frontend-app-adapter; patching the adapter's installed copy with `webpackIgnore` resolved it and the MFE rendered end to end.

## Possible Solution
Add the webpack magic comment alongside the existing Vite one (and all occurrences) so the import stays native under every bundler:
```ts
return await import(/* webpackIgnore: true */ /* @vite-ignore */ blobUrl)
```
Acceptance:
- `importBlobModule` carries both `/* webpackIgnore: true */` and `/* @vite-ignore */`.
- A webpack/rsbuild host mounts an MFE without `Cannot find module 'blob:…'`.
- No regression for Vite/Rolldown hosts.

tscbmstubp:
## Summary
Fixes #504.

`importBlobModule` in `packages/mfes/src/handler/mf-dynamic-module-ops.ts` annotated its dynamic import only with `/* @vite-ignore */`. webpack does not honor that comment: it compiles `import(<variable>)` into a dynamic context module, the minted `blob:` URL is not in the context map, and webpack's runtime throws `Cannot find module 'blob:…'` at mount time — so no MFE can load under a webpack/rsbuild host. Adding `/* webpackIgnore: true */` alongside the existing Vite comment keeps the import native under every bundler.

This is the only dynamic import in the package, so one line covers all occurrences. The `__frontx_lazy` stub is emitted as a blob source string and never processed by the host bundler, so it needs no annotation.

## Verification
- `npm run build` in `packages/mfes`: the `webpackIgnore` comment survives tsup/esbuild into both `dist/index.js` and `dist/index.cjs` (esbuild preserves comments inside `import()` arguments).
- `npm run type-check`: clean.
- `npm test`: 10 files, 68 tests passed.
- Vite/Rolldown behavior unchanged — `@vite-ignore` remains in place.
- The webpack-host acceptance path (rsbuild-built adapter mounting an MFE end to end) was confirmed by the issue reporter with this exact patch applied to the installed copy; no webpack toolchain exists in this repo to automate it.

No FEATURE/ADR change needed: the covering instruction (`inst-exec-import` in `packages/mfes/architecture/features/mfe-isolation/FEATURE.md`) is bundler-agnostic.

### DIFF
--- a/packages/mfes/scripts/verify-dist-import-pragmas.mjs
+++ b/packages/mfes/scripts/verify-dist-import-pragmas.mjs
@@ -1,0 +1,61 @@
+#!/usr/bin/env node
+
+/**
+ * Verify that the bundler pragmas on the blob-module dynamic import survive
+ * the build into both published dist formats.
+ *
+ * `importBlobModule` (src/handler/mf-dynamic-module-ops.ts) relies on two
+ * inline comments — `webpackIgnore: true` and `@vite-ignore` — to keep its
+ * `import()` native under webpack/rspack/rsbuild and Vite hosts (#504).
+ * Comments are the one part of the source a build step may legally drop:
+ * esbuild's `minify: true` strips them without a diagnostic, and nothing at
+ * runtime notices until an MFE fails to mount under a bundling host with
+ * `Cannot find module 'blob:…'`. Running the code cannot detect the pragma
+ * being stripped, so this script asserts its presence in the artifacts
+ * (same reasoning as scripts/verify-guard-configs.ts at the repo root).
+ *
+ * Wired into `npm run build` after tsup; fails the build if either pragma
+ * is missing from either format.
+ */
+
+import { readFileSync } from 'node:fs';
+import { dirname, join } from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+const distDir = join(dirname(fileURLToPath(import.meta.url)), '..', 'dist');
+const PRAGMAS = ['webpackIgnore: true', '@vite-ignore'];
+const FILES = ['index.js', 'index.cjs'];
+
+const failures = [];
+for (const file of FILES) {
+  const path = join(distDir, file);
+  let content;
+  try {
+    content = readFileSync(path, 'utf8');
+  } catch {
+    failures.push(`${file}: missing (build did not produce it)`);
+    continue;
+  }
+  for (const pragma of PRAGMAS) {
+    if (!content.includes(pragma)) {
+      failures.push(`${file}: pragma "${pragma}" not found`);
+    }
+  }
+}
+
+if (failures.length > 0) {
+  console.error(
+    'verify-dist-import-pragmas: the blob-import bundler pragmas did not survive the build:',
+  );
+  for (const failure of failures) {
+    console.error(`  - ${failure}`);
+  }
+  console.error(
+    'A bundling host will rewrite the dynamic import and MFE mount will fail (#504). Check tsup.config.ts for comment-stripping options (e.g. minify).',
+  );
+  process.exit(1);
+}
+
+console.log(
+  `verify-dist-import-pragmas: OK (${PRAGMAS.join(', ')} present in ${FILES.join(', ')})`,
+);
--- a/packages/mfes/src/handler/mf-dynamic-module-ops.ts
+++ b/packages/mfes/src/handler/mf-dynamic-module-ops.ts
@@ -100,6 +100,19 @@ export function rewriteBareSpecifier(
  *      guard below makes that invariant executable so accidental call
  *      sites that pass a path or `http:`/`file:` URL fail fast rather than
  *      silently loading untrusted code.
+ *
+ *      **BUNDLER PRAGMAS — the two inline comments in the `import()` below
+ *      are load-bearing. Do not remove or reorder them.**
+ *      `webpackIgnore: true` tells webpack/rspack/rsbuild hosts, and
+ *      `@vite-ignore` tells Vite hosts, to leave this dynamic import native
+ *      instead of rewriting it into their chunk-loading runtime. A bundler
+ *      that rewrites it makes MFE mount fail at runtime under that host
+ *      with `Cannot find module 'blob:…'` (#504). The pragmas survive the
+ *      package's current tsup/esbuild build in both output formats, but
+ *      esbuild minification (`minify: true`) strips comments without a
+ *      diagnostic — so the build script runs
+ *      `scripts/verify-dist-import-pragmas.mjs`, which fails the build if
+ *      either pragma is missing from the published dist.
  * @inputs `blobUrl` — a `blob:` URL from `URL.createObjectURL(...)` or a
  *         `data:` URL from the same code path under test mocks.
  */
@@ -117,7 +130,7 @@ export async function importBlobModule(blobUrl: string): Promise<unknown> {
   // @cpt-end:cpt-frontx-algo-mfe-isolation-trust-kernel-import:p1:inst-inspect-scheme
   // @cpt-begin:cpt-frontx-algo-mfe-isolation-trust-kernel-import:p1:inst-exec-import
   // @cpt-begin:cpt-frontx-algo-mfe-isolation-trust-kernel-import:p1:inst-return-module
-  return await import(/* @vite-ignore */ blobUrl);
+  return await import(/* webpackIgnore: true */ /* @vite-ignore */ blobUrl);
   // @cpt-end:cpt-frontx-algo-mfe-isolation-trust-kernel-import:p1:inst-return-module
   // @cpt-end:cpt-frontx-algo-mfe-isolation-trust-kernel-import:p1:inst-exec-import
 }
--- a/.codacy.yaml
+++ b/.codacy.yaml
@@ -82,7 +82,7 @@ exclude_paths:
   # dynamic import of blob URLs). Every function in it is required by a
   # custom ESLint rule to document its safety argument via `@safety-reviewed`
   # and `@why` JSDoc tags. Editing this file requires security review.
-  - "packages/screensets/src/mfe/handler/mf-dynamic-module-ops.ts"
+  - "packages/mfes/src/handler/mf-dynamic-module-ops.ts"
 
   # Monorepo package manifests.
   #
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -62,6 +62,17 @@ export default [
   js.configs.recommended,
   ...tseslint.configs.recommended,
 
+  // Package-local build scripts (e.g. packages/mfes/scripts/) run under Node,
+  // unlike the root scripts/** which is still ignored pending #483.
+  {
+    files: ['packages/*/scripts/**/*.mjs'],
+    languageOptions: {
+      globals: {
+        ...globals.node,
+      },
+    },
+  },
+
   // L0 BASE: Universal rules for all TS/TSX files
   {
     files: ['**/*.{ts,tsx}'],
--- a/packages/mfes/package.json
+++ b/packages/mfes/package.json
@@ -25,7 +25,7 @@
     }
   },
   "scripts": {
-    "build": "tsup",
+    "build": "tsup && node scripts/verify-dist-import-pragmas.mjs",
     "type-check": "tsc --noEmit",
     "type-check:test": "tsc -p tsconfig.test.json",
     "test": "npm run test:unit",


packages/mfes/src/handler/mf-dynamic-module-ops.ts
```diff
@@ -117,7 +117,7 @@ export async function importBlobModule(blobUrl: string): Promise<unknown> {
   // @cpt-end:cpt-frontx-algo-mfe-isolation-trust-kernel-import:p1:inst-inspect-scheme
   // @cpt-begin:cpt-frontx-algo-mfe-isolation-trust-kernel-import:p1:inst-exec-import
   // @cpt-begin:cpt-frontx-algo-mfe-isolation-trust-kernel-import:p1:inst-return-module
-  return await import(/* @vite-ignore */ blobUrl);
+  return await import(/* webpackIgnore: true */ /* @vite-ignore */ blobUrl);
```
itechmeat, line 133: 🟠 Both comments here are load-bearing directives to two different host bundlers, and nothing in the file says so - the JSDoc above argues safety, not this. Adding `minify: true` to the package's tsup config strips both without a diagnostic, which would silently reopen this bug for webpack and Vite alike, and arguably the reason it existed at all is that `@vite-ignore` sat here unexplained. Could the `@why` block name the concrete failure (`Cannot find module 'blob:…'` at mount under a webpack/rsbuild host) so a future editor knows not to touch them?

itechmeat, line 133: 🟡 [Not a blocker] Worth confirming the path to the reporter: publishing is version-gated and the package version is unchanged here, so merging this to develop ships nothing and #504 stays open on their side until a publish follows. They are also on `@gears-frontx/screensets` 0.2.0-alpha.0, which no longer exists - the package is `@gears-frontx/mfes` 0.3.0-alpha.2 now, so their migration is a rename plus the fix, not a version bump. Is a publish MR already planned, and should the issue stay open until it lands?

itechmeat, line 133: 🟡 [Not a blocker] Nothing exercises `importBlobModule` in the suite today, and this fix lands without a regression guard. A real webpack acceptance test is out of reach (no webpack or rspack anywhere in the repo's dependencies) and a source-text regex would only assert the line being changed - but a dist-level assertion, that `dist/index.js` and `dist/index.cjs` still contain `webpackIgnore` after a build, is cheap and catches exactly the silent-strip case. There is precedent for that shape: `scripts/verify-guard-configs.ts` exists on the reasoning that running a guard cannot detect the guard being weakened. Here or as a follow-up?

itechmeat, line 133: 🔵 [Not a blocker] Pre-existing, but it lands on the line you are touching: `.codacy.yaml` excludes `packages/screensets/src/mfe/handler/mf-dynamic-module-ops.ts`, a path that stopped existing when the package was renamed to `packages/mfes` - so this file is not actually excluded from the security scan, while its own header states that it is. The practical effect here is that the changed line may get flagged. Worth repointing the exclusion, or dropping the claim from the header.

itechmeat, line 133: 🔵 [Not a blocker] Also pre-existing, also in this file's header: it names a custom ESLint rule `local/trusted-patterns-file` as what enforces the `@safety-reviewed`/`@why` tags, and the DoD `cpt-frontx-dod-mfe-isolation-blob-core` requires that rule to exist - but a grep for `trusted-patterns` at head returns exactly one hit, the comment claiming it exists. So those tags are unenforced and that DoD is not met today. Not this PR's doing, just worth surfacing while the file is open.

itechmeat, line 133: 🔵 [Not a blocker] The change introduces a constraint on the published package - its dist has to stay a native dynamic import under any host bundler - and no artifact records it. Every bundler mention under `architecture/` is about guest-side neutrality, the opposite direction, and ADR-0011's Consequences (which enumerate the exclusions the implementation relies on) do not list this one. The PR is right that `inst-exec-import` is bundler-agnostic, but that argues for the line rather than against it: someone reading the instruction against the code has no reason to preserve the pragmas. One sentence in ADR-0011, whenever convenient.

itechmeat, line 133: ⚪ [Not a blocker] Context rather than a request: `template-shell/packages/framework/src/testing/describeBootstrapMfeContract.ts` has four dynamic imports with non-literal arguments in a helper that ships to consumers via the package's `./testing` export, so a webpack-bundled consumer of it hits the same rewrite this PR fixes. Template territory and a different package, so out of scope here - just so the class is not assumed fully swept.

itechmeat (changes requested):
A correct, minimal fix with genuinely thorough homework - the claims that matter all hold up: the pragma survives the package's actual tsup/esbuild config in both formats, the pinned Vite honours `@vite-ignore` in second position, and this really is the only executable dynamic import in the package. One thing keeps it from approve: nothing in the code says the two comments are load-bearing, and a future `minify: true` deletes both without a sound - the same silence that let this bug exist in the first place. The rest is context worth knowing rather than work: the reporter needs a publish and a package rename before #504 closes on their side, and this file's own header describes a Codacy exclusion and a lint rule that no longer exist.

packages/mfes/src/handler/mf-dynamic-module-ops.ts
tscbmstubp: Done in 96cbc89d. The `@why` block now states outright that the two inline comments are load-bearing bundler directives (`webpackIgnore: true` for webpack/rspack/rsbuild hosts, `@vite-ignore` for Vite hosts), names the concrete failure their removal causes — MFE mount failing at runtime with `Cannot find module 'blob:…'` under a bundling host (#504) — and warns that esbuild's `minify: true` strips comments without a diagnostic.

Since documentation alone is the same silence that let this happen, the build now also asserts the pragmas survive — see the reply on the regression-guard thread.

tscbmstubp: Taken here rather than as a follow-up, since it directly protects this change: `packages/mfes/scripts/verify-dist-import-pragmas.mjs` runs as part of `npm run build` (after tsup) and fails the build if `webpackIgnore: true` or `@vite-ignore` is missing from `dist/index.js` or `dist/index.cjs` — same reasoning as `scripts/verify-guard-configs.ts`, credited in the script header. Verified in both directions: passes on a clean build, and stripping the pragma from dist makes it exit 1 with a pointer at `tsup.config.ts` comment-stripping options.

tscbmstubp: Agreed on all points: merging this ships nothing on its own, so #504 should stay open until a publish lands. A version-bump + publish follow-up is the plan, and the migration note for the reporter (rename `@gears-frontx/screensets` 0.2.0-alpha.0 → `@gears-frontx/mfes`, then pick up the fixed version) is worth posting on the issue when that version exists to point at.

tscbmstubp: Fixed here — `.codacy.yaml` now excludes `packages/mfes/src/handler/mf-dynamic-module-ops.ts` instead of the dead pre-rename `packages/screensets/...` path, so the header's claim is true again and the changed line stays inside the intended trust-kernel exclusion.

tscbmstubp: Confirmed — `local/trusted-patterns-file` exists only as the comment claiming it does, so the `@safety-reviewed`/`@why` tags are unenforced and `cpt-frontx-dod-mfe-isolation-blob-core` is not met today. Leaving it out of this PR deliberately: closing it means either writing the rule or rewording the header and revisiting the DoD, both bigger than this fix. Worth its own issue alongside the #483 scripts-enforcement gap.

tscbmstubp: Fair point — the constraint (the published dist must keep this dynamic import native under any host bundler) now has an executable form in the build via the dist pragma check, but no artifact records it, and every bundler mention in `architecture/` points the other way. A sentence in ADR-0011's Consequences is the right home; queued as a follow-up rather than mixed into this fix.

tscbmstubp: Thanks for sweeping the class — noted that `describeBootstrapMfeContract.ts`'s four non-literal dynamic imports ship to consumers via the framework package's `./testing` export and would hit the same rewrite under a webpack host. Agreed it's out of scope here (template territory, different package); tracking it as a separate follow-up so it doesn't get assumed fixed by this PR.

packages/mfes/src/handler/mf-dynamic-module-ops.ts
```diff
@@ -100,6 +100,19 @@ export function rewriteBareSpecifier(
  *      guard below makes that invariant executable so accidental call
  *      sites that pass a path or `http:`/`file:` URL fail fast rather than
  *      silently loading untrusted code.
+ *
+ *      **BUNDLER PRAGMAS — the two inline comments in the `import()` below
+ *      are load-bearing. Do not remove or reorder them.**
```
itechmeat, line 105: 🔵 [Not a blocker] Small thing in an otherwise precise paragraph: order is not actually load-bearing. Vite tests its regex against the whole paren contents rather than the first comment, and I checked webpack by building it - with the pragmas in either order the import stayed native, and only the un-annotated build produced `Cannot find module 'blob:…'`. So "do not remove" holds exactly, "or reorder" forbids something safe. Worth dropping the two words, unless the intent is deliberate belt-and-braces.

packages/mfes/src/handler/mf-dynamic-module-ops.ts
```diff
@@ -100,6 +100,19 @@ export function rewriteBareSpecifier(
...
+ *      `webpackIgnore: true` tells webpack/rspack/rsbuild hosts, and
+ *      `@vite-ignore` tells Vite hosts, to leave this dynamic import native
+ *      instead of rewriting it into their chunk-loading runtime. A bundler
+ *      that rewrites it makes MFE mount fail at runtime under that host
+ *      with `Cannot find module 'blob:…'` (#504). The pragmas survive the
```
itechmeat, line 110: 🔵 [Not a blocker] The second half here restates the guard script's own header almost point for point - `minify: true`, the surviving-the-build note, the fail-the-build behaviour are all in verify-dist-import-pragmas.mjs:4-18 too. The part the source comment uniquely owes the reader is the first half: these two comments are load-bearing and removing them breaks #504. Keeping the build-tooling story with the tool that enforces it would mean the next tsup change only has to be remembered in one place. Partly this is the ADR line you deferred doing its job - no complaint about the deferral itself.

packages/mfes/scripts/verify-dist-import-pragmas.mjs
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * `importBlobModule` (src/handler/mf-dynamic-module-ops.ts) relies on two
+ * inline comments — `webpackIgnore: true` and `@vite-ignore` — to keep its
+ * `import()` native under webpack/rspack/rsbuild and Vite hosts (#504).
+ * Comments are the one part of the source a build step may legally drop:
+ * esbuild's `minify: true` strips them without a diagnostic, and nothing at
+ * runtime notices until an MFE fails to mount under a bundling host with
+ * `Cannot find module 'blob:…'`. Running the code cannot detect the pragma
+ * being stripped, so this script asserts its presence in the artifacts
+ * (same reasoning as scripts/verify-guard-configs.ts at the repo root).
```
itechmeat, line 15: ⚪ [Not a blocker] The analogy to `scripts/verify-guard-configs.ts` is fair, and worth noting that the model has a test sitting next to it while this one has none - and nothing type-checks it either, since this package's tsconfig covers only `src/**/*`. The location follows existing practice (`packages/ui-kit/scripts/`) and the script runs on every build, so a mistake in it surfaces immediately. Mentioning it rather than asking for it.

itechmeat (approved):
🟢 This round answers everything that mattered, and answers it with working code rather than a promise. The guard is the part worth calling out: it is wired into `build:packages`, which is what both CI and the publish workflow actually run, it passes on a clean build and genuinely fails when the pragmas are stripped, and it is not fooled by its own documentation - esbuild drops the JSDoc, so the only match left in dist is the real one at the import site. The `@why` paragraph and the Codacy repoint close the other two. Approving; what is left is polish, and the deferrals all read as reasonable calls with the reasoning written down. Two small asks that are not code: the description still reads like the original one-liner, and the four follow-ups would be easier to hold onto with issue numbers on them.

GeraBart (approved):

PR #577, opened 2026-08-17
feat(policy): add version-bump-on-change CI gate (#575)

### LINKED ISSUE #575: CI gate: require a version bump when a package's substantive files change
Follow-up from #574 review: the PR changed `@gears-frontx/react`'s public API and initially shipped without a version bump, and nothing caught it.

The existing policy checks are consistency checks, not change checks. `policy:version-check` validates the ecosystem packages' semver lines and the mfes→gts-plugin coupling edge; `policy:template-pin-drift` verifies pins match whatever version is on disk. A PR that changes code but bumps nothing passes all of them — the pins still "match" the stale version.

Proposal: add `scripts/version-bump-on-change-check.mjs`, wired as `policy:version-bump-on-change` into `main.yml` PR CI, following the same pattern as the pin-drift job (per ADR-0001's stance, a plain `policy:*` script — no changesets/lerna/semantic-release):

- Diff against the merge base with the PR's base branch (`git diff --name-only $(git merge-base <base> HEAD)..HEAD`).
- Group changed files by package directory: the `packages/*` ecosystem artifacts plus the `template-shell/packages/*` workspace members.
- For each package with substantive changes, fail unless its `package.json` version differs from the version at the merge base.

Decisions to settle in the implementation:

- What counts as substantive — likely exclude pure docs (`*.md`, `llms.txt`) and possibly test-only changes; a simple starting rule is anything under `src/` or `package.json` dependency changes requires a bump.
- Comparing against the merge-base version (not just "package.json changed") so a bump-then-revert within one PR still fails.
- Whether template-mfe fixture packages are exempt (they are pinned consumers, covered by `policy:template-pin-drift` once the source package bumps).

Also worth a one-line statement of the rule in CONTRIBUTING.md so contributors bump proactively instead of discovering the gate in CI.

tscbmstubp:
Closes #575.

Adds a `policy:version-bump-on-change` CI gate that fails a PR when a governed package has substantive changes without a version bump — the change-check counterpart to the existing consistency checks (`policy:version-check`, `policy:template-pin-drift`), which pass a stale version as long as pins still match it (#574).

## What's in here

- `scripts/version-bump-on-change-check.mjs` — the gate, following the pin-drift script pattern (plain `policy:*` script per ADR-0001; no changesets/lerna/semantic-release)
- `scripts/version-bump-on-change-check.test.mjs` — 23 tests against real temp git repos
- `package.json` — `policy:version-bump-on-change` script
- `.github/workflows/main.yml` — PR-only step in the `validate` job; checkout gets `fetch-depth: 0` so the merge base is computable
- `CONTRIBUTING.md` — one-line statement of the rule

## Semantics

- **Governed packages:** every non-private `@gears-frontx`-scoped package — `packages/*`, `template-shell` itself (publishes `dist-lib` from its own `src/`, exact-pinned by four template-mfe fixtures), and `template-shell/packages/*`. Discovery is reused from `scripts/template-ecosystem-packages.mjs`, not re-derived.
- **Exemption:** per-package `private: true` — excludes the `template-mfe` fixture apps (pinned consumers, covered by `policy:template-pin-drift` once the source package bumps) and `internal/*`.
- **Substantive change:** any non-doc file under the package's `src/` (colocated tests included; `*.md`/`llms.txt` excluded), or a change to a dependency field of its `package.json`.
- **Comparison:** the `package.json` version at the PR's merge base vs HEAD (blob content via `git show`), so a bump-then-revert within one PR still fails.
- **Attribution:** longest-matching root, so `template-shell/packages/react/src/*` demands a react bump, not a parent `template-shell` bump.
- Packages added or removed by the PR itself are skipped; the step runs only on `pull_request` events (no base ref on `push`).

## Deliberately out of scope (noted during review)

- `files`-field payloads outside `src/` (e.g. `packages/cyber-pilot-kit-frontx`'s `skills/`/`guidelines/`)
- semver-ordering validation (any version *change* passes, including downgrades)
- an escape-hatch/ignore mechanism for pure-mechanical `src/` changes
- concurrent-PR bump collisions (two in-flight PRs can bump to the same version; surfaces later via pin-drift/publish)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -26,6 +26,12 @@ jobs:
       uses: actions/checkout@v5
       with:
         persist-credentials: false
+        # #575: policy:version-bump-on-change needs a merge base with the
+        # PR's target branch, and a shallow (depth-1) checkout has no history
+        # to compute one from. Full history costs little here — this repo is
+        # small — and matches the same choice already made in docs.yml and
+        # publish-packages.yml for the same reason.
+        fetch-depth: 0
 
     - name: Setup Node.js ${{ matrix.node-version }}
       uses: actions/setup-node@v5
@@ -122,6 +128,25 @@ jobs:
     # files directly, not the built CLI.
     - name: Check template ecosystem pin-drift
       run: npm run policy:template-pin-drift
+
+    # #575, follow-up from #574 review: #574 changed `@gears-frontx/react`'s
+    # public API and shipped with no version bump, and neither policy check
+    # above would have caught it — both are CONSISTENCY checks (does the
+    # semver line make sense, do pins match the disk), not CHANGE checks (did
+    # a package with substantive changes bump). This diffs the PR branch
+    # against its merge base with the target branch and fails for any
+    # governed package — every `packages/*` artifact, plus `template-shell`
+    # itself and every `template-shell/packages/*` member (any non-`private`
+    # `@gears-frontx`-scoped package; `template-mfe` and its fixture apps are
+    # `private` and exempt) — whose `src/` (or package.json dependencies)
+    # changed without its own `version` changing — comparing actual ref
+    # content rather than the diff, so a version bumped then reverted within
+    # the same PR still fails. Pull-request-only: the merge base it needs is
+    # the PR's base branch, which only exists as `GITHUB_BASE_REF` on
+    # `pull_request` events.
+    - name: Check version bump on substantive package change
+      if: github.event_name == 'pull_request'
+      run: npm run policy:version-bump-on-change
 
     # #524: `npm install` reconciling an out-of-sync template lockfile drops
     # the template's `file:.` self-link entry, writing a lockfile its own
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -80,6 +80,8 @@ The project is **pre-1.0** — backward compatibility is not guaranteed.
 - **Patch bump** (`0.1.0` → `0.1.1`) — non-breaking fixes/features
 - **Alpha increment** (`alpha.0` → `alpha.1`) — each merge to develop
 
+A PR that changes non-documentation source under `src/`, or the dependency fields of `package.json`, for a governed package (non-private `@gears-frontx`-scoped packages: `packages/*`, `template-shell`, and `template-shell/packages/*`) must bump that package's own `version` in the same PR — and update every exact pin on it (see `policy:template-pin-drift`). CI (`policy:version-bump-on-change`, pull requests only) compares the version at the PR's merge base against the version at its head, so a bump that is later reverted within the same PR does not count. Packages the PR itself adds or removes are exempt, as are private packages (e.g. the `template-mfe` fixture apps), which are pinned consumers rather than published sources.
+
 ## Publishing
 
 Publishing is automated via CI/CD. On push to a publishing branch, CI detects version changes and publishes affected packages.
--- a/scripts/version-bump-on-change-check.mjs
+++ b/scripts/version-bump-on-change-check.mjs
@@ -1,0 +1,506 @@
+/**
+ * Version-Bump-On-Change CI Gate (#575, follow-up from #574 review).
+ *
+ * #574 changed `@gears-frontx/react`'s public API and shipped with no version
+ * bump, and nothing caught it. The two existing `policy:*` guards could not
+ * have caught it either, because both are CONSISTENCY checks, not CHANGE
+ * checks: `policy:version-check` validates the ecosystem packages' semver
+ * lines and the mfes->gts-plugin coupling edge in isolation, and
+ * `policy:template-pin-drift` verifies a pinned dependency range matches
+ * whatever version is on disk RIGHT NOW. A PR that edits `src/` and bumps
+ * nothing passes both: the pins still "match" the stale version, and the
+ * semver line is still internally consistent - just unchanged.
+ *
+ * This guard asks the one question neither of those does: did a package with
+ * substantive changes in this PR ALSO change its own `version`? It diffs the
+ * PR branch against the merge base with its target branch, groups the changed
+ * files by owning package directory, and fails for any package whose
+ * substantive changes were not accompanied by a version bump.
+ *
+ * ## Which directories are "a package" here
+ *
+ * A governed package is any non-`private` package whose own name is in the
+ * ecosystem's npm scope (`@gears-frontx/*`), discovered by REUSING
+ * `template-ecosystem-packages.mjs`'s existing structural discovery
+ * (`readEcosystemPackages` for `packages/*`, `readTemplateEcosystemPackages`
+ * for every template's own root identity AND every one of its workspace
+ * members - see that module's docblock for why re-deriving either walk here
+ * would be the duplicated-knowledge bug its own predecessor was rewritten to
+ * remove) rather than a second, narrower copy of it. Concretely, today:
+ *
+ *  - every `packages/*` directory (this repo's own published ecosystem
+ *    artifacts);
+ *  - `template-shell` ITSELF (`@gears-frontx/frontx-template-shell`) - it
+ *    publishes `dist-lib`, built from its own `src/`, and is exact-pinned by
+ *    four `template-mfe` fixture manifests, exactly the #574 shape this gate
+ *    exists for;
+ *  - every `template-shell/packages/*` workspace member (react, auth,
+ *    framework, i18n, state, studio - ADR-0033 relocated `@gears-frontx/react`'s
+ *    source there, and #574 is the concrete case that skipped a bump).
+ *
+ * `template-mfe` and its four `src-app/mfe_packages/*` fixture apps
+ * (`@gears-frontx/demo-mfe` and friends) ARE in the ecosystem's npm scope -
+ * scope alone does not exclude them - but every one of them declares
+ * `private: true`, and it is exactly that per-package fact, not a heuristic
+ * on `template-mfe`'s own (unscoped) root manifest, that excludes them here:
+ * they pin these packages' PUBLISHED versions rather than being one of them,
+ * and are covered once the source package bumps by
+ * `policy:template-pin-drift`, not by this gate. The same `private: true`
+ * check is what would exempt a future `template-shell/src-app/mfe_packages/*`
+ * member too, should that workspace pattern ever gain one.
+ *
+ * ## What counts as a substantive change
+ *
+ * Per file, relative to its package root:
+ *  - anything under `src/` (excluding a pure-docs file - `*.md` or
+ *    `llms.txt` - even one that happens to live inside `src/`);
+ *  - `package.json` ITSELF, but only when a dependency field
+ *    (`dependencies`/`devDependencies`/`peerDependencies`/
+ *    `optionalDependencies`) actually changed between the merge base and
+ *    HEAD - not merely appearing in the diff (a version-only edit, a
+ *    `scripts` tweak, must not itself force a second bump).
+ *
+ * Everything else - README changes, build config, a `package.json` edit that
+ * touches no dependency field - is out of scope for this gate by
+ * construction: it is simply never classified as substantive. This repo
+ * colocates tests under `src/` (e.g. `src/tokens.test.ts`), so the issue's
+ * open question about test-only changes resolves to "still substantive"
+ * here: a change to only `packages/ui-kit/src/tokens.test.ts` DOES force a
+ * bump, same as any other `src/` file - deliberately, since `src/` is the
+ * one signal this gate trusts, not a second classification of what's
+ * "inside" it.
+ *
+ * ## Why the comparison is merge-base-to-HEAD content, not "did the diff touch
+ * version"
+ *
+ * `versionAt(ref, packageRoot)` reads the actual `package.json` blob at a ref
+ * and pulls `version` out of it - it does not inspect the diff hunk. That is
+ * what makes a bump-then-revert within one PR still fail: if `version` is
+ * edited in an early commit and reverted in a later one, `git diff --name-only`
+ * may not even list `package.json` as changed at the merge base once the
+ * revert lands, and the two blobs compared here are IDENTICAL - so if `src/`
+ * also changed in that PR (a separate file), the substantive check still
+ * trips and the version comparison still correctly reports "unchanged".
+ * Comparing anything other than the two endpoints' actual content would miss
+ * that case.
+ *
+ * A package with no `package.json` at the merge base (introduced by this PR)
+ * is skipped rather than failed: a brand-new package has no prior version to
+ * have bumped FROM. Likewise a package whose `package.json` no longer exists
+ * at HEAD (removed by this PR) is skipped - there is no version left to check.
+ *
+ * CLI entry: `node scripts/version-bump-on-change-check.mjs [--base-ref <ref>]`
+ * (exit 0 on success). `--base-ref` defaults to `VERSION_BUMP_BASE_REF`, then
+ * `origin/${GITHUB_BASE_REF}` (set by GitHub Actions on `pull_request`
+ * events), then `origin/develop`. Core logic is exported for unit tests in
+ * `scripts/version-bump-on-change-check.test.mjs`.
+ */
+import { execFileSync } from 'node:child_process';
+import fs from 'node:fs';
+import path from 'node:path';
+import process from 'node:process';
+import { pathToFileURL } from 'node:url';
+import {
+  DEPENDENCY_FIELDS,
+  ecosystemScopeMatcher,
+  readEcosystemPackages,
+  readPackageManifest,
+  readTemplateEcosystemPackages,
+} from './template-ecosystem-packages.mjs';
+
+/**
+ * @typedef {{
+ *   mergeBase: (baseRef: string) => string;
+ *   listChangedFiles: (mergeBaseSha: string) => string[];
+ *   readTextAtRef: (ref: string, filePath: string) => string | null;
+ * }} GitReader
+ */
+
+/** A pure-docs file never counts as a substantive change, even under `src/`. */
+const DOCS_ONLY_EXTENSIONS = new Set(['.md']);
+const DOCS_ONLY_BASENAMES = new Set(['llms.txt']);
+
+/**
+ * @param {string} filePath repo-root-relative, posix-separated
+ * @returns {boolean}
+ */
+export function isDocsOnlyPath(filePath) {
+  return DOCS_ONLY_EXTENSIONS.has(path.posix.extname(filePath)) || DOCS_ONLY_BASENAMES.has(path.posix.basename(filePath));
+}
+
+/**
+ * Whether the package rooted at `dir` (repo-root-relative) declares
+ * `private: true` in its own manifest - the per-package fact this gate
+ * exempts on, rather than any heuristic about its parent template (see
+ * module docblock). A directory with no manifest at all governs nothing
+ * either, though every `dir` passed in here comes from
+ * `readEcosystemPackages`/`readTemplateEcosystemPackages`, both of which
+ * already guarantee the manifest they named exists and parses.
+ *
+ * @param {string} rootDir monorepo root
+ * @param {string} dir repo-root-relative
+ * @returns {boolean}
+ */
+function isPrivatePackage(rootDir, dir) {
+  const manifestPath = path.join(rootDir, dir, 'package.json');
+  if (!fs.existsSync(manifestPath)) return true;
+  return readPackageManifest(manifestPath)['private'] === true;
+}
+
+/**
+ * Every package root this gate governs, repo-root-relative and posix-joined
+ * (`packages/api`, `template-shell`, `template-shell/packages/react`, ...).
+ * See the module docblock for what qualifies and why.
+ *
+ * @param {string} rootDir monorepo root
+ * @returns {string[]} sorted
+ */
+export function discoverPackageRoots(rootDir) {
+  const ecosystem = readEcosystemPackages(rootDir);
+  const isEcosystemScopeName = ecosystemScopeMatcher(ecosystem.map(({ name }) => name));
+
+  /** @type {string[]} */
+  const candidateDirs = [
+    ...ecosystem.map(({ dir }) => `packages/${dir}`),
+    ...readTemplateEcosystemPackages(rootDir, isEcosystemScopeName).map(({ dir }) => dir),
+  ];
+
+  return candidateDirs
+    .filter((dir) => !isPrivatePackage(rootDir, dir))
+    .map((dir) => dir.split(path.sep).join('/'))
+    .sort();
+}
+
+/**
+ * Groups changed files under whichever package root owns them, picking the
+ * LONGEST matching root when more than one is a prefix - e.g.
+ * `template-shell/packages/react/src/x.ts` must attribute to
+ * `template-shell/packages/react`, not to the also-matching parent
+ * `template-shell`, now that both are governed roots (a nested-root
+ * scenario this gate did not have to consider until `template-shell` itself
+ * became one). A file that matches no root (repo tooling, docs at the repo
+ * root, an excluded private fixture tree) is silently dropped - this gate
+ * has nothing to say about it.
+ *
+ * @param {string[]} changedFiles repo-root-relative, posix-separated
+ * @param {string[]} packageRoots from `discoverPackageRoots`
+ * @returns {Map<string, string[]>}
+ */
+export function groupChangedFilesByPackageRoot(changedFiles, packageRoots) {
+  /** @type {Map<string, string[]>} */
+  const grouped = new Map();
+  for (const file of changedFiles) {
+    /** @type {string | undefined} */
+    let bestRoot;
+    for (const candidate of packageRoots) {
+      if (!file.startsWith(`${candidate}/`)) continue;
+      if (bestRoot === undefined || candidate.length > bestRoot.length) bestRoot = candidate;
+    }
+    if (bestRoot === undefined) continue;
+    const bucket = grouped.get(bestRoot);
+    if (bucket) bucket.push(file);
+    else grouped.set(bestRoot, [file]);
+  }
+  return grouped;
+}
+
+/**
+ * @param {unknown} manifest
+ * @returns {Record<string, unknown>}
+ */
+function dependencyFieldsOf(manifest) {
+  /** @type {Record<string, unknown>} */
+  const result = {};
+  if (typeof manifest !== 'object' || manifest === null) return result;
+  for (const field of DEPENDENCY_FIELDS) {
+    const value = /** @type {Record<string, unknown>} */ (manifest)[field];
+    if (typeof value !== 'object' || value === null || Array.isArray(value)) continue;
+    // Entries sorted by key so the stringify comparison below is key-order
+    // insensitive: `sort-package-json`, `npm pkg fix` and npm itself all
+    // reorder dependency entries without changing their meaning, and a
+    // reorder must not read as a dependency change.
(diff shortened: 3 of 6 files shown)


scripts/version-bump-on-change-check.mjs
```diff
@@ -0,0 +1,481 @@
...
+  // stdio[2] to the Error it throws - so the caller-facing message in
+  // `check()` loses nothing.
+  /** @param {string[]} args */
+  const run = (args) => execFileSync('git', args, { cwd: rootDir, encoding: 'utf8', stdio: ['ignore', 'pipe', 'pipe'] });
+  return {
+    mergeBase(baseRef) {
+      return run(['merge-base', baseRef, 'HEAD']).trim();
+    },
+    listChangedFiles(mergeBaseSha) {
+      return run(['diff', '--name-only', `${mergeBaseSha}..HEAD`])
```
itechmeat: 🔴 `git diff --name-only` detects renames by default and prints only the destination path, so a file moved from one package to another names only the receiving package. Moving `useSharedProperty.ts` from react to auth with no bumps anywhere fails only on auth, while react - which just lost a module from its published surface - is never mentioned; the same path silently passes a move out of `src/` within one package. Adding `--no-renames` to this call makes git print both sides, so both packages get asked to bump.

scripts/version-bump-on-change-check.mjs
```diff
@@ -0,0 +1,481 @@
...
+  } catch (error) {
+    logError(
+      `[version-bump-on-change-check] FAIL: cannot resolve a merge base with "${baseRef}" ` +
+        `(${error instanceof Error ? error.message : String(error)}). The checkout needs enough history to ` +
+        `reach it - fetch the base branch (or check out with full history) before running this check.`,
+    );
+    return 1;
+  }
+
+  const packageRoots = discoverPackageRoots(rootDir);
```
itechmeat, line 338: 🟠 If discovery returns an empty list this prints "every one of 0 governed package(s)" and exits 0 - reproduced on a repo with no `packages/` directory and a committed unbumped `src/` change. The failure cascades: no `packages/*` means the scope matcher has no scopes, which drops the template members too, so one broken path silences the whole gate. `template-pin-drift-check.mjs` already fails closed on exactly this case at lines 160-169, with a comment saying a zero-result discovery is never a silent pass - worth matching here, since silence is also what success looks like.

scripts/version-bump-on-change-check.mjs
```diff
@@ -0,0 +1,481 @@
...
+ * changing is NOT what this asks - only the four dependency fields
+ * (`DEPENDENCY_FIELDS`), because a bare version-only edit must never force a
+ * second bump of itself.
+ *
+ * @param {unknown} baseManifest
+ * @param {unknown} headManifest
+ * @returns {boolean}
+ */
+export function dependencyFieldsChanged(baseManifest, headManifest) {
+  return JSON.stringify(dependencyFieldsOf(baseManifest)) !== JSON.stringify(dependencyFieldsOf(headManifest));
```
itechmeat, line 239: 🟠 The stringify comparison is key-order sensitive: swapping two dependency entries with no semantic change reads as a dependency change and fails the PR, naming only `package.json` as substantive with no hint which field tripped it (reproduced end to end). `npm pkg fix`, `sort-package-json`, `prettier-plugin-packagejson` and npm's own occasional reordering all produce this. Sorting the entries of each dependency field before comparing would close it; note the guard test at :182-185 spreads the object, which preserves order, so it cannot catch this today.

scripts/version-bump-on-change-check.mjs
```diff
@@ -0,0 +1,481 @@
...
+    }
+    logError(
+      `\nBump the "version" in each package's package.json above (and any exact pin on it elsewhere - see ` +
+        '`npm run policy:template-pin-drift`), then rerun `npm run policy:version-bump-on-change` to confirm.',
+    );
+    return 1;
+  }
+
+  log(
+    `Version-bump-on-change check passed: every one of ${packageRoots.length} governed package(s) with ` +
```
itechmeat: 🟠 `packageRoots.length` is the number of governed roots discovered, not the number that had substantive changes, but the sentence attaches "with substantive changes" to it - so a PR touching only `scripts/` prints a claim that every governed package had substantive changes and bumped (reproduced on a repo where none changed). It is the one line a reader of a green run sees, and it reads as "these packages were bumped". The unbumped count two branches up is tracked correctly, so only the pass branch needs the wording or the count fixed.

scripts/version-bump-on-change-check.mjs
```diff
@@ -0,0 +1,481 @@
...
+ * @param {string} packageRoot
+ * @param {string[]} filesInRoot repo-root-relative, posix-separated, already
+ *   filtered to this root
+ * @param {string} baseRef the merge-base sha (or any ref `readTextAtRef`
+ *   accepts) to diff against
+ * @param {GitReader} git
+ * @returns {boolean}
+ */
+export function hasSubstantiveChange(packageRoot, filesInRoot, baseRef, git) {
+  const srcPrefix = `${packageRoot}/src/`;
```
itechmeat, line 270: 🟡 [Not a blocker] The `src/`-only rule is declared as deliberate scope, but the size of the gap is worth naming: changing `cyber-pilot-kit-frontx/skills/**`, `api/commands/*`, `ui-kit/tsup.config.ts`, `ui-kit/scripts/buildPlugin.ts` or `template-shell/src-app/**` all pass green (verified by running the gate), and the first two directories are listed in their manifests' `files`. Build config is the sharpest case - adding an entry point to `tsup.config.ts` changes a package's published surface with no bump demanded - and it is not mentioned in the body's out-of-scope list. For template-shell the rule currently watches 10 of its 98 own files. Adding root-level build configs to the substantive set would cover the worst of it without inverting the rule.

scripts/version-bump-on-change-check.mjs
```diff
@@ -0,0 +1,481 @@
...
+ * @typedef {{
+ *   mergeBase: (baseRef: string) => string;
+ *   listChangedFiles: (mergeBaseSha: string) => string[];
+ *   readTextAtRef: (ref: string, filePath: string) => string | null;
+ * }} GitReader
+ */
+
+/** A pure-docs file never counts as a substantive change, even under `src/`. */
+const DOCS_ONLY_EXTENSIONS = new Set(['.md']);
+const DOCS_ONLY_BASENAMES = new Set(['llms.txt']);
```
itechmeat, line 121: 🟡 [Not a blocker] `llms.txt` and `CLAUDE.md` are listed in the `files` array of api, auth, framework, i18n, react and state - they are the packages' agent-facing contract, not incidental prose, and #574 itself shipped +34/-2 in react's `CLAUDE.md` and +22/-1 in its `llms.txt`. A PR rewriting only `llms.txt` republishes a different tarball under the same version, so an exact pin keeps serving the old one. Excluding `.md` under `src/` is right (ui-kit's colocated component docs); it may be worth excluding docs only when they are not in the manifest's `files`.

scripts/version-bump-on-change-check.mjs
```diff
@@ -0,0 +1,481 @@
...
+      `[version-bump-on-change-check] FAIL: ${unbumped.length} package(s) have substantive changes but no ` +
+        `version bump:`,
+    );
+    for (const { packageRoot, version, files } of unbumped) {
+      logError(`  ${packageRoot} is still at ${version} despite substantive changes:`);
+      for (const file of files) logError(`    ${file}`);
+    }
+    logError(
+      `\nBump the "version" in each package's package.json above (and any exact pin on it elsewhere - see ` +
+        '`npm run policy:template-pin-drift`), then rerun `npm run policy:version-bump-on-change` to confirm.',
```
itechmeat, line 384: 🟡 [Not a blocker] Following this instruction locally does not work: bump the version, rerun as the message says, and it fails identically, still reporting the old version, because versions are read from `HEAD` blobs while the edit is uncommitted (verified). The sibling `policy:template-pin-drift` reads the working tree, so its "fix and rerun" really does confirm. Reading HEAD is the right call for a CI gate, so this is probably just a wording fix - say the bump has to be committed first.

scripts/version-bump-on-change-check.mjs
```diff
@@ -0,0 +1,481 @@
...
+    readTextAtRef(ref, filePath) {
+      try {
+        // git show's path separator is always `/`, matching the posix-joined
+        // paths this module works in throughout, regardless of host OS.
+        // stdio for stderr is piped (not inherited) so the very common
+        // "path exists on disk, but not in <ref>" case - a brand-new or
+        // removed package, handled deliberately below - never spams a
+        // passing CI run's log.
+        return execFileSync('git', ['show', `${ref}:${filePath}`], { cwd: rootDir, encoding: 'utf8', stdio: ['ignore', 'pipe', 'ignore'] });
+      } catch {
```
itechmeat, line 437: 🟡 [Not a blocker] This catch turns any `git show` failure into `null`, and a `null` at the base ref means "the PR added this package", which skips it with no log line - so a git failure on one package's manifest converts a real violation into a silent pass for that package. The comment is right that both meanings are handled the same way, and that is the issue: one of them means the gate stopped working here. I could not reproduce a natural trigger (shallow and partial clones both fail earlier and loudly at merge-base), so this is latent - a `git cat-file -e` probe with everything else rethrown would keep it that way.

scripts/version-bump-on-change-check.mjs
```diff
@@ -0,0 +1,481 @@
...
+  const log = options.log ?? console.log;
+  const logError = options.logError ?? console.error;
+  const env = options.env ?? process.env;
+  const argv = options.argv ?? process.argv.slice(2);
+
+  const baseRef =
+    options.baseRef ??
+    baseRefFromArgv(argv) ??
+    env.VERSION_BUMP_BASE_REF ??
+    (env.GITHUB_BASE_REF ? `origin/${env.GITHUB_BASE_REF}` : undefined) ??
```
itechmeat, line 486: 🔵 [Not a blocker] All 23 tests pass `baseRef` straight through the `runCli` option, so the precedence ladder (argv, `VERSION_BUMP_BASE_REF`, `origin/${GITHUB_BASE_REF}`, the `origin/develop` default) is never exercised - including the one expression CI actually depends on. A regression there turns the gate into a permanent pass or a permanent red with no test to catch either. Since `argv` and `env` are already injectable options, three assertions on which ref reaches `git.mergeBase` would cover it.

scripts/version-bump-on-change-check.mjs
```diff
@@ -0,0 +1,481 @@
...
+    if (baseVersion === headVersion) unbumped.push({ packageRoot, version: headVersion, files });
+  }
+
+  if (unbumped.length > 0) {
+    logError(
+      `[version-bump-on-change-check] FAIL: ${unbumped.length} package(s) have substantive changes but no ` +
+        `version bump:`,
+    );
+    for (const { packageRoot, version, files } of unbumped) {
+      logError(`  ${packageRoot} is still at ${version} despite substantive changes:`);
```
itechmeat, line 379: 🔵 [Not a blocker] The heading says "despite substantive changes" but the list under it is the package's whole changed-file bucket, unfiltered - a PR touching `README.md` alongside `src/x.ts` sees both, so the message teaches that the README was substantive, which is the opposite of the rule. `hasSubstantiveChange` returns a boolean and discards which file tripped it, so either it could return the matching file, or the heading could just say "changed files".

CONTRIBUTING.md
```diff
@@ -80,6 +80,8 @@ The project is **pre-1.0** — backward compatibility is not guaranteed.
 - **Patch bump** (`0.1.0` → `0.1.1`) — non-breaking fixes/features
 - **Alpha increment** (`alpha.0` → `alpha.1`) — each merge to develop
 
+A PR that changes non-documentation source under `src/`, or the dependency fields of `package.json`, for a governed package (non-private `@gears-frontx`-scoped packages: `packages/*`, `template-shell`, and `template-shell/packages/*`) must bump that package's own `version` in the same PR — and update every exact pin on it (see `policy:template-pin-drift`). CI (`policy:version-bump-on-change`, pull requests only) compares the version at the PR's merge base against the version at its head, so a bump that is later reverted within the same PR does not count. Packages the PR itself adds or removes are exempt, as are private packages (e.g. the `template-mfe` fixture apps), which are pinned consumers rather than published sources.
```
itechmeat, line 83: 🔵 [Not a blocker] Two wording mismatches, both inert today. `readEcosystemPackages` returns every `packages/*` directory with a manifest and applies no scope filter (the scope matcher only reaches the template branch), so the coded rule is "any non-private `packages/*` plus scoped template members" - an unscoped `packages/foo` would be governed contrary to this sentence. And `private: true` is presented as what excludes `internal/*`, but `internal/` is never walked by discovery at all; both its manifests happen to be private, so the outcome coincides while the stated mechanism does not.

.github/workflows/main.yml
```diff
@@ -123,6 +129,25 @@ jobs:
...
+    # itself and every `template-shell/packages/*` member (any non-`private`
+    # `@gears-frontx`-scoped package; `template-mfe` and its fixture apps are
+    # `private` and exempt) — whose `src/` (or package.json dependencies)
+    # changed without its own `version` changing — comparing actual ref
+    # content rather than the diff, so a version bumped then reverted within
+    # the same PR still fails. Pull-request-only: the merge base it needs is
+    # the PR's base branch, which only exists as `GITHUB_BASE_REF` on
+    # `pull_request` events.
+    - name: Check version bump on substantive package change
+      if: github.event_name == 'pull_request'
```
itechmeat, line 148: 🔵 [Not a blocker] There is no ignore file, no commit-message opt-out and no per-package waiver - the only flag moves the comparison point. An eslint autofix adding a line to one `src/` file in six packages fails all six (reproduced); on the real repo that is fourteen packages plus every exact pin on them plus two regenerated lockfiles, since `policy:template-pin-drift` runs in the same job. That may well be the intended cost, but it is worth an explicit decision rather than a discovery during the first repo-wide lint sweep. Related: CONTRIBUTING now requires a bump per PR while the section above it describes the alpha increment as "each merge to develop" - the two coincide only while every merge is one PR.

scripts/version-bump-on-change-check.mjs
```diff
@@ -0,0 +1,481 @@
...
+ *  - `package.json` ITSELF, but only when a dependency field
+ *    (`dependencies`/`devDependencies`/`peerDependencies`/
+ *    `optionalDependencies`) actually changed between the merge base and
+ *    HEAD - not merely appearing in the diff (a version-only edit, a
+ *    `scripts` tweak, must not itself force a second bump).
+ *
+ * Everything else - README changes, build config, a `package.json` edit that
+ * touches no dependency field - is out of scope for this gate by
+ * construction: it is simply never classified as substantive. This repo
+ * colocates tests under `src/` (e.g. `src/tokens.test.ts`), so the issue's
```
itechmeat, line 66: ⚪ [Not a blocker] The docblock justifies "a test-only change is still substantive" by saying the repo colocates tests under `src/`, but `template-shell/packages/react` keeps a root-level `__tests__/`, as do framework, mfes and template-shell itself - so a test-only PR in react is not substantive, contrary to the paragraph. Low impact since tests are not shipped, but the stated rationale does not hold for the package the gate was built around.

scripts/version-bump-on-change-check.mjs
```diff
@@ -0,0 +1,481 @@
...
+ * read directly from that ref's blob rather than from a diff - see the
+ * module docblock for why that is what makes bump-then-revert fail.
+ *
+ * @param {string} ref
+ * @param {string} packageRoot
+ * @param {GitReader} git
+ * @returns {string | null} `null` when the manifest does not exist at `ref`,
+ *   or declares no valid `version`
+ */
+export function versionAt(ref, packageRoot, git) {
```
itechmeat, line 300: ⚪ [Not a blocker] `versionAt` and `hasSubstantiveChange` are exported but imported nowhere - the test file takes five other functions and no other module imports this one - while the docblock explains the exports as unit-test surface. The sibling `template-pin-drift-check.mjs` exports exactly what its test imports; knip stays quiet here because of `ignoreExportsUsedInFile`. Either import them in the tests, which would also close some of the coverage gaps, or drop the `export`.

scripts/version-bump-on-change-check.test.mjs
```diff
@@ -0,0 +1,408 @@
...
+    git(root, ['checkout', '-q', '-b', 'feature']);
+    await writeJson(path.join(root, 'template-shell', 'packages', 'react', 'src', 'index.ts'), 'export {};\n');
+    commitAll(root, 'change react src without bump');
+
+    const { exitCode, output } = run(root);
+    expect(exitCode).toBe(1);
+    expect(output).toContain('template-shell/packages/react');
+  });
+
+  // Change #1: template-shell itself is now governed (it publishes dist-lib
```
itechmeat, line 255: ⚪ [Not a blocker] "Change #1" points at an item in a review iteration that stops existing once this merges, so a later reader cannot resolve it, and the sentence reads as an argument addressed to a reviewer rather than a statement about the system (same shape at :156 and :261). The constraint after the colon is worth keeping; just the label can go.

.github/workflows/main.yml
```diff
@@ -123,6 +129,25 @@ jobs:
     - name: Check template ecosystem pin-drift
       run: npm run policy:template-pin-drift
 
+    # #575, follow-up from #574 review: #574 changed `@gears-frontx/react`'s
```
itechmeat, line 132: ⚪ [Not a blocker] The #574 story and the consistency-versus-change distinction are now told in the module docblock, in this 15-line CI step comment, in CONTRIBUTING.md and in the commit body. The neighbouring `#493` and `#524` steps carry comparable comments, so this is house precedent rather than a deviation - just noting that four copies of one rationale are four things to keep in sync, and one home with pointers would drift less.

itechmeat (changes requested):
The gate is the right tool and it works where it counts: replaying #574's file set without the bump does fail, discovery yields exactly the declared package set, and the git mechanics hold up including the merge-ref shape GitHub checks out on a PR. What holds it back is a handful of small, verified holes rather than the design - a renamed file hides the package that lost it, an empty discovery result exits green, a reordered dependency key fails an honest PR, and the success line claims a count it never measured. All four are a few lines each. The larger conversation, which does not have to block this, is whether `src/` is the right proxy for "the published payload changed" - for template-shell it currently watches 10 of its 98 files.

scripts/version-bump-on-change-check.mjs
tscbmstubp: Fixed in e890c6d4: `listChangedFiles` now passes `--no-renames`, so both sides of a move are listed and both packages get asked to bump. Covered by a new test that `git mv`s a module verbatim (100% similarity, so default rename detection would have collapsed it) from `packages/api` to `packages/mfes` and asserts both packages fail.

scripts/version-bump-on-change-check.mjs
tscbmstubp: Fixed in e890c6d4: zero discovered roots now fails loudly, same rule (and comment rationale) as `template-pin-drift-check.mjs` — an empty discovery means the gate is broken, and its silence reads exactly like a pass. New test runs the gate on a repo with no `packages/` directory and asserts exit 1.

scripts/version-bump-on-change-check.mjs
tscbmstubp: Fixed in e890c6d4: `dependencyFieldsOf` now sorts each dependency field's entries by key before the stringify comparison, so a pure reorder no longer reads as a dependency change. The new test builds the two manifests from JSON text rather than object spread, precisely because of the order-preserving spread trap you pointed out at :182-185.

scripts/version-bump-on-change-check.mjs
tscbmstubp: Fixed in e890c6d4: the loop now counts packages that actually had substantive changes, and the pass branch reports that measured count — with a distinct wording ("none of N governed package(s) had substantive changes") when nothing substantive changed, so a scripts-only PR no longer claims 14 packages bumped. Both variants are asserted in new tests.

scripts/version-bump-on-change-check.mjs
tscbmstubp: Agreed the gap is real and build config is the sharpest edge of it — `tsup.config.ts` changing the entry-point set is a published-surface change by any reasonable definition. Not folding it into this PR: widening the substantive set changes what every future PR is charged for, and I'd rather that be its own reviewed decision than a rider on the gate's introduction. Will open a follow-up issue proposing root-level build configs (and possibly manifest-`files`-listed directories like `skills/**`) as substantive.

scripts/version-bump-on-change-check.mjs
tscbmstubp: Agreed — "in the manifest's `files`" is the right line, and it cleanly preserves the ui-kit colocated-docs exclusion while catching `llms.txt`/`CLAUDE.md` republishes. Deferring to the same follow-up as the substantive-scope thread above, since both amount to "consult the manifest's `files` instead of hardcoding `src/`".

scripts/version-bump-on-change-check.mjs
tscbmstubp: Good catch — reading HEAD is intentional (it's what makes bump-then-revert fail), so this stays a wording issue. Will fix the message to say the bump must be committed before rerunning; leaving it out of the amended commit only because the thread batch was already pushed — it'll ride with the follow-up.

scripts/version-bump-on-change-check.mjs
tscbmstubp: Agreed it's latent rather than live, and the `git cat-file -e` probe (missing-at-ref → null, anything else rethrown) is the right shape — it keeps "the gate stopped working" from wearing "the package is new" as a disguise. Deferring to the follow-up rather than the amended commit; the blockers took priority for this push.

scripts/version-bump-on-change-check.mjs
tscbmstubp: Fair — the precedence ladder is the one expression CI actually depends on and nothing exercises it. Since `argv` and `env` are already injectable, the three assertions are cheap; will add them in the follow-up alongside importing `versionAt`/`hasSubstantiveChange` directly (your export thread below), which closes both gaps at once.

scripts/version-bump-on-change-check.mjs
tscbmstubp: Agreed the heading over-claims. Of the two options, having `hasSubstantiveChange` return the matching file is the better one — it also gives the base-ref-precedence and unused-export threads a natural test surface. Deferred to the follow-up.

CONTRIBUTING.md
tscbmstubp: Both readings are correct — the sentence describes a scope filter that `readEcosystemPackages` doesn't apply, and credits `private: true` for an exclusion (`internal/*`) that discovery simply never reaches. Will reword CONTRIBUTING to state the coded rule as coded. Deferred to the follow-up.

.github/workflows/main.yml
tscbmstubp: Acknowledging this as an explicit decision rather than an accident: no waiver mechanism for now. A repo-wide autofix sweep legitimately changes every package's published payload, so fourteen bumps is the rule working as stated — but if that first sweep proves the cost wrong in practice, an opt-out should be designed deliberately (probably a commit-trailer waiver, not an ignore file, so it stays visible in review). Also agreed on the CONTRIBUTING per-PR vs per-merge tension — will align the wording in the follow-up.

scripts/version-bump-on-change-check.mjs
tscbmstubp: Right — the rationale over-generalizes from ui-kit. The rule itself ("src/ is the one signal, tests under it included") stands, but the docblock shouldn't claim colocation as a repo-wide fact when react, framework, mfes and template-shell keep root-level `__tests__/`. Will reword to present it as the rule's deliberate boundary rather than a colocation claim, in the follow-up.

scripts/version-bump-on-change-check.mjs
tscbmstubp: Will import them in the tests rather than drop the exports — as you note, that also closes real coverage gaps (`versionAt`'s null branches and `hasSubstantiveChange`'s docs-under-src short-circuit are currently only reachable end-to-end). Follow-up.

scripts/version-bump-on-change-check.test.mjs
tscbmstubp: Agreed — the "Change #1" labels are review-iteration residue that stops resolving the moment this merges. Will keep the constraints, drop the labels (here, :156 and :261) in the follow-up.

.github/workflows/main.yml
tscbmstubp: Fair observation. Given the neighbouring #493/#524 steps set the precedent, I'll keep the four copies for now but treat the module docblock as the one home if they ever need to change — happy to collapse the CI-step comment and CONTRIBUTING passage to pointers if drift actually shows up.

scripts/version-bump-on-change-check.mjs
```diff
@@ -0,0 +1,506 @@
...
+  const changedFiles = git.listChangedFiles(mergeBaseSha);
+  const grouped = groupChangedFilesByPackageRoot(changedFiles, packageRoots);
+
+  /** @type {UnbumpedPackage[]} */
+  const unbumped = [];
+  let substantiveCount = 0;
+
+  for (const [packageRoot, files] of grouped) {
+    if (!hasSubstantiveChange(packageRoot, files, mergeBaseSha, git)) continue;
+    substantiveCount += 1;
```
itechmeat, line 359: 🟡 [Not a blocker] `substantiveCount` increments before the two skip paths below it (the brand-new-package and removed-package `continue`s), so a package that was never version-checked lands in a sentence asserting it "also bumped its version". Reproduced: a PR that only adds one new governed package with `src/` prints "every one of 1 package(s) with substantive changes ... also bumped its version" with nothing bumped anywhere, and a new-plus-bumped mix prints "every one of 2" when one was checked. Moving the increment below the skips just moves the inaccuracy, so the cheapest honest fix is probably the wording - something like "either bumped its version or was itself added or removed by this PR", which is true by construction at that point since `unbumped` is empty.

scripts/version-bump-on-change-check.mjs
```diff
@@ -0,0 +1,506 @@
...
+    );
+    return 1;
+  }
+
+  const packageRoots = discoverPackageRoots(rootDir);
+  // Zero discovered roots is never a silent pass (same rule as
+  // `template-pin-drift-check.mjs`): this repo always has governed packages,
+  // so an empty result means discovery is broken - and a broken discovery
+  // would otherwise report success in exactly the same voice as a real one.
+  if (packageRoots.length === 0) {
```
itechmeat, line 343: 🔵 [Not a blocker] The delta adds a whole new refusal path, but its rationale lives only in the body comment - the docblock's discovery section still describes what qualifies as a package and never that an empty result is now a hard failure. The sibling this rule is borrowed from states its own equivalent right in its docblock ("and zero templates found is a hard failure, never a vacuous pass"), so one clause here would keep the two readable side by side and let a maintainer predict the FAIL from the docs.

itechmeat (approved):
All four blockers are properly closed, and closed at the level of behaviour rather than wording: each fix reproduces green on a real temp repo, each got a test, and all five new tests fail against the previous revision for the right reasons - so they pin the fixes instead of merely accompanying them. `--no-renames` now names both sides of a move (and catches a move out of `src/` as a bonus), the empty-discovery guard fails closed in the sibling's voice, the key sort is a pure canonicalization that leaves the check just as sharp, and the success line reports a measured count. Approving. One small thing left over from the fourth fix, and one doc clause, both noted inline and neither worth holding this up.

GeraBart (approved):

PR #579, opened 2026-08-17
feat(telemetry): let a host turn navigation capture off

hello1101n:
The navigation plugin is unconditional today — `start()` registers it and it wraps `window.history.pushState`/`replaceState` for the client's lifetime. An app that reports its own page views cannot opt out, and inherits the history patching for enrichment it does not want.

`navigationCapture` (default on) is decided where the built-ins are composed, not inside the plugin:

```ts
plugin(
  sessionPlugin(),
  devicePlugin(),
  config.navigationCapture && navigationPlugin(),
  telemetryAppInfoPlugin(),
  autocapturePlugin(),
);
```

`plugin()` already filters falsy entries so conditional registration needs no branch at the call site, so off means the plugin is never registered — nothing wrapped, no `page_view`, no `caused_by_id`. The plugin itself still reads no configuration.

```ts
createTelemetry({ appName: 'insight', appVersion: '1.0.0', navigationCapture: false });
```

**Why:** Insight's usage monitoring records visits server-side where sessions are minted and reports viewed pages from its own beacon. A second `page_view` stream and a second owner of `window.history` are cost with no return.

**Note:** this deliberately does not follow `autocapture`, which reads its flag inside its own setup. If a third built-in needs an opt-out, the better answer is one `builtins` map covering all of them, with `autocapture` folded in — tracked as #581 rather than left here.

**Artifacts:** the `inst-register-builtins` step, the built-in-context DoD and two acceptance criteria in event-collection `FEATURE.md`; the client and navigation lines in `DESIGN.md`. PRD untouched — it does not spec the `autocapture` toggle either.

**Checks:** `type-check`, `type-check:test`, `type-check:demo`, `lint`, 47 telemetry tests (4 new; 3 of them failed before the change).


packages/telemetry/architecture/features/event-collection/FEATURE.md
```diff
@@ -450,7 +450,7 @@ The system **MUST** key plugins by name so a later registration replaces an earl
...
...an event on every navigation path change including History API transitions — stamping each later record with that page view's identifier as its `caused_by_id`, so every record links to the page it occurred on — and **MUST** offer a locale plugin that reads the application's required locale source per record rather than capturing it at setup — when no locale enrichment sets a language, the device plugin fills the browser's reported one. Because the History API fires no event for `pushState` and `replaceState`, the navigation plugin **MUST** wrap those two methods to observe transitions, and **MUST** restore the original methods on teardown. Consumers should know the SDK patches `window.history`; any other code wrapping the same methods must compose with this. Where navigation capture is disabled in configuration the client **MUST NOT** register the navigation plugin at all, so nothing is wrapped, no page view is recorded and no record is attributed, and an application that owns its own page-view reporting keeps sole possession of `window.history`. The plugin itself reads no configuration; the decision is the client's, taken through the falsy-entry seam registration already provides.
```
itechmeat: 🟡 [Not a blocker] Turning the flag off also removes a session-refresh source: the navigation plugin's trigger calls refreshSession on every path change and owns the only popstate listener, while the session manager counts only scroll/keydown/click as activity. So with navigationCapture: false an app that navigates by timer or redirect can see its session lapse mid-flow. The DoD sentence reads as an exhaustive consequence list and this fourth consequence is not in it, nor in the README paragraph - worth one clause in each, or a correction if I am misreading the intent.

packages/telemetry/src/services/telemetry.ts
```diff
@@ -101,7 +101,7 @@ export function createTelemetry(configRaw: TelemetryConfig): TelemetryService {
     plugin(
       sessionPlugin(),
       devicePlugin(),
-      navigationPlugin(),
+      config.navigationCapture && navigationPlugin(),
```
itechmeat, line 106: 🟡 [Not a blocker] The builtins-map deferral is reasonable and well argued, but it lives only in the PR description - no issue, no pointer. Deferred API-shape decisions recorded only in PR bodies tend to get lost after merge: could it get a tracked home (an issue or a TODO pointing at one)?

packages/telemetry/architecture/features/event-collection/FEATURE.md
```diff
@@ -144,7 +144,7 @@ Realizes the collection and delivery requirements of the SDK and the sequence `c
    1. [x] - `p1` - Report through the logger (surfaced when verbose) that the client is already started and a new client should be built - `inst-log-restart-refused`
    2. [x] - `p1` - **RETURN** the client unchanged - `inst-return-already-started`
 3. [x] - `p1` - Set the start flag, which teardown deliberately never clears - `inst-set-started`
-4. [x] - `p1` - Register the built-in session, device, navigation, application-info and autocapture plugins, after any the caller registered so a colliding caller plugin is replaced - `inst-register-builtins`
+4. [x] - `p1` - Register the built-in session, device, navigation, application-info and autocapture plugins, after any the caller registered so a colliding caller plugin is replaced, passing navigation through as falsy where navigation capture is disabled in configuration so it is never registered - `inst-register-builtins`
```
itechmeat, line 147: 🔵 [Not a blocker] With the flag off the name `navigation` stops being reserved: the untouched comment above the plugin list (telemetry.ts line 99, "a caller plugin using a built-in name is the one that gets replaced. This is why those names are reserved") is now conditionally false, since a caller plugin under that name survives and gets set up - a registry-visible difference from autocapture: false, where the built-in still occupies its slot. Might even be useful; either way one clause on the comment or the DoD would make it deliberate.

packages/telemetry/README.md
```diff
@@ -43,6 +43,10 @@ and installs their listeners. Out of the box that gives you a `session_start` ev
 session, a `page_view` on every path change (`pushState`, `replaceState`, `popstate`), device, OS,
 client, viewport and timezone fields on every record, and captured user interactions.
 
+An application that routes page views itself can turn that one off with `navigationCapture: false`.
```
itechmeat: 🔵 [Not a blocker] With the plugin off nothing can supply attribution at all - autocapture records simply stop carrying `caused_by_id`. A host that owns page views can stamp it itself from an `event` hook (or register its own plugin under the `navigation` name), and the API is already there - worth a line next to this paragraph so the trade-off and the escape hatch are both visible.

packages/telemetry/src/utils/types.ts
```diff
@@ -33,6 +34,12 @@ export type TelemetryConfig = {
    * @default true
    */
   autocapture?: boolean;
+  /**
+   * Send a `page_view` on every path change, and stamp later records with the page they happened on.
+   * With it off the navigation plugin is not registered and `window.history` is left unwrapped.
+   * @default true
+   */
+  navigationCapture?: boolean;
```
itechmeat, line 42: 🔵 [Not a blocker] PRD's `fr-builtin-context` still states the enrichment MUST unconditionally, and the package precedent points both ways: `enabled` earned its own FR (`fr-delivery-disable`) while `autocapture` never got one - the body says PRD was left untouched for that reason. Since telemetry sits in the artifact chain with FULL traceability, was skipping the PRD deliberate, or is a qualifying clause in `fr-builtin-context` worth adding while the area is warm?

packages/telemetry/architecture/DESIGN.md
```diff
@@ -369,7 +369,7 @@ Supplies the context that makes an event stream answerable — session, device,
 
 - Session: stamps the current session onto records and emits an event when a session begins.
 - Device: derives browser, operating system and platform fields from the user agent, plus viewport and timezone.
-- Navigation: emits an event on every path change, including History API transitions and back-forward navigation.
+- Navigation: emits an event on every path change, including History API transitions and back-forward navigation. Registered only when configuration leaves navigation capture on, so switching it off costs the plugin nothing and it reads no flag of its own.
```
itechmeat: ⚪ [Not a blocker] The same three facts (not registered / history untouched / no caused_by_id) are now written out in four places - README, the JSDoc, DESIGN and the DoD - and the DoD's last sentence ("the decision is the client's, taken through the falsy-entry seam...") is mechanism, not requirement; DESIGN already owns that, and "reads no flag of its own" also duplicates the nearby normalized-config line. Trimming the DoD to the requirement and letting DESIGN carry the how would keep each artifact in its lane.

packages/telemetry/README.md
itechmeat: ⚪ [Not a blocker] Two wording nits in one place: "that one" points back across a sentence listing four things, so the referent is guesswork - naming it ("can turn page-view capture off") fixes it; and the untouched line above ("start() registers the built-in plugins - session, device, navigation, app info and autocapture") is now conditionally untrue, which the new paragraph could acknowledge in passing ("all but navigation unconditionally", or similar).

itechmeat (approved):
Clean, well-scoped change - the flag rides the falsy-entry seam the package already documents, the artifact chain moves in step with the code (down to repairing a stale test count in the README), and the new tests assert observable behaviour rather than plumbing. Nothing here blocks approval: my comments are one behavioural consequence the docs don't yet name, a couple of doc-precision points, and a request to give the deferred builtins-map idea a tracked home. All answerable in-thread.

packages/telemetry/architecture/features/event-collection/FEATURE.md
hello1101n: Confirmed: `refreshEvents` is scroll/keydown/click, and `navigation.ts` is the only other `refreshSession()` caller.
One correction — `session.ts` `onEvent` mints a new session when the stored one expired, so the effect is one visit split across two sessions, not a lapse mid-flow.
Named as the fourth consequence in the DoD and as a README bullet. c75fc58d

packages/telemetry/architecture/features/event-collection/FEATURE.md
hello1101n: Right, and it breaks a claim in my description: with the flag off nothing occupies the `navigation` name, so a caller plugin under it survives and is set up — unlike `autocapture: false`, where the built-in keeps its slot.
Now stated in the plugin-registry DoD and in the `telemetry.ts` comment. c75fc58d

packages/telemetry/src/utils/types.ts
hello1101n: Deliberate (matching `autocapture`), but the precedent is split — `enabled` has `fr-delivery-disable`, `autocapture` has nothing.
`fr-builtin-context` now adds: **MAY** let an application switch off an enrichment whose signal it already owns. c75fc58d

packages/telemetry/architecture/DESIGN.md
hello1101n: Trimmed — the DoD keeps the requirement only, DESIGN keeps the seam, and "reads no flag of its own" is gone since the boundary line above it already says that.
DESIGN also picked up the session refresh, which is what makes the fourth consequence legible. c75fc58d

packages/telemetry/README.md
hello1101n: Both fixed: the referent is named ("turn page-view capture off"), and the line above now reads "All but navigation are registered unconditionally". c75fc58d

hello1101n: Verified — `caused_by_id` is writable from the `event` hook, which is how navigation sets it.
The README bullet now names the loss and both escape hatches: stamp it from an `event` hook, or register your own plugin under the now-free `navigation` name. c75fc58d

packages/telemetry/src/services/telemetry.ts
hello1101n: #581 — carries both mechanisms, the four-edit cost per opt-out, and the open questions: rename vs alias `autocapture`, which built-ins to expose, which name-reservation behaviour the map locks in.
Linked from the description too.

GeraBart (approved):

PR #580, opened 2026-08-17
feat(telemetry): redact identifying values from recorded urls

hello1101n:
A `page_view` path and an autocaptured link url reach the collector raw. Applications put people and records in the address itself (`/invite/john@x.com`, `/users/98765`), so the url stream carries identity even where no field does — and dropping the url instead would remove the unit an adoption question is asked in.

`redactUrls` (off by default) rewrites values to a placeholder naming the shape, keeping the address groupable:

```ts
createTelemetry({
  appName: 'insight',
  appVersion: '1.0.0',
  redactUrls: true,
  sanitizeUrl: (url) => url.replace(/^\/o\/\d+/, '/o/:orgId'),
});
```

| Whole segment or query value | Becomes |
| --- | --- |
| `john@x.com` | `:email` |
| a UUID | `:uuid` |
| a JWT (`eyJ…`, two dots) | `:token` |
| 5+ digits | `:id` |
| 16+ hex chars | `:hash` |

**Deliberately narrow.** Five-or-more digits rather than any number, and JWTs anchored on the `eyJ` that base64 of `{"` always produces, so `/reports/2026`, `/docs/Chapter-2-Introduction` and `www.example.com` in an absolute href all survive. The cost is in the README rather than hidden: `/users/8` and `?name=Nikita` match nothing. That is what `sanitizeUrl` is for — it runs **first**, on the raw url, because only the host's router knows which segments are ids, and whatever it returns is still swept by the patterns. A hook that throws is logged and the patterns run anyway; a broken rule must not publish what it was written to remove.

**One policy, two callers.** `utils/url.ts` is shared by the navigation plugin and the autocapture walk, so a value in a clicked link gets the same treatment as one in a path change — `$el_attr_href` and `$external_click_url` included, query strings and hash-router fragments included. In autocapture it runs *after* the existing value-safety gate, so cards and national ids still drop the whole value rather than getting rewritten.

**Known gaps**, stated in the README: query parameter *names* are untouched, and `src`/`action` are not treated as urls, so a form action carrying an id is recorded as-is.

**Artifacts:** PRD requirement `cpt-frontx-telemetry-fr-url-redaction`, DESIGN component `cpt-frontx-telemetry-component-url-redaction` wired into `architecture/DECOMPOSITION.md`, an event-collection algorithm + DoD with `@cpt` code markers on every instruction, and the autocapture walk's anchor step.

**Checks:** `type-check`, `type-check:test`, `type-check:demo`, repo `lint`, `tsup` build, 61 tests (18 new — every pattern, both call sites, the hook order and the throw path). `studio validate` and `spec-coverage --min-coverage 38` both pass locally.

**Note:** touches the same `types.ts` / `utils.ts` / README lines as #579, so whichever lands second needs a rebase.


packages/telemetry/README.md
```diff
@@ -241,6 +243,43 @@ single-use, so the id is minted by the next client's `start()`:
...
+```ts
+createTelemetry({
+  appName: 'cloud',
+  appVersion: '1.0.0',
+  redactUrls: true,
+  sanitizeUrl: (url) => url.replace(/^\/o\/\d+/, '/o/:orgId'),
+});
+```
+
+Whatever it returns is still swept by the built-in patterns. If it throws, the error is logged and
```
itechmeat: 🟠 README (lines 66 and 276), the sanitizeUrl JSDoc and the event-collection DoD all say the hook's output "is still swept by the built-in patterns" unconditionally, but the code sweeps only when `redactUrls` is on (`url.ts:47`), and the PR's own test pins the unswept case. DESIGN.md:444 states the gated behaviour correctly, so the surfaces also disagree with each other. Either sweep whenever a hook is configured, or qualify all four sentences with "when `redactUrls` is on" - as written, a consumer reads a privacy guarantee the code does not provide.

packages/telemetry/architecture/PRD.md
```diff
@@ -297,6 +297,16 @@ The system **MUST** reduce accidental sensitive-data capture through default sup
 
 **Actors**: `cpt-frontx-telemetry-actor-end-user`
 
+#### Identifying values in recorded addresses
+
+- [x] `p2` - **ID**: `cpt-frontx-telemetry-fr-url-redaction`
```
itechmeat, line 302: 🟠 The new FR and component are wired into PRD, DESIGN and DECOMPOSITION, but neither FEATURE picked them up: event-collection's Requirements/Components lists (FEATURE.md:57/:61) still lack `…fr-url-redaction`/`…component-url-redaction`, and dom-autocapture's (:49/:53) likewise - so on the artifacts' own terms the requirement is realized by no feature and the component exercised by none, even though the algorithm and DoD live in event-collection. `cfs validate` will not catch this: it pairs markers with instructions, it does not check membership lists.

packages/telemetry/src/utils/url.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  try {
+    // @cpt-begin:cpt-frontx-telemetry-algo-event-collection-url-redaction:p2:inst-apply-host-hook
+    sanitized = sanitizeUrl(url);
+    // @cpt-end:cpt-frontx-telemetry-algo-event-collection-url-redaction:p2:inst-apply-host-hook
+  } catch (error: unknown) {
+    // @cpt-begin:cpt-frontx-telemetry-algo-event-collection-url-redaction:p2:inst-recover-host-hook
+    // Returning the raw url here would publish the values the host asked to have removed, so the
+    // built-in patterns run even where configuration leaves them off.
+    logger.logError('Telemetry sanitizeUrl threw; falling back to the built-in patterns.', error);
```
itechmeat, line 41: 🟡 [Not a blocker] Two gaps in the DoD sentence "a rewrite that throws MUST be reported and MUST NOT result in the raw url being recorded": the report goes through `logError`, which is verbose-gated and silent at the default config (the same file's wording for the analogous case is "surfaced when verbose", FEATURE.md:108); and the fallback runs the patterns over the raw url, which comes back verbatim when nothing matches - the `/o/42` from the example stays raw if the hook throws. Worth rewording the DoD to what the code guarantees, or dropping the url on a throw if the absolute promise is the intent - which was meant?

packages/telemetry/architecture/DESIGN.md
```diff
@@ -416,6 +418,35 @@ Records interaction without per-element instrumentation, which is the part of an
...
+
+Urls are recorded from two unrelated places — a path change and a clicked link — and an identifying value carries the same weight in both. One policy called by both keeps a single set of recognized shapes, rather than a rule per recording site that drifts apart.
+
+##### Responsibility scope
+
+- Applies an application-supplied rewrite to the raw url first, so a host maps its own routes against real values.
+- Reports a throwing application rewrite and applies its own patterns to the raw url instead, so a failed rewrite cannot publish what it was meant to remove.
+- Replaces whole path segments, query values and fragment segments that match a recognized identifying shape with a placeholder naming the shape.
+- Tests the percent-decoded form, so an encoded value is not missed.
+- Treats the fragment as a path, so a hash-routed application is covered.
```
itechmeat: 🟡 [Not a blocker] Navigation records `window.location.pathname` and dedupes on it - a hash-only route change never emits a page_view, and the recorded string cannot contain `#` or `?` by construction, so the fragment and query branches only ever fire on captured hrefs. The policy does treat fragments; "a hash-routed application is covered" (also in the FEATURE step, inst-split-url) overstates it - covered for links, not for hash routing. Worth saying which one it is.

packages/telemetry/README.md
```diff
@@ -241,6 +243,43 @@ single-use, so the id is minted by the next client's `start()`:
...
+the built-in patterns run on the raw url anyway - a broken rule must not publish what it was written
+to remove.
+
+Two gaps worth knowing: query **names** are never rewritten (only values), and `src` and `action`
```
itechmeat: 🟡 [Not a blocker] The two new gaps are appended inline while the README already has a `## Known gaps` section right below - they belong in that list, or the list stops being the list. The throw-fallback reasoning is also now told in five places (README, DESIGN scope, FEATURE step, DoD, and a comment in url.ts) - one home pointed at from the rest would age better. And `## Urls` reads as a subsection of `## PII` rather than a sibling.

packages/telemetry/README.md
```diff
@@ -241,6 +243,43 @@ single-use, so the id is minted by the next client's `start()`:
...
+
+`sanitizeUrl` covers what the SDK cannot recognize. It runs first, on the raw url, because your
+router is the only thing that knows which segments are ids:
+
+```ts
+createTelemetry({
+  appName: 'cloud',
+  appVersion: '1.0.0',
+  redactUrls: true,
+  sanitizeUrl: (url) => url.replace(/^\/o\/\d+/, '/o/:orgId'),
```
itechmeat, line 288: 🟡 [Not a blocker] `sanitizeUrl` receives two different shapes: navigation passes `location.pathname` (no origin, no query), autocapture passes the raw href attribute (absolute, `../relative`, `#/x`, `mailto:`). This README example anchored on `^` silently does nothing for an absolute href carrying the same `/o/42`. Nothing tells the consumer - worth a line in the JSDoc/README, or normalizing before the call.

packages/telemetry/src/plugins/autocapture/autocapture.ts
```diff
@@ -278,12 +279,13 @@ export function autocapturePlugin(): TelemetryPlugin {
 
         // @cpt-begin:cpt-frontx-telemetry-algo-dom-autocapture-ancestor-walk:p1:inst-check-anchor
         if (href) {
-          result.$el_attr_href = href;
+          const recordedHref = redactUrl(href, context);
+          result.$el_attr_href = recordedHref;
```
itechmeat, line 285: 🔵 [Not a blocker] The walk already records every non-ignored attribute raw as `$el_attr_<name>`, so a form `action`, an img `src` or a `data-return-url` land unredacted next to the redacted href. The gap is declared for src/action - but since the policy is already invoked in this very function, is there a reason not to run it over an allowlist of url-ish attribute names, here or in a stacked follow-up?

packages/telemetry/src/utils/url.ts
```diff
@@ -0,0 +1,112 @@
...
+    .join('/');
+}
+
+function redactQuery(query: string): string {
+  return query
+    .split('&')
+    .map((pair) => {
+      const separatorAt = pair.indexOf('=');
+
+      if (separatorAt === -1) {
```
itechmeat, line 82: 🔵 [Not a blocker] `?next=/users/98765` survives whole - the decoded value contains a slash, so no whole-value rule matches - and a redirect parameter is one of the most common places an id ends up in a query. It follows honestly from the whole-segment/whole-value policy, but unlike param names and src/action this case is named nowhere. If nested urls are out of scope by design, the Known gaps list seems the place to say it.

packages/telemetry/src/plugins/autocapture/autocapture.ts
```diff
@@ -278,12 +279,13 @@ export function autocapturePlugin(): TelemetryPlugin {
 
         // @cpt-begin:cpt-frontx-telemetry-algo-dom-autocapture-ancestor-walk:p1:inst-check-anchor
         if (href) {
-          result.$el_attr_href = href;
+          const recordedHref = redactUrl(href, context);
```
itechmeat, line 284: 🔵 [Not a blocker] The rewritten inst-mark-external text now describes recording through the redaction policy, but the `redactUrl` call and the `$el_attr_href` write sit above the `@cpt-begin:…inst-mark-external` marker, inside the parent inst-check-anchor region whose instruction is a bare "IF an anchor target was remembered". The region carrying the behaviour and the instruction describing it do not line up - moving the begin marker up two lines, or splitting the recording into its own numbered instruction, squares it.

packages/telemetry/architecture/DESIGN.md
```diff
@@ -416,6 +418,35 @@ Records interaction without per-element instrumentation, which is the part of an
 
 - `cpt-frontx-telemetry-component-plugins-manager` — registered and set up through it
 - `cpt-frontx-telemetry-component-events-manager` — emits captured events through it, and enriches through its hook
+- `cpt-frontx-telemetry-component-url-redaction` — rewrites an anchor's url before it is recorded
+
+#### Url Redaction
+
+- [x] `p2` - **ID**: `cpt-frontx-telemetry-component-url-redaction`
```
itechmeat, line 427: 🔵 [Not a blocker] The §3.2 component-model diagram still shows seven components - the new url-redaction component and its two declared edges (from Built-In Plugins and Autocapture) exist only in the prose below. Two lines of mermaid keep the first thing a reader sees in sync.

packages/telemetry/src/utils/types.ts
```diff
@@ -10,6 +10,8 @@ export type TelemetryConfigNormalized = {
   verbose: boolean;
   url: string;
   autocapture: boolean;
+  redactUrls: boolean;
+  sanitizeUrl?: (url: string) => string;
```
itechmeat: 🔵 [Not a blocker] The `(url: string) => string` signature is written three times - both types.ts fields plus the Pick in url.ts - while the file already names such shapes (`export type LogEvent = …` sits right below). One `SanitizeUrlFn` alias next to it collapses all three.

packages/telemetry/architecture/features/event-collection/FEATURE.md
```diff
@@ -288,6 +290,23 @@ Realizes the collection and delivery requirements of the SDK and the sequence `c
 3. [x] - `p1` - **FOR EACH** registered plugin, invoke its setup with that context - `inst-invoke-setup`
 4. [x] - `p1` - Leave any plugin registered after this pass stored but not set up - `inst-late-plugin-unset`
 
+### Url Redaction
+
+- [x] `p2` - **ID**: `cpt-frontx-telemetry-algo-event-collection-url-redaction`
```
itechmeat, line 296: ⚪ [Not a blocker] event-collection's Error Scenarios list enumerates observable degradations of exactly this shape (no window, double start, late plugin, rejected delivery) but did not gain the throwing-sanitizeUrl case - it lives only in the instruction and the DoD. One bullet closes it.

packages/telemetry/architecture/features/dom-autocapture/FEATURE.md
```diff
@@ -299,9 +299,12 @@ The system **MUST** abandon the entire event when any element from the target up
 
 The system **MUST** abandon the entire event, not merely the offending element's fields, when any element on the walked path is a password or hidden input or carries a name or identifier matching a sensitive pattern. For free-text and selection controls it **MUST** restrict attribute reading to name, identifier and accessible label. It **MUST** reject values matching payment-card and national-identifier patterns. The documentation **MUST** present this as a pattern-based safety net and not as a compliance guarantee, and **MUST** direct an application to the subtree opt-out or an element-hook veto as the authoritative control.
 
+The url a captured anchor contributes **MUST** pass through the shared url redaction policy rather than a rule of autocapture's own, so an identifying value in a link carries the same treatment as one in a path change.
```
itechmeat: ⚪ [Not a blocker] The sentence lands under the "Whole-Event Redaction" heading, whose DoD is about dropping - whole events on sensitive elements, card/national-id values. Placeholder substitution in a url is the opposite policy, so the heading no longer covers its own contents. event-collection got a properly titled DoD for this behaviour; autocapture's half may deserve the same.

packages/telemetry/src/__tests__/url.test.ts
```diff
@@ -0,0 +1,112 @@
...
+    };
+
+    expect(redactUrl('/users/john@x.com', contextFor({ sanitizeUrl }))).toBe('/users/:email');
+  });
+
+  test('reports a throwing sanitizeUrl through the logger', () => {
+    const sanitizeUrl = () => {
+      throw new Error('router lookup failed');
+    };
+    const context = contextFor({ sanitizeUrl, verbose: true });
```
itechmeat: ⚪ [Not a blocker] The test replaces `logError` with a spy on the context object, so the spy catches the call at any `verbose` - the flag has no effect on the assertion, and the line reads as if it proves the report is surfaced. Dropping the flag (or asserting through a real logger) makes the test say what it means.

packages/telemetry/src/plugins/autocapture/autocapture.ts
```diff
@@ -1,5 +1,6 @@
 import { getEventTarget, isSensitiveElement, isTextNode } from '../../utils/dom';
 import type { TelemetryData } from '../../utils/eventTypes';
+import { redactUrl } from '../../utils/url';
```
itechmeat, line 3: ⚪ [Not a blocker] Url helpers now live in two homes - `convertToURL` in plugins/autocapture/helpers.ts, `redactUrl` in utils/url.ts - and the new code hand-splits on `#`/`?` instead of using `URL`/`URLSearchParams`. If the hand-split is deliberate (`new URL` would absolutize and re-encode, changing what is recorded), a constraint comment saying so would stop the next reader from "fixing" it; merging the two homes could ride the same change.

packages/telemetry/src/utils/types.ts
```diff
@@ -33,6 +35,17 @@ export type TelemetryConfig = {
    * @default true
    */
   autocapture?: boolean;
+  /**
+   * Replace identifying values in recorded urls with placeholders: `:email`, `:uuid`, `:token`,
+   * `:id` and `:hash`. Applies to the `page_view` path and to autocaptured link urls.
+   * @default false
+   */
+  redactUrls?: boolean;
```
itechmeat, line 50: ⚪ [Not a blocker] Nothing in the repo sets `redactUrls`, and there is no other `createTelemetry` consumer here. Who is turning it on - insight-front, cloud? If adoption needs a follow-up on that side, worth a pointer so the flag does not stay dead.

itechmeat (changes requested):
Careful, honest work - the narrow-by-design policy is stated with its costs, the regexes hold up against their own README examples, the flag-off path is verified byte-identical, and the error path goes through the package logger. What blocks approval is not the code but two claim-vs-reality gaps: the docs promise the sanitize hook's output is always swept by the built-in patterns while the code sweeps only behind the flag, and the new PRD requirement and DESIGN component are picked up by no FEATURE, so the traceability chain the PR body advertises stops one link short. Both are cheap to close; the rest is questions and polish.

GeraBart (approved):

hello1101n:
Close for now, I'm focused on other tasks

PR #582, opened 2026-08-17
fix: host-chrome polish - studio defaults, stable test ids, menu grace window, shadow-root style adoption

itechmeat:
This ports the host-chrome fixes accumulated on the demo branches into develop: the studio panel no longer covers a freshly opened project, the shell's chrome exposes stable test ids for browser automation, the menu's empty-state hint waits for MFE discovery to settle instead of flashing before registration finishes, and an MFE's shadow root adopts host styles without letting the shell's preflight outrank the MFE's own CSS.

## What changes

- **The studio panel starts collapsed.** A scaffolded project used to open with the panel expanded over the app; the stored developer preference still wins when present.
- **The studio dropdown's exit animation no longer depends on cascade position.** `.studio-portal-container` scoping lifts the closing rule's specificity above the bare `tailwindcss-animate` utility. Deliberately untested: jsdom resolves neither cascade order across injected sheets nor animations, so a unit test could only assert the presence of the added string.
- **Stable test ids for the host chrome.** `studio/src/testIds.ts` publishes ids for the collapse toggle, theme trigger and options; `GlassmorphicButton` and menu items follow (`menuItemTestId(ext.id)` derives from the extension id - `presentation.route` is neither unique nor required). The `_blank-mfe` scaffold screen gets the same treatment, so every screen copied from it inherits the contract; each id is pinned by the existing HomeScreen suite, and the expand button's testid - the only entry into the now-collapsed-by-default overlay - is asserted by its own test.
- **The menu's empty state waits out a grace window.** The "No screens yet" hint renders only once `EMPTY_STATE_GRACE_MS` passes and the extension list has been read from the current registry, so it no longer flashes while MFEs are still registering or while a registry that arrives late has not been polled yet. The timer runs from mount, independent of the registry, so a shell-only seed with no MFE registry still surfaces the hint once the window closes - it is that scenario's main pointer to `frontx add`. The read is guarded: a screen extension registered without `presentation` used to throw past `setReadRegistry`, stranding the fallback along with the hint it exists to show; the read is now wrapped, with the mark applied in a `finally` since it stands for the read having happened rather than having succeeded, so settlement always completes. `EMPTY_STATE_GRACE_MS` is documented as a floor on how early the no-screens claim may be made, not a signal that discovery has settled.
- **Host styles adopt into the front of the shadow root.** `ThemeAwareReactLifecycle` inserts the host's adopted styles before the MFE's own, so the shell's preflight stops outranking an MFE's CSS. The adopted `<style>` and `<link>` pieces are collected in host-document order, interleaved as the host declares them, not grouped by kind - grouping would reorder the cascade for equal-specificity rules. Both the adopted block and the base resets upsert on remount - a marker attribute plus remove-before-insert for the former, an id lookup plus overwrite for the latter, mirroring the upsert `injectStylesheet` already does in `packages/mfes` - so a remounted container replaces its previous block instead of stacking a copy behind it.
- **`useApiStream` TSDoc reconciled with the implementation** that #483 landed (render-derived reset), plus the superseded-language-load regression test develop lacked; the commit deliberately carries both halves, its subject names them. `'connecting'` is documented as published from the render that enables the stream, with `descriptor.connect()` itself running in the post-commit effect. A key or mode change while disabled is now documented separately: it still resets `data`/`events`/`error`, but writes no `'connecting'`, and the published status stays `'idle'` until the later flip to enabled.
- **ADR-0031 moved to `packages/cli/architecture/ADR/`** per the review note on #548 (leftover of the SDLC-artifacts split): content byte-identical (`rename 100%`), seven links in the cli DESIGN rewritten, and `artifacts.toml` taught that the cli member owns ADRs by mirroring the root ADR pattern - the first package-level ADR in the repo. Root numbering now skips 0031; the explorations file's mentions of a different, superseded "ADR-0031" predate this PR and are left alone.

## Verification

- template-shell: type-check clean, lint 0 warnings, test:unit 61 files / 611 tests, arch:deps 378 modules / 989 deps, no violations
- root: type-check:all clean, lint clean, arch:check 12/12
- composed shell + MFE overlay (the Template Validate path): install, build, type-check, arch:deps, lint, test:unit all green, incl. the `_blank-mfe` suite (8 files / 21 tests)
- studio: validate 70 artifacts / 0 errors, coverage 171/171; validate-kits 2/2; spec-coverage thresholds met
- every behavioural change is mutation-checked: reverting each fix reddens exactly the test that pins it (collapsed default, menu item ids, grace gate incl. the registry-less and late-registry paths, missing-`presentation` read guard, studio ids incl. the expand testid, shadow-root ordering incl. the interleaved style/link order and remount dedup of the adopted block and base resets, superseded-load derivation, scaffold ids)

## Coordination

- Routing stays out deliberately - screen-by-URL and menu route pushes are another lane; this diff carries zero routing semantics (verified by grep and by review).
- #564 edits ADR-0031 at its root path and will meet a trivial rename conflict after this merges.
- No files are shared with #566; its vitest-config rewrite still picks up the new react-package test through the shared include glob.
- Residue `@cpt-` markers were stripped from every touched template-territory file per ADR-0033; none added.

## Summary by CodeRabbit

- **New Features**
  - Studio panels now start collapsed when no saved preference exists.
  - Menus wait briefly for extension discovery before showing an empty state.
  - Theme and menu controls provide stable identifiers for browser automation.
  - Theme triggers display registered theme names consistently.
  - Studio dropdown animations and adopted styles preserve expected ordering and transitions.

- **Bug Fixes**
  - Prevented outdated language loads from overriding current translation state.
  - Improved menu resilience when extension discovery fails.

- **Documentation**
  - Clarified API connection lifecycle behavior.
  - Documented template ownership boundaries and optional ADR traceability.


template-shell/packages/react/src/hooks/useApiStream.ts
```diff
@@ -50,36 +50,50 @@ export interface ApiStreamResult<TEvent> {
...
  * - Mount / `enabled` becomes `true` / `descriptor.key` or `mode` changes:
- *   `data`, `events`, and `error` are reset, then status goes to
- *   `'connecting'` and `connect()` is called.
- * - `enabled` becomes `false` (or starts `false`): status is forced to
- *   `'idle'` and `connect()` is never called; no other field is reset by
- *   this transition alone.
+ *   `data`, `events`, and `error` are reset and status is set to
+ *   `'connecting'` synchronously during render (not from an effect);
+ *   `connect()` is then called from an effect that runs after that render
+ *   commits.
```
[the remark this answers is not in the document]
itechmeat: Split the disabled case into its own transitions bullet: a key/mode change still resets data/events/error while disabled, but the status write is gated, so the published status stays `idle` and the skipped `connecting` lands on the later flip to enabled. 907f74d0

template-shell/src-app/app/layout/Menu.tsx
```diff
@@ -65,6 +103,19 @@ export const Menu: React.FC<MenuProps> = ({ children }) => {
...
+  // optional on the app - a project without the microfrontends plugin never
+  // gets one. Sharing the effect would leave `discoverySettled` false forever
+  // in exactly the case the hint is written for, making it unreachable by the
+  // reader who most needs it. Counting from mount costs the other end of the
+  // range: a registry that arrives after the window has closed can show the
+  // hint for the moment before its first poll lands.
+  useEffect(() => {
+    const graceTimer = setTimeout(() => setDiscoverySettled(true), EMPTY_STATE_GRACE_MS);
+    return () => clearTimeout(graceTimer);
+  }, []);
```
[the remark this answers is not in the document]
itechmeat: The empty state now needs the list to have been read from the current registry as well as the timer, so a registry arriving after the window closes no longer flashes the hint; with no registry at all the timer stays the whole gate, keeping the `frontx add` hint reachable. Regression test samples the DOM from inside the first read, which is the render that used to show it. 100d3e97

template-shell/packages/react/src/mfe/ThemeAwareReactLifecycle.tsx
```diff
@@ -123,19 +121,52 @@ export abstract class ThemeAwareReactLifecycle implements MfeEntryLifecycle<Chil
...
+        clone.textContent = el.textContent ?? '';
+        adoptedStyles.appendChild(clone);
+        return;
+      }
+      adoptedStyles.appendChild(el.cloneNode(true));
     });
+
+    // Staged in a fragment so the adopted pieces keep their host-document order
+    // relative to each other while moving as one block to the front.
+    shadowRoot.insertBefore(adoptedStyles, shadowRoot.firstChild);
```
GeraBart, line 238: Nice fix for the ordering bug, but I think it introduces a separate one: `unmount()` never removes what `adoptHostStylesIntoShadowRoot` (or `injectBaseResets`) appended, and `default-mount-manager.ts`'s `createShadowRoot` reuses `element.shadowRoot` when it already exists. So if the same container gets mounted again (e.g. navigating away from and back to a screen slot that keeps its `containerRef` div alive), this runs a second time against a non-empty shadow root with no id-based dedup.

Reproduced against real jsdom running this exact algorithm: calling it twice on one shadow root that already held an MFE `<style>` went from 3 children to 5 after the second call — each remount adds another full copy of the adopted block, and for `<link>` clones that means a repeat network fetch of the same stylesheet.

The cascade-ordering invariant itself still holds (duplicates stay ahead of the MFE style), so this doesn't undo the fix — but it's unbounded DOM/CSSOM growth per remount, and none of the 4 new test cases exercise a second invocation.

Suggest giving the adopted block a stable id and doing an upsert-or-replace on each `mount()`, mirroring the pattern `upsertStyleElement`/`injectStylesheet` already use in `packages/mfes/src/shadow/index.ts`, or clearing it explicitly in `unmount()`.

template-shell/src-app/app/layout/Menu.tsx
```diff
@@ -58,13 +100,41 @@ export const Menu: React.FC<MenuProps> = ({ children }) => {
...
+  // Both halves are required before the menu may claim there are no screens.
+  // A registry that appears only after the grace window has closed is unread at
+  // the render that introduces it, and the timer says nothing about it: settled
+  // alone would flash the hint for that one render, until the discovery effect's
+  // first read lands. Waiting for the read covers that, and costs nothing when
+  // the registry is there from mount - it is read in the mount effect, long
+  // before the window closes. With no registry at all there is nothing to wait
+  // for, so the timer is the whole gate; that is the case the hint is written
+  // for and it must stay reachable.
+  const discoveryComplete = discoverySettled && (!mfeRegistry || readRegistry === mfeRegistry);
```
GeraBart, line 200: This is a real improvement over `develop` (previously the false "No screens yet" flashed on every load; now only if registration runs long), but worth flagging that `EMPTY_STATE_GRACE_MS` is a fixed timeout, not an actual "discovery settled" signal. Extension registration is async/event-driven with no upper bound (Module Federation remote fetch under `effects.ts`), so on a cold cache or slow network it can still take longer than 2s — at which point `discoveryComplete` becomes `true` while `extensions.length === 0` is still true, and the exact false-empty-state message this PR targets can still render. The existing "still registering" test only delays by one 500ms poll tick, well inside the window, so nothing currently catches the >2s case.

Not necessarily blocking, but might be worth either wiring this to the mfe slice's actual per-extension registering/registered state (it's already tracked in `effects.ts`) instead of a fixed window, or documenting `EMPTY_STATE_GRACE_MS` explicitly as a UX floor rather than a correctness guarantee.

template-shell/packages/studio/src/CollapsedButton.tsx
```diff
@@ -46,6 +44,7 @@ export const CollapsedButton: React.FC<CollapsedButtonProps> = ({ toggleCollapse
       }}
     >
       <GlassmorphicButton
+        data-testid={STUDIO_EXPAND_TESTID}
```
GeraBart, line 47: Every other new testid in this PR (collapse, theme trigger, theme options, menu items) got a matching test assertion — this one didn't. Couldn't find any test anywhere asserting `STUDIO_EXPAND_TESTID` resolves (same gap on the `GlassmorphicButton` side that forwards it). Given the panel now defaults to collapsed, this button is the *only* entry point into the Studio overlay, so a future regression here would only surface in a browser run rather than in CI.

Worth a quick `CollapsedButton.test.tsx` asserting `screen.getByTestId('studio-expand')`, mirroring what `StudioPanel.test.tsx` already does for the collapse button.

template-shell/src-app/app/layout/Menu.tsx
```diff
@@ -58,13 +100,41 @@ export const Menu: React.FC<MenuProps> = ({ children }) => {
       const sorted = screenExts
         .sort((a, b) => (a.presentation.order ?? 999) - (b.presentation.order ?? 999));
       setExtensions(sorted);
+      // Marked here rather than beside the first `refresh()` call below so the
+      // flag cannot outrun the read it stands for. Every later poll re-sets the
+      // same reference, which React discards without a re-render.
+      setReadRegistry(mfeRegistry);
```
GeraBart: Minor/optional: `getExtensionsForDomain(...)` is cast (`as ScreenExtension[]`) with no runtime check, and `.sort` above dereferences `a.presentation.order` unconditionally. If a registered extension is ever missing `presentation`, `refresh()` throws inside this unguarded `setInterval` callback (predates this PR) — but now that also means `setReadRegistry` here never runs, so `discoveryComplete` never resolves and the menu can't even fall back to the empty-state hint anymore. Not saying this needs to hold up the PR since it's a narrow, pre-existing-cast edge case, but since you're touching this function anyway, a `try/catch` with `setReadRegistry(mfeRegistry)` in a `finally` would keep the fallback reachable.

GeraBart (commented):
Ran a deeper code-level pass on top of the existing CodeRabbit review. Two things worth a look before merge, plus a couple of smaller notes.

template-shell/packages/studio/src/CollapsedButton.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+    t: (key: string) => key,
+  }),
+}));
+
+vi.mock('./hooks/useDraggable', () => ({
+  useDraggable: () => ({
+    position: { x: 120, y: 80 },
+    isDragging: false,
+    handleMouseDown: handleDragMouseDown,
+  }),
```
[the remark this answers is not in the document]
itechmeat: Fixed - moved `handleDragMouseDown` into `vi.hoisted` so it exists before the `vi.mock('./hooks/useDraggable', ...)` factory runs, avoiding the TDZ reference. 23f8b968.

template-shell/packages/react/src/mfe/ThemeAwareReactLifecycle.tsx
itechmeat: Both the adopted block and the base resets now upsert on remount instead of stacking, mirroring the injectStylesheet pattern from packages/mfes; a test pins the second mount.

template-shell/src-app/app/layout/Menu.tsx
itechmeat: Kept the timer and documented the constant as a UX floor: bootstrap registers extensions directly without emitting the event, so registrationStates stays empty through the whole real boot and a gate on it would show the hint immediately.

template-shell/packages/studio/src/CollapsedButton.tsx
itechmeat: Added CollapsedButton.test.tsx asserting studio-expand resolves, mirroring StudioPanel.test.tsx.

template-shell/src-app/app/layout/Menu.tsx
itechmeat: Wrapped the read in try/catch with setReadRegistry in a finally so a malformed extension cannot strand the empty-state fallback; pinned by a test.

template-shell/src-app/app/layout/Menu.tsx
```diff
@@ -49,22 +104,80 @@ export const Menu: React.FC<MenuProps> = ({ children }) => {
...
-        .sort((a, b) => (a.presentation.order ?? 999) - (b.presentation.order ?? 999));
-      setExtensions(sorted);
+      try {
+        // The cast asserts `presentation` rather than checking it -
+        // `getExtensionsForDomain` returns the domain's extensions without that
+        // guarantee - so an extension registered without one throws on the
+        // `a.presentation.order` below.
+        const screenExts = mfeRegistry.getExtensionsForDomain(FRONTX_SCREEN_DOMAIN) as ScreenExtension[];
+        const sorted = screenExts
+          .sort((a, b) => (a.presentation.order ?? 999) - (b.presentation.order ?? 999));
```
gs-layer, line 144: The `try/catch` keeps the fallback reachable, but it aborts the **whole** read rather than the bad item. When the comparator throws, `setExtensions(sorted)` never runs, so `extensions` keeps its previous value (`[]` on a fresh load) while `setReadRegistry` fires in the `finally` and opens the gate. One malformed extension therefore hides every working screen *and* replaces them with a hint that is actively wrong.

Reachable in production: `src-app/app/mfe/bootstrap.ts:283-286` registers manifest-declared extensions behind only a domain-ownership filter - nothing validates `presentation`, which is what the comment above already concedes ("returns the domain's extensions without that guarantee").

Reproduced by running this file verbatim in an isolated harness: with `[Alpha, Beta, malformed]` registered, after the grace window the DOM is `No screens yet. If this project has no src-app/mfe_packages/ directory, run frontx add frontx-template-mfe...` and both valid screens are invisible (`Alpha visible? false`).

The justification - "the only thing still worth rendering once discovery is broken" - fits a broken registry, but discovery is not broken here: the other extensions were read fine. Filtering keeps them, and the empty-state fallback still works when *every* extension is malformed:

```ts
const sorted = screenExts
  .filter((ext) => ext?.presentation)
  .sort((a, b) => (a.presentation.order ?? 999) - (b.presentation.order ?? 999));
```

The existing regression test would still pass, since it pairs one malformed extension with one well-formed one; worth extending it to assert the well-formed screen stays visible.

template-shell/packages/react/src/mfe/ThemeAwareReactLifecycle.tsx
```diff
@@ -123,28 +139,94 @@ export abstract class ThemeAwareReactLifecycle implements MfeEntryLifecycle<Chil
...
+   * shadow root that `createShadowRoot` handed back intact, and appending a
+   * second copy would add one node per mount forever. Overwriting in place also
+   * keeps the position this node earned on the first mount - behind the adopted
+   * host block and behind the MFE's own CSS - which a remove-and-append would
+   * not, since the container gains nodes between mounts.
+   *
+   * Looked up by attribute-selector form rather than `getElementById`, which
+   * `ShadowRoot` has but `Element` does not, and this hook takes either.
    */
   private injectBaseResets(container: Element | ShadowRoot): void {
```
gs-layer, line 273: Reusing the resets node in place fixes the stacking, but it inverts the cascade order between the first mount and every remount - and the comment above justifying it describes a DOM sequence that does not happen.

The MFE's stylesheet is not a node the shadow root simply keeps across mounts. `MfeHandlerMF.wrapLifecycleWithStylesheets` removes it on unmount (`removeInjectedStylesheets`, MfeHandlerMF.ts:529-531) and re-injects it on the next mount via `upsertStyleElement`, which `appendChild`s a fresh `<link>` (MfeHandlerMF.ts:523-525, :578). So:

- first mount: `[adopted][isolation][MFE][resets]`
- remount: `[adopted][isolation][resets][MFE]`

Confirmed with a jsdom probe replicating that exact call sequence:

```
FIRST : [host style, #__frontx-shadow-isolation__, LINK#__frontx-mfe-runtime-style-0, #__frontx-base-resets__]
SECOND: [host style, #__frontx-shadow-isolation__, #__frontx-base-resets__, LINK#__frontx-mfe-runtime-style-0]
```

The comment at L217-219 says in-place overwriting "keeps the position this node earned on the first mount ... which a remove-and-append would not, since the container gains nodes between mounts". The container does not *gain* the MFE's nodes between mounts - it loses and re-gains them, appended at the tail. Remove-and-append is precisely what would have preserved the first-mount order here.

Observable: the resets declare `*, *::before, *::after { border-color: currentColor }` at specificity 0-0-0 and the MFE's own compiled Tailwind preflight declares the same property at the same specificity with the theme default. An MFE element with a bare `border` class renders `currentColor` on first mount and `#e5e7eb` after navigating away and back. Same flip for `* { margin: 0; padding: 0 }` and for any `:host` rule in the MFE's CSS against the resets' `:host { font-family / line-height / color / background-color }`.

The new remount tests cannot catch it: `ThemeAwareReactLifecycle.test.tsx:170-195` calls `unmount()` / `mount()` directly and never removes and re-adds the MFE stylesheet, so it models a remount shape production never produces. A case that removes the MFE `<style>` between the two mounts, mirroring `removeInjectedStylesheets`, would pin the real sequence.

template-shell/src-app/app/layout/Menu.tsx
```diff
@@ -49,22 +104,80 @@ export const Menu: React.FC<MenuProps> = ({ children }) => {
...
+      } catch (error) {
+        // Caught rather than left to escape, because `refresh` runs from two
+        // places with two different failures. From the effect body below an
+        // escaping throw takes the whole menu down. From the interval it is
+        // swallowed by the timer, but it also skips the `setReadRegistry` that
+        // used to sit at the end of this function, so `discoveryComplete` never
+        // resolves and the menu loses even its empty-state hint - the one thing
+        // still worth rendering when discovery is broken. The list keeps
+        // whatever the last successful poll read.
+        console.error('[Menu] Failed to read screen extensions from the MFE registry', error);
```
gs-layer: This sits inside the 500 ms poll with no latch, so a single malformed extension logs for the entire session. Measured in the harness: **21 `console.error` calls in 10 simulated seconds** (~2/s, unbounded). Each carries the `Error`, so devtools retains a stack per entry and the real console noise buries everything else - including the fact that the menu is showing a wrong hint.

A `useRef` latch keyed on the failure (or on `error.message`) would log once per distinct cause and keep the diagnostic value.

template-shell/packages/react/src/mfe/ThemeAwareReactLifecycle.tsx
```diff
@@ -123,28 +139,94 @@ export abstract class ThemeAwareReactLifecycle implements MfeEntryLifecycle<Chil
   /**
    * Copy all inline <style> and <link rel="stylesheet"> from the host document
    * into the shadow root so that Tailwind and component styles apply inside the MFE.
+   *
+   * The clones are inserted ahead of everything already in the shadow root, never
+   * appended, because adopted host styles are context rather than authority: where
+   * they and the MFE's own CSS declare the same property at equal specificity, the
```
gs-layer, line 164: This states the invariant as a property of the shadow root - where host CSS and the MFE's own CSS tie on specificity, the MFE wins - but it is enforced for the adopted block only. `mount()` still runs `injectBaseResets` *after* `MfeHandlerMF` injected the MFE stylesheet (L112-118), so on the first mount the resets sit last and win every 0-0-0 tie against the MFE.

Concretely, the resets' `border-color: currentColor` beats the MFE preflight's `border-color: <theme default>` - the same class of defect as the `background-color: transparent` case this method exists to fix, just arriving from the other shell-owned block.

The placement is pre-existing and I am not asking for it to change in this PR. But the PR is where the invariant gets written down, so either the resets move ahead of the MFE CSS too, or the docblock scopes the claim to the adopted block and says why the resets are deliberately exempt.

template-shell/packages/react/src/mfe/ThemeAwareReactLifecycle.tsx
```diff
@@ -123,28 +139,94 @@ export abstract class ThemeAwareReactLifecycle implements MfeEntryLifecycle<Chil
...
-    styleElements.forEach((el) => {
-      const clone = document.createElement('style');
-      clone.textContent = el.textContent ?? '';
-      shadowRoot.appendChild(clone);
-    });
-    const linkElements = document.head.querySelectorAll('link[rel="stylesheet"]');
-    linkElements.forEach((el) => {
-      const clone = el.cloneNode(true) as HTMLLinkElement;
-      shadowRoot.appendChild(clone);
+    shadowRoot.querySelectorAll(`[${ADOPTED_HOST_STYLE_ATTR}]`).forEach((el) => el.remove());
```
gs-layer: Two notes on this line.

**The `<link>` gap.** Every remount now removes the adopted `<link>` clones and creates fresh ones, so the shadow tree is momentarily without the host stylesheets while `mount()` proceeds synchronously to `createRoot(container).render(...)` (L121-129). Before this change the previous block was left in place and a duplicate appended, so host CSS was never absent. Net effect: a remount can now paint the MFE unstyled where it previously could not. Flagging as plausible rather than confirmed - whether a shadow-root `<link>` blocks rendering of its tree is engine-dependent and jsdom cannot show it. If the concern is real, keeping the old block until the new one is in place (insert first, then remove the previously marked nodes) closes it without giving up the dedup.

**The docblock at L171.** It attributes "re-resolve every cloned `<link>`" to the appending behaviour being replaced, but the new code re-clones every host `<link>` on every mount too, so it re-resolves them just the same. Only the unbounded node growth is actually fixed - worth saying so, since the sentence currently claims more than the change delivers.

template-shell/packages/studio/src/testIds.ts
```diff
@@ -0,0 +1,39 @@
...
+ * decision.
+ */
+
+/** The collapsed floating button - clicking it expands the panel. */
+export const STUDIO_EXPAND_TESTID = 'studio-expand';
+
+/** The expanded panel's header control - clicking it collapses the panel. */
+export const STUDIO_COLLAPSE_TESTID = 'studio-collapse';
+
+/** The theme switcher's trigger. Its text is the active theme's label. */
```
gs-layer: "Its text is the active theme's label" does not match what the trigger renders. `ThemeSelector.tsx:54` renders `formatThemeName(currentTheme || '')` - `currentTheme` is the registry **id** - while the options at `ThemeSelector.tsx:64` render `formatThemeName(theme.name || theme.id)`, the **name**.

The two sides of this verification contract therefore read different fields, and the new test cannot surface it: its fixture is `{ id: 'dracula-large', name: 'Dracula Large' }`, the one shape where both formattings coincide.

Failure scenario for the run this API is written for: a registry ships `{ id: 'dracula-large', name: 'Dracula (Large)' }`; the run clicks the option labelled "Dracula (Large)", reads "Dracula Large" off the trigger, and concludes the theme did not apply.

Either the doc should say the trigger carries the formatted **id**, or the trigger should resolve `currentTheme` through `themes` and render the same `name || id` the options do. The second is the one that makes the id-keyed option testids and the trigger text agree.

template-shell/packages/studio/src/sections/ThemeSelector.test.tsx
```diff
@@ -0,0 +1,42 @@
...
+  { id: 'dracula-large', name: 'Dracula Large' },
+];
+
+vi.mock('@gears-frontx/react', () => ({
+  useTheme: () => ({ currentTheme: 'default', themes, setTheme }),
+  useTranslation: () => ({ t: (key: string) => key }),
+}));
+
+vi.mock('../StudioProvider', () => ({
+  useStudioContext: () => ({ portalContainer: null }),
```
gs-layer: Mocking `portalContainer: null` sends `DropdownMenuContent` into `document.body`, so the rendered menu has no `.studio-portal-container` ancestor (confirmed in a DOM dump - it is a direct `<body>` child). That means the scoped animation rules this PR adds to `studioStyles.ts` are never exercised by any test.

The PR notes the cascade itself is untestable in jsdom, which is fair - but *selector reachability* is not. A typo in either scoped selector's escaping would leave the stuck-open-menu bug fully intact (Radix waits for an `animationend` that never fires, and the closed menu stays visible and clickable over the app) while the whole studio suite stays green.

Rendering this component with a real `.studio-portal-container` element as `portalContainer` and asserting `container.querySelector('.studio-portal-container .data-\\[state\\=closed\\]\\:animate-out')` resolves would close the gap without needing a cascade.

template-shell/src-app/app/layout/Menu.test.tsx
```diff
@@ -0,0 +1,193 @@
...
+    // Rendering at all is the first half of the claim: an escaping throw here
+    // comes out of the effect body, not the interval, and takes the tree down.
+    render(<Menu />);
+
+    await act(async () => {
+      vi.advanceTimersByTime(EMPTY_STATE_GRACE_MS);
+    });
+
+    expect(emptyState()).not.toBeNull();
+    expect(consoleError).toHaveBeenCalled();
```
gs-layer: This accepts any console error produced by the render, so it can outlive the production log it is meant to pin.

It is sound today - I removed the `console.error` line while keeping the `catch` and this test failed, so no stray React warning currently satisfies it. But any future warning from the same render (a key warning, an `act()` warning, an unknown-prop warning once `SidebarMenuButton` changes) would silently keep it green with the log gone.

```ts
expect(consoleError).toHaveBeenCalledWith(
  expect.stringContaining('[Menu] Failed to read screen extensions'),
  expect.any(Error)
);
```

template-mfe/src-app/mfe_packages/_blank-mfe/src/screens/home/HomeScreen.tsx
```diff
@@ -178,26 +194,36 @@ export const HomeScreen: React.FC<HomeScreenProps> = ({ bridge }) => {
...
+                  {language}
+                </dd>
               </div>
             </dl>
           </CardContent>
         </Card>
 
         <Card>
-          <CardContent className="p-6">{statusCardBody}</CardContent>
+          <CardContent className="p-6" data-testid="screen-status">
```
gs-layer, line 224: `screen-status` is the only new testid in this file with no assertion anywhere, although the docblock above states the contract as keeping a `screen-<control>` testid "on each interactive control **and on the status region**".

Every other id added here is pinned by the suite, so this one is the single place where a later edit can drop the attribute and leave CI green - surfacing only as a failing browser run, and inherited by every screen scaffolded from this file. One `expect(screen.getByTestId('screen-status')).toBeTruthy()` alongside the existing payload assertion covers it.

template-shell/packages/studio/src/testIds.ts
```diff
@@ -0,0 +1,39 @@
+/**
```
gs-layer, line 1: This file documents itself as the package's verification API, but it is not re-exported: `src/index.ts` exports only `StudioOverlay`, `StudioProvider`, `useStudioContext` and three types.

So a consumer of `@gears-frontx/studio` writing an unattended run cannot `import { STUDIO_EXPAND_TESTID }` and has to hardcode `'studio-expand'` - which is the drift this docblock says the constants exist to prevent. Inside the package the constants do their job; outside it they are unreachable.

Adding `export * from './testIds';` to `src/index.ts` makes the published contract importable and costs nothing (the module is constants only, no runtime).

gs-layer:
One more finding that has no line to anchor to, since it lives in a file this PR does not touch — leaving it here rather than dropping it.

**`architecture/DESIGN.md:109` — the root ADR inventory no longer maps 1:1 to `architecture/ADR/`.**

After the move, the root DESIGN still lists `cpt-frontx-adr-template-ownership-boundary-declaration` under "CLI (projects orchestration)", and it is now the only ID in that list without a corresponding file in `architecture/ADR/`. Root numbering reads 0001–0030, 0032, 0033.

Not a validation error — I confirmed `cfs validate` resolves the cross-system reference cleanly (70 artifacts, 0 errors, 0 warnings, 171/171 coverage, byte-identical to base), and the CLI's own DESIGN already carries a symmetric list of root-owned ADR IDs at `packages/cli/architecture/DESIGN.md:70-81`, so the pattern is established in both directions.

The risk is a reader, not a script: someone treating that subsection as the index of `architecture/ADR/` finds the gap between 0030 and 0032, sees no file, and concludes the record was deleted or superseded — the same misreading `architecture/explorations/2026-07-16-multi-template-composition.md:207` already invites for a different, historical "ADR-0031". One clause on the bullet (`— owned by the CLI member tree, packages/cli/architecture/ADR/`) settles it. Defensible to leave as-is if that inventory is meant to be ecosystem-wide rather than tree-scoped.

---

For context on the rest of the review: I checked the `.cf-studio/config/artifacts.toml` change by parsing both refs with `tomllib` — the single-bracket `[systems.children.autodetect.artifacts.ADR]` attaches to `systems[0].children[4]`, which is the CLI member, and `autodetect` is per-child, so the pattern reaches exactly one member and the other five are untouched. The negative control confirms the block is load-bearing: deleting it turns the validate run into `✗ 25 error(s)`, every one a `ref-no-definition` for `cpt-frontx-adr-template-ownership-boundary-declaration` — matching the count in your TOML comment exactly. Repo-wide grep at the head finds zero surviving `../../../architecture/ADR/0031-` references, and the two relative links inside the moved file (`../PRD.md`, `../DESIGN.md`) resolve to the CLI's own PRD/DESIGN at the new depth, which is where the `cpt-frontx-fr-cli-*` requirements it cites are actually defined — the move fixed those two links rather than breaking them.

template-shell/src-app/app/layout/Menu.tsx
itechmeat: Fixed: malformed extensions are filtered out before the sort, so valid screens survive a bad sibling and the catch stays for a registry broken outright; both the malformed-sibling and the all-malformed cases are pinned by tests.

template-shell/packages/react/src/mfe/ThemeAwareReactLifecycle.tsx
itechmeat: Fixed: the resets are placed in the shell-owned head before any MFE stylesheet on every mount, so first mount and remount both yield adopted-isolation-resets-MFE; pinned by a remount test that removes and re-appends the MFE stylesheet the way MfeHandlerMF does, and the comment describing the wrong sequence is rewritten.

template-shell/src-app/app/layout/Menu.tsx
itechmeat: Fixed: a ref latch keyed on the error message logs once per distinct cause, deliberately not cleared on success so a flapping registry cannot re-log at poll rate; pinned by a call-count test.

template-shell/packages/react/src/mfe/ThemeAwareReactLifecycle.tsx
itechmeat: Fixed together with the remount thread: the resets moved ahead of the MFE stylesheet on every mount, so the documented invariant now holds for both shell-owned blocks instead of being scoped down.

template-shell/packages/react/src/mfe/ThemeAwareReactLifecycle.tsx
itechmeat: Fixed both parts: the fresh adopted block is inserted before the superseded nodes are removed, so host CSS is never absent mid-remount (pinned by a mutation-observer order test), and the docblock now claims only bounded growth, not avoided re-resolution.

template-shell/packages/studio/src/testIds.ts
itechmeat: Fixed: the trigger resolves the current theme through the registry list and renders the same name-or-id the options carry, falling back to the formatted id; pinned with a fixture where id and name diverge, which also keeps both sides of the contract the verify-walk in #583 reads consistent.

template-shell/packages/studio/src/sections/ThemeSelector.test.tsx
itechmeat: Fixed: a test renders with a real studio-portal-container element and asserts the animation selectors, read out of STUDIO_CSS itself rather than retyped, resolve to the open menu - a typo on either side reddens it.

template-shell/src-app/app/layout/Menu.test.tsx
itechmeat: Fixed: the assertion now pins the exact [Menu] message prefix, an Error argument and a single call.

template-mfe/src-app/mfe_packages/_blank-mfe/src/screens/home/HomeScreen.tsx
itechmeat: Fixed: the HomeScreen test asserts screen-status resolves, beside the existing payload assertion.

template-shell/packages/studio/src/testIds.ts
itechmeat: Fixed: the constants are re-exported from the studio package index and appear in the built type declarations, so an unattended run imports them instead of hardcoding strings.

itechmeat:
@gs-layer fixed: the DESIGN bullet now names the owner (owned by the CLI member tree, packages/cli/architecture/ADR/), and the artifacts validation still reports 70 artifacts, 0 errors, 171/171 coverage. Thanks for the negative-control check on the TOML block.

GeraBart (approved):

GeraBart (approved):

PR #584, opened 2026-08-18
feat: move both template-mfe packages from vendored shadcn copies onto @gears-frontx/ui-kit

### LINKED ISSUE #551: Build scaffolded screens from @gears-frontx/ui-kit instead of vendored copies (_blank-mfe first)

itechmeat:
Closes #551.

Both MFE packages of template-mfe move off their vendored shadcn-style component copies onto the ecosystem's own `@gears-frontx/ui-kit`: `_blank-mfe` - the scaffold every new screen is copied from - and the demo-mfe showcase. The bulk of the diff is deletions: the vendored component tree, its utility layer and its toolchain dependencies leave the template, and what remains imports the kit.

## What changes

- **`_blank-mfe` builds its screen from the kit.** The vendored `button`/`card`/`skeleton` copies, `lib/utils.ts` and the tailwind/clsx/cva/radix-slot toolchain are gone from the scaffold; `HomeScreen` renders kit `Card` and `Skeleton` with a CSS module for layout. Every project seeded from the template inherits this shape.
- **demo-mfe's showcase moves to the kit and trims to what the kit ships.** 56 hand-rolled demo elements become 19, exactly one per kit barrel export; the dropped 37 demonstrated components the kit does not ship (charts, calendars, accordions, data tables and the like), so the showcase now shows what a seeded project can actually build. This is a product call folded in from the demo branch - confirmation from Андрей welcome. The vendored tree under `components/`, `hooks/` and `lib/` is deleted, 641 showcase i18n keys become 52 across all 36 locales.
- **Theming inside a shadow root is solved by two small shared modules, not by a provider.** `anchorKitThemeOnShadowHost` rewrites the kit's `:root` token declarations to `:host` (comments stripped first), because `:root` matches no node inside a shadow tree and the tokens would otherwise never resolve. `kitThemeScope` maps the shell's theme identifier onto the kit's two `data-theme` scopes with `light` as the fallback for unknown identifiers - an element carrying neither scope would inherit the developer machine's `prefers-color-scheme`, which is how a screen ends up dark inside a light shell. The transform assumes the kit's theme stylesheet keeps `:root` out of quoted values and feature queries; the exact version pin and the tests hold that assumption, and a hostile-input test pins the current behaviour so any future change to the transform is a conscious one.
- **The kit is pinned exactly `0.3.0-alpha.3`**, resolved from the registry (the issue text says alpha.1; develop has since published alpha.3 with the RSC fix, and the raise did its job - it surfaced the alpha.3 Badge API change, `shape="dot"` split into `shape` plus a boolean `dot`, fixed here). The template harness deliberately keeps the kit out of `overrides`: the registry version is the one a seeded project gets, so the harness should test exactly that.
- **Screen logic stays at develop's newer state under the kit markup.** The demo branch this ports from predates develop's bridge-swap re-read and the language-keyed direction effect, so a naive port would have regressed all five screens; here every screen keeps `readBridgeProperty`, the `prevBridge` re-read and the `dir` effect, and the bridge-swap test the demo branch had deleted is restored and extended. `ProfileDetailsCard` is the one place the demo branch's version wins wholesale: a nullable edit buffer replaces the `prevSyncKey` re-sync pattern instead of working around it.
- **Bridge-replacement is now pinned in demo-mfe too**, not only in the scaffold: the demo screens get bridge-rerender coverage asserting immediate re-read from the new bridge and exactly-once unsubscription from the old one.
- **Profile edit UI translates in every locale.** 35 of 36 profile locales were missing the six keys the edit flow calls (a pre-existing gap; this PR rewrites that screen, so it closes it): the keys are added everywhere, and the showcase trim's five orphaned keys were restored across all 36 locales the same way.

## Verification

- root: `type-check:all`, `lint`, `arch:check`, `build:packages`, `test:unit` all clean; `validate:templates` passes for both templates; `policy:template-pin-drift` and `policy:template-lockfile-selflink` clean; `npm ci` clean at root and inside template-mfe
- studio: `cfs validate` 70 artifacts / 0 errors, coverage 171/171; `cfs validate-kits`; `cfs spec-coverage`; `cfs check-language` all clean
- composed shell + MFE overlay (the Template Validate path, against registry tarballs): install, build, type-check, arch:deps, lint, test:unit all green
- live browser pass over the composed app: all five screens, theme Default→Dark, language English→العربية - zero console errors and zero warnings
- the theme mechanism is mutation-checked: breaking the `:root`→`:host` rewrite or the scope mapping reddens exactly the tests that pin them; same for the bridge-swap re-read

## The two risk questions from #551

**CSS injection timing in a federated build.** The kit's component CSS never passes through `document.head`: Vite emits it into the exposed module's stylesheet, the MFE manifest lists it as a sync asset, and the host's `wrapLifecycleWithStylesheets` awaits `injectRemoteStylesheets(container, ...)` - a `<link>` into the shadow root itself - strictly before `lifecycle.mount()` and therefore before `createRoot().render()`. The design tokens take a second path: `theme.css?inline` is injected synchronously in `initializeStyles()`, which runs before `createRoot` in the lifecycle. Both halves are ordered ahead of first paint. One pre-existing caveat, unchanged by this PR: the `<link>` append does not await the sheet's fetch completion - ordering is guaranteed, load completion is not, and that property applies identically to the vendored components this replaces.

**Theme-token source of truth.** The shell decides which palette, the kit owns the values, and the two token grammars never share a root. The shell's theme identifier arrives over the bridge, `kitThemeScope` maps it to the kit's `data-theme` scope on the screen root, and the kit's tokens live on `:host` inside the shadow tree. Loading the kit's `theme.css` at document level was rejected explicitly: the shell declares the same token names as HSL triplets for Tailwind while the kit declares complete colours, so kit values in `document.head` would invalidate every colour utility in the shell and every other MFE. For the same reason `CurrentThemeScreen` deliberately stays off the kit - it exists to display the shell's palette, and anchoring kit tokens on its host would blank every swatch.

## Coordination

- **#582 (open):** carries testid additions to the same `HomeScreen`. This branch keeps every testid and assertion #582 introduces (verified against its head), so the post-merge rebase is a near-noop; expect textual conflicts only on the two loading-state tests, where the kit's Skeleton emits hashed CSS-module classes and the `.animate-pulse` queries became `role="status"`/`aria-busy` assertions with the same intent.
- **#567 (open):** extracts shared hooks from the same demo-mfe screens; whichever merges second reconciles both sides.
- The showcase trim 56 → 19 is a product decision folded in from the demo branch - flagged for Андрей's confirmation.
- Nothing routing-related is ported, and nothing outside `template-mfe/` is touched.
- Residue `@cpt-` markers were stripped from every touched template-territory file per ADR-0033; none added.
- Follow-up worth its own change, deliberately not here: the lockfile self-link guard only recognises `file:.` specs, so it cannot notice a template losing its `file:../*` override links - this PR hit exactly that during lockfile regeneration and fixed it by hand-verified reinstall.


template-mfe/src-app/mfe_packages/demo-mfe/src/screens/uikit/UIKitElementsScreen.tsx
```diff
@@ -167,161 +181,108 @@ export const UIKitElementsScreen: React.FC<UIKitElementsScreenProps> = ({ bridge
...
+        </Card>
+      </main>
+
+      {/*
+        Toast container, mounted once for the screen. The shell may run a Toaster
+        of its own on the kit's shared manager; this one stays inside the shadow
+        root through `container`, and a duplicate viewport on that same manager
+        would render every toast twice.
+      */}
+      <Toaster container={portalContainerRef} />
```
gs-layer: **[high] `Toaster` portals into `document.body`, not this shadow root**

`<Toaster container={portalContainerRef} />` is rendered *before* the `<div ref={portalContainerRef}>` it points at (line 284). Base UI resolves the container in `useFloatingPortalNode`'s `useIsoLayoutEffect`, whose deps `[containerProp, parentPortalNode, uniqueId]` are all stable — and on the first commit React runs an earlier sibling's layout effect *before* attaching a later sibling's ref. So `container.current` is `null`, the resolver falls through to `?? document.body`, and it never re-resolves.

Concrete failure: user clicks "Show a toast". The viewport mounts into `<body>`, outside the shadow root — where neither the adopted component stylesheets nor the tokens on `:host` reach it, and where the shell declares the same token names as HSL triplets, so `var(--background)` resolves to an invalid value and drops. The toast renders unstyled and outside the screen's `data-theme` scope. This is exactly the failure the comment at line 90-96 says the ref exists to prevent.

Verified independently with a minimal React 19 render (`@testing-library/react` + jsdom): with this sibling order the earlier sibling's `useLayoutEffect` observes the later sibling's ref as `null`.

The other four portals (Dialog / Select / DropdownMenu / Tooltip) are unaffected — their popups mount on `open`, in a later commit, when the ref is already set.

Fix: move `<div ref={portalContainerRef} className={styles.portalContainer} />` above `<Toaster>`.

template-mfe/src-app/mfe_packages/_blank-mfe/src/shared/anchorKitThemeOnShadowHost.ts
```diff
@@ -0,0 +1,42 @@
...
+ * a quoted value nor a feature query names it anywhere. Two things hold that:
+ * the exact version pin in this package's `package.json`, and the tests beside
+ * this file, which assert the real imported stylesheet comes back with no
+ * `:root` left in it. A kit upgrade that introduces either context needs a real
+ * CSS parser here, not a wider regex.
+ *
+ * @param css - Stylesheet source, minified or not
+ */
+export function anchorKitThemeOnShadowHost(css: string): string {
+  return css.replace(/\/\*[\s\S]*?\*\//g, '').replace(/:root\b/g, ':host');
```
gs-layer: **[medium] The `:root:not(...)` rewrite produces a selector that never matches**

The blanket `:root` → `:host` replacement turns the kit's `:root:not([data-theme='light'])` (theme.css:406) into `:host:not([data-theme='light'])`. A shadow host is *featureless*: it is matched only by `:host`, `:host()` and `:host-context()`, so this compound selector matches nothing and the whole `@media (prefers-color-scheme: dark)` block (theme.css:405-455, dark tokens + `color-scheme: dark`) is dead inside the shadow root.

Measured in headless Chrome — shadow root with `:host{--tok:PLAIN}` plus each candidate:

| selector | computed `--tok` |
|---|---|
| `:host:not([data-theme='light'])` | `PLAIN` (no match) |
| `:host(:not([data-theme='light']))` | `FUNCTIONAL_MATCHED` |

Today this is masked because every kit screen sets `data-theme` explicitly in every render branch. It is a trap in the scaffold, though: a screen copied from `_blank-mfe` that omits `data-theme` — and copying is what this package is for — silently gets no dark tokens on a dark-mode machine, which is the very failure `kitThemeScope`'s docblock describes.

The correct output is the functional form `:host(:not([data-theme='light']))`.

Same code in `demo-mfe/src/shared/anchorKitThemeOnShadowHost.ts:49`. See also the test at `anchorKitThemeOnShadowHost.test.ts:49`, which pins the dead selector as desired behaviour and would redden on the fix.

template-mfe/src-app/mfe_packages/demo-mfe/src/screens/uikit/UIKitElementsScreen.tsx
```diff
@@ -167,161 +181,108 @@ export const UIKitElementsScreen: React.FC<UIKitElementsScreenProps> = ({ bridge
     return () => observer.disconnect();
   }, []);
 
+  const kitThemeScope = kitThemeScopeFor(theme);
```
gs-layer, line 213: **[medium] Scroll-spy observes zero nodes, so the active-item highlight is dead code**

(About the `IntersectionObserver` effect just above, closing at line 182.)

`useScreenTranslations` starts with `loading = true` (`useScreenTranslations.ts:48`), so the first commit takes the early return at line 186 and renders only the skeleton — no `[id^="element-"]` node exists yet. The effect is declared above that return, so it *does* run: `querySelectorAll` returns an empty NodeList and `observer.observe` is called zero times. With `[]` deps it never re-runs after translations resolve, and the seven categories are `React.lazy`, so their nodes appear later still.

Result: `setActiveElement` is never called, `activeElement` stays `undefined` for the lifetime of the screen, scrolling the showcase never highlights a sidebar item, and the two rules this PR adds — `.elementButtonActive` and `.elementButtonActive:hover` (`UIKitElements.module.css:255-263`) — can never apply. The `activeElement` prop threaded into `CategoryMenu` is likewise inert.

The `[]` deps predate this PR, but the PR ships the new styling and the prop plumbing on top of them. A `[translationsLoading]` dep (plus re-observing when a lazy chunk resolves) would make both live.

template-mfe/src-app/mfe_packages/demo-mfe/src/shared/kitThemeScope.ts
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * Two scopes is the whole resolution this bridge can offer. A host theme is a
+ * full palette, not a light/dark bit — `dracula` maps to the kit's dark scope
+ * and then renders in the kit's greys rather than Dracula's purples. Closing
+ * that gap means unifying the two token grammars, a decision above this
+ * template.
+ *
+ * @param hostTheme - Value of the host's shared theme property
+ */
+export function kitThemeScopeFor(hostTheme: string): 'light' | 'dark' {
```
gs-layer, line 54: **[medium] This copy of the mapping is covered by no test at all**

The PR description says: *"the theme mechanism is mutation-checked: breaking the `:root`→`:host` rewrite or the scope mapping reddens exactly the tests that pin them"*. That holds for `anchorKitThemeOnShadowHost` (its test file is duplicated into both packages), but not for `kitThemeScope`: only the `_blank-mfe` copy is exercised (`HomeScreen.test.tsx:172-202`). Nothing in `demo-mfe` asserts `kitThemeScopeFor` or the `data-theme` it produces.

Mutation check: replace `DARK_HOST_THEMES` here with an empty `Set`. With the shell on `dark`, `dracula` or `dracula-large`, all three kit-backed demo screens (helloworld, profile, uikit) render a light kit surface inside a dark shell — and the whole `demo-mfe` suite stays green.

There is also nothing guarding the two copies against drifting apart; they are kept in sync by convention only.

template-mfe/src-app/mfe_packages/demo-mfe/src/screens/profile/components/ProfileDetailsCard.tsx
```diff
@@ -39,29 +43,27 @@ export const ProfileDetailsCard: React.FC<ProfileDetailsCardProps> = ({
...
+  const isEditing = editingValues !== null;
 
   const initialValues = useMemo(() => getFormValues(user), [user]);
+  const formValues = editingValues ?? initialValues;
   const normalizedValues = useMemo(
     () => ({
-      firstName: trim(formValues.firstName),
-      lastName: trim(formValues.lastName),
-      department: trim(formValues.department),
+      firstName: formValues.firstName.trim(),
```
gs-layer, line 73: **[medium] Dropping lodash `trim` loses null-safety and turns a bad field into a screen crash**

Base (`e354ceb5`, lines 62-64) used lodash's null-safe `trim(formValues.firstName)`; this is now native `.trim()`.

`ApiUser.firstName: string` is a compiler assertion about an untyped HTTP body, not a guarantee. If `GET /me` returns `{"user":{"firstName":null,...}}`:

- before: `trim(null) === ''`, the card renders;
- now: `TypeError: Cannot read properties of null (reading 'trim')` is thrown inside the `normalizedValues` `useMemo` — i.e. during the render phase — so the whole `ProfileScreen` unmounts to the error boundary rather than one field degrading.

Same for `lastName` and `department`. If the intent was to shed the lodash dependency, `(x ?? '').trim()` keeps the old behaviour.

template-mfe/src-app/mfe_packages/demo-mfe/src/screens/uikit/UIKitElementsScreen.tsx
```diff
@@ -103,9 +119,7 @@ export const UIKitElementsScreen: React.FC<UIKitElementsScreenProps> = ({ bridge
   // Load translations
   const { t, loading: translationsLoading } = useScreenTranslations(languageModules, bridge);
 
-  // Subscribe to theme and language. Text direction is derived from
-  // `language` by the effect below only — useScreenTranslations consumes
-  // `language` for translation loading and has no role in direction.
+  // Handle theme subscription; language + direction come from useScreenTranslations
```
gs-layer: **[low] This comment reverted to the stale branch's version and is now factually wrong**

`develop` had already replaced it with the opposite statement:

```
// Subscribe to theme and language. Text direction is derived from
// `language` by the effect below only — useScreenTranslations consumes
// `language` for translation loading and has no role in direction.
```

The port reinstated the demo branch's text. `useScreenTranslations` has no role in direction at all — the `dir` effect at line 145-152 is the only writer.

This is the one place across the five screens where the port actually lost a `develop` fix; the bridge logic itself (`readBridgeProperty`, the `prevBridge` re-read, the exactly-once unsubscription, the `dir` effect) is intact everywhere, as the PR description claims.

Risk: someone reworking `useScreenTranslations` trusts the comment, leaves the effect below untouched, and ships an RTL regression.

template-mfe/src-app/mfe_packages/demo-mfe/src/screens/profile/ProfileScreen.test.tsx
```diff
@@ -111,17 +111,21 @@ async function setupProfileScreen(options?: {
       [FRONTX_SHARED_PROPERTY_LANGUAGE]: 'sv',
     },
   });
-  const { container, unmount } = render(<ProfileScreen bridge={bridgeFixture.bridge} />);
+  const { container, rerender, unmount } = render(
+    <ProfileScreen bridge={bridgeFixture.bridge} />
+  );
```
gs-layer, line 120: **[low] The `dir` assertions never cover the effect's initial pass**

(About lines 116-120 and the direction assertions that follow.)

`mockShadowHost` is installed *after* `render()`. During the first render `containerRef.current.getRootNode()` returns `document`, `'host' in document` is `false`, and the effect writes nothing — so the initial LTR assignment is asserted by nothing. Only the later switch to `'ar'` is checked.

Mutation check: change the effect's deps to `[]`, or invert the fallback to `'rtl'` — the profile, helloworld, theme and uikit suites all stay green.

The trailing `expect(rootElement?.getAttribute('dir')).toBeNull()` also passes trivially: the screen writes `dir` to the shadow *host*, never to that element.

`_blank-mfe/src/screens/home/HomeScreen.test.tsx:223-235` does it correctly — raises a real shadow root *before* render and asserts `host.dir === 'ltr'`. Worth copying into the four demo suites.

template-mfe/src-app/mfe_packages/demo-mfe/src/screens/uikit/UIKitElementsScreen.tsx
gs-layer: **[low] `Toaster` mounted without `label` / `closeLabel` on a 36-locale screen**

The kit documents both props as the strings a non-English consumer is expected to supply:

> Accessible name for the toast region landmark. Base UI puts `role="region" aria-label="Notifications"` on the viewport; without this prop a screen-reader user hears that English landmark name no matter what language the app is in.

With the bridge language on `ar`, `de` or `ja`, a screen reader announces the toast region as "Notifications" and the dismiss control as "Close toast", in English. The screen already translates 52 keys; these two have no keys yet.

template-mfe/src-app/mfe_packages/demo-mfe/src/screens/uikit/i18n/ru.json
```diff
@@ -1,643 +1,54 @@
...
-  "element.pagination.demo.current_page": "Current page:",
-  "element.link.title": "Link",
-  "element.link.description": "Hyperlink component",
-  "element.link.demo.standard": "Standard hyperlink",
-  "element.link.demo.hover": "Link with hover underline",
-  "element.link.demo.muted": "Muted link",
-  "element.menu.title": "Menu",
-  "element.menu.description": "Context or dropdown menu",
-  "element.menu.demo.note": "See the Navigation Menu example above (Nav element) for a Menubar demonstration. UIKit also exports DropdownMenu and ContextMenu - see the Actions category.",
+  "element.field.title": "Field",
```
gs-layer: **[low] The four newly added showcase keys are English in all 35 non-English locales**

`element.field.title` / `element.field.description` / `element.label.title` / `element.label.description` were added to all 36 locale files with the English string byte-for-byte.

Checked across the namespace: in each of the 35 non-English locales, 50 of the 52 values are identical to `en.json` — only `title` and `description` are actually translated. That 50 is inherited from base and is not this PR's doing, but the four new keys were written English-only, so the gap grew rather than held.

User-visible effect: switching the shell to `ru`/`ja`/`ar` on the showcase changes only the page heading and blurb; the category menu and every element title stay English.

(For contrast, the six profile keys this PR adds *are* genuinely translated in all 35 locales — none of them equals the English string.)

template-mfe/src-app/mfe_packages/demo-mfe/src/screens/profile/components/ProfileDetailsCard.tsx
```diff
@@ -75,15 +77,14 @@ export const ProfileDetailsCard: React.FC<ProfileDetailsCardProps> = ({
   const handleFieldChange =
     (field: keyof ProfileFormValues) =>
       (event: React.ChangeEvent<HTMLInputElement>) => {
-        setFormValues((current) => ({
-          ...current,
+        setEditingValues((current) => ({
+          ...(current ?? initialValues),
```
gs-layer: **[low] `?? initialValues` is unreachable**

`handleFieldChange` is bound only to the `<Input>`s inside the `<form>`, which renders only when `isEditing` is true — and `isEditing` is exactly `editingValues !== null`. So `current` is never `null` in this updater.

No runtime consequence; the cost is that the fallback reads as "a field can be edited outside edit mode", which the component's own invariant forbids, and a later refactor may preserve the dead path as if it carried meaning.

template-mfe/src-app/mfe_packages/_blank-mfe/README.md
```diff
@@ -247,7 +317,20 @@ If TypeScript cannot resolve `@gears-frontx/*` imports:
...
+is reading the shell's tokens rather than the kit's. A Tailwind colour utility
+in this package is the usual cause — see [Styling](#styling).
+
+If a `Card`'s parts are crammed together vertically — a footer button flush
+against the last field — while its left and right padding looks right, an
+element wraps the card's slots and swallows the card's gap. Make the slots
+direct children of `Card` again — see [Styling](#styling).
+
+If a component's own styles are missing, the CSS is not reaching the shadow
+root at all: confirm the component is imported statically (a lazily imported
```
gs-layer: **[low] This troubleshooting step contradicts the handler**

`MfeHandlerMF` injects **both** the sync and the async CSS assets at mount:

```ts
// packages/mfes/src/handler/MfeHandlerMF.ts:416
const stylesheetPaths = [...exposeAssets.css.sync, ...exposeAssets.css.async];
```

so a lazily imported kit component's CSS *is* in this MFE's stylesheet. A developer whose component renders unstyled follows this line, converts a `React.lazy` import to a static one, sees no change, and concludes the styling pipeline is healthy — while the real cause (a missing `exposeAssets.css` entry, or tokens not anchored on `:host`) goes uninvestigated.

template-mfe/src-app/mfe_packages/demo-mfe/vitest.config.ts
```diff
@@ -1,9 +1,35 @@
...
+  defineMfeProject(__dirname),
+  defineConfig({
+    test: {
+      /*
+       * Vitest resolves a stylesheet import to an empty string unless CSS is
+       * processed. KitThemedLifecycle re-anchors the ui-kit theme tokens onto
+       * the shadow host by rewriting the stylesheet it imports, so without this
+       * the one test covering that rewrite would assert against nothing.
+       */
+      css: true,
```
gs-layer, line 22: **[low] The two comments in this file contradict each other about CSS**

With `css: true` set here, the `server.deps.inline` comment below ("Vite … resolves a stylesheet import in a test environment to nothing") is no longer true — the two adjacent comments state opposite things about the same import.

Why it matters: a maintainer reading the `inline` block concludes CSS is still stubbed and drops `css: true` as redundant. `anchorKitThemeOnShadowHost.test.ts:77` then loses its footing — `expect(kitThemeCss).toContain(':root')` is precisely the guard that keeps the rewrite test from passing vacuously on an empty string. `css: true` is load-bearing; the comment should say so.

Identical block in `_blank-mfe/vitest.config.ts:19`.

template-mfe/src-app/mfe_packages/_blank-mfe/src/shared/anchorKitThemeOnShadowHost.ts
```diff
@@ -0,0 +1,42 @@
...
+ * What this buys instead is a constraint, and it is worth stating plainly: the
+ * rewrite is textual, not a parse, so it cannot tell a selector from the two
+ * other places `:root` may legally appear. Inside a quoted value (a string, or
+ * an attribute selector's value) it would rewrite text that is data, and inside
+ * a feature query such as `@supports selector(:root)` it would rewrite the
+ * condition being tested rather than a selector. The comment stripper has the
+ * same blind spot in reverse: a quoted value containing a comment delimiter
+ * would be eaten as prose.
+ *
+ * The pinned kit version satisfies the assumption. Every `:root` in its
```
gs-layer, line 50: **[low] The docstring claims a guard the tests cannot actually provide**

This paragraph (and the PR description) say the "no `:root` inside a quoted value or a feature query" assumption is held by two things: the exact version pin, and *"the tests beside this file, which assert the real imported stylesheet comes back with no `:root` left in it"*.

That assertion cannot detect the failure it is offered against. If a kit upgrade adds `.x::after{content:":root"}` or `@supports selector(:root){…}`, the transform silently rewrites data or a feature-query condition — and `expect(anchored).not.toContain(':root')` plus `expect(anchored).toContain(':host')` both stay green. The "conscious act" the docstring promises never happens.

A test that would hold it: assert the *number* of `:root` occurrences in the pinned stylesheet, or assert that every occurrence sits in selector position (preceded by `{`, `}`, `,` or start of input) — either fails loudly on the upgrade the paragraph is worried about.

template-mfe/src-app/mfe_packages/demo-mfe/src/screens/uikit/components/CategoryMenu.tsx
```diff
@@ -66,42 +67,42 @@ export const CategoryMenu: React.FC<CategoryMenuProps> = ({ t, activeElement, co
   };
 
   return (
-    <nav className="sticky top-4 bg-background border border-border rounded-lg p-4 max-h-[calc(100vh-2rem)] overflow-y-auto">
-      <h2 className="text-lg font-semibold mb-4">{t('title')}</h2>
-      <ul className="space-y-1">
+    <nav className={styles.menu}>
+      <h2 className={styles.menuTitle}>{t('title')}</h2>
```
gs-layer: **[low] The sidebar heading reuses the page title key**

`t('title')` is the same key the content `<h1>` renders (`UIKitElementsScreen.tsx:212`), so in `en` the user sees "UIKit Elements" twice, and a screen reader's heading list gets two identical entries at different levels while the `<nav>` landmark ends up without a distinct name.

No separate key (`menu_title`, `categories`) exists in `i18n/en.json` yet.

gs-layer:
**[medium] `demo-mfe`'s lifecycle test mocks the base class away, so kit token injection is uncovered**

Posting this as a general comment rather than inline: `template-mfe/src-app/mfe_packages/demo-mfe/src/lifecycle.test.tsx` is not part of this PR's diff, so there is no line to anchor to — but the code it fails to cover (`KitThemedLifecycle`) is new here.

That file's `vi.mock` replaces `ThemeAwareReactLifecycle` with a stub that has no `initializeStyles` at all (line 20-27), and `lifecycle-uikit` is never imported by the test. Consequently neither `KitThemedLifecycle.initializeStyles` (`shared/KitThemedLifecycle.ts:50-54`) nor the `extends KitThemedLifecycle` on the three kit lifecycles is asserted anywhere.

Mutation check: change `lifecycle-helloworld` / `lifecycle-profile` / `lifecycle-uikit` back to `extends ThemeAwareReactLifecycle`, or empty out `KitThemedLifecycle.initializeStyles`. Three of the four demo screens then mount with no kit tokens in their shadow root — every kit `var()` unresolved, components rendering unpainted — and the `demo-mfe` suite still passes.

`_blank-mfe` has the equivalent assertion (`lifecycle.test.tsx:112-126`); `demo-mfe` does not.

template-mfe/src-app/mfe_packages/demo-mfe/src/screens/uikit/UIKitElementsScreen.tsx
itechmeat: Fixed: the portal-container div renders before the Toaster, and a test with a layout-effect stand-in mirroring Base UI's resolution pins that the ref is attached by the time it is read.

template-mfe/src-app/mfe_packages/_blank-mfe/src/shared/anchorKitThemeOnShadowHost.ts
itechmeat: Fixed in both copies: a root selector with a compound tail is rewritten to the functional :host(tail) form, pseudo-elements staying outside the parens, and the test that pinned the dead selector now pins :host(:not([data-theme='light'])) plus a guard that no compound :host survives.

template-mfe/src-app/mfe_packages/demo-mfe/src/screens/uikit/UIKitElementsScreen.tsx
itechmeat: Fixed: the effect re-runs when translations resolve and a MutationObserver re-scans as the lazy categories mount; a test asserts observe is called with the element nodes that appear after loading.

template-mfe/src-app/mfe_packages/demo-mfe/src/shared/kitThemeScope.ts
itechmeat: Fixed: kitThemeScope.test.ts is duplicated into demo-mfe verbatim, the same way the module and the anchor test are - emptying DARK_HOST_THEMES now reddens it.

template-mfe/src-app/mfe_packages/demo-mfe/src/screens/profile/components/ProfileDetailsCard.tsx
itechmeat: Fixed at the boundary: getFormValues coalesces the three nullable fields to empty strings, so a null field renders instead of unmounting the screen; pinned by a null-payload test.

template-mfe/src-app/mfe_packages/demo-mfe/src/screens/uikit/UIKitElementsScreen.tsx
itechmeat: Fixed: develop's wording is restored verbatim.

template-mfe/src-app/mfe_packages/demo-mfe/src/screens/profile/ProfileScreen.test.tsx
itechmeat: Fixed in all four demo suites: a real shadow host is raised before render, the initial ltr pass is asserted, and inverting the fallback to rtl now reddens every one of them.

template-mfe/src-app/mfe_packages/demo-mfe/src/screens/uikit/UIKitElementsScreen.tsx
itechmeat: Fixed: label and closeLabel come from the screen namespace, with the two keys translated in all 36 locales.

template-mfe/src-app/mfe_packages/demo-mfe/src/screens/uikit/i18n/ru.json
itechmeat: Fixed: the four keys this PR added are really translated in all 35 non-English locales - the three cells that stay as Label in nl, id and ms are those languages' own word for it, not a copy - and a new test pins key-set parity across all 36 locales.

template-mfe/src-app/mfe_packages/demo-mfe/src/screens/profile/components/ProfileDetailsCard.tsx
itechmeat: Fixed: the updater keeps null as null and the dead initialValues fallback is gone.

template-mfe/src-app/mfe_packages/_blank-mfe/README.md
itechmeat: Fixed: the step now points at a missing assets.css entry and at tokens not anchored on :host, since the handler injects sync and async CSS alike.

template-mfe/src-app/mfe_packages/demo-mfe/vitest.config.ts
itechmeat: Fixed in both configs: the comments now say css: true is load-bearing for the pinned-stylesheet guard - confirmed by turning it off, which reddens exactly the two tests that import the real stylesheet.

template-mfe/src-app/mfe_packages/_blank-mfe/src/shared/anchorKitThemeOnShadowHost.ts
itechmeat: Fixed: the test now counts the :root occurrences in the pinned stylesheet and asserts each stands in selector position - an injected @supports selector(:root) block reddens it where the old assertion stayed green; the docstrings in both copies claim exactly that now.

template-mfe/src-app/mfe_packages/demo-mfe/src/screens/uikit/components/CategoryMenu.tsx
itechmeat: Fixed: the sidebar names itself from a new menu_title key, translated in all 36 locales and pinned by a new CategoryMenu test.

itechmeat:
@gs-layer fixed: demo-mfe's lifecycle test now imports lifecycle-uikit, pins that the three kit lifecycles extend KitThemedLifecycle (and that lifecycle-theme does not), and pins that initializeStyles re-anchors the kit tokens onto the shadow host - emptying it or reverting the extends reddens the suite.

GeraBart (approved):

PR #586, opened 2026-08-18
feat(template): adopt the design-guardrails template

tscbmstubp:
Adopts the design-guardrails template into the monorepo and hardens it through eight pilot runs until agents generate correct screens in one pass. The PR adds one new top-level template, a verification workspace package, three CI gates, and the shell fixes the pilots forced.

## The template

`template-design-guardrails/` is a new manifest-only template, identity `@gears-frontx/template-design-guardrails`. It ships `DESIGN.md` plus an AI-extension bundle: four skills (`generate-interface`, `review-interface`, `verify-interface`, `create-theme`), workflows, ten guidelines, and the contract artifacts (`design-contract.json`, `quality-gates.json`, `component-state-matrix.json`, `layout-contract.json`, `review-rubric.json`, a generated guideline routing index). Imported from the `frontx-design-guardrails` v0.1 handoff (31/31 checksums verified), with the `frontx-` prefix dropped from all identities.

The bundle went through several content audits after import. Rules that lived in two or three guidelines each now have one owner and pointers. The fallback token set, the component manifest, and kit-guaranteed behavior rows are gone: the template targets `@gears-frontx/ui-kit` specifically, so the installed kit is the only token and component authority. Hedges written for an unknown host ("if Tailwind is present") are gone for the same reason.

## Runtime verification

The template delivers its own verification machinery as a workspace package, `@gears-frontx/design-verify`, using the same subtree mechanism template-mfe uses. The shell knows only a generic convention: it loads `src/dev-entry.ts` from every installed verify package and logs whether none, a broken one, or a running one is present. A shell without the template carries nothing.

The package contains:

- A dev-only design-defect checker (tree-shaken from production): axe-core over the live document including shadow roots, plus objective layout rules. Current rule set: page horizontal scroll, clipped text, tiny text, skipped headings, control-height mismatch in shared rows, card content flush against or overflowing a drawn edge, and dead spacing classes (a utility present in the DOM that no CSS rule defines). Findings print on a `[design-defects]` console prefix; `window.__frontxDesignDefects()` re-runs on demand. The FrontX Studio panel is excluded, so every finding belongs to the app.
- A headless runner (`npm run verify:ui`) that sweeps a screens × themes × widths matrix over CDP and writes `findings.json`. It supports menu-label navigation for shells that mount screens on click (a route sweep there measures the empty shell and passes vacuously), can start and stop the dev server itself, and diffs against a prior report via `--baseline` (new / resolved / carried over). Screenshots are opt-in: findings.json is the evidence channel, and agents neither take nor read images during generation.
- An a11y gate: `eslint-plugin-jsx-a11y` on all shell JSX and a vitest-axe suite over composed ui-kit markup, with a canary test proving axe still detects a known violation so the green assertions cannot go vacuous.

## Shell fixes the pilots forced

- ui-kit components now render styled inside MFE shadow roots: the app entry imports the kit's CSS for adoption into shadow roots, shell tokens migrated from HSL fragments to the kit's full-color format, `data-theme` stamping follows the shell theme, and hover tokens derive from the active theme via color-mix. Studio's own stylesheet joined the token migration.
- The host Tailwind config now scans MFE package sources. MFEs ship no stylesheet of their own, so a spacing utility unused by host code matched no CSS rule and silently did nothing. This was the root cause of both pilot variants losing their margins; the dead-spacing-class checker rule keeps it from regressing.
- `frontxMfGts()` fails the MFE build when an expose's CSS cannot reach the shadow root, naming the offending files. Previously this failed silently and cost a pilot three fix rounds.
- A token-format CI gate scans template territory for the retired `hsl(var(--x))` vocabulary, which fails silently against the migrated tokens. Its first sweep caught one latent regression.
- The four template CI guards and the composition job now handle manifest-only templates (no root `package.json`).

## Process rules the pilots taught

Each rule below exists because a pilot run failed without it:

- Delegated screen-writing must hand over the contract files explicitly; an implementer that has not read them is regenerated, and all delegation runs in the foreground (a background subagent is orphaned when a headless session's turn ends).
- The verify-fix loop is bounded: two attempts per finding, ten per screen, stop on repeated regression, report the rest as open.
- A change not seen rendered is a hypothesis; fixes are re-verified before being reported resolved. Sweeps are tiered so intermediate fixes re-run only their own state, and the closing full pass compares against the first pass via `--baseline`.
- template-mfe's add-mfe-package skill now states what screens are built from: the ui-kit inventory in the installed `llms.txt`, under the guardrails contract when the bundle is installed. Hand-rolled look-alikes of kit components are defects. The scaffold's Bridge Info demo card is named as residue that must not ship. A parallel workflow phases the per-package steps for multi-MFE requests.

## Versions

Each changed package bumps once relative to develop (post-#584): template-shell alpha.3 to alpha.4 (MFE pins follow), react alpha.3 to alpha.4, framework alpha.2 to alpha.3, studio alpha.1 to alpha.2, template-mfe alpha.0 to alpha.1, template-design-guardrails and design-verify are new at 0.1.0-alpha.0.

## Verification

Offline install → seed → add activates all bundle entries with zero structural errors. Full CI replicated locally: root build/type-check/lint/tests/arch, all policy guards, template-shell standalone, and the composition job on a scratch replica. The runner and every checker rule were validated live against pilot apps, two-sided where possible (defect present fails, defect fixed passes). Pilot evidence: on the final template, a sonnet-medium run produced a two-screen app in one pass with a clean 8-state sweep.

## Notes for reviewers

- Rebased onto develop after #584 landed; the branch is squashed to one commit. #584's new lifecycle comments described the pre-migration HSL-triplet tokens and were updated to the migrated reality (the token-format gate caught them).
- The review skill stays advisory in v0.1; nothing here gates a merge on design findings.
- On merge, the landing SHA becomes the pilot pin.

## Summary by CodeRabbit

* **New Features**
  * Added a design guardrails template with accessibility, responsive layout, theming, interaction, and content guidance.
  * Added advisory interface review and verification workflows with visual, accessibility, and design-defect checks.
  * Added guideline indexing, token-format validation, and parallel MFE package creation workflows.
  * Added light/dark theme metadata and improved theme color handling.

* **Bug Fixes**
  * Improved CSS delivery validation to prevent unstyled federated interfaces.
  * Improved validation for documentation-only templates and manifest-only overlays.
  * Added accessible page headings, landmarks, and valid sidebar list structure.


denyakortunov:
Looks good overall. I think one important part is still missing: clearer rules for how the agent builds a full screen.

FrontX UI Kit should remain the only component source. Since it follows shadcn names and patterns, the agent can use that familiar structure, but must always use the actual FrontX components and APIs.

Before coding, it should map each UI pattern to a FrontX component and define the grid and alignments instead of guessing.

Automated checks are useful, but we still need a final visual review. The idea that shadcn patterns improve generation also needs to be tested with a few pilots.

template-design-guardrails/.frontx/ai/@gears-frontx/template-design-guardrails/skills/generate-interface/SKILL.md
```diff
@@ -0,0 +1,181 @@
...
+A delegated screen written without the contract looks plausible and fails
+review; the contract does not transfer by implication.
+
+## Plan before coding
+
+State briefly:
+
+- screen job and primary action;
+- shell and composition pattern;
+- selected UI kit components;
```
denyakortunov: Can we make this more specific? Before coding, the agent should map every visible UI pattern to a FrontX component or approved composition, and define the grid and alignments. If something is missing, it should report the gap instead of guessing or recreating it locally.

template-design-guardrails/.frontx/ai/@gears-frontx/template-design-guardrails/skills/verify-interface/SKILL.md
```diff
@@ -0,0 +1,138 @@
...
+URLs) is only for apps that actually mount on navigation: run it against an
+action-mounted shell and every sweep measures the empty shell while
+reporting clean — a vacuous verification that reads as a pass.
+
+Your evidence channel is `findings.json` and the `[design-defects]` console
+lines — that is the complete output; a finding is mapped from its rule id
+and detail, never from pixels. The runner captures no screenshots by
+default; leave it that way during generation and verification, and do not
+read screenshot files back: rendering images into your context is expensive
+and adds nothing the findings do not already state. Screenshots are visual
```
denyakortunov: I don’t think screenshots add nothing. The checker catches known technical issues, but not hierarchy, density, component choice or overall visual quality. It’s fine to skip screenshots during every fix, but we should require one final visual review before calling the screen correct.

template-design-guardrails/.frontx/ai/@gears-frontx/template-design-guardrails/guidelines/constructor-foundations.md
```diff
@@ -0,0 +1,67 @@
...
+
+`MUST`, `MUST NOT`, `SHOULD`, `SHOULD NOT`, and `MAY` are requirement levels.
+
+When sources disagree, use this order:
+
+1. Installed application behavior and product requirements.
+2. Installed FrontX template manifests and `@gears-frontx/ui-kit` documentation.
+3. Machine-readable contracts in this bundle.
+4. Guidelines in this bundle.
+5. Approved examples from the host product.
```
denyakortunov: Should approved product designs be higher in this list? The current app may contain old UI. I think approved requirements and design references should define how the screen looks, while the UI kit docs remain the source of truth for component APIs.

template-design-guardrails/.frontx/ai/@gears-frontx/template-design-guardrails/reference-artifacts/design-contract.json
```diff
@@ -0,0 +1,73 @@
...
+    "MUST": "Required; violation is a blocker or major finding according to quality-gates.json.",
+    "MUST_NOT": "Prohibited; violation is a blocker or major finding according to quality-gates.json.",
+    "SHOULD": "Expected; an exception requires a recorded product or technical rationale.",
+    "MAY": "Optional and context-dependent."
+  },
+  "uiBase": {
+    "package": "@gears-frontx/ui-kit",
+    "observedVersion": "0.3.0-alpha.3",
+    "implementation": ["React 19", "Base UI", "CSS Modules", "CVA"],
+    "inventory": "The installed llms.txt is the component inventory and API reference; this bundle does not duplicate it."
```
denyakortunov: Can we mention that the kit uses shadcn-aligned names and patterns? This gives the agent a familiar structure without making shadcn a second dependency. The actual components and APIs should still always come from FrontX, with no direct shadcn imports.

template-design-guardrails/.frontx/ai/@gears-frontx/template-design-guardrails/guidelines/typography-content-and-iconography.md
```diff
@@ -0,0 +1,85 @@
...
+- Never truncate the only primary action or state.
+- Prefer wrapping titles before shrinking type.
+- Paths and hashes may use middle truncation when the beginning and end both matter.
+- A truncated value **MUST** have a keyboard-accessible way to reveal or copy the full value.
+
+## Iconography
+
+- Icons reinforce, not replace, unfamiliar text.
+- Status tables **SHOULD** use a simple dot plus label rather than a unique decorative icon per row.
+- Chevron means disclosure, not sorting. Use a directional arrow for sort order.
```
denyakortunov: Can this follow the installed table pattern instead? Our approved tables use ChevronUp and ChevronDown for sorting. I’d keep the rule about consistent sort indicators and aria-sort, but not hard-code arrow versus chevron here.

template-design-guardrails/.frontx/ai/@gears-frontx/template-design-guardrails/skills/generate-interface/SKILL.md
```diff
@@ -0,0 +1,181 @@
...
+  reservation plus a margin for the same slot, or an invisible spacer element.
+  If intended spacing does not appear in the rendered result, diagnose why the
+  styles are not applying instead of adding more; hedges written while styles
+  are broken all fire at once when the styles start working.
+- Match control size variants inside a shared row: a button beside an input
+  must use the size variant with the same control height. Align to the input
+  element itself — top and bottom edges coinciding — not to the field block
+  around it (label above, error or help text below). Alignment comes from
+  the row's layout, never from spacer elements or eyeballed margins. The
+  button must stay beside the input at every supported width, not wrap to
```
denyakortunov: Do we need this to be absolute? The controls should have matching heights while they share a row, but an approved narrow layout may stack them. I’d let the product’s responsive pattern decide instead of always forbidding wrapping.

template-design-guardrails/.frontx/ai/@gears-frontx/template-design-guardrails/skills/generate-interface/SKILL.md
```diff
@@ -0,0 +1,181 @@
...
+or the matching vitest config for in-shell screens — and run the project's
+full type-check and test suite exactly once, when the screen is complete.
+A full-project sweep repeated per fix rebuilds every package and dominates
+iteration time without adding evidence about the changed screen.
+
+A change you have not seen rendered is a hypothesis, not a fix. Source-level
+self-review verifies that values exist; it cannot verify that they compose —
+whether gaps stack, controls align, or a stylesheet is actually delivered.
+Treat the app as runnable only when a dev server is already reachable (e.g.,
+an existing dev-mode session or open page); never start one yourself.
```
denyakortunov: Doesn’t this conflict with verify-interface, which can start and stop the server itself? I think the agent should be allowed to call the official verification runner and let it manage the server. Otherwise the final check may often remain unverified.

denyakortunov (commented):
Left a few product and design comments. Overall the direction looks good.

template-design-guardrails/.frontx/ai/@gears-frontx/template-design-guardrails/skills/generate-interface/SKILL.md
tscbmstubp: Addressed in ddf028bb — the plan step now requires a component map (every visible pattern → kit component or approved composition) plus the grid and alignment lines; anything unmapped is reported as a gap, not guessed or recreated locally.

template-design-guardrails/.frontx/ai/@gears-frontx/template-design-guardrails/skills/verify-interface/SKILL.md
tscbmstubp: Addressed in ddf028bb — the closing full pass now runs with `--screenshots` and the agent reviews them once (every screen, both themes, widest + 320px) against review-rubric.json for hierarchy, density, component choice and overall coherence. The report must carry that review to count as complete. Screenshots stay off during the fix loop.

template-design-guardrails/.frontx/ai/@gears-frontx/template-design-guardrails/guidelines/constructor-foundations.md
tscbmstubp: Addressed in ddf028bb — approved product requirements and design references now rank first in both constructor-foundations and design-contract `sourcePrecedence`; installed application behavior drops to fifth, with a sentence explaining the app may carry UI the design is replacing.

template-design-guardrails/.frontx/ai/@gears-frontx/template-design-guardrails/reference-artifacts/design-contract.json
tscbmstubp: Addressed in ddf028bb — added `uiBase.naming`: shadcn-aligned names and patterns, every component and API still from @gears-frontx/ui-kit, no direct shadcn imports.

template-design-guardrails/.frontx/ai/@gears-frontx/template-design-guardrails/guidelines/typography-content-and-iconography.md
tscbmstubp: Done — the guideline now defers to the installed table pattern and keeps the `aria-sort` requirement (ddf028bb). One heads-up: the installed @gears-frontx/ui-kit build ships ArrowUp/ArrowDown in its sortable header, not ChevronUp/ChevronDown, so the kit and the approved table designs currently disagree. Worth raising with the ui-kit owners so there's a single answer for the agent to follow.

template-design-guardrails/.frontx/ai/@gears-frontx/template-design-guardrails/skills/generate-interface/SKILL.md
tscbmstubp: Addressed in ddf028bb — heights must match while the controls share a row; whether a narrow layout stacks them is the product's responsive pattern to decide. Accidental wrapping is still a defect.

template-design-guardrails/.frontx/ai/@gears-frontx/template-design-guardrails/skills/generate-interface/SKILL.md
tscbmstubp: Addressed in ddf028bb — the agent may bring the server up through the verify runner's `--start-server` (bounded waits, stops it when done). Only an unbounded foreground dev-server command stays forbidden, which removes the conflict with verify-interface.

.github/workflows/main.yml
```diff
@@ -347,15 +368,34 @@ jobs:
...
+            # $shell below: a template like template-design-guardrails ships
+            # no root package.json (it's a docs/AI bundle) yet still owns an
+            # installable package subtree (src-app/verify_packages/*). Root
+            # package.json alone would misclassify it as manifest-only and
+            # skip the zero-count guard, so any subtree — file or directory —
+            # that contains a package.json counts as code-bearing.
+            if find "$overlay/$subtree" -not -path '*/node_modules/*' \
+              -name package.json 2>/dev/null | grep -q .; then
+              overlay_has_pkg=1
+            fi
```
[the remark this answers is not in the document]
tscbmstubp: Fixed in 39693f71 — probe is now `[ -n "$(find … -print -quit)" ]`.

scripts/token-format-check.mjs
```diff
@@ -0,0 +1,203 @@
...
+// guards: the filename is owned by `template-discovery.mjs`.
+export { MANIFEST_FILENAME };
+
+/**
+ * A custom property consumed through an HSL wrapper: `hsl(var(` / `hsla(var(`,
+ * with optional whitespace after the opening paren. `var(--x, hsl(...))` — a
+ * full-color fallback inside `var()` — does not match: there the wrapper
+ * contains no `var(`.
+ */
+export const FORBIDDEN_PATTERN = /\bhsla?\(\s*var\(/;
```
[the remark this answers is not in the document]
tscbmstubp: Fixed in 39693f71 — added the `i` flag plus `HSL(`/`Hsla(` cases in token-format-check.test.mjs.

template-design-guardrails/src-app/verify_packages/design-verify/runner/cdp.mjs
```diff
@@ -0,0 +1,174 @@
...
+    this.ws = new this.WebSocketImpl(this.wsUrl);
+    await new Promise((resolve, reject) => {
+      this.ws.onopen = resolve;
+      this.ws.onerror = () => reject(new Error(`CDP WebSocket failed to open: ${this.wsUrl}`));
+    });
+    this.ws.onmessage = (event) => this.handleMessage(String(event.data));
+    await this.send('Runtime.enable');
+    await this.send('Log.enable');
+    await this.send('Page.enable');
+  }
```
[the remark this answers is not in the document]
tscbmstubp: Fixed in 39693f71 — `rejectPending()` runs on `onerror`, `onclose` and `close()`; fake-WebSocket test covers all three.

template-design-guardrails/src-app/verify_packages/design-verify/runner/chrome.mjs
```diff
@@ -0,0 +1,176 @@
...
+      } catch {
+        return [];
+      }
+    });
+  const found = [];
+  const playwrightRoot = path.join(home, '.cache', 'ms-playwright');
+  for (const entry of listDir(playwrightRoot)
+    .filter((e) => /^chromium-\d+$/.test(e))
+    .sort()
+    .reverse()) {
```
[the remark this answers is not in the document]
tscbmstubp: Fixed in 39693f71 — `compareVersionsDesc` compares embedded numbers numerically; fixture now includes chromium-999 vs chromium-1208 and linux-99 vs linux-115.

template-design-guardrails/src-app/verify_packages/design-verify/runner/config.mjs
```diff
@@ -0,0 +1,225 @@
...
+      case '--menu':
+        config.menus = value.split(',').map((m) => m.trim()).filter(Boolean);
+        break;
+      case '--widths': {
+        const widths = value.split(',').map((w) => Number.parseInt(w.trim(), 10));
+        if (widths.some((w) => !Number.isFinite(w) || w <= 0)) {
+          errors.push(`--widths must be positive integers, got: ${value}`);
+        } else {
+          config.widths = widths;
+        }
```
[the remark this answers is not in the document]
tscbmstubp: Fixed in 39693f71 — `parseWholeInteger` requires `^\d+$`; tests cover `1.5`, `1440px`, `700ms`, `1e3`.

template-mfe/.frontx/ai/@gears-frontx/frontx-template-mfe/skills/add-mfe-package/SKILL.md
```diff
@@ -84,11 +87,38 @@ to mount into an extension domain).
...
+   `@gears-frontx/template-design-guardrails` bundle is installed in the project
+   (`.frontx/ai/@gears-frontx/template-design-guardrails/`), its `generate-interface`
+   skill and design contract govern how the screen is generated — follow them. If it
+   is not installed, say so in the plan: the screen is being generated without a
+   design contract.
+   Replace the scaffold's `HomeScreen` content wholesale: its "Bridge Info"
+   card (domain id, instance id, theme, language) is scaffold demo residue,
+   the screen equivalent of the `templateExample` flag — a shipped screen
+   that still renders it is a defect. Keep the lifecycle wiring the scaffold
+   demonstrates; drop the demo markup and its i18n keys.
```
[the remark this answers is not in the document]
tscbmstubp: Fixed in 39693f71 — the step now requires replacement keys in every shipped locale with identical key sets.

template-mfe/.frontx/ai/@gears-frontx/frontx-template-mfe/workflows/add-mfe-package-workflow.md
```diff
@@ -46,6 +46,15 @@ each step names the concrete command or file template-mfe ships.
...
    - Rename/replace `src/screens/home/HomeScreen.tsx` with the real screen.
+   - Build the UI from `@gears-frontx/ui-kit` components and semantic tokens:
+     confirm the package declares `@gears-frontx/ui-kit` in its dependencies (add
+     it if the scaffold copy predates the kit migration and lacks it) and read the
+     installed kit's `llms.txt` for the component inventory before writing markup.
+     Hand-rolling a look-alike of an existing kit component is a defect.
+   - If `.frontx/ai/@gears-frontx/template-design-guardrails/` exists in the
+     project, that bundle's `generate-interface` skill and design contract govern
+     the screen's generation — follow them. If it does not, state in the plan that
+     the screen is generated without a design contract.
```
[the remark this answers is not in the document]
tscbmstubp: Fixed in 39693f71 — step 8 runs `verify-interface` when the guardrails bundle is installed, and says so in the hand-off when it is not.

template-shell/.frontx/ai/@gears-frontx/frontx-template-shell/guidelines/mfe-package-contract.md
```diff
@@ -74,6 +74,41 @@ three, on one shared predicate per rule rather than a copy each.
...
+
+MFE screens render inside open shadow roots, and shadow roots do not see the
+host document's stylesheets by default. Three delivery paths exist; pick by
+what the package owns:
+
+1. **No package-owned CSS (the `template-mfe` examples).** The host's compiled
+   Tailwind covers `src/mfe_packages/**` content paths, and
+   `ThemeAwareReactLifecycle.adoptHostStylesIntoShadowRoot()` clones every host
+   `<style>`/`<link>` into the shadow root. A package that only uses Tailwind
+   utilities and UI-kit components needs no stylesheet of its own.
```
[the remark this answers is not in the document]
tscbmstubp: Fixed in 39693f71 — both sites now say `src-app/mfe_packages/**`.

template-shell/.frontx/ai/@gears-frontx/frontx-template-shell/guidelines/mfe-package-contract.md
```diff
@@ -74,6 +74,41 @@ three, on one shared predicate per rule rather than a copy each.
...
+   prescribes is path 2. (UI-kit CSS-modules from `node_modules` are exempt:
+   the host delivers those by style adoption, path 1.)
+
+**Never deliver Tailwind's preflight (`@tailwind base` / the `@import
+"tailwindcss"` base layer) into a shadow root.** Preflight's element resets
+(`button { background: transparent }` and friends) override UI-kit component
+CSS inside the shadow root and visibly break kit controls. A package-owned
+stylesheet destined for `initializeStyles()` must contain the utilities and
+components layers only; `injectBaseResets()` already provides the box-model
+resets a shadow root needs.
```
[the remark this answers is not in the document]
tscbmstubp: Fixed in 39693f71 — the rule is scoped to package-owned CSS (paths 2/3) and states that path 1 adopts the host preflight on purpose, pointing at the lifecycle's counter-declarations.

template-design-guardrails/src-app/verify_packages/design-verify/src/designDefects.ts
```diff
@@ -0,0 +1,683 @@
...
+      probe.className = token;
+      // gap only resolves to a length in a flex/grid formatting context.
+      probe.style.display = 'flex';
+      probe.style.flexDirection = 'column';
+      let measured = probe;
+      if (spec.kind === 'sibling') {
+        probe.appendChild(doc.createElement('div'));
+        measured = doc.createElement('div');
+        probe.appendChild(measured);
+      }
```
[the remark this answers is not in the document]
tscbmstubp: Not a bug. `dead` is `spec.properties.every(prop => zero)`, so a probe is dead only when *both* margins are zero. With Tailwind 3.4.1 `space-y-*` the measured second child has a non-zero `margin-top`, `every` returns false, and the class is correctly reported alive. Switching to `some` would invert the check and flag every live utility. Resolving.

template-shell/packages/studio/src/styles/studioStyles.ts
```diff
@@ -199,31 +199,31 @@ export const STUDIO_CSS = /* css */ `
...
 .dark .dark\\:hover\\:bg-black\\/60:hover { background-color: rgb(0 0 0 / 0.6); }
-.hover\\:text-muted-foreground:hover { color: hsl(var(--muted-foreground)); }
-:where(.hover\\:bg-accent:hover) { background-color: hsl(var(--foreground) / 0.1); }
-:where(.hover\\:bg-primary\\/90:hover) { background-color: hsl(var(--foreground) / 0.85); }
-:where(.hover\\:bg-secondary\\/80:hover) { background-color: hsl(var(--foreground) / 0.08); }
-:where(.hover\\:bg-destructive\\/90:hover) { background-color: hsl(0 60% 50% / 0.9); }
+.hover\\:text-muted-foreground:hover { color: var(--muted-foreground); }
+:where(.hover\\:bg-accent:hover) { background-color: color-mix(in oklab, var(--foreground) 10%, transparent); }
+:where(.hover\\:bg-\\[color-mix\\(in_oklab\\,var\\(--primary\\)_90\\%\\,transparent\\)\\]:hover) { background-color: color-mix(in oklab, var(--foreground) 85%, transparent); }
+:where(.hover\\:bg-\\[color-mix\\(in_oklab\\,var\\(--secondary\\)_80\\%\\,transparent\\)\\]:hover) { background-color: color-mix(in oklab, var(--foreground) 8%, transparent); }
```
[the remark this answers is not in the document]
tscbmstubp: Holding this one for a maintainer decision. The whole `STUDIO_CSS` block deliberately remaps tokens onto `--foreground` mixes (bg-accent → 10%, bg-input → 15%, border-input → 20%), so 85%/8% look like part of that palette rather than a typo — restoring 90%/80% *of `--foreground`* would make secondary hover almost opaque foreground. Will confirm intent before changing.

template-design-guardrails/.frontx/ai/@gears-frontx/template-design-guardrails/skills/generate-interface/SKILL.md
```diff
@@ -0,0 +1,188 @@
...
+## Implement
+
+Rules phrased "Never" or "Do not" are binding — blocker-mapped where `quality-gates.json` says so; rules phrased "avoid" or "prefer" are strong defaults that may yield to a documented product reason.
+
+- Import `@gears-frontx/ui-kit/theme.css` once at the application entry.
+- Compose with UI kit components and native semantic elements.
+- Use the existing app's composition styling mechanism. Never override UI kit internals.
+- Use semantic tokens for visual values. Map Tailwind utilities to tokens. Never use an arbitrary color, radius, or spacing value when a semantic token exists (design-contract prohibition; SYS-002).
+- Use the spacing scale and primary rhythm declared in `layout-contract.json`; verify against the project's Tailwind config.
+- Use one page heading and exactly one primary-styled action per local region; use at most one non-neutral semantic color per region unless each maps to a distinct state.
```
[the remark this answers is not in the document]
tscbmstubp: Holding for a maintainer decision: whether a data-series palette exemption belongs in this skill or whether generated screens don't carry charts.

scripts/generate-guideline-index.mjs
GeraBart, line 1: this script seems to be specific for the design-guardrails template. i assume it should be moved there

template-mfe/.frontx/ai/@gears-frontx/frontx-template-mfe/skills/add-mfe-package/SKILL.md
```diff
@@ -84,11 +87,41 @@ to mount into an extension domain).
...
    `createFrontX().use(effects()).use(queryCacheShared()).use(mock()).build()` so it
    joins the host's shared `QueryClient` without owning a second one.
-6. **Regenerate manifests** — run `npm run generate:mfe-manifests` so the host's
+6. **Build the screen's UI from `@gears-frontx/ui-kit`** — the shell installs the
+   kit, and its installed `llms.txt` is the component inventory. Confirm the copied
+   package declares `@gears-frontx/ui-kit` in its dependencies (add it if the
+   scaffold copy predates the kit migration and lacks it) and compose from kit
+   components and semantic tokens; hand-rolling a
+   look-alike of an existing kit component is a defect, not a style choice. If the
+   `@gears-frontx/template-design-guardrails` bundle is installed in the project
```
GeraBart: i don't think mfe template should have any knowledge about the design guardrails template.

scripts/generate-guideline-index.mjs
tscbmstubp: It reads single-tenant today, but it's a CI policy guard rather than template content, and `scripts/` is where those live: `token-format-check.mjs` and `template-pin-drift-check.mjs` also operate on template territory from here, wired as root `policy:*` npm scripts that `main.yml` runs. The template itself can't host it — it's manifest-only (no `package.json`, so no npm scripts or test runner), and its `frontx-template.json` installs only the declared exclusive subtrees, so a script parked in the template directory would be repo dev tooling living in shipped-template territory while root CI still reaches into it. `GOVERNED_BUNDLES` is an explicit opt-in list precisely so other templates that adopt guideline front matter join the same guard later (the header comment covers why it's not template discovery). Happy to move it if we'd rather break the pattern, but I'd keep it with its siblings.

template-mfe/.frontx/ai/@gears-frontx/frontx-template-mfe/skills/add-mfe-package/SKILL.md
tscbmstubp: Agreed — fixed in 891a027a. template-mfe no longer names the guardrails template anywhere; the skill and both workflows now use the same generic convention the shell applies to verify packages: any installed AI bundle shipping a `generate-interface` skill (`.frontx/ai/*/*/skills/generate-interface/`) is the design-contract bundle that governs screen generation, and the closing verify pass detects `verify-interface` the same way. The guardrails bundle satisfies the convention without being named.

tscbmstubp:
@GeraBart no conflicts

template-design-guardrails/src-app/verify_packages/design-verify/package.json
```diff
@@ -0,0 +1,27 @@
...
+  "version": "0.1.0-alpha.0",
+  "type": "module",
+  "description": "Design-verification workspace package delivered by @gears-frontx/template-design-guardrails: the dev-only runtime design-defect checker (axe-core + objective layout rules), the composed-markup axe test suite, and the jsx-a11y lint fragment. The shell wires each of these in optionally, so an app without the guardrails template carries no verification machinery.",
+  "exports": {
+    "./runtime": "./src/designDefects.ts",
+    "./eslint": "./eslint.config.mjs"
+  },
+  "scripts": {
+    "type-check": "tsc --noEmit",
+    "verify:ui": "node runner/cli.mjs",
```
itechmeat, line 13: [Not a blocker] 🟠 runner/cli.mjs carries a raw NUL byte (line 142, the pageKey separator), so git treats the file as binary and GitHub shows no diff - the largest runner file landed unreviewable. Replacing the raw byte with the escaped form (backslash-u0000, or `JSON.stringify([route, theme])`) is behavior-identical and one character wide; please push that so the file gets a reviewable text diff. Since the platform never showed it, I reviewed the file as-is: it's well-factored, args match config.mjs, cleanup ordering and exit codes are right. Four non-blocking robustness notes: (1) the state loop has no per-state try, so one stalled CDP command at state 7 of 9 unwinds past buildReport and the run exits with no findings.json at all - catching per state and recording environmentError would keep the verified states; (2) the [verify-packages] failure scan starts at beforeReload, so a failure logged during the first load is missed and the run keeps re-visiting the dead route once per width (minutes of sweepTimeout waits) instead of short-circuiting; (3) in menu mode routeSlug strips [^a-zA-Z0-9]+ with a 'root' fallback, so non-ASCII menu labels all write root-<theme>-<width>.png and silently overwrite each other; (4) the cleanup loop isn't per-try - if chrome.stop throws, the detached dev server survives the run.

template-shell/src-app/app/main.tsx
```diff
@@ -5,7 +5,9 @@ import { FrontXProvider, apiRegistry, createFrontXApp, MfeHandlerMF, gtsPlugin,
 import { themeSchema, languageSchema, extensionScreenSchema } from '@gears-frontx/frontx-template-shell';
 import { Toaster } from '@/app/components/ui/sonner';
 import { AccountsApiService } from '@/app/api';
+import '@gears-frontx/ui-kit/theme.css'; // UI-kit design tokens (imported exactly once, at the app entry)
 import './globals.css'; // Global styles with CSS variables
+import '@gears-frontx/ui-kit'; // side-effect: puts ui-kit component CSS in the host document, where MFE shadow roots adopt it (ThemeAwareReactLifecycle). Must come after globals.css so component rules beat Tailwind's preflight resets on specificity ties.
```
itechmeat, line 10: [Not a blocker] 🟠 The shell renders no kit components, yet this bare side-effect import pulls the kit's entire JS graph into the host bundle just to obtain its CSS - ui-kit's package.json declares no sideEffects field, so the bundler keeps every chunk (recharts, embla, cmdk, react-day-picker and the rest). There's genuinely no CSS-only entry today (libInjectCss ties component CSS to its JS chunks), so this is the pragmatic option - but worth a kit-side follow-up shipping a components.css next to theme.css so hosts can take the CSS without the JS?

template-shell/packages/react/src/mfe/ThemeAwareReactLifecycle.tsx
```diff
@@ -100,8 +100,14 @@ function MountRuntimeAwareProvider({
  * Styling strategy:
  * 1. adoptHostStylesIntoShadowRoot() clones all host <style> and <link> into the
  *    front of the shadow root, bringing the full compiled Tailwind CSS (including
- *    MFE utilities, since the host's content paths cover src/mfe_packages/**).
- *    Front, not back, so the MFE's own stylesheets outrank them - see the method.
+ *    MFE utilities, since the host's content paths cover src-app/mfe_packages/**) and
+ *    the ui-kit component CSS (the host entry side-effect-imports
+ *    @gears-frontx/ui-kit so that CSS exists in the host document; the kit's
+ *    CSS-module class names are stable across builds, so they match the classes
```
itechmeat: [Not a blocker] 🟡 "stable across builds" overstates it: the kit's Vite default scoped name ends in the source line number - installed 0.3.0-alpha.3 emits _badge_rbq84_9, and .badge now sits at line 35 of badge.module.css, so the next version emits a different class. Stability holds per version; what actually keeps host and MFE kits equal is the pin-drift policy, and that's worth naming here since the whole host-carries-kit-CSS design rests on it. If an independently deployed MFE ever diverges from the host's kit version, every kit component in it renders unstyled with no error - the same silence the new mf-gts guard kills elsewhere. Reword the comment; a runtime version assert on mount could be a follow-up.

template-shell/packages/framework/src/types.ts
```diff
@@ -314,6 +314,13 @@ export interface ThemeConfig {
   variables: Record<string, string>;
   /** Whether this is the default theme */
   default?: boolean;
+  /**
+   * Base appearance of the theme. When set, apply() stamps it as
+   * `data-theme` on the document root so design-system stylesheets that
+   * switch token sets on [data-theme='light'|'dark'] (e.g. the installed
+   * UI kit's theme.css) follow this theme's light/dark nature.
+   */
+  appearance?: 'light' | 'dark';
```
itechmeat, line 325: [Not a blocker] 🟡 why optional? A consumer theme without appearance makes the registry removeAttribute('data-theme'), and on a dark-OS machine the kit's `:root:not([data-theme='light'])` block then owns every token the theme doesn't define (--surface, --subtle-foreground, the sidebar family) - shell-defined names stay light, kit-only names flip dark: the exact mixed palette #584 described. All five shipped themes set it, so the hole is consumer-themes-only and silent. Stamping `appearance ?? 'light'` keeps the type backward-compatible and closes it.

template-design-guardrails/.frontx/ai/@gears-frontx/template-design-guardrails/reference-artifacts/design-contract.json
```diff
@@ -0,0 +1,74 @@
...
+  "status": "pilot",
+  "normativeLanguage": {
+    "MUST": "Required; violation is a blocker or major finding according to quality-gates.json.",
+    "MUST_NOT": "Prohibited; violation is a blocker or major finding according to quality-gates.json.",
+    "SHOULD": "Expected; an exception requires a recorded product or technical rationale.",
+    "MAY": "Optional and context-dependent."
+  },
+  "uiBase": {
+    "package": "@gears-frontx/ui-kit",
+    "observedVersion": "0.3.0-alpha.3",
```
itechmeat: [Not a blocker] 🟡 A few places in the bundle disagree with the installed kit they name as the only authority: design-contract.json's observedVersion says 0.3.0-alpha.3 while the repo ships and pins 0.4.0-alpha.1 (source-registry.json same); layout-contract.json's spacingScalePx [0,4,8,12,16,24,32,40,48,64,80] misses the kit's 20 and invents 40/48/64/80 (real scale: 4,8,12,16,20,24,32); the generate-interface "Correct" example doesn't compile against the kit (Button sizes are default|sm|lg, Input has no size prop); visual-foundations cites `color.text.primary`, a token shape that doesn't exist. One sweep against the installed kit would make the bundle honest - and is anything keeping these artifacts current going forward, the way policy:guideline-index does for the front matter?

template-design-guardrails/.frontx/ai/@gears-frontx/template-design-guardrails/reference-artifacts/design-contract.json
```diff
@@ -0,0 +1,74 @@
+{
+  "id": "constructor-design-contract",
```
itechmeat: [Not a blocker] 🟡 The guideline body is fully generic - only the filename constructor-foundations.md, its H1 and this contract id carry the company name, in an upstream repo that stays vendor-neutral. Renaming to foundations.md / a neutral id (plus the extension.json entries and a regenerated index) is cheap now and breaking after consumers start pinning these ids - worth doing before merge or right after?

template-shell/src-app/app/globals.css
```diff
@@ -5,38 +5,49 @@
 /**
  * Theme CSS variables
  * Default values for initial load; overridden dynamically by themeRegistry.apply()
+ *
+ * Deliberately NOT inside @layer base: @gears-frontx/ui-kit/theme.css declares
+ * identically named tokens (--background, --foreground, --radius-*, ...) on an
+ * unlayered :root, and unlayered declarations always beat layered ones. Keeping
```
itechmeat: [Not a blocker] 🔵 Tailwind 3.4 emits no native @layer - `@layer base` here is Tailwind's own directive that the PostCSS plugin unwraps, so the compiled output has no cascade layers and "unlayered beats layered" isn't the rule in play (verified by running postcss: no @layer survives in the emitted CSS). The move itself is fine; ordering vs theme.css was already decided by import order in main.tsx. Worth rewording the comment to the real mechanism so the next reader doesn't build on the wrong cascade model?

template-shell/tailwind.config.ts
```diff
@@ -6,9 +6,12 @@ export default {
   content: [
     './index.html',
     // Host app chrome (layout, menu, studio, components/ui). MFE packages
-    // under src-app/mfe_packages/* build their own CSS and are intentionally
-    // excluded here (scanning their node_modules/dist OOMs the host).
+    // render into the host's light DOM and ship no stylesheet of their own,
```
itechmeat: [Not a blocker] 🔵 Both halves of this sentence look wrong: MFEs mount into open shadow roots (mount() branches on `container instanceof ShadowRoot`, adoptHostStylesIntoShadowRoot delivers the styles), and they do ship stylesheets - demo-mfe alone has four *.module.css. The real reason the scan is needed: MFE builds compile no Tailwind of their own and pick utilities up from the adopted host sheet. The conclusion is right - worth stating the actual reason?

template-shell/packages/studio/src/uikit/base/button.tsx
```diff
@@ -12,13 +12,13 @@ const buttonVariants = cva(
     variants: {
       variant: {
         [ButtonVariant.Default]:
-          'bg-primary text-primary-foreground shadow hover:bg-primary/90',
+          'bg-primary text-primary-foreground shadow hover:bg-[color-mix(in_oklab,var(--primary)_90%,transparent)]',
```
itechmeat, line 15: [Not a blocker] 🔵 This PR adds --primary-hover/--card-hover to every theme and then doesn't consume them anywhere in the shell (only demo-mfe reads --primary-hover) - and the local recipes mix toward transparent while the theme tokens mix toward --background/--foreground, so two hover vocabularies land in one PR, plus the hand-escaped mirrors of these class strings in studioStyles.ts that must stay textually in sync. Fine as a faithful port of the old /0.9 alpha behavior; converging on the tokens later would be tidier - follow-up?

template-shell/src-app/app/themes/utils.ts
```diff
@@ -1,10 +1,12 @@
...
+ * Theme tokens are full CSS colors (e.g. `hsl(0 0% 9%)`), the same format
+ * @gears-frontx/ui-kit uses, so one token vocabulary drives both the shell's
+ * Tailwind utilities (`var(--primary)`) and the kit's component CSS.
  */
-export function hslToVar(color: string): string {
-  if (color === 'transparent') return 'transparent';
-  if (color.startsWith('hsl(')) return color.slice(4, -1);
+// Intentionally identity: kept as a semantic marker distinguishing color
+// tokens from other literals since the full-color-token migration.
+export function cssColor(color: string): string {
```
itechmeat, line 10: [Not a blocker] ⚪ It returns its argument - 25 call sites per theme file for an identity function kept as a "semantic marker" (the comment does say it's deliberate, so this is a question, not a demand): is the marker worth the indirection, or can the values just be values?

template-shell/src/build/mf-gts.ts
```diff
@@ -577,6 +577,107 @@ class StandaloneEsmBuilder {
...
+ * Rollup may hoist the extracted CSS into a chunk shared with other exposes
+ * or the standalone entry; the federation manifest then reports
+ * `css: { async: [], sync: [] }` for the expose, the host's stylesheet
+ * injection has nothing to inject, and the screen renders unstyled with no
+ * error anywhere. This turns that silence into a build failure.
+ *
+ * For each expose whose manifest attributes no CSS at all: walk the chunk
+ * graph reachable from its declared JS assets (static and dynamic imports)
+ * and collect the package-own CSS modules found there. An expose that does
+ * declare CSS is trusted — the guard targets the empty-attribution hoist,
```
itechmeat, line 622: [Not a blocker] ⚪ A few points worth queuing, none gating: 1) the guard trusts any expose that declares ANY css, so a second stylesheet hoisted into a shared chunk still ships half-styled and silent - the docblock owns that scope choice, but an issue would keep it visible; 2) capturedChunks is never cleared and `vite build --watch` reuses the plugin instance, so chunks from previous rebuilds accumulate - clear it in buildStart; 3) demo-mfe/src/lifecycle.test.tsx:122 still carries the pre-migration "HSL triplets" rationale this PR rewrote in KitThemedLifecycle.ts and _blank-mfe/lifecycle.tsx - missed sibling; and now that the shell stamps data-theme and carries theme.css, is the per-host token re-anchor in KitThemedLifecycle still needed at all?; 4) token-format-check matches comments too - the first constraint comment quoting the banned hsl(var()) pattern in a .ts file reds the build; 5) kit CSS ordered after globals.css also wins specificity ties against Tailwind *utilities*, not just preflight - e.g. a kit component given className="px-8" by an MFE; 6) adoptHostStylesIntoShadowRoot reclones the whole (now much bigger) kit CSS per shadow root - adoptedStyleSheets would share one parsed sheet (pre-existing, up to you); 7) two different definitions of "manifest-only template" land in this PR (any nested package.json in main.yml's overlay probe vs root-only in version-bump-on-change-check.mjs) - two answers to one question, both commented, still worth unifying; 8) design-verify's a11y.test.tsx imports @testing-library/react, react and the kit with none declared (works on hoisting only), tsconfig's include omits runner/ while runner.test.ts imports five modules from there, and "types": ["vite/client"] has no vite behind it.

itechmeat (approved):
Approving. The CSS-delivery work here is the strongest part of the PR: the new mf-gts guard turns the silent unstyled-screen failure into a build error, the Tailwind scan + dead-spacing checker close the lost-margins bug from both ends, and the cascade mechanics the comments describe mostly hold up under adversarial reading. My asks are all addressable in follow-ups or small pushes: runner/cli.mjs carries a raw NUL byte, so GitHub shows no diff for the largest runner file - I reviewed it as-is (it's solid, notes inline), but please make it diffable; a few load-bearing comments in the CSS pipeline state mechanisms that aren't the ones actually in play; and the optional `appearance` field quietly reopens the mixed-palette case for consumer themes. Details inline, everything marked [Not a blocker].

template-design-guardrails/src-app/verify_packages/design-verify/package.json
tscbmstubp: Fixed in fef88bac — the page key is now `JSON.stringify([route, theme])`, so cli.mjs is plain text again and the diff renders. The four robustness notes (per-state try, etc.) are queued as a follow-up rather than folded in here.

template-shell/packages/framework/src/types.ts
tscbmstubp: Fixed in fef88bac — apply() now stamps `appearance ?? 'light'`, so a consumer theme without the field never leaves data-theme unset; the type stays optional/backward-compatible. Comment on the field documents the default and the mixed-palette hole it closes.

template-design-guardrails/.frontx/ai/@gears-frontx/template-design-guardrails/reference-artifacts/design-contract.json
tscbmstubp: Swept in fef88bac — observedVersion/auditedKit now say 0.4.0-alpha.1; spacingScalePx is the kit's real `--space-*` scale [0,4,8,12,16,20,24,32] (layout-and-responsive.md preferred values updated to match); the Correct example uses the kit's actual Button/Input APIs (no `size` on Input, Button default/sm/lg); visual-foundations cites `var(--foreground)` / `text-foreground` instead of the invented token shape. Nothing keeps these honest automatically yet — a pin-drift-style check for observedVersion is a reasonable follow-up.

template-design-guardrails/.frontx/ai/@gears-frontx/template-design-guardrails/reference-artifacts/design-contract.json
tscbmstubp: Renamed in fef88bac — `guidelines/foundations.md`, id `foundations`, contract id `design-contract`; extension.json updated and the index regenerated (`policy:guideline-index` passes). Also dropped the two remaining 'Constructor' mentions in skill descriptions and source-registry so the bundle body is fully vendor-neutral.

template-shell/packages/react/src/mfe/ThemeAwareReactLifecycle.tsx
tscbmstubp: Reworded in fef88bac — the comment now says stability holds per kit version, names the template-pin-drift check as the actual guarantee, and spells out the silent-unstyled failure a drifted MFE would hit. Runtime version assert on mount left as a follow-up.

template-shell/src-app/app/globals.css
tscbmstubp: Reworded in fef88bac — the comment now describes the real mechanism: Tailwind 3.4 unwraps `@layer base` and hoists its contents to the `@tailwind base` position, so the winner is plain source order via main.tsx import order.

template-shell/tailwind.config.ts
tscbmstubp: Reworded in fef88bac — now states the actual reason: MFE builds compile no Tailwind of their own and pick utilities up from the host sheet adopted into their shadow roots.

template-shell/src-app/app/main.tsx
tscbmstubp: Agreed it's the pragmatic option today; kit-side `components.css` (CSS without the JS graph) queued as a follow-up.

template-shell/packages/studio/src/uikit/base/button.tsx
tscbmstubp: Agreed — faithful port for now; converging the shell hover recipes onto --primary-hover/--card-hover and dropping the hand-escaped studioStyles mirrors queued as a follow-up.

template-shell/src-app/app/themes/utils.ts
tscbmstubp: Kept as-is for this PR — the marker is what the token-format and pin-drift checks key off when they read theme files. Happy to drop it in a follow-up if the checks move to a different anchor.

template-shell/src/build/mf-gts.ts
tscbmstubp: Correction to my earlier reply: nothing reads `cssColor` — no policy script or test keys off it, it is purely the semantic marker the comment describes. Left as-is in this PR to keep the diff scoped; inlining the values across the five theme files is a clean mechanical follow-up and I lean toward doing it.

denyakortunov (changes requested):
I found three issues that should be addressed before merge:

### [P1] Regenerate the shell lockfile from the complete UI Kit dependency graph

The lock entry for `@gears-frontx/ui-kit@0.4.0-alpha.1` in `template-shell/package-lock.json` lists only `@base-ui/react` and `class-variance-authority`, although the installed package declares and imports additional runtime dependencies, including `@shadcn/react`.

In a fresh shell + guardrails composition outside the monorepo root, `npm install` leaves `@shadcn/react` absent and the design-verify a11y suite fails to resolve `@shadcn/react/message-scroller` from the UI Kit.

`Template Validate` stays green because its earlier root-level `npm ci` installs `@shadcn/react` into an ancestor `node_modules`, masking the standalone failure. Please regenerate `template-shell/package-lock.json` and make the composition check run without ancestor dependency leakage.

### [P2] Make the design-verify package's declared type-check pass and exercise it in CI

Running `npm run type-check --workspace=@gears-frontx/design-verify` in the composed tree currently fails with:

- TS7016 for the five imported `runner/*.mjs` modules;
- TS7006 for callbacks whose types are consequently lost;
- TS2550 for `Array.prototype.at`, because this package targets/includes only ES2020.

The shell's aggregate type-check currently discovers MFE packages but not `verify_packages`, so CI does not run this script. Please align the TS configuration/declarations with the runner tests and add verification workspaces to the aggregate type-check.

### [P2] Treat a requested screenshot failure as incomplete verification

In `runner/cli.mjs`, with `--screenshots`, a capture or file-write failure only sets `record.screenshotError`. `buildReport()` and `exitCode()` ignore that field, so the summary can still report every state as clean and the command can exit 0 even though a required screenshot was never produced.

The closing verification workflow requires every requested screenshot to exist and be reviewed. Please promote this to an environment failure, or count screenshot failures separately and return exit code 2 whenever screenshot capture was requested but failed.

tscbmstubp:
@denyakortunov all three addressed in 4d4d57f2 (branch squashed to one commit):

**[P1] Shell lockfile.** Confirmed: the `@gears-frontx/ui-kit@0.4.0-alpha.1` entry carried the 0.3.0-alpha.3 dependency block. Regenerated the subtree from the registry in a clean copy of `template-shell` outside the monorepo (re-adding the `file:.` self-link entry npm's reconciliation drops, #524). The lock now lists the kit's 11 runtime dependencies, and in that isolated tree `npm ci` passes, `@shadcn/react/message-scroller` resolves from the kit, and the composed design-verify suite passes (50 tests, a11y included). `Template Validate` no longer installs inside the checkout: both the standalone and the composition steps copy the template to `$RUNNER_TEMP` (without `node_modules`) before `npm ci` / `npm install`, so an ancestor `node_modules` can no longer supply what the lockfile doesn't.

**[P2] design-verify type-check.** `tsconfig.json` now includes `runner/` with `allowJs` (the runner tests see the JSDoc-typed `.mjs` modules), lib `ES2022` for `Array.prototype.at`, and `node` types; `npm run type-check --workspace=@gears-frontx/design-verify` passes in the composed tree. The shell's aggregate type-check (`run-mfe-type-checks.ts`) scans `src-app/verify_packages` alongside `src-app/mfe_packages` via a new `discoverTypeCheckProjects`, so CI runs it; three tests cover the merged discovery, including the missing-`type-check`-script refusal.

**[P2] Screenshot failures.** `buildReport()` counts `screenshotFailures`, excludes such states from `statesClean`, and `summaryLines()` prints a `SCREENSHOT <route> <theme> <width>px: <error>` line per state; `exitCode()` returns 2 whenever a requested screenshot failed, regardless of `--fail-on-findings`. Exit-code docs in `cli.mjs`, `--help`, and the verify-interface skill updated to match; test added.

GeraBart (approved):

template-design-guardrails/src-app/verify_packages/design-verify/runner/report.mjs
```diff
@@ -0,0 +1,185 @@
...
+    for (const finding of state.findings ?? []) {
+      const entry = {
+        route: state.route,
+        theme: state.theme,
+        width: state.width,
+        id: finding.id ?? 'unknown-rule',
+        target: finding.target,
+        detail: finding.detail,
+      };
+      map.set(findingKey(state, finding), entry);
```
denyakortunov: **[P1] Preserve duplicate findings in baseline comparisons**

`collectFindings()` stores findings in a `Map` keyed by route/theme/width/id/target. Multiple instances of the same defect with the same generated target overwrite each other. In a baseline with one instance and a current report with two, `diff.new` is empty and `--fail-on-findings` exits successfully. Please compare findings as a multiset, preserve counts, or include a stable occurrence identity.

template-design-guardrails/src-app/verify_packages/design-verify/runner/report.mjs
```diff
@@ -0,0 +1,185 @@
...
+  const after = collectFindings(current);
+  const added = [];
+  const resolved = [];
+  let carriedOver = 0;
+  for (const [key, entry] of after) {
+    if (before.has(key)) carriedOver += 1;
+    else added.push(entry);
+  }
+  for (const [key, entry] of before) {
+    if (!after.has(key)) resolved.push(entry);
```
denyakortunov: **[P1] Do not resolve findings in states that were not successfully rechecked**

A baseline finding is marked resolved whenever its key is absent from the current findings map. This also happens when the corresponding current state has an `environmentError` or was omitted by a narrower matrix. Only report a finding as resolved when the same route/theme/width state completed successfully and the finding was actually re-observed as absent; otherwise report it as unverified.

template-design-guardrails/.frontx/ai/@gears-frontx/template-design-guardrails/guidelines/visual-foundations-and-tokens.md
```diff
@@ -0,0 +1,84 @@
...
+- section title;
+- component title;
+- body;
+- label;
+- supporting metadata;
+- code/machine metadata.
+
+Minimums:
+
+- ordinary body and controls: 14 px minimum in compact density, 16 px minimum in comfortable density; at the narrow viewport tier use the 16 px floor regardless of density;
```
denyakortunov: **[P1] Align the typography contract with the installed type ramp**

These minimums cannot currently be satisfied while using the installed typography roles: the shipped ramp resolves to body 15 px, label 13 px, metadata 12 px, and mono 11 px, while this rule requires 16 px for body and controls at narrow widths. The runtime checker simultaneously flags only values below 10 px. In the composed template I observed 13 px action labels and 11 px metadata at 360 px, all reported as clean. Please reconcile the documented floors, the installed tokens, and the checker—ideally with role- and viewport-aware verification.

template-design-guardrails/src-app/verify_packages/design-verify/src/designDefects.ts
```diff
@@ -0,0 +1,683 @@
...
+    }
+  }
+  return findings;
+}
+
+function collectLayoutFindings(): DesignDefectFinding[] {
+  const findings: DesignDefectFinding[] = [];
+
+  const scroller = document.scrollingElement;
+  if (scroller && scroller.scrollWidth > scroller.clientWidth + OVERFLOW_TOLERANCE_PX) {
```
denyakortunov, line 608: **[P1] Detect unusable narrow layouts without horizontal overflow**

Horizontal overflow alone does not establish a usable responsive state. In the assembled template, shrinking the Current Theme screen to 360 px while navigation is expanded leaves the H1 and body copy only 8 CSS px wide, yet `scrollWidth` and `clientWidth` are both 360, so this check returns clean. Please either auto-collapse the auxiliary navigation in the shell or add a verification signal for pathological primary-content width/text measure so this state cannot pass the closing sweep.

template-design-guardrails/.frontx/ai/@gears-frontx/template-design-guardrails/guidelines/visual-foundations-and-tokens.md
```diff
@@ -0,0 +1,84 @@
...
+
+1. **Foundation** — raw color, dimension, font, radius, shadow and motion values.
+2. **Semantic** — text, surface, border, action, focus, status and layout roles.
+3. **Component** — only when a component requires a stable role not expressible by semantic tokens.
+
+Do not introduce component tokens simply to preserve accidental CSS values.
+
+## Spacing
+
+The spacing system (4 px base, 8 px primary rhythm, the 0–80 scale and its usage bands) is defined once, in the layout guideline's spatial system; the machine-readable values live in `layout-contract.json`. Apply it, do not restate it.
```
denyakortunov: **[P2] Remove the conflicting 0–80 spacing scale**

“0–80 scale” contradicts both `layout-contract.json`, whose complete scale is `[0, 4, 8, 12, 16, 20, 24, 32]`, and the layout guideline. This can lead an agent to introduce unsupported 40–80 px values despite the token-only rule. Please name the actual installed scale here or defer to the contract without restating a different range.

denyakortunov (changes requested):
The adoption is directionally strong, and the previously requested lockfile, type-check, and screenshot-failure fixes are present. I still found five issues that can make the guardrails report false confidence or give agents contradictory design instructions.

I also assembled the shell, MFE, and design-guardrails templates and reviewed the rendered wide and 360 px states. The main CI dependency-audit failure appears inherited from `develop`, so I am not attributing it to this PR, although the branch remains red.

template-design-guardrails/src-app/verify_packages/design-verify/runner/report.mjs
tscbmstubp: Fixed in 07c0c6e0. `collectFindings()` now builds a multiset (one entry per identity carrying `count`), and `diffReports()` compares counts: a higher count is `new` with `count` = the extra instances, a lower count is `resolved`/`unverified` by the difference, and `carriedOver` counts instances rather than keys. Summary lines show multiplicity (`NEW … card-missing-padding @ div.card (x2)`). Test: baseline with one instance vs run with two → 1 new, exit 1 under `--fail-on-findings`; two→one → 1 resolved.

template-design-guardrails/src-app/verify_packages/design-verify/runner/report.mjs
tscbmstubp: Fixed in 07c0c6e0. `diffReports()` now returns a third bucket, `unverified`: a baseline finding is only `resolved` when the same route/theme/width state is present in the current report **and** completed without `environmentError`; if the state errored or was not in the current matrix, the finding is reported as `UNVERIFIED <state>: <rule> @ <target>` and never as resolved. Test covers all three cases (re-observed clean, environment error, omitted state). Exit code is unchanged for unverified alone: the environment-error case already exits 2, and a deliberately narrower matrix is the caller's choice with the summary naming what was left unconfirmed.

template-design-guardrails/.frontx/ai/@gears-frontx/template-design-guardrails/guidelines/visual-foundations-and-tokens.md
tscbmstubp: Reconciled in 07c0c6e0, making the installed ramp the authority so the three sources cannot disagree again: (1) the guideline no longer states 14/16 px floors — it maps roles to the kit's `--text-*-size` tokens (body copy = body role; controls, actions, navigation and form labels = label role or larger; metadata = meta; nothing visible below mono) and adds "roles do not step down at narrow widths — meet the viewport with layout, never smaller text"; (2) the checker reads the floors from `:root` at sweep time instead of a hard-coded 10 px: `tiny-text` fires below `--text-mono-size` (11 px as shipped) for any visible text, and a new `small-control-text` rule fires below `--text-label-size` (13 px as shipped) for text inside a button, link, form control, label or control role — both fall back to the shipped values when the tokens are absent; (3) the verify-interface mapping names both rules and the tokens they read. Viewport-awareness comes from the matrix: the floors are enforced at every width swept, and the guideline forbids stepping roles down at the narrow tier. The 11 px metadata you observed at 360 px is the mono role and is now the documented floor; the 13 px action labels are the kit's label role, which the guideline now names as the floor for interactive text — if the product wants larger controls at narrow widths, that is a kit ramp decision the checker will follow automatically via the token.

template-design-guardrails/src-app/verify_packages/design-verify/src/designDefects.ts
tscbmstubp: Not addressed yet — this one has a scope decision in it. The detection rule (a `narrow-content` finding when the primary content region or a heading inside it renders below a width floor, so the 8 px column you saw cannot pass) fits this PR; auto-collapsing the auxiliary navigation at the narrow tier is a template-shell layout change with its own blast radius that I would take as a follow-up unless you want it here. Which do you want in this PR?

template-design-guardrails/.frontx/ai/@gears-frontx/template-design-guardrails/guidelines/visual-foundations-and-tokens.md
tscbmstubp: Fixed in 07c0c6e0: the spacing section now defers to `spacingScalePx` in `layout-contract.json` as the complete list of permitted steps ("and no other value is a spacing step") and no longer names a range.

denyakortunov (commented):
UI/UX review of the assembled shell and example MFEs at desktop, tablet, and 320–360 px widths in both themes. The existing unresolved narrow-layout navigation finding remains the main responsive blocker. I found three additional non-cosmetic accessibility and interaction gaps:

1. **[P1] Keep the active screen title as the single page H1** — `template-shell/src-app/app/layout/Header.tsx:38`

   In the assembled shell the accessibility tree exposes both `FrontX` and the screen title (`Current Theme`, `User Profile`, etc.) as level-1 headings. Before a screen is selected, `<main>` is still blank, so the hidden heading makes `page-has-heading-one` pass without giving the user a meaningful page state. Please keep the visible active-screen/empty-state title as the page `h1` and expose the product name as non-heading brand text (or a labelled banner). This also aligns the shell with the new “Use one page h1” guideline.

2. **[P1] Add programmatic current-state requirements for navigation** — `template-design-guardrails/.frontx/ai/@gears-frontx/template-design-guardrails/reference-artifacts/component-state-matrix.json:5`

   The matrix covers tabs, trees, and tables but not primary/local navigation. In the composed shell the selected “Current Theme” item has `data-active="true"` while both `aria-current` and `aria-pressed` are absent, so the current destination is conveyed only by the blue fill. Please add a navigation family/requirement (`aria-current="page"` or the pattern-appropriate selected state, plus selected + focus/hover checks) and keep it as an application obligation rather than assuming the kit guarantees it.

3. **[P1] Make the Studio panel usable without a pointer and within narrow viewports** — `template-shell/packages/studio/src/StudioPanel.tsx:60`

   At 360 CSS px the panel opens at its 400 px default width and is clamped to `x=20`, leaving 60 px off-screen; opening also unmounts the trigger and drops focus to `<body>`. Both move and resize are mouse-only (the resize handle is exposed as `role="button"`/`tabIndex=0` but has no keyboard action). This conflicts with the new core rules that all actions work without a pointer, dragging has a single-pointer alternative, and floating UI must not obscure focus. Please clamp size as well as position, move focus into the panel and return it on collapse, and provide keyboard/single-pointer move/resize controls.

I’m leaving this as a comment review rather than “Request changes” because the design-guardrails workflow is advisory in v0.1, but these should be addressed before treating the template as accessibility-complete.

PR #591, opened 2026-08-20
feat(ui-kit): full shadcn/ui port on Base UI + components browser demo

itechmeat:
## What

Full port of the shadcn/ui component set onto Base UI in `@gears-frontx/ui-kit` (63 components, CSS Modules + design tokens, no Tailwind and no Radix imports in kit source), plus the `demo/` browser sandbox that consumes the package strictly through its public `exports` as an acceptance check. All component docs are colocated (`<component>.md`) and aggregated for AI consumers via `llms.txt`.

## State

- 28 commits, rebased onto `develop`; all DCO-signed.
- Battle-tested against a real four-screen demo app built entirely from this kit; the fixes that came out of that exercise are part of the branch.
- Reviewed by two independent single-pass reviews (Codex / gpt-5.6-sol and opencode / glm-5.3): 17 confirmed findings fixed in the final wave (AlertDialog close behavior, Slider default thumbs, hydration-safe Sidebar, DatePicker month sync + strict typed-date parsing, Carousel prop composition and subscription cleanup, escaped Chart CSS injection, restored Recharts stock-stroke retint, DataTable pageSize sync, overridable UI strings, client-boundary docs guard, typeset docs, ADR-0022 wording).
- Test suite: 909 unit tests green, `test:consumer` green (vite, esbuild barrel/subpath, webpack, types), type-check includes the demo workspace.

## Notes for reviewers

- All 16 `@radix-ui/*` entries in the lockfile are transitives of `cmdk` (used by the Command component); the kit has no direct Radix imports and no unused dependencies. Replacing `cmdk` with a Base UI implementation is tracked as an open decision.
- `packages/ui-kit/demo` is not published to npm (`files` whitelist) and mirrors the `packages/telemetry/demo` precedent.
- ui-kit's Studio-kit artifact chain (`.cf-studio-kit.toml`) is stated as planned in ADR-0022; the registry ignore records that debt.


packages/ui-kit/src/components/marker/marker.tsx
```diff
@@ -0,0 +1,74 @@
...
+    },
+  },
+  defaultVariants: {
+    variant: 'default',
+  },
+});
+
+export interface MarkerProps extends useRender.ComponentProps<'div'>, VariantProps<typeof markerVariants> {}
+
+export function Marker({ className, variant = 'default', render, ...props }: MarkerProps) {
```
GeraBart, line 42: LOW: `variant = 'default'` is a default parameter, so it only applies on `undefined`. An explicit `variant={null}` (permitted by cva's own `VariantProps` type) leaves `data-variant` unset (React drops `null` attributes) while `markerVariants({ variant: null })` still resolves the `default` CSS class — a class/attribute mismatch. `item.tsx:58` already guards the equivalent case with `variant ?? 'default'`; consider the same guard here for consistency. Only reachable via an explicit `null`, not omission — non-blocking.

packages/ui-kit/src/components/native-select/native-select.tsx
```diff
@@ -0,0 +1,51 @@
...
+export type NativeSelectOptionProps = ComponentProps<'option'>;
+
+export function NativeSelectOption({ className, ...props }: NativeSelectOptionProps) {
+  return <option className={cx(styles.option, className)} {...props} />;
+}
+
+export type NativeSelectOptGroupProps = ComponentProps<'optgroup'>;
+
+export function NativeSelectOptGroup({ className, ...props }: NativeSelectOptGroupProps) {
+  return <optgroup className={cx(styles.option, className)} {...props} />;
```
GeraBart, line 50: LOW: `NativeSelectOptGroup` reuses `styles.option` (the leaf `<option>` class) instead of a dedicated optgroup class. Works today, but is a naming/intent mismatch that could misapply option-specific styling to optgroups later. Non-blocking nice-to-have.

GeraBart (approved):
Reviewed all 63 ported components across correctness, style, performance, test coverage, and security — solid work overall, no blocking issues. Two low-severity, non-blocking cosmetic nits left as inline comments. The `Template Validate` CI failure (unpublished `@gears-frontx/ui-kit@0.4.0-alpha.0`) is expected — this repo publishes packages automatically on merge.

packages/ui-kit/src/components/marker/marker.tsx
itechmeat: @GeraBart Fixed - `variant ?? 'default'` now guards the explicit-null case the same way `item.tsx` does (059f315e). New test covers `variant={null}`. https://github.com/constructorfabric/gears-frontx/pull/598

packages/ui-kit/src/components/native-select/native-select.tsx
itechmeat: @GeraBart Fixed - `NativeSelectOptGroup` now styles on its own `.optgroup` class instead of reusing `.option` (bed828c0). https://github.com/constructorfabric/gears-frontx/pull/598

Issue #590, opened 2026-08-19
[BUG] Actions chains emitted from a nested MFE don't reach shell-level targets


GeraBart:
It seems that in the current implementation of mfes, actions delivery works only on a singular parent-child level. If a child has an extensions domain, and thus becomes a parent for its own children, actions chains delivery mechanism doesn't seem to work across nesting levels.

PR #592, opened 2026-08-21
Fix: cross layers actions delivery

GeraBart:
## Summary by CodeRabbit

- **New Features**
  - Enabled nested microfrontends to communicate and dispatch actions across multiple hosting levels.
  - Added demo workflows to mount the Widgets Host, Widget A, and send a nested ping.
  - Preserved bridge state across remounts while distinguishing inactive from permanently removed extensions.
  - Added localized labels for the nested-widget workflow.

- **Bug Fixes**
  - Improved concurrent loading, lifecycle readiness, cleanup, and fallback handling.
  - Prevented stale registrations and in-flight actions after failed mounts or unmounts.
  - Isolated widget state across remounts and sibling instances.


template-mfe/src-app/mfe_packages/demo-mfe/src/screens/helloworld/i18n/ar.json
```diff
@@ -6,6 +6,7 @@
   "navigation_description": "انقر فوق الزر أدناه للانتقال إلى شاشة السمة برمجياً.",
   "go_to_theme": "الذهاب إلى شاشة السمة",
   "open_profile_refresh": "Open Profile & Refresh",
+  "mount_widgets_host_and_ping": "Mount Widgets Host + Widget A + Ping (nested)",
```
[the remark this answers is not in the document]
GeraBart: Not a defect — this matches an existing, established convention in this same file, not something new introduced by this PR. The sibling key `open_profile_refresh` (already in every one of these locale catalogs before this change) is *also* left as literal English in every non-English locale — this PR's `mount_widgets_host_and_ping` key was added the same way, for consistency with that precedent, not by oversight. Leaving as-is; a batch translation pass (if wanted) should cover both keys together as a separate, deliberate effort rather than diverging just this one new key from its sibling's established pattern.

packages/mfes/src/runtime/DefaultMfeRegistry.ts
```diff
@@ -200,6 +329,337 @@ export class DefaultMfeRegistry extends MfeRegistry {
...
+      // copy of this package (`inst-mint-escalation-on-link`).
+      escalate: (chain) => {
+        if (revoked) {
+          return Promise.reject(
+            new Error(`Inbound bridge link for '${extensionId}' has been revoked.`)
+          );
+        }
+        tagArrivalEdge(chain.action, childBridge);
+        return this.executeActionsChain(chain);
+      },
```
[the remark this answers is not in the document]
GeraBart: Confirmed as a real bug, fixed. Both the downward forwarding-entry delivery (`link.edge.onActionsChain`) and the upward `escalate` closure funneled into the public, failure-swallowing `executeActionsChain`, so `CrossHopRoute.send`'s promise never rejected on a genuine cross-hop failure — the sending hop's `fallback` never fired, exactly as described. Added a private `executeActionsChainOrThrow` (awaits the mediator's `ChainResult`, throws when `!completed`) and wired it into exactly these two cross-hop delivery points, leaving every other `executeActionsChain` call site (top-level entry points with no fallback of their own) untouched. Added two regression tests in `cross-nesting-reachability.test.ts`: `(f2)` for escalation-path failure, `(f3)` for forwarding-entry-path failure — both assert the sender's `fallback` actually fires. No public-surface change (private method); `dist/index.d.ts` diffed clean.

packages/mfes/src/mediator/actions-chains-mediator.ts
```diff
@@ -315,15 +358,60 @@ export class DefaultActionsChainsMediator extends ActionsChainsMediator {
...
+      timeout =
+        incomingBudget !== undefined
+          ? Math.max(0, incomingBudget - elapsed)
+          : Math.max(0, chainTimeout - elapsed);
+      const route = resolved;
+      invoke = () =>
+        new Promise<void>((resolve, reject) => {
+          const chain: ActionsChain = {
+            action: { ...action, timeout },
+          };
```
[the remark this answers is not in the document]
GeraBart: Investigated against ADR-0007 and the FEATURE's DoD for `inst-cross-hop-timeout` before changing anything. Concluded this is working as specified, not a defect:

- `Action` (`packages/mfes/src/types/index.ts`) is a real, publicly-exported, cross-realm-serialized shape — it is exactly what travels across `route.send()` to a possibly different independently-loaded copy of this package. Carrying the remaining budget out-of-band (e.g. a WeakMap keyed by action object, analogous to `tagArrivalEdge`) would not survive that trip: a WeakMap in one loaded copy is invisible to another copy that receives the forwarded/escalated action. So an out-of-band channel would need to be a *new field on the wire shape itself*, which the ADR/FEATURE's MFES-6 zero-public-surface-growth constraint (explicitly named for this exact mechanism) forecloses. Reusing `action.timeout` as the sole channel is the only spec-compliant option available, not an oversight.
- The FEATURE's own DoD text for this instruction is unconditional: on a cross-hop resolution, it directs discarding whatever the local per-action timeout resolution would have produced and using the decremented remaining chain budget instead — no carve-out for "this happens to be the first hop and the action happened to carry an explicit timeout."
- Practically, the elapsed time between chain dispatch and this first `executeAction` call is on the order of a few ms, so the described effect (an author's explicit first-hop timeout getting decremented before use) is bounded to a negligible amount, not a meaningful behavioral defect.

Left unchanged — no code or spec change made. Happy to revisit if there's a concrete scenario where this actually produces a wrong observable outcome, since the trade-off is real, but as designed it's intentional.

packages/mfes/src/runtime/default-mount-manager.ts
```diff
@@ -221,6 +359,30 @@ export class DefaultMountManager extends MountManager {
...
+          for (const relink of adopters) {
+            relink(null);
+          }
+        }
+        // Deliberately NOT deleted from the map — the next mount of this
+        // extension needs it to re-offer a fresh link to the same registry.
+        // @cpt-end:cpt-frontx-algo-mfe-host-communication-registration-propagation:p2:inst-unlink-on-retraction
+      }
+      // @cpt-end:cpt-frontx-algo-mfe-host-communication-registration-propagation:p2:inst-retract-advertisements
+
```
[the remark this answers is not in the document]
GeraBart: Investigated whether moving retraction/adopter-unlink after `lifecycle.unmount()` (in a `finally`) would actually be safer — concluded the current up-front ordering is intentional and correct, not a defect:

- Retraction only tears down the *inbound-bridge-link* (cross-hop escalation/forwarding routing through that specific bridge) — it does not dispose the `childBridge` object or its ability to dispatch actions locally. A mounted extension's own direct `bridge.executeActionsChain(...)` calls route through the bridge-factory's separate callback, unaffected by inbound-bridge-link retraction. So `unmount()` doing its own local cleanup dispatch is unaffected either way.
- The only realistic scenario retraction-after-unmount would help is an extension whose `unmount()` itself needs to *escalate* through a nested registry it hosts during its own teardown — a narrow, non-standard case.
- Moving retraction to run after `unmount()` would reopen exactly the collision risk the current ordering exists to prevent (a stale forwarding entry surviving into a subsequent remount), and would be strictly worse if `unmount()` hangs — a scenario this test suite already exercises (`ACTION_HANG`) — since the stale entry would then live for the full duration of the hang instead of being retracted immediately.

Left unchanged — the current ordering is deliberate and already documented in a code comment at the retraction call site.

packages/mfes/src/runtime/DefaultMfeRegistry.ts
```diff
@@ -200,6 +329,337 @@ export class DefaultMfeRegistry extends MfeRegistry {
...
+   * target actually re-propagates further.
+   */
+  private repropagateThroughInboundBridge(): void {
+    for (const [targetId, actionTypeIds] of this.advertisableTargets) {
+      this.propagateAdvertisementUpward(targetId, actionTypeIds);
+    }
+    for (const [targetId, entry] of this.forwardingEntries) {
+      this.propagateAdvertisementUpward(targetId, entry.actionTypeIds);
+    }
+  }
```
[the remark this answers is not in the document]
GeraBart: Addressed via the sibling finding on this same failure-swallowing path (line 454) — that one turned out to be a real bug (cross-hop delivery used the failure-swallowing public `executeActionsChain` instead of a throwing variant), now fixed with a private `executeActionsChainOrThrow`. The two new regression tests `(f2)`/`(f3)` in `cross-nesting-reachability.test.ts` assert the behavior that actually matters here — that the SENDER's `fallback` fires on a cross-hop failure — by dispatching entirely through the public registry API (`registry.executeActionsChain`), matching this test suite's existing convention of testing only through the public mount/dispatch path rather than reaching into the mediator directly. Since `DefaultMfeRegistry.executeActionsChain` intentionally returns `Promise<void>` (as you note) rather than exposing `ChainResult`, asserting a raw `ChainResult.completed === false` would require bypassing the registry and calling the mediator directly, which this suite deliberately avoids. Given the fallback-fired assertion already proves the fix at the level a real caller observes it, I'm treating a direct mediator-level assertion as optional and not adding it — but no objection if you'd still like one as a supplementary test.

template-mfe/src-app/mfe_packages/demo-mfe/src/lifecycle-widgets-host.tsx
```diff
@@ -195,24 +235,141 @@ async function bootstrapWidgetsRuntime(
...
+   * `mount_ext` dispatches for the same extension — one from this pass, one
+   * from a chain's own `next` continuation racing in right after `mount()`
+   * resolves — can trip "lifecycle not loaded" in `mountExtension`. Waiting
+   * for this pass to finish before unblocking `mount()` means any later
+   * `mount_ext` for an extension this pass already mounted lands on the
+   * cheap, safe `mountState === 'mounted'` early-return in `mountExtension`
+   * instead of racing a second real load/mount — without requiring any
+   * change to shared framework mount-concurrency handling.
+   */
+  readonly onDomainAttached: () => void;
```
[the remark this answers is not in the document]
GeraBart: Fixed — updated the doc comment on `onDomainAttached` to describe the actual current state: `loadExtension` now safely shares its in-flight promise across concurrent callers, and the remaining risk is one step later — `mountExtension` itself isn't serialized past its `mountState === 'mounted'` early-return, so two concurrent `mountExtension` calls for the same never-mounted extension can both pass that check before either finishes.

template-mfe/src-app/mfe_packages/demo-mfe/src/lifecycle-widgets-host.tsx
```diff
@@ -335,12 +506,119 @@ function WidgetsHostScreen(): React.ReactElement {
...
+    // drive its own loading/error UI without re-triggering bootstrap.
+    this.bootstrapPromise = bootstrapWidgetsRuntime(this.widgetsApp);
+
+    let resolveDomainAttached!: () => void;
+    this.domainAttachedPromise = new Promise<void>((resolve) => {
+      resolveDomainAttached = resolve;
+    });
+    this.onDomainAttached = resolveDomainAttached;
+
+    super.mount(container, bridge, mountContext);
```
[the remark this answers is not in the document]
GeraBart: Checked the base class: `ThemeAwareReactLifecycle.mount()` (`template-shell/packages/react/src/mfe/ThemeAwareReactLifecycle.tsx`) is declared and implemented as returning `void`, not `Promise<void>` — there is no promise from `super.mount(...)` to capture. This is a false positive; left the call as-is.

template-mfe/src-app/mfe_packages/demo-mfe/src/lifecycle-widgets-host.tsx
```diff
@@ -335,12 +506,119 @@ function WidgetsHostScreen(): React.ReactElement {
...
+    // closes BOTH races: `registerDomain` making the domain routable
+    // (`bootstrapPromise`), and `DefaultExtensionMounter.attach(root)` giving
+    // it somewhere to actually mount into (`domainAttachedPromise`) — see
+    // each field's doc comment. If `bootstrapPromise` rejects,
+    // `Promise.all` rejects immediately without waiting on
+    // `domainAttachedPromise` (which would otherwise never resolve, since
+    // `WidgetsHostScreen` never renders `ExtensionDomainSlot` on the error
+    // path).
+    await Promise.all([this.bootstrapPromise, this.domainAttachedPromise]);
   }
```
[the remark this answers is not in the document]
GeraBart: Fixed — added a 15s timeout racing each of `bootstrapPromise` and `domainAttachedPromise` independently (not one shared timeout), so a stall is reported against the specific step that stalled (manifest bootstrap vs. domain-slot attach) with a clear, actionable error message, rather than hanging `mount()` — and therefore the whole calling chain, per this PR's own mount-completion contract — indefinitely.

template-mfe/src-app/mfe_packages/widgets-fixture-a/src/lifecycle.tsx
```diff
@@ -44,9 +53,14 @@ function generateRandomHex(): string {
...
-// by the ping handler when invoked.
-let lastPingObserver: ((value: string) => void) | null = null;
+// Per-mount last-ping value. The handler always writes here, independent of
+// whether a subscriber is wired yet, so registering the handler synchronously
+// in `mount()` (required so a chained `next` step can reach it as soon as
+// `mount()` resolves) can never race the DOM's observation of a ping: a ping
+// that lands before `WidgetA` subscribes is still visible the moment it does,
+// via the initial-state read below, instead of depending on ordering.
+let lastPingValue: string | null = null;
+let lastPingSubscriber: ((value: string) => void) | null = null;
```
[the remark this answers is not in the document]
GeraBart: Confirmed real, fixed. Alpha/beta are already isolated from each other via separate module evaluations (per this file's own documented blob-URL-isolation invariant), but that invariant doesn't protect a genuine remount of the SAME extension: `DefaultMountManager` reuses the already-loaded module (`loadState === 'loaded'` skips reloading), so a shared module-level value could leak a previous mount's last-ping into a fresh mount's first render, and one mount's cleanup could clear a live sibling's subscriber. Applied your proposed fix essentially as-is: `lastPingValue`/`lastPingSubscriber` replaced with `Map<string, ...>` keyed by `bridge.instanceId`, with the initial-state read and the effect's re-sync/subscribe/cleanup all scoped to that instance's own entry (including deleting the instance's entry, not just its subscriber, in cleanup, and guarding the subscriber removal against a stale closure clearing a different live mount's subscriber for the same id).

template-mfe/src-app/mfe_packages/widgets-fixture-a/src/lifecycle.tsx
```diff
@@ -72,15 +93,48 @@ interface WidgetAProps {
...
+    lastPingSubscribers.set(instanceId, setLastPing);
     return () => {
-      lastPingObserver = null;
+      // Only remove this instance's own subscriber -- guard against a stale
+      // closure clearing a different, still-live mount's subscriber for the
+      // same instanceId (e.g. React re-invoking effects in development).
+      if (lastPingSubscribers.get(instanceId) === setLastPing) {
+        lastPingSubscribers.delete(instanceId);
+      }
+      lastPingValues.delete(instanceId);
```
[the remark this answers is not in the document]
GeraBart: Investigated — not reachable in this codebase as it currently stands. `WidgetsFixtureALifecycle.mount()` creates its own React root via a bare `createRoot(container)` call inside `ThemeAwareReactLifecycle.mount()`, with no `<StrictMode>` wrapper anywhere in that call chain — `<StrictMode>` only wraps the shell's own separate top-level root (`template-shell/src-app/app/main.tsx`), and React 18+ `createRoot()` roots are independent: Strict Mode's double-invoke-effects behavior doesn't propagate from an ancestor DOM tree into a nested, independently-created root. So the synchronous double-invoke-effects sequence this finding depends on never actually happens for this component.

That said, the underlying inconsistency you spotted is real and worth keeping in mind: the effect's `lastPingValues.delete(instanceId)` cleanup is unconditional, while the adjacent `lastPingSubscribers` cleanup right next to it is correctly guarded against a stale-closure clobber. Leaving as-is since it's not currently reachable, but flagging it as a defensive-hardening candidate if this component's root is ever wrapped in Strict Mode in the future.

packages/mfes/__tests__/registration-propagation/cross-nesting-reachability.test.ts
```diff
@@ -0,0 +1,1121 @@
...
+    const failureLogged = errorSpy.mock.calls.some((call: unknown[]) =>
+      call.some((arg: unknown) =>
+        String(arg).includes('Actions chain failed') ||
+        String(arg).includes('No handler found') ||
+        String(arg).includes('BRIDGE_INACTIVE') ||
+        String(arg).includes('inactive')
+      )
+    );
+    expect(failureLogged).toBe(true);
+    expect(leafCounter.count).toBe(0);
```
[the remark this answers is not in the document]
GeraBart: Confirmed as a genuine test-quality gap and fixed — thanks for catching this. The permissive `'No handler found'` branch could mask a real regression (silent retraction-on-mount-failure) as a false pass. One correction to the suggested fix text: the literal string `'BRIDGE_INACTIVE'` never appears in the logged message (only lowercase `inactive` does — `BRIDGE_INACTIVE` is only the error's `.code` property, never interpolated into the message), so I mirrored the sibling test `(h)`'s already-correct two-assertion pattern instead: assert `inactive`/`BRIDGE_INACTIVE` IS logged, and separately assert `No handler found` is NOT logged.

packages/mfes/architecture/features/mfe-host-communication/FEATURE.md
```diff
@@ -116,46 +118,89 @@ This feature details the host–MFE dispatch mechanism and the child-facing brid
...
+   1. [x] - `p1` - Child bridge invokes the injected mediator-registration callback with the action type identifier and handler instance, keyed to the extension's own GTS identifier; the registration survives the bridge's deactivation and is released only at that extension's permanent unregistration. The bridge wraps the registered handler in an internal activity gate, so an invocation arriving while the bridge is inactive never reaches the handler - `inst-fwd-reg-handler`
 3. [ ] - `p1` - **IF** the child registers a child domain for cross-runtime action forwarding - `inst-child-reg-domain`
    1. [x] - `p1` - Child bridge invokes the injected child-domain-registration callback with the domain identifier - `inst-fwd-reg-domain`
    2. [x] - `p1` - Parent runtime registers a catch-all forwarding handler in the mediator, keyed to the child domain identifier - `inst-register-catchall`
    3. [x] - `p1` - The catch-all forwarding handler wraps any incoming action in an actions chain and delivers it through the bridge transport to the child runtime's mediator - `inst-catchall-forward`
```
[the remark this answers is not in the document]
GeraBart: Investigated — the specific claim doesn't hold: `registerChildDomain`/`unregisterChildDomain` are real, public, exported methods on the concrete `ChildMfeBridgeImpl` (not the abstract `ChildMfeBridge` base class), exercised directly in `packages/mfes/__tests__/bridge/cross-runtime-routing.test.ts` — step 3 accurately describes a real, deliberate, tested feature and shouldn't be removed.

You did correctly sense a real inconsistency, though — it's on the other side: step 6 (and the matching DoD sentence) claimed the child-facing bridge surface is 'exactly four methods' without ever clarifying that's scoped to the ABSTRACT `ChildMfeBridge` class specifically (the type the host actually hands to `mount()`), not the concrete implementation's full public surface, which additionally carries `registerChildDomain`/`unregisterChildDomain` for callers who narrow to the concrete type. Fixed by rescoping that wording (this file's step 6 + DoD, plus a matching clarification in ADR-0008) rather than touching step 3.

packages/mfes/src/handler/types.ts
```diff
@@ -35,8 +38,10 @@ export abstract class ParentMfeBridge {
  * Provided to child MFEs for communication with the host.
  */
 export abstract class ChildMfeBridge {
-  abstract readonly domainId: string;
-  abstract readonly instanceId: string;
+  /** The GTS id of the domain the extension is mounted into. */
+  abstract readonly extDomainId: string;
+  /** The extension's own GTS id. */
+  abstract readonly extensionId: string;
```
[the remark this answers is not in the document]
GeraBart: Confirmed real — thanks for catching this. 4 test files still had `ChildMfeBridge`-shaped mocks using the old `domainId`/`instanceId` names (`template-shell/packages/react`'s `hooks.test.tsx`, `queryHooks.helpers.tsx`, `queryClientContext.test.ts`, and `template-mfe`'s `HomeScreen.test.tsx`), all now renamed to `extDomainId`/`extensionId`.

Also root-caused why this was invisible to CI: `template-shell/packages/react/tsconfig.test.json` was extending the shell app's root tsconfig (which has no `jsx` option) instead of its own sibling config, so `type-check:test` for this workspace failed wholesale with unrelated JSX noise before ever reaching these real errors — fixed the `extends` target too.

Not adding a migration note per your suggestion — this repo is pre-1.0 with no established migration-doc convention (no MIGRATION.md anywhere, CHANGELOG.md isn't maintained per-change); breaking changes here are tracked via the required version-bump policy plus the living ADR text, which already documents the current shape.

scripts/pin-template-ecosystem-to-local.mjs
```diff
@@ -0,0 +1,275 @@
...
+  const manifestEdits = [];
+  for (const [relFile, subs] of byFile) {
+    const manifestPath = path.join(templateDir, relFile);
+    const manifestDir = path.dirname(manifestPath);
+
+    const edits = subs.map((sub) => {
+      const localAbsDir = path.join(repoRoot, 'packages', sub.localDir);
+      const fileSpec = `file:${toPosixRelative(path.relative(manifestDir, localAbsDir))}`;
+      return { field: sub.field, packageName: sub.packageName, pinnedVersion: sub.pinnedVersion, fileSpec };
+    });
```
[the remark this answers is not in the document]
GeraBart: Confirmed real — thanks for catching this. Fixed by adding a narrow allowlist local to this one script (`INSTALLED_DEPENDENCY_FIELDS = ['dependencies', 'devDependencies', 'optionalDependencies']`), rather than editing the shared `DEPENDENCY_FIELDS` constant — that constant is used by two other scripts (`template-pin-drift-check.mjs`, `version-bump-on-change-check.mjs`) which legitimately DO care about `peerDependencies` pins for their own purposes, and it's byte-equality-tested against `packages/cli/src/manifest/validate-content-self-containment.ts`, so narrowing it would have broken that contract. This mirrors the existing `INSTALLED_DEPENDENCY_FIELDS` precedent already in `template-lockfile-selflink-check.mjs` for the same reason. Also confirmed this was a latent/defensive fix, not an active break — no template package.json today pins an ecosystem package as an exact-version peer dependency (they're all ranges like `*`/`>=0.2.0-0`, which the pin-site scanner already ignores). Added regression tests (a peer-only pin fails closed with no substitution, and a peer pin alongside a real dependency pin is left untouched while the dependency is correctly localized).

scripts/pin-template-ecosystem-to-local.mjs
```diff
@@ -0,0 +1,275 @@
...
+  const logLines = [];
+
+  for (const { manifestPath, relFile, edits, overrides } of manifestEdits) {
+    const manifest = JSON.parse(fs.readFileSync(manifestPath, 'utf8'));
+    const reportedFile = path.join(path.relative(repoRoot, path.dirname(manifestPath)), relFile);
+
+    for (const { field, packageName, pinnedVersion, fileSpec } of edits) {
+      manifest[field][packageName] = fileSpec;
+      logLines.push(`${packageName} ${pinnedVersion} -> ${fileSpec} (${reportedFile} / ${field})`);
+    }
```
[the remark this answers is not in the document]
GeraBart: Confirmed and fixed exactly as proposed — thanks. `manifestPath` is already `<repoRoot>/<templateDirName>/<relFile>`, so `path.relative(repoRoot, manifestPath)` alone gives the correct path for both the root and nested cases; the root case only happened to look right before by coincidence. Dropped `relFile` from this function's destructuring since it's unused here now (kept on the `ManifestEdit` type and in `planPinLocalization`, where it's still needed for the root-manifest `overrides` gate). Also added the test coverage you flagged as missing — a nested-manifest case asserting the log line reports the path exactly once, verified to fail against the pre-fix code with the exact duplicated-segment string you described.

packages/mfes/src/runtime/inbound-bridge-link.ts
```diff
@@ -0,0 +1,317 @@
...
+   */
+  escalate(chain: ActionsChain): Promise<void>;
+}
+
+interface LinkCarryingBridge {
+  [LINK_PROPERTY_KEY]?: InboundBridgeLink;
+}
+
+/**
+ * @cpt-begin:cpt-frontx-algo-mfe-host-communication-registration-propagation:p2:inst-adopt-ambient-bridge
```
itechmeat: 🟠 [Not a blocker] The `inst-adopt-ambient-bridge` region's `@cpt-end` sits at line 244, inside the doc block of `adoptAmbientInboundBridgeLink` (declared at 255) - so the marked region covers `registerInboundBridgeLink`/`unregisterInboundBridgeLink` (a different concern) and stops before the body of the function that actually realizes the instruction. The pair validates green, but the region attests to the wrong code - worth re-scoping while the file is fresh.

packages/mfes/src/runtime/inbound-bridge-link.ts
```diff
@@ -0,0 +1,317 @@
...
+   * Identity token for loop containment: the same `ChildMfeBridge` instance
+   * used both as this registry's inbound bridge and, from the parent's side,
+   * as the key recorded alongside any forwarding entry created for a target
+   * this registry (or one of its descendants) advertised. Compared by
+   * reference, never invoked directly.
+   */
+  readonly edge: ChildMfeBridge;
+
+  /**
+   * @cpt inst-propagate-upward / inst-collision-check / inst-collision-reject / inst-record-forwarding-entry / inst-repropagate-upward
```
itechmeat: 🟡 [Not a blocker] This `@cpt inst-...` form (space, bare instruction names, no owning spec id) matches none of the traceability tooling - every regex keys off the `@cpt-` hyphen, so these 10 new sites (mediator 102/114, DefaultMfeRegistry 412/559/607/635, inbound-bridge-link 171/179/187/245) read as traceability but are inert comments. The form exists nowhere at merge-base - worth settling now: either plain prose or real `@cpt-begin/@cpt-end` pairs, before it spreads.

packages/mfes/__tests__/registration-propagation/cross-nesting-reachability.test.ts
```diff
@@ -0,0 +1,1120 @@
...
+    });
+    registry0.registerDomain(makeDomain(D0), new GenericDomainFactory());
+
+    await registry0.registerExtension(makeExtension(REVOKE_EXT, D0, REVOKE_ENTRY));
+    const mounter0 = registry0.getMounter(D0);
+    mounter0.attach(document.createElement('div'));
+    await mounter0.mount(REVOKE_EXT, document.createElement('div'));
+
+    expect(capturedBridge).toBeDefined();
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
itechmeat: 🟡 [Not a blocker] The only two lint suppressions in ~2700 lines of new test code (here and at 942-943), and both have a mechanical fix: the production code reads the same symbol property via `(bridge as unknown as LinkCarryingBridge)[LINK_PROPERTY_KEY]`, so the same typed cast works in the test without `any`.

packages/mfes/src/handler/mfe-bridge-factory-default.ts
```diff
@@ -20,6 +20,6 @@ export class MfeBridgeFactoryDefault extends MfeBridgeFactory<ChildMfeBridgeImpl
   }
 
   dispose(bridge: ChildMfeBridgeImpl): void {
-    bridge.cleanup();
+    bridge.destroy();
```
itechmeat, line 23: 🟡 [Not a blocker] The rename touched this factory's teardown half, but `create()` (line 18, unchanged) still does a bare `new ChildMfeBridgeImpl(...)` - and `active` now starts `false` with only `DefaultRuntimeBridgeFactory.acquireBridge` ever calling `activate()`, so anything built through this factory rejects every chain with `BRIDGE_INACTIVE` forever. Nothing in `src/` calls `bridgeFactory.create`, so this looks like dead public surface - if so, deciding its fate (drop or activate) beats renaming half of it. Same for its docs: `MfeBridgeFactory.create`'s `instanceId` param doc in handler/types.ts still says "Unique instance ID" while the argument lands in what is now `extensionId`.

template-mfe/src-app/mfe_packages/demo-mfe/src/lifecycle-widgets-host.tsx
```diff
@@ -195,24 +235,141 @@ async function bootstrapWidgetsRuntime(
...
+   * itself only happens after `bootstrap` resolves), and its root-attach
+   * effect runs on a LATER React commit than the microtask that resolves
+   * `mount()`'s own promise and immediately drives the chain's `next` node.
+   * Awaiting this signal too — not just `bootstrap` — is what closes that
+   * second race.
+   *
+   * This component deliberately delays calling it until AFTER its own
+   * auto-mount-on-attach pass (below) has settled for every extension
+   * currently registered on this domain — not the instant `attach()`
+   * returns. `DefaultMountManager.loadExtension` does not correctly await a
```
itechmeat: 🟡 [Not a blocker] This paragraph says `loadExtension` "returns immediately once loadState === 'loading' rather than awaiting the in-flight load" - but default-mount-manager.ts:150-158 in this same PR now awaits `inFlightLoadsByExtension`. The wait itself still earns its keep (it closes the separate attach-on-a-later-commit race), but the justification documents a bug this PR fixed - the text is stale against its own diff.

packages/mfes/src/runtime/DefaultExtensionMounter.ts
```diff
@@ -63,13 +71,38 @@ export class DefaultExtensionMounter extends ExtensionMounter {
       );
     }
 
-    await this.mountManager.mountExtension(extensionId, container);
+    const inFlight = this.inFlightMountsByExtension.get(extensionId);
+    if (inFlight) {
+      return inFlight;
```
itechmeat: 🟡 [Not a blocker] The in-flight dedupe here is keyed on extension id only, while `mount(extensionId, container)` takes a container - a second concurrent caller with a different container gets the first caller's promise, resolves successfully, and its container is never appended: a silent no-op that looks like success. Is a second container for the same extension impossible by construction (strategies own container creation), or should this case reject?

packages/mfes/src/runtime/DefaultMfeRegistry.ts
```diff
@@ -200,6 +340,340 @@ export class DefaultMfeRegistry extends MfeRegistry {
...
+   * registry's inbound bridge. `undefined` when this registry holds no
+   * inbound bridge (it is the shell). Arrival-edge tagging is NOT done here
+   * — it happens inside `link.escalate` itself, minted by the PARENT
+   * registry at link time (`buildInboundBridgeLinkFor`), so that the tag is
+   * written and later read by the same (parent) copy of this package
+   * regardless of which copy this (child) registry belongs to.
+   *
+   * @cpt-begin:cpt-frontx-algo-mfe-host-communication-mediator-dispatch:p1:inst-escalation-lookup
+   */
+  private resolveEscalationRoute(_targetId: string): CrossHopRoute | undefined {
```
itechmeat: 🟡 [Not a blocker] `_targetId` is ignored, so tier-5 escalation fires for every unresolved dispatch in any non-root registry - a typo'd target in a nested MFE now crosses every bridge up to the shell before failing, and comes back as a generic cross-hop error instead of `NoHandlerForActionTargetError`. Loop containment holds (traced the arrival-edge exclusion end to end), so this is cost and diagnosability, not correctness - is unconditional escalation deliberate because the child cannot know what an ancestor holds? A one-line confirmation closes this.

template-mfe/src-app/mfe_packages/demo-mfe/src/lifecycle-widgets-host.tsx
```diff
@@ -122,6 +127,41 @@ function createWidgetsHostApp(): ReturnType<ReturnType<typeof createFrontX>['bui
...
+ * that adopts no inbound bridge — degrading every future mount of this
+ * extension to root-registry behavior for good, regardless of any fix to
+ * WHERE the nested registry construction happens relative to `mount()`.
+ * This placeholder never touches `microfrontends()`, so it never calls
+ * `mfeRegistryFactory.build()` — leaving that one slot free for
+ * `DemoMfeWidgetsHostLifecycle.mount()`'s own, later, synchronous call to
+ * the real `createWidgetsHostApp()` to win it from inside the rendezvous
+ * window instead.
+ */
+function createWidgetsHostAppShell(): ReturnType<ReturnType<typeof createFrontX>['build']> {
```
itechmeat, line 157: 🟡 [Not a blocker] Every nested host now has to construct a `microfrontends()`-free decoy app in its constructor, purely so the module-level `mfeRegistryFactory` singleton's one build slot survives until `mount()`. That is not a demo detail - it is a framework constraint every consumer hits the first time they nest, and the demo currently teaches the workaround. Fixing the singleton here would widen the blast radius into template territory, fair - but then the constraint deserves a place in the framework docs and a tracked follow-up, not a 25-line comment in one demo file. Is one planned?

packages/mfes/src/runtime/DefaultMfeRegistry.ts
```diff
@@ -567,6 +1050,33 @@ export class DefaultMfeRegistry extends MfeRegistry {
...
+   * to run (e.g. a mounted extension's own bridge dispatch, or the
+   * lifecycle-stage action runner). Cross-hop delivery is different: the
+   * SENDER already has a `fallback` (or lack of one) it needs to observe the
+   * real outcome to act on, so swallowing the failure here would silently
+   * turn a failed cross-hop dispatch into an apparent success at the sender.
+   */
+  private async executeActionsChainOrThrow(chain: ActionsChain): Promise<void> {
+    const result = await this.mediator.executeActionsChain(chain);
+    if (!result.completed) {
+      throw new Error(
```
itechmeat: 🔵 [Not a blocker] The PR adds `BridgeInactiveError` with `code = 'BRIDGE_INACTIVE'` and exports it from the barrel, but every hop rebuilds the failure as a plain `Error` from `result.error` (a string; same shape at 421, 459, 582, 1277) - so a consumer one nesting level up can never `instanceof` it or read `.code`, only string-match the message. That is also the structural reason the new tests assert on message text. `ChainResult.error` being a string is the pre-existing contract, understood - but the two link-lifecycle rejections are exactly the ones a consumer's fallback logic will want to branch on. Worth a follow-up thought on carrying the code across the hop?

packages/mfes/src/mediator/actions-chains-mediator.ts
```diff
@@ -14,13 +14,20 @@
...
 import {
   ActionsChainsMediator,
   ActionHandler,
   type ChainResult,
   type ChainExecutionOptions,
 } from './types';
 
+/** Narrows a resolved handler to the cross-hop (forwarding-entry/escalation) shape. */
+function isCrossHopRoute(resolved: ActionHandler | CrossHopRoute): resolved is CrossHopRoute {
+  return typeof (resolved as CrossHopRoute).send === 'function';
```
itechmeat: 🔵 [Not a blocker] Narrowing on `typeof resolved.send === 'function'` - `send` is a name an application ActionHandler subclass could plausibly own, and such a handler would silently take the cross-hop branch and then explode on `registerInFlight`. A symbol brand (or a name nobody would collide with) is cheaper than that debugging session, and `'send' in resolved` would also drop the bare cast. Routes are only minted inside the registry today, so the odds are low - just cheap to close.

packages/mfes/__tests__/registration-propagation/cross-nesting-reachability.test.ts
```diff
@@ -0,0 +1,1120 @@
...
+const MOUNT_EXT = 'mock.action.v1~mount_ext.v1~';
+const UNMOUNT_EXT = 'mock.action.v1~unmount_ext.v1~';
+const ACTION_ROOT = 'mock.action.v1~action_root.v1~';
+const ACTION_LEAF = 'mock.action.v1~action_leaf.v1~';
+const ACTION_HANG = 'mock.action.v1~action_hang.v1~';
+const ACTION_COLLIDE = 'mock.action.v1~action_collide.v1~';
+const ACTION_UNRESOLVABLE = 'mock.action.v1~action_unresolvable.v1~';
+const ACTION_ASYNC = 'mock.action.v1~action_async.v1~';
+
+function createMockPlugin(entries: Map<string, MfeEntry>): TypeSystemPlugin {
```
itechmeat, line 64: 🔵 [Not a blocker] `createMockPlugin`, `makeDomain`, `makeEntry`, `makeExtension`, `actionChain`, `NoopHooks`, `GenericDomainImpl`, `GenericDomainFactory` are copy-pasted near-identically into all four new test files - roughly 150 lines x4 of the 2700 added, and the main reason each file tops 500 lines. `cross-copy-boundary` genuinely needs its own copies (its harness instantiates against two independently loaded module copies), but the other three could share one `__tests__/fixtures/` module. Fine as a follow-up.

.github/workflows/template-drift.yml
```diff
@@ -80,11 +80,70 @@ jobs:
     - name: Build workspace packages
       run: npm run build:packages
 
-    # `dev:template:link` replaces installed directories inside the template's
-    # node_modules — the pinned tree has to exist before it can be repointed.
-    - name: Install template from its lockfile
+    # `npm ci` here cannot install a pin whose tarball is not published yet, and
```
itechmeat: 🔵 [Not a blocker] Lines 83-107 and 130-144 are a design document inside a yml - the npm/cli#524 analysis, the "41 of 546 packages resolved differently" measurement, the version-vs-publish standoff. All worth keeping, but in CONTRIBUTING.md (which the comment already points at), with one line per flag here pointing back - the next person editing this workflow needs the three commands, not the essay. Same pattern elsewhere in the PR: ~60 comment lines for the 3 `WidgetsHostScreenProps` props and the 40-line design narrative opening inbound-bridge-link.ts - rationale belongs in DESIGN/ADR, code keeps the constraint plus a pointer.

.gitignore
```diff
@@ -19,6 +19,9 @@ public/shared
 .__mf__temp
 .mf
 
+# QA browser-verification screenshots (session evidence, not product content)
+.qa-evidence/
```
itechmeat: 🔵 [Not a blocker] Agent/QA session artefacts in the repo's ignore list - nothing in the repo produces this directory, and it is not part of this PR's purpose. Drop it, or if the QA flow is real, land it with the flow.

packages/mfes/src/runtime/default-mount-manager.ts
```diff
@@ -39,6 +46,57 @@ export class DefaultMountManager extends MountManager {
...
+   */
+  private readonly childBridgesByExtension = new Map<string, ChildMfeBridge>();
+
+  /**
+   * The re-link callback of whichever registry adopted the inbound-bridge
+   * link of the CURRENT (or most recent) mount of each extension, keyed by
+   * extension id — retained across unmount/remount cycles, not cleared on
+   * retraction, and deleted only by `releaseExtension`. Since the link is
+   * minted once at first mount and lives for the extension's whole
+   * registration lifetime (framing correction, §0), this map is populated
```
itechmeat: ⚪ [Not a blocker] "framing correction, §0" resolves to nothing in the repo (second site at line 311) - it is a note to whoever edited the diff, not to a reader of the system. The surrounding sentences already state the constraint, so the citation can simply go.

packages/mfes/architecture/features/mfe-host-communication/FEATURE.md
```diff
@@ -218,9 +265,22 @@ The system **MUST** provide a child bridge exposing exactly `executeActionsChain
...
 - [x] Chain execution follows the `next` continuation on success and the `fallback` continuation on failure, both within per-action and whole-chain timeout bounds
 - [x] In-flight action tracking prevents unregistration of a target's handlers while actions for that target are pending
 - [x] Action admission is delegated to the injected type-system provider; no type-format literals appear in the mediator
-- [x] The child bridge surface is exactly `executeActionsChain`, `subscribeToProperty`, `getProperty`, and `registerActionHandler`; the parent bridge surface is exactly `instanceId` and `dispose()`
+- [ ] The child bridge surface is exactly the four capability methods `executeActionsChain`, `subscribeToProperty`, `getProperty`, and `registerActionHandler` plus exactly two readonly identity properties, `extDomainId` (the GTS identifier of the domain the extension is mounted into) and `extensionId` (the extension's own GTS identifier); the parent bridge surface is exactly `instanceId` and `dispose()`, unchanged regardless of nesting depth
```
itechmeat: ⚪ [Not a blocker] This criterion was `[x]`; the only change adds the two identity properties, and cross-nesting-reachability.test.ts:390 ("(e) ...exactly 4 methods + 2 identity properties...") verifies it directly - yet the box flipped to `[ ]`, as are all 11 new criteria. If the checkbox means "confirmed", the artifact now understates what is delivered - what does the box mean here, and should the delivered ones be checked?

architecture/ADR/0007-action-dispatch-and-chaining.md
```diff
@@ -26,13 +26,14 @@ date: 2026-06-05
...
 
 ## Decision Drivers
 
 * Decoupled coordination — senders address a target by identifier, never by holding a reference to it (anchors `cpt-frontx-fr-mfe-host-communication`).
 * Deterministic routing — a `(targetId, actionTypeId)` pair resolves to at most one handler, with a defined fallback tier when no specific handler exists.
 * Cross-runtime forwarding — actions targeting a child runtime's domains must be forwardable even though the parent does not know the child's action vocabulary at registration time.
+* Reachability across nesting depth without probing — a target nested arbitrarily many runtime boundaries away must be reachable in a single addressed hop-by-hop path, not through fan-out to every descendant; this reinforces, for a reason not weighed at the time, the existing exclusion of the publish/subscribe alternative — so depth-spanning delivery must be achieved by carrying routing state along the mediator chain rather than by broadcasting.
```
itechmeat: ⚪ [Not a blocker] "for a reason not weighed at the time" narrates the ADR's own drafting history - the rest of both ADR rewrites is clean on this. A decision doc reads as if the current decision had always been the decision; the phrase deletes without loss.

packages/mfes/src/bridge/ParentMfeBridge.ts
```diff
@@ -10,8 +10,8 @@
 
 import { ParentMfeBridge } from '../handler/types';
 import type { ActionsChain, SharedProperty } from '../types';
-import type { ChildMfeBridgeImpl } from './ChildMfeBridge';
-import { BridgeDisposedError } from './errors';
+import { ChildMfeBridgeImpl } from './ChildMfeBridge';
```
itechmeat: ⚪ [Not a blocker] This changed from `import type` to a value import, but `ChildMfeBridgeImpl` is used only in type positions in this file (field at 28, ctor param at 63, return at 72) - no `instanceof`, no construction. The value import adds a runtime edge for nothing, and no lint catches it.

packages/mfes/src/runtime/DefaultMfeRegistry.ts
```diff
@@ -200,6 +340,340 @@ export class DefaultMfeRegistry extends MfeRegistry {
...
+    // @cpt-end:cpt-frontx-algo-mfe-host-communication-registration-propagation:p2:inst-repropagate-upward
+    // @cpt-end:cpt-frontx-algo-mfe-host-communication-registration-propagation:p2:inst-no-collision
+    return true;
+  }
+
+  /**
+   * Compose and propagate an advertisement for a locally-admitted target
+   * upward through this registry's inbound bridge, if it has one.
+   *
+   * @cpt-begin:cpt-frontx-algo-mfe-host-communication-registration-propagation:p2:inst-compose-advertisement
```
itechmeat: ⚪ [Not a blocker] This pair's `@cpt-end` is the first line of `propagateAdvertisementUpward`'s body (531), so the region spans the closing `*/`, the signature, and nothing else. The instruction is genuinely realized at 841-848 and 1130-1139, which carry their own correct pairs - this third, empty pair can simply go.

itechmeat (approved):
Approving - the direction is right and the execution shows real care: the loop-containment story holds under tracing, the revocation ordering is correct, the cross-copy rendezvous reasoning checks out, and the CodeRabbit rounds were closed with genuine fixes and honest rebuttals rather than rubber stamps. Everything below is non-blocking: a few traceability-marker hygiene items worth settling before they spread, some edge-case questions that a short answer closes, and a recurring pattern of design rationale living in code comments where a doc pointer would serve the next reader better. None of it should hold the merge.

packages/mfes/src/runtime/inbound-bridge-link.ts
GeraBart: Fixed — re-scoped the marker region. `inst-adopt-ambient-bridge`'s `@cpt-begin`/`@cpt-end` now wraps `adoptAmbientInboundBridgeLink`'s actual body; `registerInboundBridgeLink`/`unregisterInboundBridgeLink` now carry their own correct pairs (`inst-mint-escalation-on-link`, `inst-retract-advertisements`).

packages/mfes/src/runtime/inbound-bridge-link.ts
GeraBart: Fixed — all 10 sites converted: field/interface-method declarations became plain prose (the real implementations already carry correct hyphenated `@cpt-begin/@cpt-end` pairs elsewhere), and the ones lacking a real pair (`retractPropagatedTarget`, `retractInboundBridgeLinkFor`, `resolveForwardingEntryRoute`, and the `ChildMfeBridge` abstract class for `inst-inbound-bridge-internal`) now have one.

packages/mfes/__tests__/registration-propagation/cross-nesting-reachability.test.ts
GeraBart: Fixed — hoisted `LINK_PROPERTY_KEY` to one module-level const, added a local `LinkCarryingBridge` interface, replaced both `as any` casts with `as unknown as LinkCarryingBridge`, dropped both eslint-disable comments.

packages/mfes/src/handler/mfe-bridge-factory-default.ts
GeraBart: Investigated deletion, but `MfeBridgeFactoryDefault`/`MfeHandler.bridgeFactory` turns out to be spec-mandated public API — referenced by a checked-off DoD and acceptance criterion in mfe-registry/FEATURE.md, ADR-0006, and re-exported downstream in template-shell/packages/framework and packages/react. Deleting it needs an ADR-0006 revision reconciling handler-provided vs. registry-owned bridge construction, which is out of scope for this PR (issue #590) — flagged as a follow-up with a pointer comment at handler/types.ts:198. Fixed the in-scope mechanical piece: the stale `instanceId`-named JSDoc on `create()`'s param now correctly describes it as the extension's GTS id (`extensionId`).

template-mfe/src-app/mfe_packages/demo-mfe/src/lifecycle-widgets-host.tsx
GeraBart: Fixed — rewrote the paragraph to describe the actual remaining race (the ExtensionDomainSlot root-attach effect landing on a later React commit than the chain's `next` continuation), removed the false claim about `loadExtension` not awaiting in-flight loads. Also fixed the broken indentation nearby.

packages/mfes/src/runtime/DefaultExtensionMounter.ts
GeraBart: Fixed, but not as a graceful conflict — verified container is never externally specified (every mount strategy creates it internally via `hooks.create`), so two different containers for the same in-flight extension can only mean a bug in the calling strategy. `DefaultExtensionMounter.mount` now throws a hard internal-invariant error immediately in that case, naming the extension id. Added regression tests for both the throw and the legitimate same-container dedup path.

packages/mfes/src/runtime/DefaultMfeRegistry.ts
GeraBart: Fixed — confirmed this is deliberate: by the time the escalation tier runs, the target has already been checked against every local resolution path, so escalation is correctly target-blind (a nested registry can't know what its ancestor holds). Dropped the unused parameter entirely rather than leaving an underscore that reads like a bug, and documented the target-blindness explicitly.

template-mfe/src-app/mfe_packages/demo-mfe/src/lifecycle-widgets-host.tsx
GeraBart: Good catch — this constraint was genuinely undocumented. A full singleton redesign is out of scope for this PR (touches template-shell/packages/framework and the mfes registry-factory cache pattern more broadly), but extended the existing `inst-nested-registry-lifetime-scope` instruction in FEATURE.md with the cache-consumption consequence and the resulting "defer build() until inside mount()" requirement, so it's now specified rather than just worked around in this one demo.

packages/mfes/src/runtime/DefaultMfeRegistry.ts
GeraBart: Went the opposite direction from carrying the code across — traced every consumer of `ChainResult.error` and confirmed the fallback mechanism is deliberately failure-cause-agnostic everywhere (only "did it fail" drives behavior, never the specific error). Preserving identity across hops would serve no actual consumer, so removed `ChainResult.error` entirely instead of adding a `cause` field, and simplified the cross-hop throw to a shared contentless sentinel. Two pre-existing tests that asserted on error-message text were rewritten to assert outcome/mechanism instead (e.g. spying on the actual rejection path for the inactive-vs-missing-handler distinction, rather than parsing log text).

packages/mfes/src/mediator/actions-chains-mediator.ts
GeraBart: Fixed properly rather than patched — converted `CrossHopRoute` from a plain object-literal interface to a real class, and `isCrossHopRoute` now uses `instanceof` instead of duck-typing on `send`. An app `ActionHandler` subclass can now safely define its own `send` method with no collision risk. Added a regression test for exactly that case.

.github/workflows/template-drift.yml
GeraBart: Fixed — moved the npm/cli#524 analysis, the runner-version facts, and the 41/546 measurement into a new "Why template drift CI installs without a lockfile" subsection in CONTRIBUTING.md. The workflow now carries 2-line pointers instead of the full essay. Also rewrote the alpha version-bump paragraph in CONTRIBUTING.md to state the standing rule rather than narrate it as resolved history.

.gitignore
GeraBart: Fixed — removed the `.qa-evidence/` entry and comment; confirmed nothing in the repo produces or reads that directory.

packages/mfes/src/runtime/default-mount-manager.ts
GeraBart: Fixed — removed the dangling "(framing correction, §0)" / "(§0)" parentheticals at all three sites; confirmed no such plan document exists for this PR and the surrounding sentences read correctly without them.

packages/mfes/architecture/features/mfe-host-communication/FEATURE.md
GeraBart: Fixed, and audited the whole surrounding block rather than just this one line: 7 criteria flipped to checked (implementation+test confirmed), 3 had over-claiming text corrected first then checked (the active/inactive-surface claim scoped to the abstract contracts, the diagnostic-logging condition corrected, the revoked-link claim carved out for `retractAdvertisement`'s legitimate no-op), and 4 are left unchecked as genuine test-coverage gaps rather than assumed done.

architecture/ADR/0007-action-dispatch-and-chaining.md
GeraBart: Fixed — removed "for a reason not weighed at the time" from the decision-driver bullet.

packages/mfes/src/bridge/ParentMfeBridge.ts
GeraBart: Fixed — reverted to `import type` for `ChildMfeBridgeImpl`; confirmed all usages in the file are type-only positions.

packages/mfes/src/runtime/DefaultMfeRegistry.ts
GeraBart: Fixed — removed the empty pair around `propagateAdvertisementUpward`'s signature; confirmed the two substantive pairs for the same instruction elsewhere in the file remain intact.

packages/mfes/__tests__/registration-propagation/cross-nesting-reachability.test.ts
GeraBart: Agreed this is worth doing, but not in this PR — extracting a shared `__tests__/fixtures/` module for `cross-nesting-reachability.test.ts` and its siblings (excluding `cross-copy-boundary.test.ts`, which genuinely needs its own isolated copies) is a reasonable follow-up cleanup. Not addressing it here to keep this review-response batch scoped to the findings that were actually fixed.

Issue #594, opened 2026-08-25
fix(template-mfe): react-hooks/set-state-in-effect lint failure in widgets-fixture-a/lifecycle.tsx breaks develop CI


tscbmstubp:
## Problem

`develop`'s `Template Validate` CI job (and any PR merging current `develop`) fails on lint:

```
template-shell/src-app/mfe_packages/widgets-fixture-a/src/lifecycle.tsx
  121:5  error  Error: Calling setState synchronously within an effect can trigger cascading renders
  react-hooks/set-state-in-effect
```

This is on `develop` itself (confirmed the file is unchanged/identical across recent branches), so it currently breaks CI for any PR that merges or rebases onto `develop` — it surfaced on PR #567 after merging `develop` in, even though that PR never touched this file.

## Where

`template-mfe/src-app/mfe_packages/widgets-fixture-a/src/lifecycle.tsx` (overlaid into `template-shell` at CI time), inside `WidgetA`'s effect:

```tsx
const [lastPing, setLastPing] = React.useState<string | null>(
  () => lastPingValues.get(instanceId) ?? null,
);

React.useEffect(() => {
  // Re-sync against this instance's own value the moment this effect
  // commits, THEN subscribe for pings that arrive after that. ...
  setLastPing(lastPingValues.get(instanceId) ?? null);   // <- flagged line (121)
  lastPingSubscribers.set(instanceId, setLastPing);
  return () => { /* ... */ };
}, [bridge, instanceId]);
```

The effect's own comment explains *why* it re-syncs synchronously (to catch a ping that lands between the initial-state seed and subscription), so this isn't a careless `setState`-in-effect — the `react-hooks/set-state-in-effect` rule doesn't know that. Needs an intentional fix (e.g. re-deriving during render like the pattern already used elsewhere in this PR's `useBridgeProperty`/`useHostDirection` hooks, or an eslint-disable with a comment explaining the deliberate re-sync) rather than a mechanical suppression.

## Impact

Blocks `Template Validate` on `develop` and on any PR that merges/rebases current `develop` in, independent of that PR's own diff.

PR #598, opened 2026-08-27
fix: green develop after the ui-kit port - follow-ups and template hygiene

itechmeat:
## Description

`develop`'s Template Validate job has been red since the ui-kit 0.4 port landed. The composed shell+overlay tree stopped type-checking the moment the ecosystem's `@gears-frontx/ui-kit` moved to 0.4.0-alpha.0: the four uikit demo screens in `demo-mfe` were still written against the pre-0.4 `Field` (which took `name`/`disabled`), `FieldError` (which took `match`), and `Badge` (which took `dot`/`muted`) - none of which the kit ships anymore. Two composed-overlay files also carried lint debt (`set-state-in-effect` in the widgets fixture's ping indicator, `no-useless-assignment` in the demo screens' `CategoryMenu`) that only surfaces once the overlay is linted as part of the assembled tree.

Separately, @GeraBart left two low-severity notes on #591 that never got addressed: `Marker` doesn't absorb an explicit `variant={null}` the way `item.tsx` already guards the equivalent case, and `NativeSelectOptGroup` reuses the `<option>` class instead of having its own.

And a batch of small template-hygiene fixes has been sitting unmerged on `demo/uikit-inbox` - a footer strip that renders when nothing feeds it, a missing `build:packages` pre-script wiring, the host chrome actions the framework's screen domain didn't declare yet, and a seeded mock-user name that disagreed with what a workspace microfrontend already showed for the same person.

## What changed

Migrated the four demo screens to the kit's actual 0.4 API rather than papering over the type errors - every label now points at its control's own id (`Field` carries no id-wiring anymore), the profile form's inputs own their `disabled` state, the email demo reads validity on blur instead of relying on `FieldError`'s removed `match`, and the badge row demonstrates what the port actually added (weights, a link-rendered badge) instead of the removed `dot`/`muted`. Cleared the `CategoryMenu` lint debt - its element lookup collapsed into the ternary it always was. The widgets fixture's `set-state-in-effect` was fixed on `develop` in the meantime (#594) with the same `useSyncExternalStore` shape and a stricter cleanup guard, so this PR no longer touches that file.

Fixed both of @GeraBart's notes: `Marker` resolves an explicit `null` variant to `'default'` up front, in both the CSS class and the `data-variant` attribute, the same way `item.tsx` already does; `NativeSelectOptGroup` now styles on its own `.optgroup` class. Both closed out with a test. Since this changed `packages/ui-kit`'s `src/`, it also carries the version bump `policy:version-bump-on-change` requires (0.4.0-alpha.1 -> 0.4.0-alpha.2 - develop published alpha.1 in the meantime) and the matching pin bump on its two consumers, kept in the same commit so neither gate is red in between.

Cherry-picked five of the hygiene fixes from `demo/uikit-inbox` (the sixth - an eslint-ignore for `template-design-guardrails/` - was left out: that directory doesn't exist on current `develop`, so the fix is moot there). Composing them onto the current tree surfaced two things that weren't visible on the source branch in isolation: `demo-mfe` ships its own parallel accounts mock, and it silently agreed with the shell's seeded identity only because both happened to say "Demo User" - once the shell's rename to "Alex Rivera" landed, the composed tree's contract test failed on the mismatch, so the same rename was carried into `demo-mfe`'s copy. And the manifest change that lets `template-mfe` claim its five packages individually instead of the whole `mfe_packages/` directory introduced the template's first file-shaped `exclusiveSubtree` entry (`README.md`) - which the CI composition script had only ever handled as a directory (`[ -d ... ]`, `cp -R src/. dest/`), so it failed the job before composing anything. Fixed the script to branch on what the tree actually has. A companion CLI test also needed updating: the conflict-checker never merges overlapping claims, so a synthetic contestant against `template-mfe`'s now-five subtrees correctly reports five conflicts instead of the one it used to, and the test's expectation is now derived from the real on-disk manifest (the same pattern the suite's own Fixture 5 already uses for the shell) instead of a snapshot.

Review follow-ups: the residue `@cpt-` markers in the two touched shell files are stripped in a marker-only commit, as ADR-0033 asks of any change that modifies an already-marked template file; the email demo's validation messages go through `t()` like the rest of the screen, keyed by the validity reason so a language switch re-renders them; and `template-shell`'s manifest version is bumped alongside `template-mfe`'s.

Verified locally: `packages/ui-kit` (vitest 910/910, build, type-check); a full local equivalent of "Validate shell + overlay composition" - fresh `npm install`, `build`, `type-check`, `arch:deps`, `lint`, `test:unit` - built against the locally packed `ui-kit` output at the bumped version (it isn't published yet - see below); the CLI suite (507/507); and all three `policy:*` gates plus `npm ci --dry-run` at the root.

## Risks / open questions

`@gears-frontx/ui-kit@0.4.0-alpha.2` isn't published yet - it publishes automatically on merge to `develop` (`publish-packages.yml`). Until then, the registry-install step of Template Validate's "Validate shell + overlay composition" job can't resolve the new pin from the real npm registry, the same two-phase gap this repo's own history shows for 0.4.0-alpha.0 itself (bumped in `88e77b9b`, the matching pin caught up separately once it was actually published, in `83c2a635`). That step should go green once this PR merges and the publish workflow runs; nothing else in it depends on the unpublished version.

Closes review threads: https://github.com/constructorfabric/gears-frontx/pull/591#discussion_r3869081148, https://github.com/constructorfabric/gears-frontx/pull/591#discussion_r3869081156

## Summary by CodeRabbit

* **New Features**
  * Added host controls for changing the application theme and collapsing the menu.
  * Added email validation on blur, with localized messages across supported languages.
  * Improved form accessibility by linking labels and inputs consistently.
  * Updated UI kit demonstrations for badge tones, weights, and link rendering.

* **Bug Fixes**
  * Ensured null variants use consistent default styling.
  * Corrected optgroup styling in native selects.
  * Updated footer rendering behavior.
  * Prevented template overlays from overwriting existing files.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -356,6 +356,17 @@ jobs:
             # being one claim — so this branches on which the tree actually
             # has, rather than assuming a directory the way `-d` alone did.
             if [ -d "$overlay/$subtree" ]; then
+              # The ground has to be free before anything is written, the way
+              # `refuseUnlessGroundFree` requires it of a real `frontx add`:
+              # every copy below is an overwrite, so a destination that already
+              # holds content means the shell and this overlay both claim it and
+              # the composition under test is not the one a developer would get.
+              # Asked before `mkdir -p`, which would otherwise create the very
+              # path being asked about.
+              if [ -e "$shell/${subtree%/}" ]; then
+                echo "::error::$overlay's exclusiveSubtree '$subtree' is already occupied in $shell"
+                exit 1
+              fi
               echo "==> Overlaying $overlay/$subtree"
               # Copy the *contents* into the destination: `cp -R src dest`
               # nests under dest when dest already exists, which would bury
@@ -363,6 +374,10 @@ jobs:
               mkdir -p "$shell/${subtree%/}"
               cp -R "$overlay/$subtree/." "$shell/${subtree%/}/"
             elif [ -f "$overlay/$subtree" ]; then
+              if [ -e "$shell/$subtree" ]; then
+                echo "::error::$overlay's exclusiveSubtree '$subtree' is already occupied in $shell"
+                exit 1
+              fi
               echo "==> Overlaying $overlay/$subtree (file)"
               mkdir -p "$shell/$(dirname "$subtree")"
               cp "$overlay/$subtree" "$shell/$subtree"
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -349,28 +349,41 @@ jobs:
           overlay_count=0
           while read -r subtree; do
             [ -n "$subtree" ] || continue
-            # A declared subtree that isn't there means the manifest and the
-            # tree disagree.
-            if [ ! -d "$overlay/$subtree" ]; then
+            # A declared subtree is usually a directory, but 0d22fe81 started
+            # declaring template-mfe's own README.md as its own
+            # exclusiveSubtree entry, alongside the four MFE package
+            # directories, once "the whole mfe_packages directory" stopped
+            # being one claim — so this branches on which the tree actually
+            # has, rather than assuming a directory the way `-d` alone did.
+            if [ -d "$overlay/$subtree" ]; then
+              echo "==> Overlaying $overlay/$subtree"
+              # Copy the *contents* into the destination: `cp -R src dest`
+              # nests under dest when dest already exists, which would bury
+              # the packages a level deeper than the count below looks.
+              mkdir -p "$shell/${subtree%/}"
+              cp -R "$overlay/$subtree/." "$shell/${subtree%/}/"
+            elif [ -f "$overlay/$subtree" ]; then
+              echo "==> Overlaying $overlay/$subtree (file)"
+              mkdir -p "$shell/$(dirname "$subtree")"
+              cp "$overlay/$subtree" "$shell/$subtree"
+            else
+              # A declared subtree that isn't there means the manifest and
+              # the tree disagree.
               echo "::error::$overlay declares exclusiveSubtree '$subtree', which does not exist"
               exit 1
             fi
-            echo "==> Overlaying $overlay/$subtree"
-            # Copy the *contents* into the destination: `cp -R src dest` nests
-            # under dest when dest already exists, which would bury the
-            # packages a level deeper than the count below looks.
-            mkdir -p "$shell/${subtree%/}"
-            cp -R "$overlay/$subtree/." "$shell/${subtree%/}/"
             # Each MFE ships a standalone package-lock.json whose file:
             # targets point into template-mfe/packages/*, which does not
             # exist. npm ignores nested locks inside a workspace, but drop
             # them so nothing reads them — from the subtree just copied, not
-            # a hardcoded mfe_packages path.
+            # a hardcoded mfe_packages path. A no-op for a file subtree that
+            # isn't itself named package-lock.json.
             find "$shell/${subtree%/}" -not -path '*/node_modules/*' \
               -name package-lock.json -delete
             # Counted from the subtree just copied, not a hardcoded
             # `src-app/mfe_packages` path — the manifest already told us
-            # where this overlay's content landed.
+            # where this overlay's content landed. A no-op (0) for a file
+            # subtree that isn't itself package.json.
             subtree_count="$(find "$shell/${subtree%/}" -not -path '*/node_modules/*' \
               -name package.json 2>/dev/null | wc -l)"
             overlay_count=$((overlay_count + subtree_count))
--- a/packages/cli/src/__tests__/template-split.e2e.test.ts
+++ b/packages/cli/src/__tests__/template-split.e2e.test.ts
@@ -294,14 +294,20 @@ describe("Fixture 4 — conflict-check on REAL templates: a synthetic 'mfe-dup' 
       // claim collides with each of them separately, and a report naming only
       // one would leave the rest for the developer to rediscover.
       const mfeSubtrees = readManifest(TEMPLATE_MFE_DIR).ownershipBoundaries.exclusiveSubtrees;
+      const swallowed = mfeSubtrees.filter((subtree) =>
+        pathWithinSubtree(subtree, 'src-app/mfe_packages/'),
+      );
+      // A derived oracle with no floor under it would pass against a manifest
+      // that had stopped declaring anything under `src-app/mfe_packages/`: an
+      // empty expectation matches an empty report, and the multi-conflict
+      // property the comment above describes would go untested. More than one
+      // is the standing fact this case exists for.
+      expect(swallowed.length).toBeGreaterThan(1);
       expect(dupAddResult.conflicts).toEqual(
-        mfeSubtrees
-          .filter((subtree) => pathWithinSubtree(subtree, 'src-app/mfe_packages/'))
-          .map((subtree) => ({
-            ground:
-              subtree === 'src-app/mfe_packages/' ? subtree : `src-app/mfe_packages/ overlaps ${subtree}`,
-            contestants: ['frontx-template-mfe-dup', TEMPLATE_MFE_IDENTITY],
-          })),
+        swallowed.map((subtree) => ({
+          ground: `src-app/mfe_packages/ overlaps ${subtree}`,
+          contestants: ['frontx-template-mfe-dup', TEMPLATE_MFE_IDENTITY],
+        })),
       );
     }
 
--- a/packages/cli/src/__tests__/template-split.e2e.test.ts
+++ b/packages/cli/src/__tests__/template-split.e2e.test.ts
@@ -286,9 +286,23 @@ describe("Fixture 4 — conflict-check on REAL templates: a synthetic 'mfe-dup' 
     if (dupAddResult.ok) return;
     expect(dupAddResult.reason).toBe('conflict');
     if (dupAddResult.reason === 'conflict') {
-      expect(dupAddResult.conflicts).toEqual([
-        { ground: 'src-app/mfe_packages/', contestants: ['frontx-template-mfe-dup', TEMPLATE_MFE_IDENTITY] },
-      ]);
+      // Derived from the real, on-disk template-mfe manifest — same pattern as
+      // Fixture 5's shell case below — rather than pinned to a snapshot: mfe
+      // claims five separate exclusive subtrees nested under
+      // 'src-app/mfe_packages/' (refactor(template-mfe): claim the packages it
+      // ships, not the whole mfe_packages directory), so the dup's one broad
+      // claim collides with each of them separately, and a report naming only
+      // one would leave the rest for the developer to rediscover.
+      const mfeSubtrees = readManifest(TEMPLATE_MFE_DIR).ownershipBoundaries.exclusiveSubtrees;
+      expect(dupAddResult.conflicts).toEqual(
+        mfeSubtrees
+          .filter((subtree) => pathWithinSubtree(subtree, 'src-app/mfe_packages/'))
+          .map((subtree) => ({
+            ground:
+              subtree === 'src-app/mfe_packages/' ? subtree : `src-app/mfe_packages/ overlaps ${subtree}`,
+            contestants: ['frontx-template-mfe-dup', TEMPLATE_MFE_IDENTITY],
+          })),
+      );
     }
 
     expect(listRealFiles(targetDir).sort()).toEqual(filesBefore);
--- a/packages/ui-kit/src/components/marker/marker.tsx
+++ b/packages/ui-kit/src/components/marker/marker.tsx
@@ -39,7 +39,15 @@ const markerVariants = cva(styles.marker, {
 
 export interface MarkerProps extends useRender.ComponentProps<'div'>, VariantProps<typeof markerVariants> {}
 
-export function Marker({ className, variant = 'default', render, ...props }: MarkerProps) {
+export function Marker({ className, variant, render, ...props }: MarkerProps) {
+  // cva's own VariantProps types this `| null` (a consumer can pass
+  // `variant={null}` to explicitly reset to the default), which a default
+  // parameter value alone doesn't absorb — only `undefined` does. cva reads
+  // that null as "emit no variant class at all" while `data-variant` would
+  // still be dropped by React, so the marker would come out with neither the
+  // default class nor the attribute. Resolved once here, same as item.tsx,
+  // so class and attribute always agree.
+  const resolvedVariant = variant ?? 'default';
   return useRender({
     defaultTagName: 'div',
     render,
@@ -49,8 +57,8 @@ export function Marker({ className, variant = 'default', render, ...props }: Mar
     // shadow-proofing needed anywhere a `data-*` prop must survive
     // mergeProps.
     props: {
-      ...mergeProps<'div'>({ className: markerVariants({ variant, className }) }, props),
-      'data-variant': variant,
+      ...mergeProps<'div'>({ className: markerVariants({ variant: resolvedVariant, className }) }, props),
+      'data-variant': resolvedVariant,
     },
   });
 }
--- a/template-mfe/src-app/mfe_packages/demo-mfe/src/screens/uikit/components/FormElements.tsx
+++ b/template-mfe/src-app/mfe_packages/demo-mfe/src/screens/uikit/components/FormElements.tsx
@@ -50,11 +50,24 @@ const REGION_ITEMS = [
   { value: 'us-east', label: 'Virginia' },
 ];
 
+/**
+ * Which of the two `ValidityState` flags the demo reacts to, kept as a reason
+ * rather than as the message it resolves to: the host can switch language while
+ * an error is on screen, and a stored message would keep the language it was
+ * produced in until the field is blurred again.
+ */
+type EmailErrorReason = 'required' | 'malformed';
+
+const EMAIL_ERROR_KEYS: Record<EmailErrorReason, string> = {
+  required: 'form_email_required',
+  malformed: 'form_email_invalid',
+};
+
 export const FormElements: React.FC<FormElementsProps> = ({ t, portalContainer }) => {
   const [region, setRegion] = useState('eu-central');
   const [plan, setPlan] = useState('free');
   const [notifications, setNotifications] = useState(true);
-  const [emailError, setEmailError] = useState<string | null>(null);
+  const [emailError, setEmailError] = useState<EmailErrorReason | null>(null);
 
   /**
    * Field derives no validation of its own, so the demo runs the check the
@@ -66,9 +79,9 @@ export const FormElements: React.FC<FormElementsProps> = ({ t, portalContainer }
   const validateEmail = (event: React.FocusEvent<HTMLInputElement>) => {
     const { validity } = event.currentTarget;
     if (validity.valueMissing) {
-      setEmailError('Email is required.');
+      setEmailError('required');
     } else if (validity.typeMismatch) {
-      setEmailError('That does not look like an email.');
+      setEmailError('malformed');
     } else {
       setEmailError(null);
     }
@@ -109,7 +122,9 @@ export const FormElements: React.FC<FormElementsProps> = ({ t, portalContainer }
             <FieldDescription id="form-demo-email-description">
               We only use it for the invoice.
             </FieldDescription>
-            <FieldError id="form-demo-email-error">{emailError}</FieldError>
+            <FieldError id="form-demo-email-error">
+              {emailError ? t(EMAIL_ERROR_KEYS[emailError]) : null}
+            </FieldError>
           </Field>
         </div>
       </ElementDemo>
--- a/template-mfe/src-app/mfe_packages/demo-mfe/src/screens/uikit/components/FormElements.tsx
+++ b/template-mfe/src-app/mfe_packages/demo-mfe/src/screens/uikit/components/FormElements.tsx
@@ -76,7 +76,7 @@ export const FormElements: React.FC<FormElementsProps> = ({ t, portalContainer }
    * blur rather than on every keystroke so a half-typed address is not
    * flagged as malformed while it is still being typed.
    */
-  const validateEmail = (event: React.FocusEvent<HTMLInputElement>) => {
+  const validateEmail = (event: React.SyntheticEvent<HTMLInputElement>) => {
     const { validity } = event.currentTarget;
     if (validity.valueMissing) {
       setEmailError('required');
@@ -84,6 +84,19 @@ export const FormElements: React.FC<FormElementsProps> = ({ t, portalContainer }
       setEmailError('malformed');
     } else {
       setEmailError(null);
+    }
+  };
+
+  /**
+   * A message already on screen has to come off the moment the field is
+   * corrected, so once there is one, every keystroke re-runs the same check.
+   * The gate is what keeps the blur-only rule intact for a field that has not
+   * been flagged yet: with no error showing there is nothing to clear, and
+   * validating from the first character is exactly what reading on blur avoids.
+   */
+  const revalidateEmail = (event: React.ChangeEvent<HTMLInputElement>) => {
+    if (emailError) {
+      validateEmail(event);
     }
   };
 
@@ -118,6 +131,7 @@ export const FormElements: React.FC<FormElementsProps> = ({ t, portalContainer }
                   : 'form-demo-email-description'
               }
               onBlur={validateEmail}
+              onChange={revalidateEmail}
             />
             <FieldDescription id="form-demo-email-description">
               We only use it for the invoice.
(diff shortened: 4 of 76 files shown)


template-shell/package.json
```diff
@@ -27,9 +27,11 @@
     "src-app/mfe_packages/*"
   ],
   "scripts": {
+    "prebuild:mfes": "npm run build:packages",
```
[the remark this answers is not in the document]
itechmeat: Fixed - dropped the redundant explicit `build:packages` from `dev`; `build:mfes`'s own `prebuild:mfes` hook still runs it once (296881a1).

template-shell/src-app/app/layout/Footer.tsx
```diff
@@ -16,7 +16,14 @@ export const Footer: React.FC<FooterProps> = ({ children }) => {
...
 
-  if (!visible) {
+  // The strip is a container and carries nothing of its own, so with no
+  // children it renders a 40px band of border and background across the whole
+  // viewport that no screen can fill: `Layout` passes it none, and the layout
+  // plugin exposes no footer event a microfrontend could hide it with. Absence
+  // of children is therefore the only signal available, and it is the right
+  // one - `visible` stays the explicit switch for a footer that does have
+  // content.
+  if (!visible || React.Children.count(children) === 0) {
```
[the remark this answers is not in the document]
itechmeat: Fixed - switched to `Children.toArray(children).length === 0` and added a regression test for the falsy-conditional-child case (39da0b23).

tscbmstubp:
# Code Review: PR #598

**PR**: fix: green develop after the ui-kit port - follow-ups and template hygiene
**Author**: @itechmeat
**Prompt**: Code Review (+ FrontXCodeReviewChecks FX-001..003)
**Review Decision**: none (no human review submitted)

Head: `39da0b23` (16 commits, 31 files, +473/-128). Review mode: Full (>200 LOC, touches ecosystem `ui-kit`/`cli` and template territory).

---

## Verdict: ⚠️ REQUEST CHANGES

The code itself is sound and both bot threads were addressed, but two touched template files still carry residue `@cpt-` markers, which ADR-0033 requires stripping in the same change (FX-002, HIGH by project rule), and the PR cannot go green before merge because it pins an unpublished `ui-kit` version.

---

## Reviewer Comment Analysis

### @coderabbitai (bot) — replies by @itechmeat (author)

| # | Concern | Relevance | Addressed? |
|---|---------|-----------|------------|
| 1 | [`dev` runs `build:packages` twice once `prebuild:mfes` exists](https://github.com/constructorfabric/gears-frontx/pull/598#discussion_r3869715018) | **Valid** — npm `pre*` hook duplicates the explicit call | ✅ Yes — `dev` no longer calls `build:packages` (296881a1); thread resolved |
| 2 | [`Children.count(false)` is 1, so a falsy conditional child re-enables the empty footer band](https://github.com/constructorfabric/gears-frontx/pull/598#discussion_r3869715026) | **Valid** — real behaviour gap | ✅ Yes — `Children.toArray(children).length === 0` + regression test (39da0b23); thread resolved |

Carried-over concerns from #591 (@GeraBart, `Marker variant={null}`, `NativeSelectOptGroup` class): both fixed here with tests — verified in the diff.

No other human reviewer comments.

---

## Own Findings

### FrontX checks (FX) 

**FX-001 Constraints** ✅ — `ui-kit` and `cli` changes stay inside their packages; no ecosystem→template import added. CLI change is a comment + test-only; CLI-1 unaffected.

**FX-002 Traceability** ❌ **HIGH**
- Two modified template files still carry residue `@cpt-` markers. ADR-0033 (`cpt-frontx-adr-template-territory-traceability`, Confirmation section): "a change modifying an already-marked template file removes every residue marker that file carries".
  - `template-shell/src-app/__test-utils__/mockUserFixture.ts` lines 1, 3, 35 (`cpt-frontx-dod-api-communication-base-service`)
  - `template-shell/src-app/app/mfe/bootstrap.ts` lines 1, 201, 266, 368, 423 (`cpt-frontx-flow-request-lifecycle-query-client-lifecycle`, `cpt-frontx-dod-mfe-registry-mfe-schema-registration`)
  - No *new* markers were added (0 `@cpt-` lines in the diff) — good.
  - Fix: delete those 8 marker comment lines, ideally in a separate marker-only commit so the strip stays distinguishable from the substantive change (as the ADR asks).
- `cfs validate`: CI "Studio Validate Artifacts" ✅ on head. (Local run in a scratch worktree could not be trusted — env-only failure; not counted.)

**FX-003 Architecture** ✅ — CI `validate` job (runs `npm run arch:check`) passed on `39da0b23` for both Node versions. Local run unreliable in the scratch worktree; CI result used as evidence.

### Correctness ✅ (one MINOR)

Verified against current sources:
- `Marker`: `variant ?? 'default'` applied to both class and `data-variant`; matches `item.tsx` pattern. Correct.
- `NativeSelectOptGroup` own `.optgroup` class; test asserts class present and `.option` absent. Correct.
- Demo screens vs. ui-kit 0.4: `Badge` variants `accent/secondary/outline/destructive` exist (`badge.tsx:36-47`); `Field` is layout-only and CSS keys on `[data-disabled='true']`/`[data-invalid='true']` — `data-disabled={isSaving || undefined}` renders `"true"` so it matches; `FieldError` renders nothing when `children` is `null`, so no stray `role="alert"` node. `SelectTrigger` forwards `id`. All correct.
- `bootstrap.ts` chrome actions: `ExtensionDomain.actions: string[]` exists (mfes d.ts:420); `app.actions.changeTheme({themeId})` and `toggleMenuCollapsed({collapsed})` match framework types; schemas mirror `mount_ext.v1.json`; `extensionsActions` deliberately untouched. Handlers resolve rather than throw on bad payload with a warn — reasonable for a host-chrome request.
- CI overlay script: file-shaped subtree branch correct (`mkdir -p $(dirname)`, `cp`); `find <file> -name package-lock.json` is a safe no-op.
- `frontx-template.json`: five narrow claims match the on-disk `mfe_packages/` listing exactly; no overlap with the shell manifest's root `README.md` claim.
- CLI e2e test now derives expectations from the real manifest and uses the `${a} overlaps ${b}` ground format from `scaffold/conflict.ts:190`. Passes locally on head (13/13).
- `Footer`: `React` is imported; `toArray` fix correct.
- **MINOR (pre-existing parity, not a regression)** — `widgets-fixture-a/src/lifecycle.tsx`: the `useSyncExternalStore` unsubscribe deletes `lastPingValues[instanceId]`. React re-runs subscribe on every `subscribe` identity change and double-invokes it in StrictMode dev, so a ping that landed before that resubscribe is wiped and the snapshot flips to `null`. The old effect cleanup did the same, so behaviour is unchanged; worth a follow-up to move the value cleanup into `unmount()` (the file's own comment already gestures at that).

### Code Style & Idiomatic Patterns ✅
Comments are heavy but explain non-obvious decisions (cva `null`, `Children.count`, uSES). `CategoryMenu` simplification is a clean collapse. Nit: `FormElements.tsx` hardcodes English validation strings (`'Email is required.'`) in a screen that otherwise uses `t()`.

### Performance ✅
`dev` no longer double-builds packages. uSES snapshot is a primitive, so no tearing/identity churn.

### Test Coverage ✅
New/updated tests for every behavioural change: Marker null variant, optgroup class, Footer (4 cases incl. regression), bootstrap chrome-actions registration, mock-user rename in both trees, CLI conflict derivation. Local: CLI suite file 13/13 on head; ui-kit tests could not be executed in the scratch worktree (env) — CI `validate` passed, author reports 910/910.

### Security ✅
No secrets, no injection surfaces. Chrome-action payloads are type-checked before reaching `app.actions`. Schemas make payload fields `required`, so the mediator rejects malformed actions.

### Mistakes & Potential Misbehaviors ⚠️
- **CI is red by construction**: Template Validate fails with `npm error notarget No matching version found for @gears-frontx/ui-kit@0.4.0-alpha.1` — the pin points at a version that only publishes *after* merge to `develop`. The PR body documents this and cites precedent (`88e77b9b` → `83c2a635`). This is a repo process gap, not a code bug, but it means the branch cannot be verified green before merge; a maintainer must knowingly accept it.
- Version bumps: `ui-kit` 0.4.0-alpha.0→alpha.1 with both consumer pins in the same commit ✅; `cli` 0.3.0-alpha.2→alpha.3 for a comment-only `src/` change (policy-driven, harmless); `template-mfe` manifest 0.1.0-alpha.0→alpha.1 ✅. `template-shell` manifest not bumped despite substantive shell changes — the `policy:version-bump-on-change` gate passed, so not blocking; noting the asymmetry only.

---

## Summary

| Area | Rating |
|------|--------|
| Correctness | ✅ Verified against ui-kit/framework/mfes sources |
| Conformance (FX) | ❌ FX-002: 8 residue `@cpt-` markers left in 2 touched template files |
| Style | ✅ Minor i18n nit |
| Performance | ✅ Removes a double build |
| Tests | ✅ Every change tested; CLI 13/13 local, rest via CI |
| Security | ✅ No issues |
| Reviewer concerns | ✅ 2 of 2 valid concerns addressed |
| Risk | 🟡 Low-medium — code low; process risk from the unpublished pin |

## Recommendation

1. **Minimum (blocking)**
   - Strip the residue markers from `template-shell/src-app/__test-utils__/mockUserFixture.ts` (lines 1, 3, 35) and `template-shell/src-app/app/mfe/bootstrap.ts` (lines 1, 201, 266, 368, 423), in a marker-only commit (ADR-0033 / FX-002).
   - Get an explicit maintainer acknowledgement that Template Validate will stay red until `ui-kit@0.4.0-alpha.1` publishes on merge, per the `83c2a635` precedent — or publish the alpha first and re-run CI.
2. **Recommended**
   - Route the `FormElements.tsx` validation messages through `t()` like the rest of the screen.
3. **Nice-to-have**
   - Follow-up: move `lastPingValues` cleanup in `widgets-fixture-a` out of the uSES unsubscribe into `unmount()` so StrictMode/resubscribe cannot drop a delivered ping.
   - Consider bumping `template-shell`'s `frontx-template.json` alongside its content change for parity with `template-mfe`.

itechmeat:
@tscbmstubp Markers stripped in a separate commit, validation messages go through `t()`, shell manifest bumped. The unpublished pin stays as described - it publishes on merge, same as alpha.1 did; that call is the maintainer's. `widgets-fixture-a` is no longer touched here: develop landed the same `useSyncExternalStore` fix (#594) with the stricter cleanup guard, so the follow-up is moot.

template-shell/src-app/app/mfe/chrome-actions.ts
```diff
@@ -0,0 +1,70 @@
...
+ * rejects an action it cannot validate.
+ */
+export const CHROME_ACTION_SCHEMAS: JSONSchema[] = [
+  {
+    $id: `gts://${CHROME_SET_THEME}`,
+    $schema: 'https://json-schema.org/draft/2020-12/schema',
+    type: 'object',
+    properties: {
+      type: { 'x-gts-ref': '/$id' },
+      target: { 'x-gts-ref': 'gts.frontx.mfes.ext.domain.v1~*' },
```
tscbmstubp: **Target wildcard admits domains that have no chrome handler.** The schema's `target` is `gts.frontx.mfes.ext.domain.v1~*`, but `CHROME_SET_THEME` / `CHROME_SET_MENU_COLLAPSED` handlers are registered only in `ScreenDomainImpl`; `OptionalDomainImpl` (bootstrap.ts:152-168) registers none. A sidebar/popup/overlay MFE that targets its own domain passes GTS validation (the entry-declaration check is a no-op for domain targets), then `resolveHandler` misses and the chain completes with `NoHandlerForActionTargetError` (actions-chains-mediator.ts:59).

Either narrow the `x-gts-ref` to the screen domain type so the mismatch fails at validation with a clear message, or register the chrome handlers in `OptionalDomainImpl` too and spread the actions into all four declarations.

template-shell/src-app/app/layout/Footer.tsx
```diff
@@ -16,7 +16,19 @@ export const Footer: React.FC<FooterProps> = ({ children }) => {
...
+  // viewport that no screen can fill: `Layout` passes it none, and the layout
+  // plugin exposes no footer event a microfrontend could hide it with. Absence
+  // of children is therefore the only signal available, and it is the right
+  // one - `visible` stays the explicit switch for a footer that does have
+  // content.
+  // `Children.count` treats a falsy conditional child (`{enabled &&
+  // <FooterContent />}` with `enabled` false) as one slot, not zero, so it
+  // would let the empty band this check exists to suppress back in the
+  // moment a caller writes the pattern. `Children.toArray` drops those slots
+  // before counting.
```
tscbmstubp: **This gate makes the footer permanently null in the shell and turns `setFooterVisible` into a dead control.** `Layout.tsx` renders `<Footer />` with no children, so the `visible` branch can never be reached. The comment's premise ('the layout plugin exposes no footer event') does not hold: `app.actions.setFooterVisible` exists (framework `plugins/layout.ts:128`, `types.ts:138`, re-exported from `@gears-frontx/react`). A project calling it gets nothing rendered and no warning.

Secondary: `Children.toArray` keeps `''`, so `<Footer>{status ?? ''}</Footer>` still renders the empty 40px strip. ui-kit's `button.tsx:94-101` and `input.tsx:40-41` use `.some(child => child !== '')` for exactly this reason.

Suggest either having `Layout` omit `<Footer>` until something supplies content, or adding a footer chrome action wired to the existing `setFooterVisible` and keeping `visible` as the single switch.

packages/ui-kit/src/components/marker/marker.tsx
```diff
@@ -39,7 +39,15 @@ const markerVariants = cva(styles.marker, {
 
 export interface MarkerProps extends useRender.ComponentProps<'div'>, VariantProps<typeof markerVariants> {}
 
-export function Marker({ className, variant = 'default', render, ...props }: MarkerProps) {
+export function Marker({ className, variant, render, ...props }: MarkerProps) {
+  // cva's own VariantProps types this `| null` (a consumer can pass
+  // `variant={null}` to explicitly reset to the default), which a default
+  // parameter value alone doesn't absorb — only `undefined` does. cva reads
+  // that null as "emit no variant class at all" while `data-variant` would
+  // still be dropped by React, so the marker would come out with neither the
```
tscbmstubp, line 47: **Same null-variant bug remains in two sibling components.** `Bubble` (bubble.tsx:64-67) and `AttachmentMedia` (attachment.tsx:108-109) have the identical `variant = 'default'` default-param + `data-variant={variant}` shape, typed via cva's `VariantProps` (which admits `| null`). `variant={null}` on either renders with neither the default class nor the attribute.

The same `?? 'default'` / `?? 'icon'` resolution fixes both; a kit-local `Exclude<..., null>` on `VariantProps` would close it for every future cva component instead of per-file.

template-mfe/src-app/mfe_packages/demo-mfe/src/screens/uikit/components/FormElements.tsx
```diff
@@ -66,16 +98,33 @@ export const FormElements: React.FC<FormElementsProps> = ({ t, portalContainer }
...
+              type="email"
+              required
+              placeholder="you@company.com"
+              aria-invalid={emailError ? true : undefined}
+              aria-describedby={
+                emailError
+                  ? 'form-demo-email-description form-demo-email-error'
+                  : 'form-demo-email-description'
+              }
+              onBlur={validateEmail}
```
tscbmstubp, line 133: **Error state is never cleared while editing.** `setEmailError` is only called from `validateEmail` (onBlur). Blur the empty field -> 'Email is required.'; type `a@b.co` -> the field stays red with `aria-invalid=true` and the error id still in `aria-describedby` until focus leaves again.

A minimal fix that keeps the 'don't flag half-typed input' intent: `onChange={(e) => { if (emailError) validateEmail(e); }}` (re-validate only while an error is showing). This is a demo screen, but it is the reference for hand-wiring aria-invalid/aria-describedby on `Field`.

template-shell/package.json
tscbmstubp, line 30: **`npm run build` now builds the workspace packages twice.** `build` is still `npm run build:packages && npm run build:mfes && ...`; with `prebuild:mfes` added, `build:mfes` fires `build:packages` a second time (six workspace tsup builds plus `prebuild:packages`). This hits CI directly: `main.yml` lines 289 and 421 and `template-drift.yml` line 110 all run `npm run build`. `dev` was updated to drop its explicit call; `build` was not.

Either drop `npm run build:packages &&` from `build`, or drop the pre-hooks and keep the explicit chains.

template-shell/src-app/app/mfe/bootstrap.ts
```diff
@@ -93,13 +96,52 @@ class HostContainerHooks implements ContainerHooks {
...
     ctx.registerHandler(
       FRONTX_ACTION_MOUNT_EXT,
       ActionHandler.fromFunction((_t, p) => this.strategy.mount(p as ActionPayload)),
     );
+    // The host chrome a mounted screen may drive. Both resolve rather than
+    // throw on a payload or a plugin that is not there: a chrome request the
+    // shell cannot honour should leave the screen running, not fail its chain.
+    ctx.registerHandler(
+      CHROME_SET_THEME,
+      ActionHandler.fromFunction((_t, p) => {
```
tscbmstubp, line 148: **The `typeof` payload guards are unreachable.** The mediator calls `this.typeSystem.register(action)` (actions-chains-mediator.ts:235) before any handler runs, and gts-plugin's `register` (plugin.ts:178-196) runs `validateInstance` and throws on failure. The schemas in chrome-actions.ts declare `required` + `type` for `themeId` / `collapsed`, so a missing or mistyped field never reaches this code. The branches, their warn strings, and the comment ('resolve rather than throw on a payload ... that is not there') describe behaviour that cannot occur.

Keep only the 'no plugin on this host' guard, which the schema cannot express.

.github/workflows/main.yml
```diff
@@ -349,28 +349,41 @@ jobs:
...
+            # has, rather than assuming a directory the way `-d` alone did.
+            if [ -d "$overlay/$subtree" ]; then
+              echo "==> Overlaying $overlay/$subtree"
+              # Copy the *contents* into the destination: `cp -R src dest`
+              # nests under dest when dest already exists, which would bury
+              # the packages a level deeper than the count below looks.
+              mkdir -p "$shell/${subtree%/}"
+              cp -R "$overlay/$subtree/." "$shell/${subtree%/}/"
+            elif [ -f "$overlay/$subtree" ]; then
+              echo "==> Overlaying $overlay/$subtree (file)"
```
tscbmstubp, line 381: **PLAUSIBLE (latent, not triggerable today):** the file-subtree branch copies with a bare `cp` and no check that the destination is absent, while the step's comment says it mirrors the assembler's ownership semantics. The CLI's `refuseUnlessGroundFree` would refuse the same write. If template-shell ever ships `src-app/mfe_packages/README.md` (it has no `src-app/mfe_packages` today), CI overwrites it silently and passes where `frontx add` reports a conflict. The directory branch has the same pre-existing gap.

A `[ -e "$shell/$subtree" ]` error check before both copies would make the 'mirrors the assembler' claim true.

packages/cli/src/__tests__/template-split.e2e.test.ts
```diff
@@ -286,9 +286,23 @@ describe("Fixture 4 — conflict-check on REAL templates: a synthetic 'mfe-dup'
...
-        { ground: 'src-app/mfe_packages/', contestants: ['frontx-template-mfe-dup', TEMPLATE_MFE_IDENTITY] },
-      ]);
+      // Derived from the real, on-disk template-mfe manifest — same pattern as
+      // Fixture 5's shell case below — rather than pinned to a snapshot: mfe
+      // claims five separate exclusive subtrees nested under
+      // 'src-app/mfe_packages/' (refactor(template-mfe): claim the packages it
+      // ships, not the whole mfe_packages directory), so the dup's one broad
+      // claim collides with each of them separately, and a report naming only
+      // one would leave the rest for the developer to rediscover.
+      const mfeSubtrees = readManifest(TEMPLATE_MFE_DIR).ownershipBoundaries.exclusiveSubtrees;
```
tscbmstubp, line 296: **PLAUSIBLE:** the derived expectation has no lower bound on how many `mfe_packages` claims it expects. The earlier `expect(dupAddResult.reason).toBe('conflict')` (line 287) prevents an empty-expectation pass, but if the manifest shrank to a single `mfe_packages` entry the test would still pass while no longer proving the 'reports every nested claim, not just one' behaviour this change exists for. Fixture 5 guards its derivation with `toContain`; Fixture 4 does not.

Suggest `expect(mfeSubtrees.filter(...).length).toBeGreaterThan(1)` (or a `toContain` on a known entry). The `subtree === 'src-app/mfe_packages/' ? subtree : ...` branch is also unreachable against the real manifest and can go.

tscbmstubp (commented):
Automated review (medium effort, 8 finder angles, 1-vote verify). Six findings CONFIRMED, two marked PLAUSIBLE.

template-shell/src-app/app/mfe/chrome-actions.ts
itechmeat: Fixed in 6d0d92c7 - both schemas target the screen domain type, so an off-domain chain now fails at validation naming the mismatch.

template-shell/src-app/app/layout/Footer.tsx
itechmeat: Fixed in c73d88e1 - Layout no longer mounts an empty Footer, and visible is back as the component's single switch.

packages/ui-kit/src/components/marker/marker.tsx
itechmeat: Fixed in 58eb70c3 - same resolution in Bubble and AttachmentMedia, each with a null-variant test. Left the generic null-excluding helper for a separate pass over every cva component.

template-mfe/src-app/mfe_packages/demo-mfe/src/screens/uikit/components/FormElements.tsx
itechmeat: Fixed as suggested (8453950f).

template-shell/package.json
itechmeat: Fixed - build no longer calls build:packages explicitly (054215cc).

template-shell/src-app/app/mfe/bootstrap.ts
itechmeat: Fixed in 6d0d92c7 - only the no-plugin guards remain; the narrowing TypeScript still needs lives in helpers that throw, naming the mediator gate that owns the invariant.

.github/workflows/main.yml
itechmeat: Fixed - both copies now refuse an occupied destination before writing (ed4e5d15).

packages/cli/src/__tests__/template-split.e2e.test.ts
itechmeat: Fixed in 3e91e824 - a greater-than-one floor on the derived list, and the unreachable branch is gone.

template-shell/src-app/app/mfe/bootstrap.test.ts
```diff
@@ -133,4 +151,92 @@ describe('bootstrapMFE (host-app)', () => {
...
+        const actionTypeId = chromeActions[constantName];
+
+        // An empty `actions` surface is the shell that opted into neither the
+        // themes nor the layout plugin — the only case these handlers still
+        // guard against.
+        const handlers = await chromeHandlersFor({});
+
+        await expect(
+          handlers.get(actionTypeId)?.handleAction(actionTypeId, payload),
+        ).resolves.toBeUndefined();
```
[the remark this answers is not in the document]
itechmeat: Fixed in e635bd2d - a narrowing helper asserts registration for all four cases and names the missing id. One correction: Vitest 4 already refused the undefined in .resolves ("You must provide a Promise to expect()"), so the case failed rather than passed vacuously - just with a message saying nothing about registration; the helper closes that trap for any future assertion that does tolerate undefined.

template-shell/src-app/app/mfe/chrome-actions.ts
```diff
@@ -0,0 +1,79 @@
...
+ * aimed at one of those passes admission and is refused two layers later by
+ * `NoHandlerForActionTargetError`, which names a missing handler rather than
+ * the domain the caller should have targeted; named here, the mismatch is
+ * reported against the pattern at the moment the action is validated.
+ *
+ * These must reach the type system before `registerDomain`, because the
+ * mediator resolves an action's schema from its `type` at dispatch time and
+ * rejects an action it cannot validate.
+ */
+export const CHROME_ACTION_SCHEMAS: JSONSchema[] = [
```
GeraBart, line 47: gts schemas should be defined in json files, not inline ts files

itechmeat: Fixed in feb342d1 - both schemas moved to json files following the mount_ext.v1.json convention; a test pins their ids and targets to the exported constants.

GeraBart (approved):

PR #599, opened 2026-08-27
fix: semgrep exceptions addressed (closes #597)

### LINKED ISSUE #597: Remove semgrep-suppression files; the exclusion rationale references code that doesn't exist (root-cause fix)
## Problem

A corporate security scanner (SecForge, running Semgrep OSS) flags any
change to `.semgrepignore` as a **Required** finding needing manual
Security Team sign-off — sign-off the team will not grant. The original
`.semgrepignore` (and one `// nosemgrep` comment in
`scripts/check-test-dependency-versions.mjs`) exists to suppress
path-injection false positives on test files that build fixture paths via
`path.join(mkdtempSync-root, 'literal')`.

The exclusion rationale in `.semgrepignore`/`.codacy.yaml` claims two
"hardened helper" functions (`joinUnderRoot`, `resolvePathUnderProjectRoot`)
already exist at `packages/cli/src/utils/{fs,project}.ts` — **neither the
functions nor that directory exist anywhere in the repo.** The suppression
is papering over code that was never actually written, and touching
`.semgrepignore` at all is blocked by policy going forward, so the
suppression-based approach is a dead end. This needs an actual fix to the
underlying pattern, not another suppression.

GeraBart:
Closes #597 

## Summary by CodeRabbit

* **New Features**
  * Added safeguards for constructing paths within trusted directories.
  * Added validation for traversal attempts, absolute paths, invalid separators, null bytes, and root escapes.

* **Bug Fixes**
  * Improved filesystem test safety with root-constrained path handling.

* **Tests**
  * Added comprehensive coverage for valid paths, rejected inputs, and boundary conditions.
  * Expanded static analysis coverage across test and development files.


.codacy.yaml
```diff
@@ -24,8 +24,9 @@ exclude_paths:
 
   # Dev-only scripts — spawn() and fs operations take paths from the
   # monorepo's own config, not user input. Not exposed to runtime.
+  # (#597 cleanup removed the duplicate exclusion from .semgrepignore;
```
tscbmstubp: While touching this file: the block further down (~L55–81) still excludes `packages/cli/src/utils/fs.ts` and `packages/cli/src/utils/project.ts` and describes `joinUnderRoot` / `resolvePathUnderProjectRoot` as the trust kernels. Neither file nor function exists in the repo — this is the stale rationale #597 asks to fix. Remove those entries or point them at `internal/test-support/path-guard.ts`.

internal/test-support/__tests__/path-guard.test.ts
```diff
@@ -0,0 +1,160 @@
...
+
+  it('accepts nested valid segments joined in sequence', () => {
+    const root = makeRoot();
+
+    expect(joinWithinRoot(root, 'nested', 'fixture.txt')).toBe(
+      path.join(root, 'nested', 'fixture.txt'),
+    );
+  });
+
+  // F-008: segments that merely *contain* '.' or '..' as a substring (not an
```
tscbmstubp: Nit: `F-008`, `F-002`, `G-004`, "Phase 2" are planning-artefact IDs that mean nothing to a reader of the repo. Keep the behavioural sentence, drop the IDs.

internal/test-support/__tests__/path-guard.test.ts
```diff
@@ -0,0 +1,160 @@
...
+// `joinWithinRoot` is the #597 fix's real path-containment helper, now
+// implemented at `../path-guard.ts`: it lets the ~11 test files across
+// packages/* that build fixture paths under a `mkdtempSync` root with
+// unguarded `fs.*` calls stop tripping the corporate semgrep scanner's
+// path-injection rule — without a `.semgrepignore`/`// nosemgrep`
+// suppression (any edit to `.semgrepignore` requires Security Team sign-off
+// the team will not grant). `.semgrepignore` and `.codacy.yaml` used to
+// reference `joinUnderRoot`/`resolvePathUnderProjectRoot` as if they existed
+// in `packages/cli/src/utils/fs.ts`; that file does not exist. This helper
+// is the real one those comments should have pointed to.
```
tscbmstubp, line 15: This comment correctly notes that `.codacy.yaml` references helpers that don't exist — but the PR only removes them from `.semgrepignore`. `.codacy.yaml` still carries them (see my comment there).

tscbmstubp (changes requested):
Reviewed the helper, the migration, and CI.

**What's good**
- `joinWithinRoot` is correct: per-segment screening plus a `path.relative` containment check that correctly accepts `..hidden`. I ran the 14 tests locally against vitest 4.1.4 — all pass.
- Dep-cruiser edges, workspace wiring, exact `vitest` pin, and the cli/kit version bumps with their `core.toml` / `.cf-studio-kit.toml` pins are consistent.

**Blocking**
1. **The root cause of #597 is still in `.codacy.yaml`.** Lines ~55–81 still exclude `packages/cli/src/utils/fs.ts` and `packages/cli/src/utils/project.ts` and describe `joinUnderRoot` / `resolvePathUnderProjectRoot` as trust kernels. None of those files or functions exist. Please remove that block or rewrite it to point at `internal/test-support/path-guard.ts`.
2. **Two fixture-path test files are not migrated:** `packages/cli/src/adapters/__tests__/fs-read-content-items.test.ts` and `packages/cli/src/adapters/__tests__/provenance-io.test.ts`. Both build paths under a `mkdtempSync` root with `fs.*` calls. With `.semgrepignore` gone they will now surface as findings.

**Non-blocking**
- The failing `Template Validate` job (demo-mfe type errors) also fails on `develop`'s last three runs — pre-existing, not this PR.
- `package-lock.json` flips ~40 `"peer": true` flags — npm normalisation, harmless noise.

.codacy.yaml
```diff
@@ -35,6 +36,8 @@ exclude_paths:
 
   # CLI scaffold templates — code meant to be copied into new projects,
   # not to be linted in the kit itself.
+  # (#597 cleanup removed the duplicate exclusion from .semgrepignore;
+  # this entry is now the sole remaining exclusion mechanism for this path.)
```
itechmeat: 🟡 [Not a blocker] Two things in this block: `packages/cli/template-sources/` does not exist at head or at merge-base - same class as the `packages/cli/scripts/**` entry deleted above, so it can go rather than get a comment. And the "duplicate exclusion / sole remaining mechanism" wording (here and at lines 27-28) treats Codacy and the corporate Semgrep as one mechanism; after `.semgrepignore` is gone there is no Semgrep-side exclusion left for these paths at all, which is the point of #597 but reads as the opposite. Worth folding into the same .codacy.yaml pass the existing review already asks for.

internal/test-support/path-guard.ts
```diff
@@ -0,0 +1,89 @@
...
+ */
+export function joinWithinRoot(root: string, ...segments: string[]): string {
+  for (const segment of segments) {
+    if (!isSafePathSegment(segment)) {
+      throw new Error(`joinWithinRoot: unsafe segment: ${JSON.stringify(segment)}`);
+    }
+  }
+
+  const resolvedRoot = path.resolve(root);
+  const joined = path.resolve(resolvedRoot, ...segments);
```
itechmeat, line 78: 🟡 [Not a blocker] Was the corporate scan run against this branch? This line is `path.resolve` with a spread variable - the same shape the rule fires on - and `internal/**` is in no exclude list now that `.semgrepignore` is gone, while the old `.codacy.yaml` gave a trust-kernel exclusion to a helper that never existed. Concentrating the finding into one reviewed kernel is the right remediation; a line in the description saying the scan came back clean would close the question.

internal/depcruise-config/layer-constants.cjs
```diff
@@ -71,7 +71,7 @@ const PROJECTS_ORCHESTRATION_PACKAGES = Object.freeze(['cli', 'cyber-pilot-kit-f
 // even though the edge only exists at lint/build time. Exempt from the member
 // artifact chain and the publication gate, with that scope stated in the
 // DESIGN rather than implied. Directory names under `internal/`.
-const BUILD_INTERNALS_PACKAGES = Object.freeze(['eslint-config', 'depcruise-config']);
+const BUILD_INTERNALS_PACKAGES = Object.freeze(['eslint-config', 'depcruise-config', 'test-support']);
```
itechmeat, line 74: 🟡 [Not a blocker] DESIGN.md §1.3 defines build internals as packages that "exist only to configure the build" (line 139, and "configuration packages" in the table at 173), and the comment right above this line says the exemption's scope is stated in the DESIGN rather than implied - but `test-support` ships executable code used at test time and DESIGN is untouched here. Could the definition widen to build- and test-time internals in the same PR, so the guard and the architecture agree on what the bucket means?

internal/test-support/path-guard.ts
```diff
@@ -0,0 +1,89 @@
...
+ * Calling with zero segments is an intentional no-op: it returns
+ * `path.resolve(root)` unchanged.
+ *
+ * @throws {Error} with a message matching `/unsafe segment|escapes root/i`
+ *   if any segment is unsafe or the joined result escapes `root`.
+ */
+export function joinWithinRoot(root: string, ...segments: string[]): string {
+  for (const segment of segments) {
+    if (!isSafePathSegment(segment)) {
+      throw new Error(`joinWithinRoot: unsafe segment: ${JSON.stringify(segment)}`);
```
itechmeat: 🟡 [Not a blocker] Throwing is the right strategy here (an unsafe literal segment in a fixture is programmer error), but the form binds eleven tests to message text through the `/unsafe segment|escapes root/i` regex the JSDoc promotes to a contract - rewording either message breaks them all. A small `PathGuardError` with `code: 'UNSAFE_SEGMENT' | 'ESCAPES_ROOT'`, asserted on `code`, is what the house rules ask for and costs a few lines.

internal/test-support/path-guard.ts
```diff
@@ -0,0 +1,89 @@
...
+  // containing a path separator, null byte, or ':' — no segment can ever
+  // combine with others to produce a `joined` path outside `resolvedRoot`,
+  // so this branch is not currently reachable/exercised by any test.
+  //
+  // The separator-qualified `..`-prefix check below (`!== '..'` and
+  // `!startsWith('..' + path.sep)`) is also a deliberate correction of the
+  // reference implementation this helper is modeled on
+  // (`resolvePackageJsonPathWithinRoot` in
+  // `scripts/check-test-dependency-versions.mjs`), which uses a naive
+  // substring-only `startsWith('..')` that would false-positive-reject a
```
itechmeat: 🔵 [Not a blocker] This comment names the scripts copy as the flawed original, but the same `startsWith('..')` check also lives in shipped code - `packages/cli/src/adapters/fs-installed-content-path.ts:25`, inside the boundary-invariant `@cpt` region - and neither copy is touched here. Fixing production is outside #597, fair; but could the comment name both, and the script either import this helper or get a pinned follow-up, so a known bug is not left documented only in its copy?

internal/test-support/path-guard.ts
```diff
@@ -0,0 +1,89 @@
+import path from 'node:path';
```
itechmeat, line 1: 🔵 [Not a blocker] The module opens on an import; everything durable about it (test-only, carries no `@cpt` marker, traces to no instruction, does not defend against symlinks) lives in the test file's 26-line preamble where a reader of packages/cli's tests never looks. Moving that half into a module doc block here would also shrink the preamble already flagged in the existing review.

internal/test-support/path-guard.ts
```diff
@@ -0,0 +1,89 @@
...
+  for (const segment of segments) {
+    if (!isSafePathSegment(segment)) {
+      throw new Error(`joinWithinRoot: unsafe segment: ${JSON.stringify(segment)}`);
+    }
+  }
+
+  const resolvedRoot = path.resolve(root);
+  const joined = path.resolve(resolvedRoot, ...segments);
+  const relativeToRoot = path.relative(resolvedRoot, joined);
+  // Defense-in-depth: this containment check guards against a *future*
```
itechmeat, line 80: 🔵 [Not a blocker] The comment says the containment branch below is unreachable given the segment screen and that no test exercises it. Either is fine as a design choice, but the house rule wants even an "unreachable" guard to have a test proving it refuses - driven through an internal seam, say - and the rationale essay belongs in the PR description rather than the file. Keep the check with a test, or drop it; (up to you).

internal/depcruise-config/layer-constants.cjs
```diff
@@ -159,7 +159,11 @@ const ALLOWED_ECOSYSTEM_EDGES = Object.freeze({
   'ui-kit': Object.freeze({ runtime: Object.freeze([]), dev: Object.freeze([]) }),
 
   // ---- Projects orchestration ----
-  cli: Object.freeze({ runtime: Object.freeze([]), dev: Object.freeze([]) }),
+  cli: Object.freeze({
+    runtime: Object.freeze([]),
+    // Test-only path-containment helper, exercised by the CLI's own test suite.
+    dev: Object.freeze(['@gears-frontx/test-support']),
```
itechmeat, line 165: 🔵 [Not a blocker] `dev` here only groups devDependencies for the manifest check in package-edge-tests.ts - no depcruise rule, no `no-restricted-imports` entry names `test-support`, and `pkgTargets()` enumerates `packages/*` only, so a `path-guard` import from `packages/cli/src/adapters/*.ts` passes every check green and breaks the published build. Either a small depcruise rule (from `packages/*/src` excluding tests, to `test-support`) or a note in the PR that the edge is unenforced - the comment alone claims more than the repo holds.

packages/cli/src/adapters/__tests__/local-fetch.test.ts
```diff
@@ -133,25 +134,25 @@ describe('offline e2e — frontx install + seed assemble the real template-shell
       expect(seedResult.appliedTemplates).toEqual([templateIdentity]);
 
       // Representative files inside declared exclusive subtrees materialize as REAL on-disk files.
-      expect(fs.existsSync(path.join(targetDir, 'package.json'))).toBe(true);
-      expect(fs.existsSync(path.join(targetDir, 'src', 'index.ts'))).toBe(true);
-      expect(fs.readFileSync(path.join(targetDir, 'src', 'index.ts'), 'utf-8')).toBe(
+      expect(fs.existsSync(joinWithinRoot(targetDir, 'package.json'))).toBe(true);
+      expect(fs.existsSync(joinWithinRoot(targetDir, 'src', 'index.ts'))).toBe(true);
+      expect(fs.readFileSync(joinWithinRoot(targetDir, 'src', 'index.ts'), 'utf-8')).toBe(
```
itechmeat, line 139: 🔵 [Not a blocker] The added line routes one `readFileSync` through the guard while line 140 right below feeds the same sink `path.join(TEMPLATE_SHELL_DIR, 'src', 'index.ts')` (same at 100). If `TEMPLATE_SHELL_DIR`-rooted reads and the `mkdtempSync(path.join(os.tmpdir(), ...))` seeds are deliberately outside the helper's contract - trusted root, only segments screened - one sentence saying so would save the next reader the question; as it stands the pair on adjacent lines looks like a stopped-halfway sweep.

packages/cli/src/adapters/__tests__/fs-content-store.test.ts
```diff
@@ -44,8 +45,8 @@ describe('FsContentStore', () => {
     store.write('my-template', JSON.stringify({ $frontxTemplateFiles: bundle }));
 
     const installedPath = resolveInstalledContentPath(root, 'my-template');
-    expect(fs.readFileSync(path.join(installedPath, MANIFEST_FILENAME), 'utf-8')).toBe(bundle[MANIFEST_FILENAME]);
-    expect(fs.readFileSync(path.join(installedPath, 'src/index.ts'), 'utf-8')).toBe(bundle['src/index.ts']);
+    expect(fs.readFileSync(joinWithinRoot(installedPath, MANIFEST_FILENAME), 'utf-8')).toBe(bundle[MANIFEST_FILENAME]);
```
itechmeat, line 48: 🔵 [Not a blocker] Adding to the two unmigrated files already listed: `packages/cli/src/__tests__/helpers/template-split-fixtures.ts` is the same class - `mkdtempSync(path.join(os.tmpdir(), prefix))` at 84, `readFileSync(path.join(templateDir, MANIFEST_FILENAME))` at 94, `path.join(root, relativeDir)` into `readdirSync` at 104. Being a `.ts` helper rather than `.test.ts`, the old `.semgrepignore` globs never covered it either, so it is the one file in the set that was already exposed and stays so.

internal/test-support/tsconfig.json
```diff
@@ -0,0 +1,16 @@
+{
+  "compilerOptions": {
+    "target": "ES2022",
+    "module": "ESNext",
+    "moduleResolution": "bundler",
+    "outDir": "./dist",
```
itechmeat: ⚪ [Not a blocker] `outDir`/`declaration`/`declarationMap` are set but there is no `build` script and `exports` points at the raw `.ts`, so these options are dead - both sibling internals ship built output. Separately, `exports` has no `"."` entry unlike both siblings, so the bare `@gears-frontx/test-support` specifier does not resolve. Fine for a private vitest-only package if intended; just flagging the divergence.

internal/test-support/__tests__/path-guard.test.ts
```diff
@@ -0,0 +1,160 @@
...
+    expect(joinWithinRoot(root, 'v1.2.3')).toBe(path.join(root, 'v1.2.3'));
+  });
+
+  it('rejects a ".." segment', () => {
+    const root = makeRoot();
+
+    expect(() => joinWithinRoot(root, '..')).toThrow(UNSAFE_SEGMENT_ERROR);
+  });
+
+  // F-002(a): a dangerous segment appearing after a valid one must still be
```
itechmeat: ⚪ [Not a blocker] Same as the note at line 63: `F-002(a)` here, `F-002(b)` at 132 and `G-004` at 153 are planning-artefact IDs a repo reader cannot resolve - one sweep covers all four.

internal/test-support/path-guard.ts
```diff
@@ -0,0 +1,89 @@
...
+ * Modeled on `isSafePathSegment()` in
+ * `scripts/check-test-dependency-versions.mjs` (~L158), extended with an
+ * explicit `:` rejection: on POSIX hosts `path.relative`/`path.resolve`
+ * based containment checks do not treat `:` specially (colons are legal in
+ * POSIX filenames), so a Windows drive-relative segment such as `C:temp.txt`
+ * would otherwise slip past character screening on this dev machine. This
+ * check rejects any segment containing `:` regardless of host OS.
+ */
+function isSafePathSegment(segment: string): boolean {
+  if (typeof segment !== 'string' || segment.length === 0) {
```
itechmeat, line 31: ⚪ [Not a blocker] The `typeof` half of this check is unreachable under the type system and nothing says which untyped boundary it guards (a JS caller?). Either drop it or name the boundary in a few words.

internal/test-support/__tests__/path-guard.test.ts
```diff
@@ -0,0 +1,160 @@
...
+  // F-002(b): a Windows drive-relative segment contains no path separator and
+  // is not '.', '..', or empty, so a *separator-only* screen would let it
+  // through. It is actually caught by the explicit `:` rejection in
+  // `isSafePathSegment` (character screening), not by the containment check
+  // — segments can never contain a path separator, so no segment can ever
+  // combine with others to escape `root` via containment alone. Built with
+  // `path.win32` explicitly so this assertion is platform-independent (a
+  // literal `'C:temp.txt'` string behaves the same on every OS the test
+  // suite runs on).
+  it('rejects a drive-relative-style segment via character screening (":" rejection)', () => {
```
itechmeat, line 152: ⚪ [Not a blocker] `via character screening (":" rejection)` describes the implementation; the first half of the title already states the behaviour and condition. The rest of the file's titles are fine.

itechmeat (approved):
Approving on the substance: a real containment helper with its own tests in place of a suppression file is exactly the root-cause fix #597 asked for, the version bumps and pins are tooling-mandated rather than noise, and the migration inside the touched files is careful. The two open points in the existing review still need to land before merge and I will not repeat them; my comments extend them slightly (one more dead .codacy.yaml entry, one more unmigrated fixture file) and otherwise ask about the new package's documentation home, its error contract, and whether the corporate scan was actually run against the branch. None of it changes the direction.

.codacy.yaml
GeraBart: Fixed in b59a5437 — removed all four dead `exclude_paths` entries and their fabricated `joinUnderRoot`/`resolvePathUnderProjectRoot` rationale (`packages/cli/src/utils/fs.ts`, `packages/cli/src/utils/project.ts`, `packages/cli/src/commands/ai/sync.ts`, `packages/cli/src/generators/project.ts`). Confirmed none of those paths exist in the repo. You're right that the original PR only cleaned up `.semgrepignore` and missed this block entirely.

internal/test-support/__tests__/path-guard.test.ts
GeraBart: Fixed in b59a5437 — dropped the `F-008`, `F-002(a)`, `F-002(b)`, `G-004` planning-artifact ID prefixes and the outdated "Phase 2" forward-reference framing, keeping each comment's actual behavioral explanation.

internal/test-support/__tests__/path-guard.test.ts
GeraBart: Fixed in b59a5437, alongside the `.codacy.yaml` cleanup — removed the stale `joinUnderRoot`/`resolvePathUnderProjectRoot` references there too, not just from `.semgrepignore`.

.codacy.yaml
GeraBart: Fixed in b59a5437 — removed the `packages/cli/template-sources/**` entry (confirmed gitignored/untracked, same class as the already-removed `packages/cli/scripts/**`), and reworded the `scripts/**` comment: it no longer frames Codacy's exclusion as "the sole remaining mechanism" alongside Semgrep — after `.semgrepignore` is gone there's no Semgrep-side exclusion left for this path at all, which was the actual point. Good catch.

packages/cli/src/adapters/__tests__/fs-content-store.test.ts
GeraBart: Fixed in the review-comments commit — migrated `template-split-fixtures.ts`'s `readManifest` to `joinWithinRoot`, and left the recursive `listRealFiles` path.join alone with a comment explaining why (it accumulates a genuinely dynamic multi-segment path across recursive calls, not individually-known segments). Thanks for catching this one, it slipped through because the original migration only searched `.test.ts`/`.test.tsx` files.

internal/test-support/__tests__/path-guard.test.ts
GeraBart: Fixed alongside the F-008/F-002 cleanup on the other thread — F-002(a), F-002(b), and G-004 are all gone now, same pass.

internal/depcruise-config/layer-constants.cjs
GeraBart: Fixed — added `frontx-test-support-test-only` to `.dependency-cruiser.cjs`: forbids any non-test file under `packages/*/src` from importing `@gears-frontx/test-support`, closing the gap between the manifest-level `dev` edge and actual enforcement. Verified it fires (added a throwaway production import, confirmed the rule flagged it, reverted) and that it doesn't false-positive on the legitimate test-file imports.

internal/test-support/path-guard.ts
GeraBart: Fixed — added a `PathGuardError` class with a `code: 'UNSAFE_SEGMENT' | 'ESCAPES_ROOT'` field. All 14 tests now assert on `.code` instead of matching the message regex, so the message text is free to change without breaking every consumer.

internal/test-support/tsconfig.json
GeraBart: Fixed — removed the dead `outDir`/`declaration`/`declarationMap` options (no build script, nothing consumes `dist/`) and added the missing `"."` export entry pointing at `./path-guard.ts`, matching the sibling `internal/*` packages' convention.

internal/test-support/path-guard.ts
GeraBart: Fixed — widened the comment to name both flawed copies: `resolvePackageJsonPathWithinRoot` in `scripts/check-test-dependency-versions.mjs` and `assertWithinRoot` in `packages/cli/src/adapters/fs-installed-content-path.ts` (production code, inside the `cpt-frontx-algo-template-resolution-bounded-update` boundary invariant). Left both untouched per your note that fixing production is out of scope for #597 — the comment now says so explicitly instead of only documenting the bug in this file's own copy.

PR #600, opened 2026-08-28
fix: render chart's per-instance style via JSX children

### LINKED ISSUE #597: Remove semgrep-suppression files; the exclusion rationale references code that doesn't exist (root-cause fix)

GeraBart:
## Summary by CodeRabbit

* **Security**
  * Improved chart styling safeguards by rendering generated CSS through React-managed style content.
  * Preserved sanitization of CSS identifiers and values to help prevent injection risks and hydration mismatches.

* **Documentation**
  * Expanded guidance on CSS injection behavior, server-side escaping, and client-side handling.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9164,7 +9164,7 @@
     },
     "packages/cli": {
       "name": "@gears-frontx/cli",
-      "version": "0.3.0-alpha.2",
+      "version": "0.3.0-alpha.3",
       "license": "MIT",
       "bin": {
         "frontx": "dist/cli.js"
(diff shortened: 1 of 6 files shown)


itechmeat (approved):
The change is correct and safe, and it was the last such site in the repo. To merge, bump the `ui-kit` version and update the two matching pins in `template-mfe`.

PR #585, opened 2026-08-18
docs(routing): routing member artifacts — navigation substrate with a pluggable engine provider

gs-layer:
## What

Replaces the routing member's addressing model with one query-only grammar for every domain in the tree, at any depth and any parameter count. The path stays the shell's own; the package reads and writes only the query string. Each mounted extension gets one entry: a repeated domain key means several extensions in that domain, and a dotted key names a domain nested inside another extension's own zone. The model reference link, exactly as it reads in the address bar:

```
/en?screen=dashboard;orientation=left&sheet=tenant-details;tenantId=456&sheet=user-contacts;contactId=123;view=active&widgets=line-a;range=7d&widgets=line-b;range=30d&widgets=pie;metric=revenue
```

Documentation only — this PR adds no source under `packages/*/src`.

## Why

The prior model reserved the path for one privileged domain per zone and used two different parameter encodings depending on whether a domain was that privileged one. Neither rule had a principled basis. The grammar adopted here treats every domain the same: no privileged domain per zone, the path returned to the shell as its own territory, and every parameter readable in place rather than nested inside a second encoding. The result is one uniform topology, `<shell-path>?[domain=extension;param=value;...]*`, for a screen, a sheet, a widget, or a domain nested inside any of them.

## What changed

- Four member ADRs: occupant reference boundary and mount-trigger ownership carried over; the addressing decision rewritten around the new grammar; identity stability adapted for this package.
- Core PRD and DESIGN rewritten on top of the query-only base, with the addressing decision as the single source of the grammar and DESIGN reduced to a summary that cites it.
- Core FEATUREs: prefix-resolution replaced by the grammar's codec algorithms (parse, serialize, name validity, domain-key composition); the route-ownership signal now observes one domain at a time and reports an ordered entry list with a diff (added, removed, payload-changed, replaced, reordered).
- Provider docs: an entry address (domain key and extension) replaces the base-path parameter; the provider builds a virtual location from a reserved `route` parameter and projects it either onto a mounted extension's own entry or, standalone, directly onto the page.
- Root ADR 0009: the URL-projection paragraph rewritten in place to point at the member's own addressing decision instead of restating an older grammar.
- Studio config: `artifacts.toml` ADR pattern relaxed for the two routing members.
[...]

PR #602, opened 2026-08-31
Feat: routing prd + design

GeraBart:


gs-layer (commented):
Good work, and the two problems you're solving are real ones. Here's where we agree and where we don't.

## We agree with

**Mount trigger ownership.** This corrects us. We had said a menu click should navigate and the mount should follow from the URL. You're right that this creates a second writer next to the mediator, and mount races were where our own state machine collected most of its bugs. Actions originate mounts, navigation replays history — better decision.

**Occupant reference boundary.** Data pairs passed as an argument, adaptation in glue. This is our own boundary written down as a decision. One ask: point at the field-level shape instead of repeating it, so there's one owner for the schema.

**Member ADRs as the place for these.** We were wrong to question this. `artifacts.toml` on develop already says it: owning an ADR is a member's option, and only DECOMPOSITION is forbidden to a member. ADR-0031 under CLI is the precedent. Your placement is right — it just needs an ADR pattern registered for the routing children, otherwise nothing discovers the files.

**Identity from `route`, not the GTS id.** Right, and the argument is decisive: a GTS id carries a version, so shipping a new version of the same screen would break every bookmark and nothing in the diff would look wrong. This closes a follow-up we had left open.

**Route support scope.** "All screens have routes, no modals do" is a property of the type, not of an instance, so checking it at registration is the right place.

## We don't agree with: moving everything into the query string

The grammar itself is sound. Two delimiters, an ancestor's match provably failing against a descendant's key, and the uniqueness proof in ADR 0003 — all of that holds. Our objection is about where the hierarchy lives.

Your model doesn't remove the hierarchy from the URL. It moves it out of the pathname and into a private syntax inside the query string. `screen.tenants.tabs::contacts` is still a path — just written in a format only this library understands, in a part of the URL that servers, caches, CDNs, analytics and log tools treat as unordered opaque pairs. The pathname is the one part they all already understand.

The second problem is the line in your DESIGN that says an application is free to put "something cosmetic" in the pathname. That gives two outcomes, both bad. Either the pathname stays empty for the most common state in a console — one screen open, address is `/` — or someone fills it for looks, and now there are two versions of which screen is open, and they drift. Keeping the pathname authoritative for the main chain means there's only one.

On uniformity: check our current head (`1d813af1`, newer than the base you rebased from). The carrier generalization there already unified the mechanism — one observer parameterized by its carrier, one resolution primitive, one back-projection helper. What's left isn't two mechanisms. It's one mechanism over two carriers, and only for where the main chain gets written.

## What we're taking from you

Your key composition rule is better than ours and we want it. On our side, keeping parallel-axis keys unique is a convention we ask consumers to follow. Your version makes it structural, and it works on parallel-axis keys without touching the pathname. You'd be co-author of that part.

## Your two requirements, closed on our side

**Concurrent occupancy.** Our one-entry limit was a deferral with a written trigger, and your dashboard example looks like a real consumer — please confirm. Lifting it needs no new grammar: repeated query keys are standard, and the discriminator is already there, because a parallel key's value starts with the occupant's declared segment. Closing one of several means replacing the entry whose value starts with that segment.

The real cost is set-shaped transitions — added and removed, instead of previous and new. That cost belongs to the requirement, not to either model, and right now neither of us has paid it: with several keys sharing a domain path, your transition shape and your close-one-of-several recipe aren't specified either.

**Nested extension domains.** Already supported since `9d287d27` and `063400dd` — resolution runs level by level, and each level hands its nested domains a base. Your composition rule improves how those keys are named; it isn't what makes nesting work.

## Three things we'd ask you to put back

These look like side effects of moving content, not decisions:

1. **The three FEATUREs.** LAYER-2 requires PRD, DESIGN and at least one FEATURE per member, and `artifacts.toml` still marks FEATURE as required for both routing members — which is why Studio Validate Artifacts fails here. They also hold the field-level schemas and the anchor points for future `@cpt` markers in code.
2. **The DECOMPOSITION anchors.** All 26 routing references are gone. Root DESIGN §1.5 needs them while the installed kit does.
3. **The root ADR 0009 note.** It currently describes both models at once — it still prefers the axial projection — and it links to a FEATURE file this PR deletes.

## Questions

1. Who actually needs concurrent projection? We'd rather lift our deferral against a named consumer. And would one key per domain with a composite value serve them as well as several keys?
2. In a composed app, what does a provider-built router match against, if the substrate never reads the pathname? Your tanstack DESIGN covers standalone; composed is where our two-package split assumed a pathname remainder under a basepath.
3. If only actions can originate a mount, what happens on a plain in-app `<a href>` click? That's a new URL, not a history replay, so nothing would mount — unless every screen switch becomes an onClick that dispatches an action, which gives up native links.
4. When a parent occupant's parameter changes — `tenantId=456` to `789` — your composition rule leaves the child domain's key unchanged, so a nested occupant stays mounted through a full change of data context. Intended?
5. Domain names look unconstrained: ADR 0001 limits the alphabet for occupant identity, but a domain named with `::` in it would forge a boundary. Also unstated: are keys matched raw or after percent-decoding?

Details are in the inline comments, and the plan for doing the work is in a separate comment below.

packages/routing/architecture/DESIGN.md
```diff
@@ -0,0 +1,507 @@
...
+
+The pathname is not part of this model at all. It is purely an application-level concern: an application remains free to put something cosmetic there, and to scope a mounted microfrontend's own router to a `basepath` within it, but this package's own resolution primitive never treats a pathname segment as a declared prefix and this package never writes the pathname to reflect occupancy. That is a real, deliberate loss of the readable pathname address the retracted model produced, accounted for where §4's first worked example lands. A domain that never projects into the URL at all simply stays runtime state this package takes no part in — projection is always an opt-in act of that domain's own consuming level, never a default this package imposes. A hash carries whatever the application chooses, ungoverned by anything in this model. How many occupants a domain may hold mounted at once is decided elsewhere and stays outside this mechanism's own scope (`cpt-frontx-adr-extension-domain-occupancy`, URL projection of domain occupancy); this mechanism only addresses whatever cardinality that decision admits.
```
gs-layer, line 56: This paragraph is where we disagree, and it states the cost honestly, which we appreciate.

Two things follow from "the pathname is not part of this model at all" that we don't think we should accept:

**Hierarchy moves out of the carrier built for it.** `screen.tenants.tabs::contacts` is still a hierarchical path — expressed in a private syntax inside a carrier the rest of the stack treats as unordered opaque pairs. The pathname is understood by server rewrite, caches, CDN path rules, analytics, log parsing and by a human reading the address bar; a query string is routinely truncated, reordered or dropped by that same stack.

**"Something cosmetic there" is a second account of the same fact.** Either the pathname stays permanently empty for the most common state in a console (one screen open → `/`), or somebody fills it for looks and it drifts from the authoritative account in the query. Keeping the pathname authoritative for the main chain removes that second account by construction.

What we do want from this section is the composition rule itself, applied to parallel-axis keys — see the note on ADR 0003.

packages/routing/architecture/ADR/0003-domain-occupancy-addressing-granularity.md
```diff
@@ -0,0 +1,148 @@
...
...declared-prefix conflict check — including between two mounts of the identical microfrontend entry, since each mount is its own distinct registration with its own independently declared prefix, never one registration shared between them (DESIGN §2.3, O3b); and (3) root-zone domain-name uniqueness, the host's own single choice at the tree's root. By induction on tree depth: a root domain's own domain-path is unique by fact (3); a nested domain's own domain-path is composed from its enclosing occupant's own domain-path (unique by the inductive hypothesis), that occupant's own occupant-identity (unique within its own domain by fact (2)), and its own locally-chosen name (unique among its own siblings by fact (1)) — so two nested domains can share an identical domain-path only if every one of those three already-guaranteed-distinct components collides simultaneously, which the three facts jointly rule out, including across two domains that happen to reuse an identical locally-chosen name under two different parents. Combined with occupant-identity uniqueness within a domain (fact 2 again, at the final segment), the pair — and therefore the composed key — is unique across the whole tree.
```
gs-layer, line 56: This induction is the part of the proposal we want to adopt, and we'd credit you as co-author of it.

On our side, uniqueness of parallel-axis keys is a binding-obligation convention ("prefix your key with your zone"). Your three local facts plus the ancestor-can't-shadow-descendant property make it structural instead, and it transplants onto parallel-axis keys without touching the pathname. After that, both carriers follow one principle — identity is the declared segment, hierarchy is composition.

Also: this section settles the same-entry question convincingly (each mount is its own registration with its own declared prefix), so the "two instances of one entry are indistinguishable" objection doesn't apply — worth keeping that paragraph wherever this lands.

One gap: the alphabet restriction in ADR 0001 covers occupant-identity, but a domain's locally-chosen name looks unconstrained here. A domain named with `::` in it would forge a boundary delimiter and make an ancestor's key parse as a descendant's. Related and unstated: whether a candidate key is matched raw or after percent-decoding (`%3A%3A` → `::`).

packages/routing/architecture/ADR/0002-mount-trigger-ownership.md
```diff
@@ -0,0 +1,122 @@
...
...cold load and replays history restoration**, because it is the only option that keeps the runtime's occupancy authority single-writer while still resolving correctly on cold load, reload, and back/forward. Originating a new mount or unmount intent — one not yet represented in any history entry — happens exclusively through actions-chains post-boot; navigation is never the origin of new intent once the application has booted. Restoring — a back/forward step, or any navigation moving the location to a state a history entry already represents — is a different thing entirely: it causes each affected level to re-resolve and its consumer to reconcile its mounted set, exactly as `cpt-frontx-routing-seq-deep-link-cold-mount`'s own back/forward fan-out already shows ("report transition, if ownership-relevant"), and exactly as DESIGN §4's own worked example already shows ("a back step taken right after opening the modal ... the console's own glue unmounts the modal in response to the modal domain's own reported transition, not through any action of this library's own"). Restoring is not a second occupancy driver alongside actions-chains; it is that same driver's own prior decision, replayed.
```
gs-layer, line 46: Accepted, and it corrects us: we had earlier recommended that a menu click navigate and the mount follow from the URL. Your point that this puts a second writer beside the mediator is right, and occupancy races were where our own state machine collected most of its findings.

One case we'd like specified before this lands: an ordinary in-app `<a href>` click. That's a pushState of a URL not yet represented in any history entry — so by this decision it originates no mount, and it isn't a replay either. Either every screen switch becomes an onClick dispatching an action (which gives up native links, though cmd-click still works as a cold load), or there's a third category here that the record doesn't name yet.

Second, smaller: cold load with several occupants already in the URL. The record says navigation handles cold-load resolution, and reconciliation acts on "already-authorized states" — what authorizes an unfamiliar deep link carrying three occupants?

template-shell/packages/framework/architecture/ADR/0001-occupant-identity-stability.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Decision Outcome
+
+Chosen option: **the extension's own `route` property as identity**, because it is the only one of the two that stays stable across exactly the kind of change a deep link or bookmark must survive: a new version of a microfrontend or extension. A GTS id is versioned by construction — `gts.frontx.mfes.ext.extension.v1~frontx.screensets.layout.screen.v1~frontx.demo.screens.helloworld.v1` names a specific version of a specific derived type — so shipping a new version of that same conceptual screen changes the id string a URL segment would have resolved against, silently breaking every previously working or bookmarked route pointing at it, with no code anywhere having done anything a reviewer would flag as wrong. `route` carries no version information at all: it is a human-authored, presentation-level path segment (`packages/mfes/src/types/index.ts`, `ExtensionPresentation.route`) that a redeployment of the same conceptual screen has no structural reason to change, and that already exists on `ScreenExtension` today without requiring a new field.
```
gs-layer, line 44: The substance is right and closes a follow-up we had left open. The versioned-GTS-id argument is the decisive one: a redeployment of the same conceptual screen would change the id a URL resolved against and silently break every bookmark, with nothing in the diff a reviewer would flag.

The placement is the problem. `template-shell/**` is template territory, which `cpt-frontx-adr-template-territory-traceability` puts outside the ecosystem artifact universe — those globs are in the `artifacts.toml` ignore list, so this file is invisible to validation. Your own `packages/routing/architecture/ADR/0001` cites `cpt-frontx-framework-adr-occupant-identity-stability`, which leaves an authoritative decision depending on a non-authoritative document — and this is the decision bookmarks rest on.

Where we'd put it: this is a glue-level decision about which existing `mfes` field to use, changing nothing in `mfes` itself, so a routing member ADR or DESIGN §2.3 Obligations fits. `0002-route-support-scope` is different — it asks the runtime to validate a per-type invariant at registration, so that one belongs with `mfes`.

Worth checking whether these two files are also what turns `Template Validate` red.

architecture/ADR/0009-extension-domain-occupancy.md
```diff
@@ -97,6 +97,8 @@ The present concrete instantiation ships three strategy classes — `ConcurrentM
...
...rallel axis, regardless of which of these strategies it runs. Either way, a navigation act is an entry into the same mount mechanism this decision governs, not a second occupancy mechanism running alongside it. The URL is the source of truth for that reflection, and occupancy is its derivative; the reverse mapping — recovering the URL from an occupancy change — applies only when a mount is triggered by something other than navigation. The routing package publishes that reflection as a signal, not as a guarantee it keeps itself: its Route Ownership Signal feature ([packages/routing/architecture/features/route-ownership-signal/FEATURE.md](../../packages/routing/architecture/features/route-ownership-signal/FEATURE.md)) exposes the owner-resolution primitive, an observable owner-change signal, and a URL back-projection helper for either projection, but the binding mechanics — actually mounting or unmounting an occupant in response to that signal, and resolving a race between two mounts — are performed by the host's own glue on top of this decision's mount strategies and registry, not by the routing package. This decision continues to own the mount strategies and the cardinality matrix.
```
gs-layer, line 100: This note now carries both models at once. It still says the axial projection is preferred and describes a parallel axis as zero-or-one — the model your own member DESIGN retracts — while also lifting the concurrent deferral.

It also cites `packages/routing/architecture/features/route-ownership-signal/FEATURE.md`, which this same PR deletes, so the reference dangles.

And the direction: the note leans on `packages/routing/architecture/ADR/0003-...` — a member-owned identifier cited from a root artifact, which LAYER-3 asks us to avoid. ADR 0002 handles the same situation by naming the member and its path instead of its identifiers; the same shape would work here.

gs-layer:
Plan, plainly: we'll pull your work into #585 instead of reworking it here, so the artifact chain and the review threads stay in one place. You'll be co-author on the commits that carry your files.

**What we take from you:**
- All three routing ADRs — occupant reference boundary, mount trigger ownership, and the key composition grammar.
- Identity from `route` instead of the GTS id. The versioning argument settles it.
- Route support scope as a per-type invariant checked at registration.
- Your key composition rule for parallel-axis keys. Ours was a naming convention; yours makes it structural. This is the best part of the proposal.

**What we keep from #585:**
- The pathname stays the address for the main screen chain. We don't want occupancy moved fully into the query string — the pathname is the one carrier the whole stack already understands, and a "cosmetic" pathname becomes a second version of the same truth.
- The three FEATUREs and the DECOMPOSITION anchors. LAYER-2 requires them, and they're where the field-level schemas and the future `@cpt` code markers live.

**Concurrent occupancy** we'll close on our side too. It doesn't need a new grammar: repeated query keys are standard, and the discriminator already exists because a parallel key's value starts with the occupant's declared segment. The real work is set-shaped transitions (added/removed instead of previous→new) — and that cost belongs to the requirement, not to either model.

**Where the framework ADRs go:** `route-support-scope` asks the runtime to validate something at registration, so it belongs with `mfes`. `occupant-identity-stability` only picks an existing `mfes` field, so it fits as a routing ADR. Either way they need to leave `template-shell/`, which validation ignores as template territory.

Questions are in the inline comments. The three I'd most like answered: what carrier a provider-built router matches in a composed app if the substrate never reads the pathname; what happens on a plain `<a href>` click when only actions can originate a mount; and why a nested domain's key stays the same when the parent's parameter changes.

packages/routing/architecture/ADR/0003-domain-occupancy-addressing-granularity.md
GeraBart: Good catch on the domain-name gap — here's a fix that closes it and answers the percent-decoding question at the same time.

**Proposal: give an extension domain its own optional `route` property**, exactly mirroring `ExtensionPresentation.route` on extensions. When a domain declares one, that value — not the domain's own GTS-derived identity — is what gets used as its own segment in the compound key. Same reasoning as `cpt-frontx-framework-adr-occupant-identity-stability`: a domain's own GTS id is versioned, so a redeploy that bumps it would change every already-composed key for every occupant nested inside that domain, silently breaking every bookmark pointing anywhere inside that subtree. A human-authored `route` value has no structural reason to change on redeploy.

**The domain's own key segment is its bare `route` value, not a composed ancestor chain.** An earlier draft of this reply composed a nested domain's key from its full ancestor path (`screen.dashboard.widgets`) the way ADR 0003 currently does. We're revising that: the key uses the domain's own `route` value alone (`widgets`), accepting that two *unrelated* domains anywhere in the tree choosing the identical `route` value would collide, as a negligible risk against real composed-name authoring practice — in exchange for materially shorter keys. That's a deliberate scope change from ADR 0003's current uniqueness proof (which guarantees global uniqueness from purely local facts, with no registry) to a stated, accepted obligation instead (domain `route` values must be chosen distinct application-wide, the consumer's own responsibility, the same class of obligation this record's own O1/O2 already are). One thing gets simpler as a result, not harder: prefix shadowing. With a bare, uncomposed name, a level's own key-prefix has no structural relationship to any ancestor's name at all, so the boundary-delimiter-immediately-after-the-name rule alone perfectly separates any two *different* names (`widgets::` can never accidentally match `widgets-detail::x`); the only residual risk is two domains choosing the exact same name, which is exactly the risk being accepted above.

Once a domain's own key segment comes from `route` instead of an unconstrained string, it goes through the identical lexical rule occupant-identity already has (`cpt-frontx-routing-adr-occupant-reference-boundary`'s validator): non-empty, drawn from an explicit allowed character set that excludes both reserved delimiters (`.` and `::`). That's the direct fix for the forgery you flagged. And matching stays raw, never post-percent-decode — same rule, extended rather than duplicated, so `%3A%3A` is just three literal characters to the matcher, not a decoded `::`.

**Why this stays in the query string, and doesn't reopen giving any one occupant a pathname slot:** a domain-level `route` value is still just one more segment feeding the same compound-key grammar — it changes *where the name comes from*, not *which carrier carries it*. Putting even one occupant's chain on the pathname means that occupant is addressed by a different rule than every other occupant in the tree — the exact special-casing this mechanism exists to avoid. It also runs into a structural wall regardless of preference: a pathname is one linear sequence, and concurrent occupancy is a branching tree (`screen` and `sheet` are siblings, `widgets` alone holds three simultaneous occupants) — you cannot fit a branching structure into one linear carrier without either privileging a branch or reducing the pathname to a delimited list that only looks like a path. There's no version of "give the pathname to one occupant" that survives contact with more than one occupant being open at once, so the `route` property belongs on the same query-string mechanism every other domain and occupant already uses, not on the pathname.

**Worked example, with domain `route` values now explicit:**

- `screen` domain → `route: "screen"`
- `sheet` domain → `route: "sheet"`
- `widgets` domain (nested inside `dashboard`'s own zone) → `route: "widgets"`

The visible key strings for this example are shorter than a composed-ancestor-chain scheme would produce, because each key now carries only the domain's own bare name, never its ancestry.

**How the tree actually gets walked to build these keys — breadth-first, level by level:**

1. **Level 0**: the shell zone's two root domains, `screen` and `sheet`, are visited together. `screen`'s winning occupant is `dashboard`, contributing `screen::dashboard=orientation%3Dleft`. `sheet` holds two concurrent occupants, contributing `sheet::tenant-details=tenantId%3D456` and `sheet::user-contacts=contactId%3D123`.
2. **Level 1**: only once `dashboard` exists does its own zone — and the `widgets` domain nested inside it — come into being at all. Causally, `widgets` still can't exist before `dashboard` mounts; but its own key segment no longer references `dashboard` or `screen` at all — it's simply its own declared `route` value. Its three concurrent occupants contribute `widgets::line-a=range%3D7d`, `widgets::line-b=range%3D30d`, `widgets::pie=metric%3Drevenue`.

Composing in this order — outer domains first, nested domains only once their enclosing occupant has actually resolved — is still why no advance knowledge of the tree's shape is needed to *produce* the keys in the right order, even though the keys themselves no longer encode that order once written.

**Full assembled URL:**

```
/?screen::dashboard=orientation%3Dleft&sheet::tenant-details=tenantId%3D456&sheet::user-contacts=contactId%3D123&widgets::line-a=range%3D7d&widgets::line-b=range%3D30d&widgets::pie=metric%3Drevenue
```

**Where each `route` value comes from, concretely:**

- An **occupant**'s (extension's) `route` comes from its own `ExtensionPresentation.route` — one per mount, since each mount is its own distinct registration.
- A **domain**'s `route` comes from the new, analogous property on the domain's own declaration — set once by whoever composes that domain into the tree (e.g. `dashboard`'s own author, when declaring that its zone contains a `widgets` domain), not per-mount, since a domain is the slot itself, not an occupant of it.

Happy to fold this into ADR 0001 as an amendment if this direction looks right to you.

gs-layer: Thanks for the revision — it sharpened the question. Let me answer it whole, since the disagreement is now precisely one thing: **the form of the address**. Both models cover the same needs — several domains per zone, nesting to any depth, N occupants in one domain, several instances of one entry, per-occupant parameters, deep link / reload / back-forward, version-stable identity, actions as the single mount writer. Nothing below is about capability.

## On the revised proposal (bare domain names)

Dropping the ancestor composition for shorter keys costs more than it saves, and your own text shows each cost:

- **The uniqueness proof becomes a convention.** You say it yourself: a "deliberate scope change from ADR 0003's current uniqueness proof … to a stated, accepted obligation". That obligation — distinct names application-wide — is the convention-based uniqueness you criticized in our model. On this point the two models are now equal, and the one thing yours did strictly better is gone.
- **The convention is unenforceable here.** Registries form a tree: a child registers its own domains in its own registry, and the parent can't enumerate them (that's why `ChildDomainForwardingHandler` exists). "Distinct application-wide" has no place to be checked. Your three *local* facts weren't elegance — they were the only kind of uniqueness this runtime can verify.
- **Collisions aren't "two unrelated domains happening to pick the same name".** Same-named domains under different parents are the normal shape of a component tree — every screen has a `tabs`, every panel has a `toolbar`. ADR 0003 §56 covered exactly that case and you removed the cover. For your own headline case it's guaranteed, not negligible: `line-a` and `line-b` each hosting a nested `legend` domain both write `legend::…`. Composed, they were `widgets.line-a.legend` and `widgets.line-b.legend`.
- **Inert garbage becomes wrong mounts.** Your rules forbid cleaning unknown keys and treat an unread key as inert. With composed keys a stale `screen.dashboard.widgets::pie` is read by nobody once the screen changes. With bare names, `tabs::contacts` survives a screen switch and is read by the next screen's `tabs` domain — if it has a `contacts` tab, it mounts the wrong state silently.
- **Hierarchy leaves the address entirely.** Your DESIGN says hierarchy is carried by traversal order and composition; the stack doesn't preserve order (your own point), and composition is now removed. You note it in passing — "the keys themselves no longer encode that order once written" — but for a library whose job is projecting the domain tree into the address, that's the job.

The motivation is key length. That cost is specific to query-only addressing: every key repeats its full ancestry. A pathname doesn't have the problem — each level adds one segment. The fix removed the model's main structural guarantee to pay for a cost the model created.

## On the pathname argument

"A branching tree can't fit one linear carrier" is true and isn't what we do. One chain goes in the pathname; everything else goes in the query. Your dashboard example in our model:

```
/dashboard?orientation=left
  &sheet=tenant-details/456
  &sheet=user-contacts/123
  &screen.dashboard.widgets=line-a/7d
  &screen.dashboard.widgets=line-b/30d
  &screen.dashboard.widgets=pie/revenue
```

Against yours:

```
/?screen::dashboard=orientation%3Dleft&sheet::tenant-details=tenantId%3D456
  &sheet::user-contacts=contactId%3D123&widgets::line-a=range%3D7d
  &widgets::line-b=range%3D30d&widgets::pie=metric%3Drevenue
```

| Piece | What it is |
|---|---|
| `/dashboard` | the exclusive `screen` domain's occupant — pathname, read natively by the engine |
| `orientation=left` | the screen's own parameter — the engine's own search param |
| `sheet=…` ×2 | concurrent domain at level 0 — repeated key, standard `getAll()` |
| `tenant-details/456` | occupant segment plus its local path; `456` is parsed by that occupant's engine |
| `screen.dashboard.widgets=…` ×3 | nested concurrent domain — **your** composition rule, first version |
| `line-a/7d`, `line-b/30d` | two instances of one entry, parameters inside each own value |

Every need on the list is met. The boundary between domain and occupant is `=` — the one the URL already has. Identity sits in the value, where it always did, so no second delimiter exists, and every key is `[a-z0-9.-]` — nothing gets percent-encoded. (`::` does: `URLSearchParams` writes `screen%3A%3Adashboard`, which a raw matcher then can't recognize as its own key — that's the tension between your "raw matching" and your grammar.)

On "special-casing": the main screen chain is privileged in the product, not by our mechanism. The screen domain has swap semantics with no `unmount_ext`; `setActiveScreen` is a first-class notion; menus, breadcrumbs and page analytics all assume one active screen. One mechanism on our side too — one observer parameterized by carrier, one primitive, one helper. The only difference between the two models is one bit: where the main chain is written. Your point that our DESIGN justifies the "one axial domain" rule technically rather than from the product is fair, and we'll fix that text.

## What query-only actually costs

- **Relative addressing stops existing.** RFC 3986 resolves relative references over path segments. `<a href="contacts">` from `/tenants/ABC/general` gives `/tenants/ABC/contacts` with no JavaScript. From `/?screen::tenants=…` it gives `/contacts`. And a reference carrying a query replaces the whole query — `<a href="?tabs::contacts=">` erases the screen, the modal and the sheets. Every link becomes JS-computed full state. `<form method="GET">` erases everything too.
- **Router engines are pathname-first.** Route trees, nested routes, `useParams`, `to="./contacts"` — all path segments. With the pathname at `/`, the engine has nothing to match, and your DESIGN leaves the composed-mode carrier "deliberately left open". That's the package we split out for exactly those mechanics.
- **Private syntax in a carrier the stack reorders, truncates and drops**, versus hierarchy in the carrier the whole stack already reads.
- **A "cosmetic" pathname** is either permanently `/` or a second account of the same fact that drifts.

## What we concede

- **The strategy seam.** If an exclusive domain becomes concurrent, its projection moves from pathname to query. You don't have that seam. For screens it's a redesign; for nested tabs it's a plausible evolution.
- **Server rewrite** stays our deployment obligation; yours doesn't need it.
- **An axial occupant's own parameters live in the flat namespace** (`?orientation=left`) and could collide with a root-level parallel domain's name. Our one convention where yours was structural — until the revision.
- **Parallel axes have no relative addressing either.** We pay for modals and sheets what you pay for everything.
- **Set-shaped transitions for concurrent domains** are unspecified on both sides.

## What we're taking from you, with credit

Identity from `route`; mount-trigger ownership; occupant reference boundary; route-support scope as a per-type invariant; and the **key composition rule in your first version** — composed from the enclosing occupant's full address path, with the three local facts. Applied to parallel-axis keys it gives structural uniqueness without the length cost, because the main chain carries its depth in the pathname. On the domain-`route` field: right idea, but it's a change to a GTS-versioned `mfes` contract, so that decision belongs there — and the alphabet gap closes without it, by running your validator on the domain-key argument at observer creation.

Still open from before, and they matter most for your model: what a provider-built router matches in a composed app; what a plain `<a href>` click does when only actions may originate a mount; and whether a nested occupant is meant to survive a full change of its parent's parameter unchanged.

gs-layer: Following up here: we adopted your query-only topology, with one change to how it's encoded. Every domain now gets semicolon-separated entries (`domain=extension;param=value;...`) instead of the double-colon compound key, so a parameter reads in place rather than nested and percent-encoded twice.

The full record now lives in https://github.com/constructorfabric/gears-frontx/pull/585, as ADR 0003 with the complete grammar and eight worked examples. Your ADR 0001 (occupant reference boundary) and ADR 0002 (mount-trigger ownership) were carried over with your decision-makers and a co-authorship trailer, then adapted to the routing member's own voice and to the new grammar (the identity lexical rule moved into ADR 0003), and dated with this branch.

What differs from your draft: no `::` delimiter, a dotted key only where a domain is genuinely nested inside another extension's zone (not for every ancestor in the chain), `widgets` treated as a peer domain rather than nested under `screen`, and an explicit statement of which party owns which part of the URL.

Would appreciate your eyes on #585 when you have a moment.

PR #603, opened 2026-08-31
fix(template-shell): drop private template re-exports so framework and react install standalone

### LINKED ISSUE #601: [BUG] `@gears-frontx/framework` cannot be published/installed standalone — hard runtime dependency on the private `frontx-template-shell`
## Summary

`@gears-frontx/react` and `@gears-frontx/framework` cannot be consumed as regular npm packages outside the `template-shell` scaffold. `framework`'s main entry point unconditionally imports three symbol groups from `@gears-frontx/frontx-template-shell` — a `"private": true` package only resolvable via `file:../..` inside the `template-shell` workspace. Any consumer who installs `@gears-frontx/framework` (or `@gears-frontx/react`, which requires `framework` as a peer) from a registry, without also being inside the `template-shell` monorepo, gets a package whose compiled entry point fails to load (`MODULE_NOT_FOUND: @gears-frontx/frontx-template-shell`).

We want to mirror `@gears-frontx/react` + `@gears-frontx/framework` (and their real dependency closure) into our internal Artifactory npm registry so teams can build FrontX apps without going through the CLI's GitHub-tarball scaffolding flow. This is currently blocked by the above.

## What we verified

- `template-shell/packages/framework/package.json` declares `@gears-frontx/frontx-template-shell` in **both** `devDependencies` (as `"file:../.."`) and `peerDependencies` (`"*"`, `optional: false`).
- `template-shell/package.json` (the `frontx-template-shell` package itself) is `"private": true`, has no `files` allowlist, and its own `dependencies` are the full demo-shell application's dependency set (`@radix-ui/*`, `lucide-react`, `tailwind-merge`, `lodash`, `react`, `react-dom`, etc.) — i.e. it is the scaffolded application root, not a scoped library.
- `template-shell/packages/framework/src/index.ts` imports **at the top-level main entry** (not a test-only subpath):
  - line 258: `export { LayoutDomain } from '@gears-frontx/frontx-template-shell';` (comment: "app-layer domain placement values")
  - lines 434–443: `export { RestMockPlugin, SseMockPlugin, MockEventSource } from '@gears-frontx/frontx-template-shell';` and the matching `RestMockConfig | SseMockConfig | SseMockEvent` types — comment: *"Protocol-specific mock plugins (relocated from `@gears-frontx/api` to the template in Phase 9's mock separation; `@gears-frontx/api` ships no mocks)."*
  - `template-shell/packages/framework/src/gts/index.ts:12`: `export { themeSchema, languageSchema, extensionScreenSchema } from '@gears-frontx/frontx-template-shell';` — comment: *"These schemas were relocated to the template package in Phase 4 (F3 GTS extraction)."*
- We built `framework` in an isolated copy of `template-shell` (fresh `npm ci`, no `.git`): the build **fails outright** (`tsup`/`tsc`: `Cannot find module '@gears-frontx/frontx-template-shell'`) until `template-shell`'s own root `npm run build:package` (producing `dist-lib/`) is run first — confirming this is a real build-order dependency, not just a type-only/erased import.
- After a successful build, `packages/framework/dist/index.cjs` literally contains `require('@gears-frontx/frontx-template-shell')` — a real runtime `require`, not something a bundler elided.
- `@gears-frontx/api` and `@gears-frontx/mfes`, by contrast, are genuinely standalone: zero `@gears-frontx/*` runtime deps (`api`'s own `CLAUDE.md`: *"It ships zero `@gears-frontx/*` runtime dependencies and can be used independently"*), `mfes`'s only peer is `@gears-frontx/gts-plugin`. Both already publish cleanly to the public registry today.

## Why this matters

Per `architecture/ADR/0033-template-territory-traceability.md`, template subtrees are intentionally kept outside the ecosystem's own specified/published artifact universe — we understand `react`/`framework`/`auth`/`state`/`i18n` currently living in `template-shell/packages/*` is a deliberate choice, not an oversight. What we're flagging is narrower: two commit-history moves (Phase 4's GTS-schema extraction, Phase 9's mock-plugin separation) relocated symbols that `framework`'s own public entry point still re-exports today, out of a package (`frontx-template-shell`) that was never meant to be an installable dependency. That combination is what makes `framework`'s compiled output non-functional for any consumer outside `template-shell`, regardless of what packaging/distribution route (npm scope, private registry, GitHub template flow) is used for `framework` itself.

## What we're asking for

Any of the following would unblock us — we don't have a strong preference, upstream is better placed to judge which fits the template/ecosystem boundary:

1. **Extract the specific re-exported surface into a small, real, published package** (`LayoutDomain`, `RestMockPlugin`, `SseMockPlugin`, `MockEventSource`, `RestMockConfig`/`SseMockConfig`/`SseMockEvent` types, `themeSchema`/`languageSchema`/`extensionScreenSchema`) — e.g. back into `@gears-frontx/api` (mock plugins) and a new minimal package for `LayoutDomain`/GTS schemas — so `framework` depends on something that actually exists in a registry.
2. **Make these symbols optional in `framework`'s public surface** (lazy/dynamic import, or split into a `@gears-frontx/framework/template-shell` subpath that isn't touched unless a consumer explicitly imports it), so the main entry point (`createFrontX`, `createFrontXApp`, the plugin system, `@gears-frontx/react`'s hooks) loads without `frontx-template-shell` present.
3. **Confirm this is out of scope / won't be changed**, and tell us whether there's a supported way to obtain a *build-only* copy of `frontx-template-shell`'s `dist-lib` output (e.g. a documented internal package name, even if unpublished to the public registry) that we're allowed to vendor into our own build of `framework` — so we know we're not fighting an unsupported path if we solve it entirely on our side.

## Non-goals

- We are not asking to change how the CLI/template-scaffolding flow (`frontx install github:...`) works — that path is unaffected either way.
- We are not asking to publish the demo shell application itself, or to move layout/theme/i18n/auth wholesale out of template territory — only the handful of symbols `framework`'s *own* main entry currently re-exports from the private template root.

## How we'd confirm it's resolved

`npm install @gears-frontx/react @gears-frontx/framework` (or their internally-republished equivalents) in a bare project outside `template-shell`, then `import { createFrontXApp } from '@gears-frontx/framework'` / `import { FrontXProvider } from '@gears-frontx/react'` load without `MODULE_NOT_FOUND`.

tscbmstubp:
Fixes #601.

## Problem

`@gears-frontx/framework`'s main entry re-exported `LayoutDomain`, `RestMockPlugin`/`SseMockPlugin`/`MockEventSource` (+ their config types), and `themeSchema`/`languageSchema`/`extensionScreenSchema` from `@gears-frontx/frontx-template-shell` — a `"private": true` package resolvable only through the `file:../..` workspace link. `@gears-frontx/react` re-exported the same surface. The compiled entry points carried a real runtime `require('@gears-frontx/frontx-template-shell')`, so any consumer installing the packages from a registry got `MODULE_NOT_FOUND`. This also inverted the ADR 0033 boundary: a publishable library had a runtime dependency pointing up into template territory.

## Change

- Remove the app-layer re-exports from `framework`'s and `react`'s entry points (option beyond the issue's list: nothing in the ecosystem consumed them — every consumer was template-territory code).
- Drop the `frontx-template-shell` peer dependency from `framework`. The `file:../..` devDependency stays: framework's own tests still exercise the mock plugins and schemas, and devDependencies never reach consumers.
- Delete `framework/src/gts/index.ts` (existed only to re-export the template's schemas).
- Repoint all consumers (shell `src-app`, `template-mfe` MFEs, framework tests) to import the symbols from `@gears-frontx/frontx-template-shell` directly — same resolution as before in the shell and in scaffolded apps. `demo-mfe`/`_blank-mfe` move that dep from devDependencies to dependencies since it is now a runtime import.
- Bump `@gears-frontx/framework` → 0.2.0-alpha.3 and `@gears-frontx/react` → 0.2.0-alpha.4; update exact pins in `template-mfe` MFE packages and the `template-shell` lockfile surgically (`template-mfe`'s lockfile is left alone, matching 821887dd / 83c2a635 practice).

## Verification

- `framework` and `react` build; their dists contain **zero** `frontx-template-shell` imports (sourcemap comment mentions only).
- Framework tests 306/306, react tests 150/150, shell app `tsc` clean, `cfs validate` passes.
- Issue's acceptance criterion: packed `framework`, `react`, and their dependency closure with `npm pack`, installed them in a bare project outside the workspace — `import`/`require` of `createFrontXApp`, `createFrontX`, `FrontXProvider` all load (ESM and CJS).
- `demo-mfe` unit tests fail identically on `develop` (unresolvable `@shadcn/react/message-scroller` from the local ui-kit dist — pre-existing dev-harness issue, unrelated).

## Summary by CodeRabbit

* **Breaking Changes**
  * Framework and React packages no longer re-export schemas, layout utilities, or mock plugins.
  * Import these features directly from the template shell package.

* **Updates**
  * Updated framework and React package versions.
  * Updated application templates and fixtures to use the new package boundaries.
  * Improved package setup for runtime template shell support.


tscbmstubp:
CI status after the DCO re-push: DCO now passes; everything is green except **Template Validate**, which fails on pre-existing `demo-mfe` type-check errors against ui-kit 0.4.0-alpha.1 (`FieldProps.name`, `Badge.dot`, `muted` button variant — `ProfileDetailsCard.tsx`, `FormElements.tsx`, `DataDisplayElements.tsx`, `OverlayElements.tsx`). None of these files are touched by this PR, and `develop`'s own Main CI has failed on the same step ("Validate shell + overlay composition") for its last three commits, starting with 821887dd's ui-kit pin bump. The demo-mfe/ui-kit API drift needs its own fix on `develop`.

GeraBart:
Filed two follow-up issues from review findings on this PR:

- #605 — `demo-mfe`/`_blank-mfe` in `template-mfe` now carry a runtime dependency on `@gears-frontx/frontx-template-shell` (moved to `dependencies`, direct imports of `RestMockPlugin`/`themeSchema`/etc.). Per the architect, `mfe` and `shell` were split into two templates specifically so MFEs can live in their own repo independent of a shell — this PR's fix reintroduces that coupling one level down from where #601 found it.
- #606 — two substantively-modified files (`template-shell/packages/framework/src/index.ts`, `template-shell/packages/react/src/index.ts`) still carry pre-existing residue `@cpt-` markers that `cpt-frontx-adr-template-territory-traceability` requires to be stripped in the same change.

Neither issue affects the fix this PR actually needs to ship (`@gears-frontx/framework`/`@gears-frontx/react` installing standalone from a registry — verified: zero `frontx-template-shell` references remain in either package's `dependencies`/`peerDependencies` or runtime source at head). See PR approval comment for merge rationale.

GeraBart (approved):
Approving on the merge urgency: this fixes a real, customer-facing break — an enterprise consumer installing `@gears-frontx/react`/`@gears-frontx/framework` from a registry hit `MODULE_NOT_FOUND` because their compiled entry points required the private `@gears-frontx/frontx-template-shell` package. Verified independently: at head, neither package's `dependencies`/`peerDependencies` nor runtime source references `frontx-template-shell` anymore — the fix is real and isolated to `template-shell/packages/framework` and `template-shell/packages/react`.

Two issues found during review do **not** block this specific fix but must be addressed ASAP in follow-up PRs:

- #605 — this PR's fix reintroduces a template-to-template coupling one level down: `template-mfe`'s `demo-mfe`/`_blank-mfe` now hard-depend on `@gears-frontx/frontx-template-shell`, breaking the ability to apply `template-mfe` to a repo without `template-shell` — which is the reason those two templates were split in the first place.
- #606 — two substantively-modified files leave pre-existing residue `@cpt-` markers unstripped, per `cpt-frontx-adr-template-territory-traceability`.

Merging now given the customer-facing urgency; please prioritize #605 and #606 right after.

PR #604, opened 2026-08-31
feat: adopt the Studio blue rebrand for ui-kit theme tokens

SysoevAndrey:
## What

Adopts the **Studio blue rebrand** for the ui-kit theme tokens, from the mockups' token spec frame **"02 · Tokens, type & elevation"** (node `40001232:5254`), which supersedes the violet "Studio / shadcn" collection `theme.css` previously implemented.

### Palette (both themes)

- Brand axis onto the blue ramp: `--primary #0065e3`, `--primary-hover / --accent-foreground #2668c5`, `--accent #e6effa`; light page onto slate-50 `#f1f5f9`; dark mode onto the slate ramp (surfaces `#0f172a`, borders `#334155`, foreground `#ffffff`) — including two collapses the frame draws explicitly (dark `surface == surface-elevated`, `border == border-strong`).
- Entries the frame doesn't draw are re-derived per the file's `derived:` convention (card/popover mirror surface; `secondary` stepped one neutral away from the now-page-equal `muted`; sidebar/code follow their same-role tokens; ring family and `--link-foreground` from the brand ramp — measured ratios at each token).
- `--popover-shadow` adopts the frame's Studio/Elevation/200 geometry (0 8 24 @ 12% foreground); two new palette-only entries: `--scenario-canvas` (moved into the `color/` namespace by the rebrand) and `--artifact-accent-icon`.
- Token **names** stay unprefixed per the shadcn convention even though the frame labels its WEB syntax `var(--color-*)` — renaming the public token surface would break every consumer for zero semantic gain (recorded in the `theme.css` header).

### Typography

The frame's **Studio/Type** ramp, carried on the existing `--text-*` role names (roles are the public API; values moved, names didn't):

| role | was | now | source style |
|---|---|---|---|
| `display` | 28/34 | **26/32** | Page Title |
| `heading-1` | 20/28 | 20/28 (kept) | no ramp counterpart — kit interpolation |
| `heading-2` | 16/24 | 16/24 (already matched) | Section Title |
| `body` | 15/20 | **14/20** | UI Primary Regular |
| `label` | 13/16 · 500 | **12/16 · 500** | UI Secondary Medium |
| `meta` | 12/16 | 12/16 (already matched) | UI Secondary Regular |
| `mono` | 11/16 | **10/14** | System Micro |

All trackings normalized to the ramp's 0. Evidence for the control axis comes from the live "Phase 3 · Remaining core controls" frame (`40001414:5973`): drawn Tabs trigger binds UI Secondary Medium, drawn compact Button binds UI Secondary Semi Bold, drawn Badge sits on the metrics Badge already had.

### WCAG deviations from the drawn values

Standing ruling: a drawn value failing its contrast floor gets a corrected value in code, recorded at the token. Exactly four deviations, all 12px status labels over their own `-soft` fills (worst case across soft/card/page):

| token | drawn | shipped | worst case |
|---|---|---|---|
| light `--success` | `#059669` (3.43:1) | `#047857` | 5.00:1 |
| light `--warning` | `#f59e0b` (1.96:1) | `#b45309` | 4.58:1 |
| light `--danger` | `#e11d48` (4.28:1) | `#be123c` | 5.72:1 |
| dark `--danger` | `#e11d48` (3.63:1) | `#fb7185` | 6.34:1 |

Light `--info` passes as drawn (4.59:1) and is kept. `--destructive` tracks the drawn red (`#e11d48`, its text is the white on-color at 4.70:1) and now deliberately diverges from the corrected `--danger`.

Same ruling applied in components: Avatar's `solid` treatment labels each status tone with the tone's own `-soft` value instead of the mockup's all-white binding (~1.5–2:1 in dark); Badge's `link` variant moved onto `--link-foreground` (dark `--primary` is 3.72:1 on the page); Badge's long-standing tone-label contrast debt is thereby resolved — recomputed table in `badge.module.css`.

### Docs

`design-notes.md` gains the rebrand entry (deviations table, evidence, open designer questions); `sidebar.md`'s stale 15px note updated; demo chart palette key `violet` → `brand`.

## Deliberately out of scope

- Elevation scale beyond the popover shadow (Elevation/100/300/Dock top — no kit consumer yet).
- Button label weight: the drawn compact button binds the **600** cut while Button reads `--text-label-weight` (500), which also dresses menus/tooltips/table headers — per-component question for the components pass.

## Open questions for design

1. Pin AA-passing status values in the Figma file (the four deviations above).
2. Confirm the dark `surface`/`border` collapses are intent rather than spec-sheet shorthand.
3. Rule the button label weight (500 vs the drawn 600).

## Verification

- All **909** unit tests pass — token block parity, AA floors (`--link-foreground`, `--subtle-foreground`) and Button focus-ring contrast recompute from the live CSS.
- Contrast matrix for every changed pair computed with the same WCAG math the tests use.
- Build + demo type-check clean; visual smoke of the kitchen-sink demo (tokens/typography/button/badge/avatar/table pages, both themes) in headless Chrome.

Heads-up for `insight-front`: token **values** changed (no names removed; two added), and type roles `body`/`label`/`mono` are 1–3px smaller — visual change only, no API change.

## Summary by CodeRabbit

* **Style**
  * Refreshed Studio Blue theme colors across light and dark modes, including sidebar styling and chart palettes.
  * Improved destructive-state colors and hover treatments for links, badges, menus, buttons, alerts, and form errors.
  * Added visible keyboard focus rings and refined avatar status colors.

* **Accessibility**
  * Improved contrast for status, link, error, sidebar, and destructive states across themes.

* **Documentation**
  * Updated design guidance and component notes covering branding, typography, chart colors, and accessibility standards.

### DIFF
--- a/packages/ui-kit/demo/examples/chart.tsx
+++ b/packages/ui-kit/demo/examples/chart.tsx
@@ -37,18 +37,23 @@ import { Row, Section } from '../shared';
  * from the ChartConfig below. The palette entries themselves are kit theme
  * tokens: the kit defines no --chart-1..5 ramp, and a literal hex would only
  * hold for one theme.
+ *
+ * No --info slot: after the rebrand --primary and --info are near-identical
+ * blues in light mode (~1.02:1 apart), so a palette carrying both would let
+ * one chart draw two indistinguishable series. The fifth slot is the
+ * theme's neutral slate instead, which every other slot separates from.
  */
 const PALETTE = {
   brand: 'var(--primary)',
-  blue: 'var(--info)',
+  slate: 'var(--muted-foreground)',
   green: 'var(--success)',
   amber: 'var(--warning)',
   rose: 'var(--destructive)',
 } as const;
 
 const chartConfig = {
-  desktop: { label: 'Desktop', color: 'var(--primary)' },
-  mobile: { label: 'Mobile', color: 'var(--info)' },
+  desktop: { label: 'Desktop', color: PALETTE.brand },
+  mobile: { label: 'Mobile', color: PALETTE.green },
 } satisfies ChartConfig;
 
 const themedChartConfig = {
@@ -69,7 +74,7 @@ const data = [
 /* ------------------------------------------------------------------ */
 
 const stageConfig = {
-  prospect: { label: 'Prospect', color: PALETTE.blue },
+  prospect: { label: 'Prospect', color: PALETTE.slate },
   engaged: { label: 'Engaged', color: PALETTE.green },
   customer: { label: 'Customer', color: PALETTE.amber },
   'at-risk': { label: 'At risk', color: PALETTE.brand },
@@ -149,7 +154,7 @@ function DonutChartExample() {
 /* ------------------------------------------------------------------ */
 
 const resolvedConfig = {
-  chat: { label: 'Chat', color: PALETTE.blue },
+  chat: { label: 'Chat', color: PALETTE.slate },
   mail: { label: 'Mail', color: PALETTE.green },
   tasks: { label: 'Tasks', color: PALETTE.amber },
 } satisfies ChartConfig;
@@ -169,7 +174,7 @@ const resolvedPerDay = [
 /* ------------------------------------------------------------------ */
 
 const recordsConfig = {
-  companies: { label: 'Companies', color: PALETTE.blue },
+  companies: { label: 'Companies', color: PALETTE.slate },
   opportunities: { label: 'Opportunities', color: PALETTE.amber },
   people: { label: 'People', color: PALETTE.green },
 } satisfies ChartConfig;
@@ -188,7 +193,7 @@ const recordsCreated = [
 /* ------------------------------------------------------------------ */
 
 const newContactsConfig = {
-  inbound: { label: 'Inbound', color: PALETTE.blue },
+  inbound: { label: 'Inbound', color: PALETTE.slate },
   total: { label: 'Total', color: PALETTE.brand },
 } satisfies ChartConfig;
 
--- a/packages/ui-kit/demo/examples/chart.tsx
+++ b/packages/ui-kit/demo/examples/chart.tsx
@@ -40,12 +40,16 @@ import { Row, Section } from '../shared';
  *
  * No --info slot: after the rebrand --primary and --info are near-identical
  * blues in light mode (~1.02:1 apart), so a palette carrying both would let
- * one chart draw two indistinguishable series. The fifth slot is the
- * theme's neutral slate instead, which every other slot separates from.
+ * one chart draw two indistinguishable series. The neutral fifth slot is
+ * --foreground, NOT --muted-foreground: the latter is the charts' own
+ * chrome color (axes, ticks, legend and tooltip labels — see
+ * chart.module.css), and a data series in the chrome color reads as
+ * furniture. Ink separates from that chrome in both themes (1.72:1 light /
+ * 2.56:1 dark) and from every other slot.
  */
 const PALETTE = {
   brand: 'var(--primary)',
-  slate: 'var(--muted-foreground)',
+  ink: 'var(--foreground)',
   green: 'var(--success)',
   amber: 'var(--warning)',
   rose: 'var(--destructive)',
@@ -74,7 +78,7 @@ const data = [
 /* ------------------------------------------------------------------ */
 
 const stageConfig = {
-  prospect: { label: 'Prospect', color: PALETTE.slate },
+  prospect: { label: 'Prospect', color: PALETTE.ink },
   engaged: { label: 'Engaged', color: PALETTE.green },
   customer: { label: 'Customer', color: PALETTE.amber },
   'at-risk': { label: 'At risk', color: PALETTE.brand },
@@ -154,7 +158,7 @@ function DonutChartExample() {
 /* ------------------------------------------------------------------ */
 
 const resolvedConfig = {
-  chat: { label: 'Chat', color: PALETTE.slate },
+  chat: { label: 'Chat', color: PALETTE.ink },
   mail: { label: 'Mail', color: PALETTE.green },
   tasks: { label: 'Tasks', color: PALETTE.amber },
 } satisfies ChartConfig;
@@ -174,7 +178,7 @@ const resolvedPerDay = [
 /* ------------------------------------------------------------------ */
 
 const recordsConfig = {
-  companies: { label: 'Companies', color: PALETTE.slate },
+  companies: { label: 'Companies', color: PALETTE.ink },
   opportunities: { label: 'Opportunities', color: PALETTE.amber },
   people: { label: 'People', color: PALETTE.green },
 } satisfies ChartConfig;
@@ -193,7 +197,7 @@ const recordsCreated = [
 /* ------------------------------------------------------------------ */
 
 const newContactsConfig = {
-  inbound: { label: 'Inbound', color: PALETTE.slate },
+  inbound: { label: 'Inbound', color: PALETTE.ink },
   total: { label: 'Total', color: PALETTE.brand },
 } satisfies ChartConfig;
 
--- a/packages/ui-kit/demo/examples/chart.tsx
+++ b/packages/ui-kit/demo/examples/chart.tsx
@@ -39,7 +39,7 @@ import { Row, Section } from '../shared';
  * hold for one theme.
  */
 const PALETTE = {
-  violet: 'var(--primary)',
+  brand: 'var(--primary)',
   blue: 'var(--info)',
   green: 'var(--success)',
   amber: 'var(--warning)',
@@ -72,7 +72,7 @@ const stageConfig = {
   prospect: { label: 'Prospect', color: PALETTE.blue },
   engaged: { label: 'Engaged', color: PALETTE.green },
   customer: { label: 'Customer', color: PALETTE.amber },
-  'at-risk': { label: 'At risk', color: PALETTE.violet },
+  'at-risk': { label: 'At risk', color: PALETTE.brand },
   churned: { label: 'Churned', color: PALETTE.rose },
 } satisfies ChartConfig;
 
@@ -189,7 +189,7 @@ const recordsCreated = [
 
 const newContactsConfig = {
   inbound: { label: 'Inbound', color: PALETTE.blue },
-  total: { label: 'Total', color: PALETTE.violet },
+  total: { label: 'Total', color: PALETTE.brand },
 } satisfies ChartConfig;
 
 const newContacts = [
@@ -255,7 +255,7 @@ function AreaSparkline() {
 /* ------------------------------------------------------------------ */
 
 const trafficConfig = {
-  visits: { label: 'Visits', color: PALETTE.violet },
+  visits: { label: 'Visits', color: PALETTE.brand },
 } satisfies ChartConfig;
 
 const traffic = [
--- a/packages/ui-kit/design-notes.md
+++ b/packages/ui-kit/design-notes.md
@@ -540,7 +540,13 @@ Architecture's build bullet).
      caught the light half going stale), restoring the pre-rebrand
      tinted-panel-on-lighter-page relationship; the light active item
      improves (1.07:1 → 1.23:1 vs the panel) and every sidebar text seat
-     re-measures clear. This is a fifth deviation from the
+     re-measures clear. A follow-up review round then caught the knock-on:
+     `--sidebar-border` derived from `--border`, whose light value IS the
+     new panel fill — separators, the menu-sub rail and the floating
+     outline vanished at 1.00:1. The border now derives from
+     `--border-strong` (1.20:1 light / 1.41:1 dark against the panel;
+     dark's collapsed border family already was border-strong), and
+     tokens.test.ts pins the pair alongside the muted guard. This is a fifth deviation from the
      drawn frame — visibility, not WCAG-text, so it sits outside the
      status-token table above — and joins the open designer question on
      whether the drawn collapses are intent.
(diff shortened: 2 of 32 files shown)


packages/ui-kit/src/styles/theme.css
```diff
@@ -142,71 +163,104 @@
...
+  --text-mono-size: 0.625rem;
+  --text-mono-line-height: 0.875rem;
   --text-mono-weight: 400;
   --text-mono-tracking: 0px;
 }
 
 :root,
 [data-theme='light'] {
-  --background: #f8fafc;
+  --background: #f1f5f9;
```
itechmeat, line 250: 🔴 The rebrand makes --muted equal its own backdrop in both themes: light --background moved onto #f1f5f9 (already --muted's value), and dark --muted moved onto #0f172a (the card/surface/popover value) - 1.00:1 where the old palette kept 1.05/1.23. Everything filled with --muted then visually disappears: Skeleton on the bare page, Toggle hover, outline/ghost Button hover, and a dozen more consumers. The --secondary comment in this same diff already applies the exact "a fill equal to the page would vanish" argument and steps it off the ramp - should --muted get the same one-step treatment (or, if the collapse is intended, a comment saying so plus a tokens.test.ts guard pinning the intent)?

packages/ui-kit/src/components/badge/badge.module.css
```diff
@@ -123,13 +131,16 @@ a.variantSecondary:hover {
...
+ * (still what button.module.css's own destructive hover uses) darkens
+ * correctly in light mode, where --foreground is dark, but LIGHTENS in dark
+ * mode instead, since --foreground flips near-white there — against the
+ * fixed --destructive-foreground label, that drops dark-mode hover
+ * to about 3.91:1. --destructive-hover is a literal, same hex both modes
+ * (see theme.css), so the direction never depends on which theme is active.
  */
 a.variantDestructive:hover {
-  background-color: color-mix(in oklab, var(--destructive) 90%, var(--foreground) 10%);
+  background-color: var(--destructive-hover);
```
itechmeat, line 147: 🟠 The badge comment says the mix-toward-foreground recipe is "still what button.module.css's own destructive hover uses" - so the diff documents a known-broken sibling without fixing it: measured, white on that mix in dark is ~3.90:1 (the button comment's 7.64:1 is a pre-rebrand number). Same edge for the --danger sweep: context-menu and dropdown-menu destructive items still paint --destructive-based mixes that now land at ~4.17:1 highlighted in dark, under the floor. The new token guards test token pairs, not the mixes those files paint, so nothing catches either case - would you extend the sweep to Button and the two menus here (it's a few lines), or pin a TODO + follow-up so the broken siblings don't ship silently?

packages/ui-kit/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@gears-frontx/ui-kit",
-  "version": "0.4.0-alpha.1",
+  "version": "0.4.0-alpha.2",
```
itechmeat: [Not a blocker] 🔴 The bump itself is required by the version gate - the missing step is the publish, so the templates only install once 0.4.0-alpha.2 is actually on npm. One coordination note: open PR #598 pins the same 0.4.0-alpha.2 with different content, so whichever lands second needs another bump or a shared publish plan - what's the ordering here?

packages/ui-kit/src/components/button/button.module.css
```diff
@@ -206,7 +206,7 @@
  * is what promotes it to --foreground) — measured against the page
  * background it actually sits on, since ghost has no fill of its own:
  * 4.55:1 light / 6.62:1 dark, both clearing the 4.5:1 AA floor for this
- * 13px label text, so the alignment carries no WCAG cost.
+ * 12px label text, so the alignment carries no WCAG cost.
```
itechmeat, line 212: [Not a blocker] 🟡 A few ratio comments the rebrand invalidated are still standing, including in blocks this diff edits: this ghost comment got its px updated but keeps the old palette's 4.55/6.62 (now 9.45/7.68), the link block below still says 4.51/3.78/4.98 and "same violet family", avatar.module.css's neutral-solid line claims 11.5/6.6 (I measure 9.45/6.96), the badge headroom sentence contradicts its own list ten lines up (spread is 4.59-5.72, info tightest), and theme.css describes dark --muted as page-equal when it's card-equal. Untouched tabs/table ratio blocks went stale the same way. Worth one sweep so the numbers readers verify against are the live ones?

packages/ui-kit/src/components/badge/badge.module.css
```diff
@@ -168,28 +187,37 @@ a.variantLink:hover {
...
  *
- * KNOWN CONTRAST DEBT, not an oversight: at 12px/500 these labels need
- * 4.5:1 (WCAG 1.4.3) and four pairs miss it — light success 3.43:1,
- * warning 2.90:1, danger 4.28:1, info 3.73:1, and dark danger 3.80:1.
- * The shortfall is in the token pairs themselves, which the mockup defines
- * and this file may not redefine, so it is raised for a design decision
- * rather than papered over with a local literal. Every other pair clears
- * AA (light accent 6.02:1, dark success 8.05:1 / warning 9.16:1 /
- * info 8.80:1 / accent 12.83:1), as does Neutral via --secondary.
+ * Contrast: at 12px/500 these labels need 4.5:1 (WCAG 1.4.3). The debt
```
itechmeat, line 193: [Not a blocker] 🟡 badge.md still publishes the contrast caveat this PR retracts ("four of the drawn tone pairs currently fall short... 3.43/2.90/4.28/3.73/3.80") - and the kit's own step-5 rule syncs <name>.md with any contract change, while the guard suite only checks CSS. Update the doc in the same pass?

packages/ui-kit/src/tokens.test.ts
```diff
@@ -575,6 +575,92 @@ describe('theme tokens', () => {
...
       });
+
+      // The status tokens are 12px label text in the kit — Badge's tone
+      // variants and Avatar's soft treatment paint them over their own
+      // -soft fill, and Field/Alert/Toast/Attachment/Bubble/Questionnaire
+      // error text paints --danger over whatever page/card/panel the
+      // component sits on — so each status color must clear the 4.5:1 AA
+      // floor against ALL of those backdrops, not only the one a single
+      // component happens to test. This is the guard the original Studio
+      // values shipped without (the PR #604 review's F-002/F-004): a hue
```
itechmeat: [Not a blocker] 🔵 Comments in tokens.test.ts, badge.test.tsx and design-notes.md cite "the PR #604 review's F-002..F-005", but no such review exists on the PR - it was an offline round, so a reader in six months can't resolve the codes. The findings are already spelled out in prose in the design-notes entry; point the comments there (or inline one clause of what the finding was) instead of the codes?

packages/ui-kit/design-notes.md
```diff
@@ -91,8 +91,13 @@ the `sidebar` / `data-table` exclusions above intact.
...
   `prefers-color-scheme`; shadcn-style token structure carrying the Studio
-  palette from the F-mockups Figma file (variable collection "Studio /
-  shadcn"). Component CSS consumes only these variables — the theme file is
+  BLUE palette from the mockups' token spec frame "02 · Tokens, type &
+  elevation" (node 40001232:5254), which superseded the original violet
+  "Studio / shadcn" collection — see the "Studio blue rebrand" entry in
+  step 4 below. The token NAMES stay unprefixed per the shadcn convention
+  even though that frame labels its WEB syntax `var(--color-*)`: renaming
+  the public token surface would break every consumer for zero semantic
+  gain (a standing ruling, recorded in theme.css's header). Component CSS consumes only these variables — the theme file is
```
itechmeat: [Not a blocker] 🔵 The rebrand entry is appended while paragraphs it supersedes still stand as current above it ("near the violet button", "violet/red read better against the page"), and the header edit splices into the middle of the old sentence with a broken wrap. Could the touched sections be rewritten in place (or the superseded paragraphs explicitly marked), so the file reads one way?

packages/ui-kit/demo/examples/chart.tsx
```diff
@@ -39,7 +39,7 @@ import { Row, Section } from '../shared';
  * hold for one theme.
  */
 const PALETTE = {
-  violet: 'var(--primary)',
+  brand: 'var(--primary)',
```
itechmeat, line 51: [Not a blocker] 🔵 After the rebrand, the demo palette's `brand` (--primary #0065e3) and `blue` (--info #2563eb) are ~1.02:1 apart, and newContactsConfig/stageConfig put both into the same chart - two series nobody can tell apart. Move one side of each pairing onto a distinguishable slot?

packages/ui-kit/src/styles/theme.css
```diff
@@ -216,146 +270,192 @@
...
+   * code block sits on. */
   --code-background: #f1f5f9;
   --code-foreground: #334155;
 
+  /* Palette-only, new with the rebrand (no kit consumer yet): the workflow
+   * scenario canvas fill — previously a `--studio-*` export, moved into
+   * the collection's color/ namespace by the rebrand, which is the line
+   * that admits it here (see header) — and the accent tone artifact cards
+   * draw their icon with. */
+  --scenario-canvas: #e2e8f0;
```
itechmeat: [Not a blocker] 🔵 The header's old line kept exactly these on the consumer side ("they name one consumer's workflow surfaces rather than any kit semantic"), and the only thing that changed is which Figma folder they sit in. Is the namespace move alone enough to admit them into the kit's public token surface, or should the app still carry them?

packages/ui-kit/src/components/badge/badge.module.css
```diff
@@ -168,28 +187,37 @@ a.variantLink:hover {
...
+ * underline as this file's affordance for "hover, no color change".
  */
 .variantSuccess {
   background-color: var(--success-soft);
   color: var(--success);
 }
 
 a.variantSuccess:hover {
-  background-color: color-mix(in oklab, var(--success-soft) 90%, var(--foreground) 10%);
+  text-decoration: underline;
```
itechmeat, line 224: [Not a blocker] ⚪ The tone hovers now underline, but .variantLink sets text-underline-offset: 4px and the tone variants don't - the hover underline sits at a different height across the same component. Worth aligning?

itechmeat (changes requested):
This is careful, well-evidenced token work: I independently recomputed a large sample of the claimed contrast ratios and the core corrections all hold (the four status-color fixes, the --destructive/--danger split, Avatar solid labels, Badge focus and hover), and the new guard tests are a real improvement. Requesting changes narrowly, on one substantive point: the rebrand collapses --muted onto its own backdrop in both themes, so Skeleton and every --muted hover state visually disappear - details inline, together with a parity question about the accessibility sweep stopping at the diff's edge. Everything else is marked [Not a blocker]. Once the --muted question is settled I'm happy to approve right away.

packages/ui-kit/package.json
SysoevAndrey: I'll bump the version again if needed — whichever PR lands second gets a new bump.

packages/ui-kit/src/styles/theme.css
SysoevAndrey: Removed both tokens in 0abca2a9. Nothing consumed them and they were never published, so nothing breaks. They stay on the app side.

packages/ui-kit/src/styles/theme.css
```diff
@@ -216,143 +272,193 @@
...
-  /* derived (dark only): follows dark --muted-foreground. */
-  --sidebar-foreground: #64748b;
+  /* Sidebar Navigation block. Read by the kit's Sidebar parts and offered
+   * for a consumer's own navigation chrome alike. The rebrand frame draws
+   * no sidebar entries, so the whole block is derived from the same-role
+   * tokens above: sidebar follows --muted, its foreground follows
+   * --muted-foreground, the active-item fill takes --surface (light) /
+   * --selection-subtle's value (dark), its foreground follows --foreground,
+   * and the border follows --border. */
+  --sidebar: #e2e8f0;
```
itechmeat, line 454: [Not a blocker] 🟠 The sidebar-follows-muted fix mechanically collapsed the light pair: --sidebar is now #e2e8f0 and --sidebar-border stayed #e2e8f0, so SidebarSeparator, SidebarMenuSub's guide rail, the floating variant's border and the outline menu-button ring all sit at 1.00:1 on the panel and vanish in light mode (1.13:1 before this commit; dark is fine at 1.41:1). Same vanish pattern the muted step-off just solved, and the new muted guard doesn't cover this pair - worth stepping --sidebar-border off the panel (e.g. onto --border-strong's value) and adding the pair to the guard?

packages/ui-kit/src/components/dropdown-menu/dropdown-menu.module.css
```diff
@@ -168,23 +168,26 @@
...
 .variantDestructive {
-  color: color-mix(in oklab, var(--destructive) 90%, var(--foreground) 10%);
+  color: var(--danger);
 }
 
 .variantDestructive:focus,
 .variantDestructive[data-highlighted] {
-  background-color: color-mix(in oklab, var(--destructive) 10%, transparent);
-  color: color-mix(in oklab, var(--destructive) 90%, var(--foreground) 10%);
+  background-color: var(--danger-soft);
```
itechmeat: [Not a blocker] 🟡 The text half of the menu fix is right, but the highlight fill lost ground: --danger-soft vs --popover is 1.099:1 light / 1.047:1 dark - weaker than both the mix it replaced (1.173/1.057) and the sibling rows' --accent highlight (1.161/1.117). Since the text is --danger at rest and highlighted alike, the fill is the only focus cue, so keyboard-landing on a destructive item in dark is a ~1.05:1 change. Same in context-menu.module.css:211, and no test covers fill-vs-popover. Maybe keep --danger text but take a stronger highlight fill (or reuse the accent-strength recipe)?

packages/ui-kit/demo/examples/chart.tsx
```diff
@@ -37,18 +37,23 @@ import { Row, Section } from '../shared';
...
+ * No --info slot: after the rebrand --primary and --info are near-identical
+ * blues in light mode (~1.02:1 apart), so a palette carrying both would let
+ * one chart draw two indistinguishable series. The fifth slot is the
+ * theme's neutral slate instead, which every other slot separates from.
  */
 const PALETTE = {
-  violet: 'var(--primary)',
-  blue: 'var(--info)',
+  brand: 'var(--primary)',
+  slate: 'var(--muted-foreground)',
```
itechmeat: [Not a blocker] 🔵 The slate slot fixes the --primary/--info collision, but --muted-foreground is the charts' own chrome color - axes, ticks, legend and tooltip labels all paint it (chart.module.css:33,131,183,221), so in Stacked bars / Multi-line / Combo a data series is now exactly the axis-and-label color and the legend swatch matches the legend text. A neutral not shared with the chrome would keep the series legible?

itechmeat (approved):
🟢 Thorough follow-up pass: all ten review points are genuinely closed, not just patched over - I recomputed the contrast math on the new head and every claimed number checks out (muted separation 1.13/1.23 light and 1.35/1.22 dark now beats even the pre-rebrand baseline), the destructive sweep reached Button and both menus with real regression guards, the ratio comments and docs were swept holistically, and all 919 tests pass. Approving. Three small observations inline, none blocking - the sidebar one is worth a quick look since it re-creates the same vanish pattern the muted fix just solved.

GeraBart (approved):

PR #607, opened 2026-09-01
Fix: 21 days old packages pinning

GeraBart:
## Summary by CodeRabbit

* **Releases**
  * Updated alpha versions across framework, API, state, internationalization, React, authentication, and micro-frontend packages.
  * Updated application templates and example modules to use compatible package versions.
* **Improvements**
  * Standardized dependency and tooling versions for more consistent installations.
  * Added UI kit support to the template shell.
* **Documentation**
  * Clarified dependency version-pinning guidelines.
  * Updated the SDK package publishing sequence.


packages/api/package.json
```diff
@@ -38,12 +38,12 @@
...
     "axios": "1.14.0",
-    "tsup": "^8.0.0",
-    "typescript": "^5.4.2",
+    "tsup": "8.5.1",
+    "typescript": "5.9.3",
     "vitest": "4.1.4"
   },
   "peerDependencies": {
-    "axios": ">=1.0.0 <1.14.1"
+    "axios": "1.14.0"
```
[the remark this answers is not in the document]
GeraBart: Intentional, not an oversight: this repo pins every third-party dependency to an exact version rather than a range, including peerDependencies like this one. Ranges aren't used here by design, so this isn't going to be relaxed.

template-shell/packages/react/package.json
```diff
@@ -37,25 +37,25 @@
...
+    "tsup": "8.5.1",
+    "typescript": "5.9.3",
     "vitest": "4.1.4"
   },
   "peerDependencies": {
     "@gears-frontx/framework": "*",
-    "@tanstack/react-query": ">=5.0.0",
-    "react": "^19.2.4",
-    "react-redux": ">=8.0.0"
+    "@tanstack/react-query": "5.101.4",
```
itechmeat, line 58: 🔴 The react package's @tanstack/react-query peer is now exact and non-optional, but all four template MFEs still pin 5.90.21 and the shell root still floats on ^5.90.0. I resolved the composed template tree (npm 11.6.2): no ERESOLVE - npm silently nests instead - and HEAD ends up with three react-query copies (5.102.8 root / 5.101.4 under packages/react / 5.90.21 under the MFE), four query-core, three toolkit, versus 2/2/1 before the PR. QueryClient contexts are module-scoped singletons, so a useQuery from @gears-frontx/react cannot see a provider the MFE mounts. Note framework's query-core peer can only ever be met by nesting (nobody in a template declares query-core directly), and the repo's own FEATURE.md:191 requires a semver range instead of an exact pin on the governed mfes->gts-plugin edge precisely so no duplicate runtime is forced. Cheapest fix in this PR: move the four MFE pins and the shell root's ^5.90.0 to 5.101.4 so the tree resolves to one query runtime - or, if module federation shared is expected to dedupe at runtime, record that assumption next to the peers.

template-shell/packages/react/package.json
```diff
@@ -37,25 +37,25 @@
...
+    "@types/react": "19.2.18",
+    "@types/react-redux": "7.1.34",
     "@vitejs/plugin-react": "4.3.4",
     "@vitest/coverage-v8": "4.1.4",
     "jsdom": "26.1.0",
-    "react": "^19.2.4",
-    "react-redux": "^9.1.0",
-    "tsup": "^8.0.0",
-    "typescript": "^5.4.2",
+    "react": "19.2.4",
```
itechmeat, line 49: 🔴 Commit 2687a36 did fix the lock desync (the CodeRabbit thread is marked addressed), but the follow-up "self packages bumps" commit edited manifests without a reinstall, so at HEAD template-shell/package-lock.json contradicts them again: the lock's root deps still say api 0.3.0-alpha.1 / gts-plugin 0.3.0-alpha.3 / mfes 0.3.0-alpha.3 against alpha.2/alpha.4/alpha.4 in the manifest, node_modules/@gears-frontx/api still resolves the alpha.1 tarball with the old axios peer range, and - added by this PR - packages/react/node_modules/react sits at 19.2.8 against the exact 19.2.4 pin. That nested react is provably stale: a clean re-resolve of these manifests produces no nesting at all, `npm install --package-lock-only` over the committed lock yields zero diffs (incremental reconcile will never prune it), and `npm ci` installs it silently - so our own tests run against 19.2.8 while apps get 19.2.4. For a determinism PR the lock contradicting the manifests defeats the point; one full regen of template-shell/package-lock.json closes it.

packages/api/package.json
```diff
@@ -38,12 +38,12 @@
     "@gears-frontx/eslint-config": "0.2.0-alpha.0",
     "@vitest/coverage-v8": "4.1.4",
     "axios": "1.14.0",
-    "tsup": "^8.0.0",
-    "typescript": "^5.4.2",
+    "tsup": "8.5.1",
+    "typescript": "5.9.3",
```
itechmeat, line 44: [Not a blocker] 🟡 Six packages move to typescript 5.9.3 + tsup 8.5.1 while framework and auth stay on exact 5.4.2 + 8.0.0, the four template MFEs stay on 5.4.2, cli/telemetry/cyber-pilot-kit keep ^5.4.2/^8.0.0, studio ^8.5.0, and the repo root 5.4.2 - so the root lock now hoists tsup 8.5.1 and typescript 5.4.2 with three nested 5.9.3 copies. Is the split intentional (e.g. framework/auth left back because a TS bump changes the emitted .d.ts and needs its own verification)? If so, one line in the PR body settles it; if not, aligning the monorepo on one typescript and one tsup would keep the next package from picking a version at random.

template-shell/package.json
```diff
@@ -66,9 +66,9 @@
     "test:unit:watch": "vitest --watch"
   },
   "dependencies": {
-    "@gears-frontx/api": "0.3.0-alpha.1",
-    "@gears-frontx/gts-plugin": "0.3.0-alpha.3",
-    "@gears-frontx/mfes": "0.3.0-alpha.3",
+    "@gears-frontx/api": "0.3.0-alpha.2",
+    "@gears-frontx/gts-plugin": "0.3.0-alpha.4",
+    "@gears-frontx/mfes": "0.3.0-alpha.4",
```
itechmeat, line 74: [Not a blocker] 🟡 The "pin every third-party dep exactly, peers included" rule is applied to packages/* and template-shell/packages/*, but the shell root - the one manifest a seeded project inherits as its own - still floats: @tanstack/react-query ^5.90.0 (the very range that pulls 5.102.8 and creates the third copy from the peer thread above), gts-ts ^0.3.0 against the exact 0.3.0 in gts-plugin, tsup ^8.0.0, typescript ^5.4.2, plus the radix/iconify/clsx/sonner/zod block. Separately, the root's exact @reduxjs/toolkit 2.11.2 violates state's and react's new exact 2.12.0 peer from day one, so npm nests 2.12.0 under both. If the shell root is a follow-up sweep, a pointer is enough - but the ^5.90.0 one is load-bearing for the query-runtime issue and belongs in this pass.

template-shell/package.json
```diff
@@ -66,9 +66,9 @@
     "test:unit:watch": "vitest --watch"
   },
   "dependencies": {
-    "@gears-frontx/api": "0.3.0-alpha.1",
-    "@gears-frontx/gts-plugin": "0.3.0-alpha.3",
-    "@gears-frontx/mfes": "0.3.0-alpha.3",
+    "@gears-frontx/api": "0.3.0-alpha.2",
```
itechmeat, line 72: [Not a blocker] 🟡 template-shell/package.json's dependency fields change in this PR with no version bump, and the version-bump gate passes: CONTRIBUTING.md:83 and the gate's own docblock (version-bump-on-change-check.mjs:33-36) list template-shell as governed, but the manifest is "private": true and isPrivate() drops it - the gate honestly reports 7 packages, not 8. Consumer impact is nil (the registry only has 0.1.0-alpha.0 and the composed app overrides the pin with file:.), so this is a docs-vs-code truth question: either teach the gate about private governed packages (then this PR owes a bump), or stop claiming template-shell in CONTRIBUTING and the docblock.

packages/gts-plugin/package.json
```diff
@@ -25,13 +25,13 @@
     "test:unit:watch": "vitest --watch"
   },
   "dependencies": {
-    "@gears-frontx/mfes": "0.3.0-alpha.3",
-    "@globaltypesystem/gts-ts": "^0.3.0"
+    "@gears-frontx/mfes": "0.3.0-alpha.4",
+    "@globaltypesystem/gts-ts": "0.3.0"
```
itechmeat, line 29: [Not a blocker] 🔵 The "everything third-party is pinned exactly, peers included" rule currently lives in one place - your reply in this PR's CodeRabbit thread; CONTRIBUTING's Versioning section doesn't have it. The title's "21 days" isn't findable either (no cooldown/minimumReleaseAge/renovate config anywhere, and the PR body is fully bot-generated), while the chosen pins are mixed-age: react-query 5.101.4 is six weeks behind (5.102.8 is five days old - plausibly inside a cooldown), axios 1.14.0 is from March (1.20.0 exists), react 19.2.4 is from January while 19.2.8 already sits in our own lock. And FEATURE.md's ecosystem-distribution rule bans exact pins on the governed edge for exactly the duplicate-runtime reason. One CONTRIBUTING section - what gets pinned exactly, what the cooldown is, how that coexists with the governed-edge ban - would close both this PR's questions and the next argument before it starts.

packages/gts-plugin/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@gears-frontx/gts-plugin",
-  "version": "0.3.0-alpha.3",
+  "version": "0.3.0-alpha.4",
```
itechmeat, line 3: [Not a blocker] ⚪ CONTRIBUTING.md:100-103's publish order lists state/mfes/api/i18n -> framework -> react -> studio/cli, but gts-plugin isn't in it - and it's published, bumped in this PR, and sits on the mfes edge. Worth adding its slot while the file is warm.

itechmeat (changes requested):
The direction is right and the intra-ecosystem part is done carefully: all seven bumps are propagated to every consumer with no orphaned pin, the root lockfile is consistent, and the governed mfes -> gts-plugin caret is preserved. Two things keep me from approving as submitted, both cheap to fix: the new exact runtime peers make the repo's own composed templates install three copies of react-query (a silently split QueryClient context - npm nests instead of erroring, I verified the resulting tree), and template-shell/package-lock.json at HEAD contradicts the manifests again after the last commit edited them without a regen. Everything else is marked [Not a blocker]; happy to approve once the templates resolve to one query runtime and the lock is regenerated.

template-shell/packages/react/package.json
GeraBart: Fixed: aligned @tanstack/react-query to a single exact 5.101.4 across template-shell/package.json, all 4 template-mfe fixtures, and this package — same treatment already applied to @reduxjs/toolkit below. Verified via a real npm install (temporarily local-linking the unpublished ecosystem packages) that the tree now resolves to one copy each, no nesting.

template-shell/packages/react/package.json
GeraBart: Fixed: regenerated template-shell/package-lock.json using this repo's own pin-template-ecosystem-to-local.mjs technique (temporarily rewrite the api/mfes/gts-plugin pins to local file: paths, install for real, then restore the manifest) — the stale nested react@19.2.8 is gone, confirmed a single deduped 19.2.4 entry. The remaining known limitation (api/mfes/gts-plugin can't resolve from the registry pre-merge) now fails loudly on a real npm ci instead of silently installing something stale, which is the best achievable state before those packages are published.

packages/api/package.json
GeraBart: Not intentional — aligned framework and auth to typescript 5.9.3 / tsup 8.5.1 (matching the other six packages), plus the shell root and the 4 template-mfe fixtures. cli/telemetry/cyber-pilot-kit-frontx/studio are outside this PR's dependency closure so left untouched, but happy to fold those in separately if wanted.

template-shell/package.json
GeraBart: Swept the remaining floating ranges in template-shell/package.json to exact pins (radix/iconify/clsx/sonner/tailwind-merge/zod/gts-ts/tsup/typescript/module-federation-vite/testing-library-user-event/esbuild/vite), including the react-query one — confirmed it's now 5.101.4 everywhere in the tree, closing the duplicate-runtime issue you flagged above.

template-shell/package.json
GeraBart: Good catch, and confirmed accurate — the gate's isPrivate() check does skip template-shell even though the docs/comments claim it's governed. Didn't change the gate logic itself in this PR (that's a separate design call: teach the gate about private-but-governed packages, or narrow the docs to match what it actually checks), but I don't want to silently leave the discrepancy undocumented either — flagging for a maintainer decision rather than picking one unilaterally in a dependency-pinning PR.

packages/gts-plugin/package.json
GeraBart: Added a Dependency Pinning subsection under Versioning in CONTRIBUTING.md covering the exact-pin rule, the 21-day eligibility window, and the one explicit exception (the mfes->gts-plugin edge staying a range). Left the specific pin ages as-is per your note (react-query's 5101.4 vs a fresher-but-too-young 5.102.8, axios/react being much older) since that's just each package's own release cadence, not something to normalize.

packages/gts-plugin/package.json
GeraBart: Added @gears-frontx/gts-plugin to the Publish Order list in CONTRIBUTING.md, right after @gears-frontx/mfes (matches the actual publish-ecosystem job ordering, since gts-plugin exact-depends on mfes).

CONTRIBUTING.md
```diff
@@ -82,6 +82,14 @@ The project is **pre-1.0** — backward compatibility is not guaranteed.
 
 A PR that changes non-documentation source under `src/`, or the dependency fields of `package.json`, for a governed package (non-private `@gears-frontx`-scoped packages: `packages/*`, `template-shell`, and `template-shell/packages/*`) must bump that package's own `version` in the same PR — and update every exact pin on it (see `policy:template-pin-drift`). CI (`policy:version-bump-on-change`, pull requests only) compares the version at the PR's merge base against the version at its head, so a bump that is later reverted within the same PR does not count. Packages the PR itself adds or removes are exempt, as are private packages (e.g. the `template-mfe` fixture apps), which are pinned consumers rather than published sources.
 
+### Dependency Pinning
+
+Third-party dependencies are pinned to exact versions so installs are deterministic.
+
+- **Exact pins everywhere** — every non-`@gears-frontx` dependency in `packages/*`, `template-shell`, and `template-shell/packages/*` declares an exact version, never a range.
```
itechmeat, line 89: 🟠 [Not a blocker] This line says "declares an exact version, never a range", and eight devDependencies added by this same PR are `"@types/node": "^24.0.0"` (api, gts-plugin, mfes, auth, framework, i18n, react, state) - all inside the directories this bullet names, and the "One exception" bullet covers only the mfes->gts-plugin edge. The caret also structurally defeats the 21-day rule: ^24.0.0 resolves to whatever 24.x is newest at install time (today 24.13.3, 58 days old - fine; tomorrow possibly a same-morning release). Either pin 24.13.3 in all eight, or name @types/node as a second exception with one line of why - both make the doc and the tree agree.

template-shell/packages/auth/package.json
```diff
@@ -29,8 +29,9 @@
   },
   "dependencies": {},
   "devDependencies": {
-    "tsup": "8.0.0",
-    "typescript": "5.4.2"
+    "@types/node": "^24.0.0",
+    "tsup": "8.5.1",
```
itechmeat, line 33: 🟠 [Not a blocker] The lock regen didn't fully take: packages/auth/node_modules/tsup in template-shell/package-lock.json is still 8.0.0 against this manifest's 8.5.1, with 27 stale transitives riding along (esbuild 0.19.12 x24 platform binaries and friends) - `npm ls tsup` on the committed tree answers `invalid: "8.5.1" from packages/auth` and exits ELSPROBLEMS, while a delete-and-reinstall gives 532 packages instead of 559 with one clean tsup. The @gears-frontx resolved entries (alpha.1/alpha.3 under root deps saying alpha.2/alpha.4) genuinely can't sync until the alphas publish, so both halves fold naturally into that inevitable post-publish lock refresh - just worth folding the tsup nest in deliberately rather than discovering it later, since the thread reply says "regenerated from scratch" and this part survived it.

itechmeat (approved):
🟢 All of round 1 is properly closed, and I verified the closures rather than trusting the manifests: a fresh template-shell install and a composed-template build (shell + all four MFEs + the ecosystem via file: overrides) both resolve to exactly one copy of react-query / query-core / toolkit / react / react-dom / react-redux, with npm ls showing every MFE deduped and zero peer warnings; the phantom nested react@19.2.8 is gone. The CONTRIBUTING section, the publish-order slot, the shell-root sweep and the typescript/tsup alignment all landed too, and every third-party pin I spot-checked (25 of them) satisfies the new 21-day rule. Approving. Two non-blocking asks are inline (the ^24.0.0 ranges vs the rule this PR writes, and a tsup 8.0.0 nest that survived the lock regen), plus one that can't be an inline comment because the file is deleted: architecture/explorations/2026-07-16-multi-template-composition.md is removed in this PR (-220 lines) with no stated reason - develop still has it, it's the only file in architecture/explorations/ so the directory itself disappears, and four agent definitions under .claude/agents/ reference that directory as a live location. If it's obsolete I'm not arguing - just not here: revert it in this PR and retire it in a separate one-liner with the why.

template-mfe/src-app/mfe_packages/widgets-fixture-a/package.json
```diff
@@ -13,24 +13,24 @@
...
+    "@gears-frontx/mfes": "0.3.0-alpha.4",
+    "@gears-frontx/gts-plugin": "0.3.0-alpha.4",
     "@gears-frontx/frontx-template-shell": "0.1.0-alpha.4",
-    "@gears-frontx/api": "0.3.0-alpha.1",
-    "@gears-frontx/i18n": "0.2.0-alpha.0",
-    "@reduxjs/toolkit": "^2.11.2",
+    "@gears-frontx/api": "0.3.0-alpha.2",
+    "@gears-frontx/i18n": "0.2.0-alpha.1",
+    "@reduxjs/toolkit": "2.12.0",
     "react-redux": "^9.2.0"
```
tscbmstubp, line 34: 🟡 [Not a blocker] `react-redux` stays `^9.2.0` here, in `widgets-fixture-b` and in `_blank-mfe`, while `@gears-frontx/react` now declares an exact, non-optional `react-redux: 9.3.0` peer. Today both resolve to 9.3.0, so the lock is clean. The next 9.x release flips that: the fixture's caret takes the new version and npm nests a second `react-redux` under `@gears-frontx/react`, which is the same split-context shape this PR just closed for react-query. Pinning `9.3.0` in the three fixtures keeps them in step with the peer. (`demo-mfe`'s `concurrently ^9.2.1` and `@testing-library/user-event ^14.6.1` are harmless floats, mentioned only for the sweep.)

tscbmstubp (changes requested):
Verified the dependency work independently at 1ed8b44: all 26 new third-party pins are at least 21 days old (youngest: sonner 2.0.8 at 28d, esbuild 0.28.2 at 29d), the `mfes -> gts-plugin` peer stays a caret, all eight bumps reach every consumer manifest and all three lockfiles, and the template-mfe and root locks resolve one copy each of react-query / query-core / toolkit / react-redux. The new `@reduxjs/toolkit` and `react-dom` peers back real imports, and the api `lodash` devDependency matches the existing `lodash/assign.js` import in `SseProtocol.ts`.

One thing I'm holding on: the `architecture/explorations/2026-07-16-multi-template-composition.md` deletion in 1ed8b44 that @itechmeat raised in the approval body. `architecture/` is the repo's stated authority tree, so a dependency PR shouldn't retire an artifact without a reason. Restoring it from `develop` clears this.

The open `@types/node` and `tsup` threads stand as written; I added one note to the `@types/node` thread and one new inline comment on the fixtures.

FYI on the red `Template Validate` check: it fails with `ETARGET` on `@gears-frontx/api@0.3.0-alpha.2` because the alphas aren't published yet. The docblock in `scripts/pin-template-ecosystem-to-local.mjs` says this check is expected to stay red until publish, so merging means accepting it, and a post-publish lock refresh for template-shell is owed.

CONTRIBUTING.md
tscbmstubp: Adding one data point to this thread: the same sentence is also contradicted by `template-shell/packages/studio/package.json`, which this PR doesn't touch and which carries thirteen caret ranges (`react ^19.2.4`, `tsup ^8.5.0`, `typescript ^5.4.2`, the radix/clsx/tailwind-merge block, `@types/*`). So even with `@types/node` pinned, the bullet as written isn't true of the tree. Either sweep `packages/studio` in this PR or narrow the scope sentence to the directories that actually comply, and note studio as a follow-up.

tscbmstubp (approved):

PR #608, opened 2026-09-01
fix(cli): make upgrade provenance write part of the atomic apply boundary

### LINKED ISSUE #506: [BUG] Make upgrade application atomic through the provenance write

tscbmstubp:
Fixes #506.

## Problem

`applyChangeSet` ran the final `deps.writeProvenance(...)` after the try/catch boundary that protects project-file mutations. A filesystem failure there threw past the promised `ApplyResult` failure path after approved project files were already changed, leaving the repository partially upgraded.

## Change

- **`packages/cli/src/upgrade/apply.ts`** — moved the provenance write (`inst-app-update-prov`) inside the existing `try` block. `.frontx/provenance.json` was already snapshotted at the top of the function, so the existing `catch` now restores it along with every mutated project file and returns typed `{ok: false}`. The success path (multi-record provenance array from #500, ADR-0021 semantics) is byte-identical apart from indentation.
- **`packages/cli/src/__tests__/upgrade.test.ts`** — regression test: applies a change set that mutates `src/App.tsx`, then fails `writeProvenance`; asserts the project file and provenance file are restored byte-for-byte and `applyChangeSet` returns `{ok: false}` instead of throwing. The test fails on `develop` and passes here.
- **`packages/cli/architecture/features/upgrade-changeset/FEATURE.md`** — `inst-app-update-prov` renumbered under `inst-app-try`; rollback contract wording, DoD, and acceptance criteria updated to cover the provenance-write failure. `cfs validate`: 0 errors, coverage 171/171.

## Verification

- Upgrade test files: all pass (incl. new regression test); full `packages/cli` suite: 412 passed, with only the pre-existing `@gears-frontx/test-support/path-guard` resolution failures identical on `develop`.
- `tsc --noEmit`: no new errors. `eslint` on changed files: clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Upgrade operations now roll back all modified project files when provenance updates fail.
  * Previous provenance data is restored when an upgrade cannot be completed.
  * Upgrade failures now return a structured error instead of leaving partial changes or unexpectedly throwing.
  * Prevented partially applied upgrades when storage errors occur after files have been modified.

* **Tests**
  * Added coverage for failures occurring after project files have been modified, including restoration of provenance data.

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@gears-frontx/cli",
-  "version": "0.3.0-alpha.4",
+  "version": "0.3.0-alpha.5",
   "description": "Template resolution CLI for FrontX SDK — zero bundled template content",
   "type": "module",
   "main": "./dist/index.cjs",
--- a/packages/cli/src/upgrade/apply.ts
+++ b/packages/cli/src/upgrade/apply.ts
@@ -121,8 +121,8 @@ export async function applyChangeSet(
   // nothing has been written yet, so `{ok:false}` is returned directly, per
   // ApplyResult's contract, with the project untouched — never a thrown
   // error surfacing after entries have already been rewritten on disk with
-  // no way back (`inst-app-restore-on-error`'s snapshot only covers the
-  // for-each-entry loop below, not a failure ahead of it).
+  // no way back (`inst-app-restore-on-error`'s restore only covers the
+  // `try` block below, not a failure ahead of it).
   let existingRecords: ProvenanceRecord[];
   try {
     existingRecords = parseProvenanceRecordSet(provContent, provPath);
@@ -176,6 +176,29 @@ export async function applyChangeSet(
       // @cpt-end:cpt-frontx-algo-upgrade-changeset-apply:p1:inst-app-apply-entry
     }
     // @cpt-end:cpt-frontx-algo-upgrade-changeset-apply:p1:inst-app-for-each-entry
+
+    // @cpt-begin:cpt-frontx-algo-upgrade-changeset-apply:p1:inst-app-update-prov
+    // Provenance is a SET — one record per applied template (ADR-0019) — never
+    // a single whole-repository record. `existingRecords`/`targetIndex` were
+    // validated above, before any file was written (see the precondition
+    // block for why validation stays outside this `try`). Only the record for
+    // the template this upgrade targets changes; every other applied
+    // template's record is re-serialized untouched (ADR-0021 Confirmation
+    // (d): "the other applied template and its provenance record are
+    // unaffected").
+    //
+    // Invariant: the `writeProvenance` FILESYSTEM WRITE is not pure — it can
+    // fail exactly like any project-file write can (issue #506) — so it MUST
+    // stay inside this `try`, on the same rollback boundary as
+    // `inst-app-for-each-entry`. A failure here lands in the `catch` below,
+    // which restores every snapshotted file — `provPath` included,
+    // snapshotted at the top of this function — so the prior provenance
+    // bytes come back exactly like any rolled-back project file.
+    const updatedRecords = existingRecords.map((record, index) =>
+      index === targetIndex ? { ...record, scaffoldedFromVersion: changeSet.targetVersion } : record,
+    );
+    await deps.writeProvenance(provPath, JSON.stringify(updatedRecords, null, 2));
+    // @cpt-end:cpt-frontx-algo-upgrade-changeset-apply:p1:inst-app-update-prov
   } catch (err) {
     // @cpt-end:cpt-frontx-algo-upgrade-changeset-apply:p1:inst-app-try
     // @cpt-begin:cpt-frontx-algo-upgrade-changeset-apply:p1:inst-app-catch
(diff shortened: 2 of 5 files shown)


packages/cli/src/upgrade/apply.ts
```diff
@@ -176,6 +176,35 @@ export async function applyChangeSet(
...
+    // already validated above, before any file was written, so building
+    // `updatedRecords` is a pure substitution into an already-known-good
+    // array — it cannot itself throw. Only the record for the template this
+    // upgrade targets changes; every other applied template's record is
+    // re-serialized untouched (ADR-0021 Confirmation (d): "the other applied
+    // template and its provenance record are unaffected").
+    //
+    // The `writeProvenance` FILESYSTEM WRITE right below is not pure — it
+    // can fail exactly like any project-file write can (issue #506) — so,
+    // unlike the computation above it, it stays inside this `try`, on the
```
itechmeat: [Not a blocker] 🟡 The move makes three neighbor comments literally false: apply.ts:124 ("inst-app-restore-on-error's snapshot only covers the for-each-entry loop below") and the "only wraps the for-each-entry loop" wording at upgrade.test.ts:420 and :487 - the try now also covers the provenance write. Their precondition-failure argument still stands directionally, but the sentences contradict the code, and both files are already edited by this PR. Worth fixing in the same pass?

packages/cli/src/upgrade/apply.ts
```diff
@@ -176,6 +176,35 @@ export async function applyChangeSet(
...
+    // same rollback boundary as `inst-app-for-each-entry`. #500 removed two
+    // earlier throw paths that used to run ahead of the first write, but
+    // left this final write outside the boundary: a failure here used to
+    // throw after approved project files were already changed, leaving the
+    // repository partially upgraded and bypassing `ApplyResult`'s
+    // `{ok:false}` failure path. Now a `writeProvenance` failure lands in
+    // the same `catch` below, which restores every snapshotted file —
```
itechmeat: [Not a blocker] 🟡 On CodeRabbit's open "Major": it's technically accurate - the catch wraps each restore in its own try/catch that swallows the error and then unconditionally returns "Apply failed and was restored", so a failed restore is reported as a successful one. But that block is byte-for-byte identical on develop, #506's acceptance criteria don't cover restore-failure reporting, and fixing it means changing ApplyResult's surface (which would also collide with #598). This PR only widens the radius marginally - the provenance file becomes one more restore candidate, versus the strictly worse pre-PR behavior of throwing past ApplyResult entirely. I'd file it as its own follow-up issue rather than gate this fix - the comment's "restores every snapshotted file" is best-effort in reality.

packages/cli/src/upgrade/apply.ts
```diff
@@ -176,6 +176,35 @@ export async function applyChangeSet(
...
+    // `updatedRecords` is a pure substitution into an already-known-good
+    // array — it cannot itself throw. Only the record for the template this
+    // upgrade targets changes; every other applied template's record is
+    // re-serialized untouched (ADR-0021 Confirmation (d): "the other applied
+    // template and its provenance record are unaffected").
+    //
+    // The `writeProvenance` FILESYSTEM WRITE right below is not pure — it
+    // can fail exactly like any project-file write can (issue #506) — so,
+    // unlike the computation above it, it stays inside this `try`, on the
+    // same rollback boundary as `inst-app-for-each-entry`. #500 removed two
```
itechmeat: [Not a blocker] 🔵 The invariant half of the new comment block earns its place (the write is impure, it lives on the rollback boundary - that stops the next reader from "optimizing" it back out). The history half ("#500 removed two earlier throw paths", "used to throw...") reads like PR-description material rather than code. If you trim it, also weigh the purity paragraph above: with the map now inside the try, "it cannot itself throw" stopped being load-bearing - the ADR-0021 sentence and the why-validation-stays-outside link are the parts worth keeping. Up to you.

packages/cli/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@gears-frontx/cli",
-  "version": "0.3.0-alpha.3",
+  "version": "0.3.0-alpha.4",
```
itechmeat: [Not a blocker] 🔵 The bump is required by the version gate, but open PR #598 also claims cli 0.3.0-alpha.4 - whichever of the two lands second will need to re-bump. Just flagging for ordering, nothing to change here unilaterally.

itechmeat (approved):
Clean, well-scoped fix - approving. The provenance write now genuinely shares the rollback boundary with the project-file loop, the regression test proves the restore actually overwrites a corrupted-in-place provenance file rather than passing vacuously, the @cpt marker chain and FEATURE renumbering stay consistent, and the only caller never depended on the old ordering. The notes inline are all non-blocking: a few neighbor comments the move made stale, a follow-up worth filing on the best-effort restore reporting (pre-existing, not this PR's bug), and one version-number coordination point with another open PR.

tscbmstubp:
Pushed edb19ac1 addressing the non-blocking notes: reworded the three stale "only wraps the for-each-entry loop" comments (apply.ts precondition block, upgrade.test.ts (j)/(k) and (l)/(m) preambles) and trimmed the new provenance comment to its invariant half. The restore-failure reporting item is filed as a separate issue (linked above). Version collision with #598 left as-is per the ordering note.

GeraBart (approved):

PR #611, opened 2026-09-03
[DO NOT MERGE] Machine-readable component contract for Button: semantic overlay compiled with facts extracted from the source

itechmeat:
## The problem

Agents consume UiKit knowledge as prose. READMEs, doc comments and demo screens describe what a component means, when to reach for it and what not to do with it, and none of that is checked against the component. Prose drifts the moment the code moves: a variant gets renamed, a default flips, a prop is removed, and the description keeps reading as though nothing happened. There is no artifact an agent can load that is both semantic enough to guide a choice and validated enough to trust.

## What this demo materializes

One contract per component, compiled from two halves that never restate each other.

The hand-written half is a semantic overlay (`button.contract.yaml`) carrying only what the code cannot express: the component's intent as a single sentence, when to use it, when not to and what to reach for instead, composition rules, named invariants, and anti-patterns paired with their replacement. Machine-extractable facts are banned from the overlay by construction, and the compiler rejects an overlay that tries to restate one. One fact, one owner.

The extracted half comes from the source. An extractor on the TypeScript compiler API reads `button.tsx` and its `cva()` call to recover the variant axes, their values, the defaults and the component's own props, and it reports anything it could not read rather than guessing.

The compiler merges the two into a canonical JSON Schema 2020-12 document with a GTS id. Structure worth noting:

- The contract structure itself is a GTS meta-schema type (`ui-component.meta.json`), so a contract is an instance of a declared type rather than an ad hoc shape. The compiled instance validates against that metamodel under Ajv's 2020 build.
- Cross-references are typed, not decorative. An `instead` target in `dont_use_when` is a GTS component type id (`gts.frontx.uikit.component.navigation_menu.v1~`), not a display name a reader has to resolve by hand. A name is a string nothing can check; an id is checked against what the kit actually ships.
- Good and bad canonical examples travel inside the contract, and the bad ones must say why they are bad.

A conformance vitest (`button.contract.test.ts`) fails the build whenever code and contract disagree, across 11 checks: every `cva` axis and value mirrored in both directions, `defaultVariants` mirrored, the overlay referencing only props that exist in code, a grammatical GTS `$id` and instance id, the extractor reporting nothing unreadable, the instance validating against the metamodel, the instance pointing at the props schema it was compiled with, every `dont_use_when` alternative resolving to a component directory the kit really has, every good example parsing as TSX, and every bad example carrying a reason.

## The tsconfig restructuring that supports it

The contract test imports from `scripts/`, which no project covered, and that exposed a gap worth closing on its own terms.

`packages/ui-kit/tsconfig.json` becomes a solution-style root: `files: []` plus `references`, compiling nothing itself. What it buys is editor routing. The language server sends each open file to the first referenced project whose `include` covers it, so sources, build tooling and colocated tests each get a real project with this package's real compiler settings instead of an inferred one on compiler defaults. Reference order matters and sources stay first.

The shipping-sources project moves to the new `tsconfig.src.json`, which keeps the declaration-emission boundary intact: it is scoped to `src/**/*` with `rootDir: ./src` and excludes tests, spec files and `__test-utils__`, and `vite-plugin-dts` now reads it (see `scripts/buildPlugin.ts`) to decide what lands in `dist/`. `tsconfig.test.json` and `tsconfig.tools.json` extend it for their compiler settings and widen `rootDir` to the package root.

This also closes a real hole: `vitest.config.ts` was type-checked by nothing. It was outside the sources project by design and was never listed in the tooling project, so `npm run type-check` skipped it and an editor fell back to compiler defaults on the one config file that decides how the suite runs. It is now in `tsconfig.tools.json` alongside `vite.config.ts`.

## This is a demo, not a rollout

Stated plainly so nothing here reads as a proposal to adopt: this exists to give the contracts idea something concrete to argue about. One component (Button) carries a contract. The meta-schema is marked `0.1.0-demo`. Nothing else in the kit is touched, no generation step is wired into the build, and no consumer reads a contract yet. If the shape is wrong, the cost of changing it is one component.

## Verification

Run in `packages/ui-kit` with the full change in place:

- `npm run type-check` passes (`tsc -p tsconfig.src.json --noEmit` and `tsc -p tsconfig.tools.json`).
- `npm run type-check:test` passes, exercising the restructured test project.
- `npx vitest run`: 67 files, 920 tests, all passing.
- `npm run build` succeeds, declarations emitted in 2260ms, 191 `.d.ts` in `dist`.
- Zero test declarations in `dist`: `find dist \( -name '*.test.*' -o -name '*.spec.*' -o -name '__test-utils__' \)` returns nothing, so the emission boundary survived the tsconfig move.

From the repo root:

- `npm run arch:check`: all 12 checks pass. The unused-exports numbers match the `develop` baseline exactly (1 unused file, 2 unused exports, 1 unused exported type), verified by running the same check on a pristine `develop` tree.
- `node scripts/version-bump-on-change-check.mjs --base-ref origin/develop` passes.

## Risks and limits

- The overlay is parsed with the `yaml` package (YAML 1.2, declared as a dev dependency of this package), so bare `no`/`on` stay strings instead of turning into booleans.
- Compiled contracts are not in this package's `files` allowlist, so they do not reach consumers. Shipping a `dist/contracts` copy, and diffing it against a fresh compile for freshness, is intended CI behaviour and not what happens today.
- Examples are syntax-checked only. The conformance test parses each good example as TSX, which catches malformed code but not a prop that does not exist or has the wrong type. Checking that needs a `tsc` program against the kit's built declarations, so the demo stops at parse.
- `knip.json` gains one line. The solution-style root no longer carries `rootDir`/`outDir`, so knip stopped inferring `src/index.ts` as the package entry and reported every `public.ts` as unused. Naming the entry explicitly restores the previous result; the file is strict JSON validated by a `check-json` hook, so the reasoning lives here rather than in a comment.
- The `ui-kit` version bump to `0.4.0-alpha.3` is included because the change guard treats the new `src/` files as substantive (develop claimed `alpha.2` in the meantime).
## CI status, explained

Template Validate is red because this branch pins `@gears-frontx/ui-kit@0.4.0-alpha.3`, which only gets published once the version lands on develop: the job packs template-shell as a standalone package and runs `npm ci` against the public registry, which fails on the unpublished version. Any PR that bumps the kit version stays red on that check until publication; develop itself is green. The unit-test red is the deliberate kit-wide completeness guard described below; `validate (24.14.x)` shows as failed only because the matrix cancels it when 25.x fails. TypeScript validation, lint and every other unit test are green.

## Deliberately red: the completeness guard

`packages/ui-kit/src/contracts.test.ts` applies the contract definition-of-done to every directory under `src/components`: the hand-written overlay `<name>.contract.yaml` exists, the two compiled artifacts next to it are byte-identical to a fresh recompile, the instance validates against the component metamodel, and the compiled axes still mirror the `cva()` call in the code.

Exactly 1 of the 63 components satisfies it today (Button, all six checks green). The other 62 fail with one named line each: `component "<name>": no contract overlay (<name>.contract.yaml) - full anatomy reference: src/components/button`. So the unit-test CI step is red by construction: 62 failed, 6 passed.

That is the point of this branch. The red wall is the live demonstration that the validator runs kit-wide and names precisely what is missing, and the green Button block next to it reads as the full anatomy checklist a covered component has to satisfy. The guard is deliberately unconditional here - no ratchet, no allowlist. A production rollout would phase the same gate in diff-aware, gating touched components first and widening as overlays land; the kit-wide run stays available as a coverage report rather than a blocking gate, so nobody inherits a wall of red they did not touch.

### DIFF
--- a/packages/ui-kit/scripts/contracts/compile.ts
+++ b/packages/ui-kit/scripts/contracts/compile.ts
@@ -34,15 +34,17 @@
 //        The instance path is outPath with `.json` swapped for
 //        `.instance.json`; both files are written together.
 import { mkdirSync, readFileSync, writeFileSync } from 'node:fs';
-import { createRequire } from 'node:module';
 import { basename, dirname, join } from 'node:path';
 import { fileURLToPath, pathToFileURL } from 'node:url';
 
-import { extractComponent, type Extraction } from './extract';
+// YAML 1.2, where a bare `no`/`off`/`yes`/`on` stays a string. Under the
+// YAML 1.1 rules js-yaml implements, those spellings parse as booleans, so
+// an overlay writing one unquoted (a `hint: no`, an anti-pattern `dont: on`)
+// would reach the compiler as `false`/`true` and fail the metamodel's
+// `type: string` far from the line that caused it.
+import { parse as parseYaml } from 'yaml';
 
-// Demo shortcut: js-yaml is what the workspace already hoists. Production
-// pins the `yaml` package (YAML 1.2) so bare `off`/`no` stay strings.
-const yaml = createRequire(import.meta.url)('js-yaml') as { load: (text: string) => unknown };
+import { extractComponent, type Extraction } from './extract';
 
 export interface Examples {
   good: { title: string; code: string }[];
@@ -193,7 +195,7 @@ export function propsSchemaId(component: string): string {
 
 function loadOverlay(component: string): Overlay {
   const path = join(kitRoot, 'src', 'components', component, `${component}.contract.yaml`);
-  const overlay = yaml.load(readFileSync(path, 'utf8')) as Overlay & Record<string, unknown>;
+  const overlay = parseYaml(readFileSync(path, 'utf8')) as Overlay & Record<string, unknown>;
 
   const shadowed = MACHINE_OWNED.filter((key) => key in overlay);
   if (shadowed.length > 0) {
--- a/packages/ui-kit/scripts/contracts/compile.ts
+++ b/packages/ui-kit/scripts/contracts/compile.ts
@@ -1,7 +1,9 @@
 // Contract compiler: merges code-extracted facts with the hand-written
 // overlay into two GTS-typed artifacts.
 //
-//   <out>.json          - props schema, $id gts://gts.frontx.uikit.component.<c>.v1~
+//   <out>.json          - props schema, a DERIVED type whose $id chains the
+//                         abstract base component type:
+//                         gts://gts.frontx.uikit.base.component.v1~frontx.uikit.component.<c>.v1~
 //   <out>.instance.json - contract instance, typed by the metamodel
 //                         gts://gts.frontx.uikit.meta.component.v1~
 //
@@ -9,6 +11,19 @@
 // the metamodel is a GTS type, so a malformed contract fails the same
 // validator that checks component props, not a review comment. The two stay
 // linked through the instance's props_schema field.
+//
+// A component's props schema is not a standalone schema that happens to look
+// like its neighbours: it derives from base.component.json and composes one
+// shared passthrough type per element kind (passthrough.dom-button.json for
+// anything backed by a <button>). Those two are hand-written source, not
+// compiled output - they describe no component's code, so there is nothing to
+// extract for them; they live next to this file and are read here and by the
+// conformance test. What the derivation buys is closure: because the
+// forwarded DOM props are declared by a schema this one $refs, the derived
+// type can set `unevaluatedProperties: false` and reject a typo'd kit prop
+// without also rejecting className, aria-* or data-*. `additionalProperties`
+// could not do that job - it cannot see through $ref or allOf, so it would
+// reject every inherited prop.
 //
 // The compiled JSON is the canonical contract every consumer reads (Ajv,
 // gtsPlugin.registerSchema, projections, the lint). In the real build it is
@@ -59,9 +74,19 @@ export interface ContractInstance extends Omit<Overlay, 'component'> {
 }
 
 export interface ContractProperty {
-  type: string;
+  // Absent for a slot: a ReactNode or render prop has no JSON Schema type.
+  // The property is still declared (with annotations only, no assertions) so
+  // that `unevaluatedProperties: false` counts it as evaluated and lets it
+  // through - a contract that rejected `icon` would be wrong, not strict.
+  // Its real type stays in x-uikit.slots, where the lint and tsc read it.
+  type?: string;
   enum?: string[];
   default?: string;
+  description?: string;
+}
+
+export interface SchemaRef {
+  $ref: string;
 }
 
 export interface CompiledContract {
@@ -69,7 +94,15 @@ export interface CompiledContract {
   $schema: string;
   title: string;
   type: 'object';
+  // Base first: gts-ts treats the FIRST $ref in allOf as the parent of the
+  // derived type (store.findParentRef), which is what makes the chained $id
+  // above and the schema body agree about who the parent is.
+  allOf: SchemaRef[];
   properties: Record<string, ContractProperty>;
+  // Closes the kit's own surface. Must be `unevaluatedProperties`, not
+  // `additionalProperties`: the latter only sees sibling `properties` and
+  // would reject everything reached through the allOf refs above.
+  unevaluatedProperties: false;
   'x-uikit': {
     metamodel: string;
     slots: Record<string, { typeText: string; optional: boolean }>;
@@ -85,6 +118,50 @@ const METAMODEL_VERSION = '0.1.0-demo';
 
 // Type id of ui-component.meta.json; prefixes every instance id.
 const METAMODEL_TYPE_ID = 'gts.frontx.uikit.meta.component.v1';
+
+// Type id of base.component.json - the parent of every component props
+// schema, and the first segment of their chained ids.
+//
+// Namespace `base`, type `component`, NOT the shorter
+// `gts.frontx.uikit.component.v1~` the review sketched: gts-ts strips the
+// fixed `gts.` scheme prefix and then requires each segment to carry 5 or 6
+// dot-tokens (vendor.package.namespace.type.vMAJOR[.minor], see
+// Gts.parseSegment). `frontx.uikit.component.v1` is four, so that id parses
+// as "Too few tokens"; `frontx.uikit.base.component.v1` is five and parses
+// as vendor=frontx, package=uikit, namespace=base, type=component, v1.
+export const BASE_TYPE_ID = 'gts://gts.frontx.uikit.base.component.v1~';
+
+// Shared passthrough type per element kind. Everything the kit backs with a
+// <button> composes this one; a link-like component would chain a
+// dom_anchor type instead of restating the same attributes.
+export const PASSTHROUGH_TYPE_ID = 'gts://gts.frontx.uikit.passthrough.dom_button.v1~';
+
+// The hand-written normative types this compiler composes. Read from disk
+// rather than inlined so the artifacts, the conformance test and Ajv all see
+// one copy of each.
+const SCHEMA_DIR = dirname(fileURLToPath(import.meta.url));
+
+export function loadBaseSchema(): Record<string, unknown> {
+  return JSON.parse(readFileSync(join(SCHEMA_DIR, 'base.component.json'), 'utf8')) as Record<string, unknown>;
+}
+
+export function loadPassthroughSchema(): Record<string, unknown> {
+  return JSON.parse(readFileSync(join(SCHEMA_DIR, 'passthrough.dom-button.json'), 'utf8')) as Record<
+    string,
+    unknown
+  >;
+}
+
+// Props the passthrough type already owns. A component that re-declares one
+// of them in its own Props interface (Button re-declares `className`,
+// because Base UI's Props omits it) must not get a second, competing
+// declaration in its own schema: one prop, one owner. They stay allowed -
+// the passthrough $ref evaluates them.
+function passthroughOwnedProps(): Set<string> {
+  const passthrough = loadPassthroughSchema();
+  const properties = (passthrough.properties ?? {}) as Record<string, unknown>;
+  return new Set(Object.keys(properties));
+}
 
 // Machine-owned fields the overlay must not restate - one fact, one owner.
 const MACHINE_OWNED = ['axes', 'props', 'defaults', 'variants'];
@@ -106,8 +183,12 @@ function gtsToken(component: string): string {
 
 // The props schema's own type id - the value the instance's props_schema
 // points at, so the two artifacts can never drift apart by hand.
-function propsSchemaId(component: string): string {
-  return `gts://gts.frontx.uikit.component.${gtsToken(component)}.v1~`;
+//
+// A GTS derived-type id is the parent id followed by the derived segment,
+// each segment closed by `~`: the component is not a namesake of the base
+// type, it is a child of it, and the id says so without a lookup.
+export function propsSchemaId(component: string): string {
+  return `${BASE_TYPE_ID}frontx.uikit.component.${gtsToken(component)}.v1~`;
 }
 
 function loadOverlay(component: string): Overlay {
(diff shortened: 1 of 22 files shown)


packages/ui-kit/src/components/button/button.contract.json
```diff
@@ -0,0 +1,124 @@
+{
+  "$id": "gts://gts.frontx.uikit.component.button.v1~",
```
GeraBart: I would suggest to create a base type called "component":
`gts.frontx.design.uikit.component.v1~`
Then "button" would become its instance or derived type.

packages/ui-kit/src/components/button/button.contract.json
```diff
@@ -0,0 +1,124 @@
...
+      "type": "string"
+    },
+    "loading": {
+      "type": "boolean"
+    },
+    "focusableWhenDisabled": {
+      "type": "boolean"
+    }
+  },
+  "x-uikit": {
```
GeraBart, line 47: Sorry, not getting it. What is "x-uikit"? I mean, it's not a property, and not a known gts trait.

packages/ui-kit/scripts/contracts/compile.ts
```diff
@@ -0,0 +1,224 @@
...
+import { mkdirSync, readFileSync, writeFileSync } from 'node:fs';
+import { createRequire } from 'node:module';
+import { basename, dirname, join } from 'node:path';
+import { fileURLToPath, pathToFileURL } from 'node:url';
+
+import { extractComponent, type Extraction } from './extract';
+
+// Demo shortcut: js-yaml is what the workspace already hoists. Production
+// pins the `yaml` package (YAML 1.2) so bare `off`/`no` stay strings.
+const yaml = createRequire(import.meta.url)('js-yaml') as { load: (text: string) => unknown };
```
tscbmstubp: js-yaml resolves YAML 1.1 booleans by default, so a value like `default: no` or a variant named `on` becomes a boolean without an error. Pin the `yaml` package (YAML 1.2) when this leaves demo status; it removes a whole class of silent mistakes in overlays that no test would catch.

packages/ui-kit/src/components/button/button.contract.json
tscbmstubp, line 1: Is it committed just for the demo?

itechmeat: Yes, we will not merge this PR, just for demo. But let's review as usual :)

packages/ui-kit/src/components/button/button.contract.json
```diff
@@ -0,0 +1,124 @@
+{
+  "$id": "gts://gts.frontx.uikit.component.button.v1~",
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "title": "UiKit button contract",
+  "type": "object",
+  "properties": {
```
tscbmstubp, line 14: Right now, the prop set is open by accident. className, loading, and focusableWhenDisabled are listed as properties, the Base UI inheritance shows up as passthrough strings, and the schema declares neither additionalProperties nor unevaluatedProperties. So Ajv passes `variannt="ghost"` without a word.

Open is the right default. A Button that rejects aria-label is useless. But it has to be a decision, not an omission. What I would put in the metamodel:

- kit-level props are normative and closed
- DOM and Base UI passthrough is allowed through one shared passthrough schema per element kind
- the lint reports anything outside both as "unchecked", never as passing

Concretely, the compiled schema would look like this:

```
{
  "$id": "gts://gts.frontx.uikit.component.v1~frontx.uikit.button.v1~",
  "allOf": [
    { "$ref": "gts://gts.frontx.uikit.component.v1~" },
    { "$ref": "gts://gts.frontx.uikit.passthrough.dom_button.v1~" }
  ],
  "properties": {
    "variant": { "type": "string", "enum": ["default","secondary","destructive","outline","ghost","link"] },
    "size":    { "type": "string", "enum": ["default","sm","lg"] },
    "loading": { "type": "boolean" }
  },
  "unevaluatedProperties": false
}
```

```
{
  "$id": "gts://gts.frontx.uikit.passthrough.dom_button.v1~",
  "properties": { "className": {"type":"string"}, "disabled": {"type":"boolean"}, "onClick": {}, "id": {"type":"string"} },
  "patternProperties": { "^aria-": {}, "^data-": {} }
}
```

It has to be unevaluatedProperties, not additionalProperties. additionalProperties only sees the properties written in the same object and cannot look through $ref or allOf, so it would reject every passthrough prop. unevaluatedProperties (2019-09+) treats a property as evaluated if any subschema claimed it and rejects only what nothing did.

packages/ui-kit/src/components/button/button.contract.json
tscbmstubp: x-uikit is a private keyword carrying the hand-written half of the contract: intent, use_when, composition, invariants, deprecations, coverage, examples. JSON Schema treats it as an annotation, and the demo validates it against a separate metamodel type through a second file, button.contract.instance.json.

That is what x-gts-traits already does. Type-level metadata, declared once by x-gts-traits-schema on an ancestor, validated when the derived type is checked. So the fix for your first comment fixes this one too:

- the metamodel (ui-component.meta.json) moves into x-gts-traits-schema on the abstract base component type
- each component carries its overlay in x-gts-traits instead of x-uikit
- button.contract.instance.json goes away, the overlay has one compiled form

I checked gts-ts 0.3.0 as installed. Trait validation (OP#13) is there: GtsStore.validateSchemaTraits walks the allOf chain, merges trait values, applies defaults, validates with Ajv, checks required trait properties and const locks. GTS.validateEntity runs it for schemas. No upgrade needed for the mechanism. What is missing is a caller. gts-plugin's registerSchema() only registers and never validates against parent or traits, so a broken trait block would register silently today. The conformance test can call validateEntity on every compiled schema at build time without touching the plugin. Whether registerSchema should also validate at runtime is your call under ADR 0005.

One question I cannot answer myself. The spec describes traits as system behavior for processing instances, with retention rules and processing directives as examples. Most of our overlay fits that: the plan validator and the lint read composition, deprecations and coverage to accept or reject a plan node. Examples and the prose in invariants do not fit. Nothing processes them, they are documentation.

Will a GTS runtime ever read traits and act on them? If traits stay validate-and-store, the whole overlay can live there. If a runtime will act on them one day, we should keep the documentation fields in our own keyword and use traits only for what a validator reads.

packages/ui-kit/src/components/button/button.contract.json
itechmeat: Done in 25c0f869: the kit now ships an abstract base component type and Button's props schema is a derived type - chained id, allOf over the base plus a shared DOM passthrough schema, unevaluatedProperties: false.

The base id landed as `gts.frontx.uikit.base.component.v1~` rather than the shorter `gts.frontx.uikit.component.v1~`: gts-ts strips the fixed `gts.` scheme prefix and then wants 5 or 6 dot-tokens per segment (vendor.package.namespace.type.vMAJOR[.minor], `Gts.parseSegment`), and `frontx.uikit.component.v1` is four, so it fails as "Too few tokens". Your own `gts.frontx.design.uikit.component.v1~` does parse - frontx / design / uikit / component / v1 is exactly five. Kept the demo's existing frontx.uikit namespace rather than frontx.design.uikit - happy to rename if design belongs in the segment.

packages/ui-kit/src/components/button/button.contract.json
itechmeat: Implemented in 25c0f869: kit-level props are closed via unevaluatedProperties: false, DOM/Base UI passthrough goes through one shared schema per element kind (aria-*/data-* via patternProperties), and the conformance test now proves variannt fails while aria-label and data-* pass.

packages/ui-kit/scripts/contracts/compile.ts
itechmeat: Agreed - switched the demo to the yaml package (YAML 1.2) in 84f975b4.

packages/ui-kit/src/components/button/button.contract.json
itechmeat: x-gts-traits does look like the right home - thanks for checking gts-ts 0.3.0. Two ways we could go: (a) the whole overlay moves into traits if they stay validate-and-store; (b) hybrid - fields the plan validator and lint read (composition, deprecations, coverage, dont_use_when) become traits, while documentation-only fields (examples, invariant prose) keep a separate keyword. We lean (b) since it stays valid under either answer to the runtime question, but other splits are welcome. Meanwhile the conformance test now calls GTS.validateEntity at build time (25c0f869), closing the silent-registration gap. @GeraBart the ADR 0005 calls are yours: will a runtime ever act on traits, and should registerSchema validate on registration?

tscbmstubp:
Now that props come from the code, I want to ask where the truth should live. It decides what the conformance test can catch.

Today, the extractor copies props and cva axes from button.tsx into the compiled schema. So the contract matches the code by construction. Remove a variant from cva() and the contract follows without a sound. Consumers get a breaking change, and nothing in CI notices. What we call "contract matches code" is really "the compiled file is fresh".

Two options.

A. Code is the truth (what we have). The extractor writes props into the schema, the YAML holds meaning only. To catch breaking changes, we diff the compiled schema against its committed previous version with gts-ts checkCompatibility and demand a major bump on an incompatible verdict. It works. It also means the compiled JSON has to stay in git, so every contract change is one YAML diff plus JSON diffs of the same fact.

B. YAML is the truth, TypeScript is generated from it. The YAML declares props and axes next to the meaning. The build emits button.contract.d.ts with the variant and size unions and the props interface, and the component imports that instead of writing its own. The cva() literal stays in code, it has to, it maps values to CSS classes. But it is typed against the generated union, so a missing or extra key fails tsc. The extractor keeps its job in reverse: it checks that cva() and the component agree with the YAML instead of copying from them. Compiled JSON becomes plain build output and moves to dist. The compatibility check runs on the YAML between base and PR. We need a small vocabulary for what JSON Schema cannot type: a slot for ReactNode, a handler for functions. The extractor's generate mode writes the first props section for all 63 components once, then switches to verify. This is the end state the proposal already names.

I prefer B. One authored file per component. Removing a prop is a YAML diff someone has to mean. Compiled JSON leaves git. And the compiler's "one fact, one owner" check flips direction. Today it throws when the YAML states a prop, because the code owns props. Under B, the YAML owns them, so the error becomes the opposite: the code declares a prop that the YAML does not.

tscbmstubp:
On ui-component.meta.json, the use_when field

A concern from today's meeting that I think is right: component prose cannot list use cases exhaustively, and trying makes it either incomplete or product-specific. use_when with minItems: 1 and the description "positive selection scenarios" reads as if it should be complete.

Proposal: shrink what the component level claims to what is intrinsic to the component.

- intent: one sentence, as now
- typical_uses (rename of use_when): one to three archetypes, no pretence of coverage. A selection-card anchor, not a list
- dont_use_when: keep, this is the valuable part. The confusable neighbours of a component (Button vs Link vs Switch vs Toggle) are a small stable set, and the typed instead makes each entry checkable
- invariants and anti_patterns: keep, they are about this component's own API

Everything that reads like a scenario moves up a level. Use cases are what blocks are: a login form or a list toolbar is the situation, and the block family card is where "when to use" has real breadth. Cross-component rules (one primary action per region, same role uses the same component and variant everywhere) belong in the guardrails template, next to the composition-patterns guideline and a role -> component+variant table that the lint can check. Product conventions stay in the project.

Reading order for an agent is then top-down: template rules, block family card, component contract last to get the props right. The component contract gets smaller, which is what we want for tokens and for staying true.

tscbmstubp:
General comment on the term "block"

Since the question came up: why "block" and not "higher-order component"?

They differ in delivery, not size. A higher-order component is still a component: it lives in the npm package, it is imported, and fixes arrive through a version bump. A block is copied into the project and then belongs to the project. DESIGN treats that as the line between a library (consumed, never copied) and a template (copied, owned by the receiver). Blocks are on the template side, so calling them components would put a copied artifact under the word this repo uses for a dependency.

What follows from that:

- blocks are meant to be edited. A login form gets product fields, validation, copy. As a component, it would sprout configuration props for every case or get forked anyway
- blocks carry use-case states (loading, empty, error, permission-denied), not element states
- blocks go through the template mechanism: manifest, provenance, upgrade, dependency report, one CLI transaction. Components need none of it
- shadcn already uses the word this way (registry:ui, registry:component, registry:block), and Shadcn Studio's catalog is block families with numbered variations. The same word keeps the registry projection readable

The line is not always clean. Field wires label, input, description, and error, and it is a component because no product would want to fork it. Rule of thumb: if a fork would be embarrassing, it is a component; if a fork is the expected use, it is a block.

itechmeat:
Closing: superseded by #616, the mergeable follow-up built after this review. Still worth picking up from here: the open ADR 0005 question to @GeraBart on whether a runtime will ever act on traits and whether registerSchema should validate on registration, plus the review threads from @GeraBart and @tscbmstubp on the base component type and closed prop sets.

PR #614, opened 2026-09-03
feat(cli): implement template registration against one project state file (ADR 0016-0033, 0042)

gs-layer:
Implements the template-registration redesign whose artifacts were accepted in #564 — ADRs 0016–0033 and 0042, the CLI package's PRD/DESIGN, and its six FEATUREs — turning them into a working CLI. 52 commits, 193 files, mostly `packages/cli` (135) and `packages/cyber-pilot-kit-frontx` (28).

## What lands

The whole registration lifecycle against one project state file: `install`, `register`/`unregister`, `assemble`, `apply`, `seed`, `upgrade` (with `--restore`), `delete`, `ownership add|remove|list`, `list`, and `validate --project`. Every command answers in the uniform `{ok, data}` / `{ok, error:{code, message, details}}` envelope ADR-0042 fixes, drawing `error.code` from its one closed vocabulary, and exits non-zero on every `ok: false`.

Structural decisions carried out rather than approximated:

- **One shared resolver** pins every remote origin to the immutable value its fetch settled on, records a `path:` origin exactly as given, and fabricates no pin for a local folder that has no publication to pin against (ADR-0017).
- **Whole-file upgrades** (`ADD`/`REPLACE`/`REMOVE`/`KEEP_LOCAL`/`UNCHANGED`) with one-generation reversal, staged writes, and a doubly-changed file refused whole rather than merged (ADR-0021).
- **Six-term effective ownership** computed once from the candidate manifest, shared verbatim by `apply`, `delete`, `ownership` and the changeset engine — never a second, independently derived boundary.
- **A trust gate on AI extensions**, so an extension bundle is discovered and activated only for an identity the project actually registered.

## Traceability

`cfs validate` reports **0 errors, 0 warnings** at 192/192 instruction coverage, with all six CLI-package FEATUREs closed. The gate went from 438 errors at the start of this branch to zero.

Worth saying plainly for review: a marker proves a citation exists, never that the code does what the cited text says. Several defects here were found by reading code against the spec that owns the concept, not by hunting markers — among them `apply` copying a template's own `frontx-template.json` into every target (the payload definition excludes it), `list --json` reporting a git ref under `version` where two other sets in the same envelope reported a manifest version, and a resolver that fabricated an address for any unknown host and would have stored the typed, moving ref as an origin.

## Verification

- `cfs validate` 0/0, coverage 192/192
- `npm test` exit 0 across nine projects (CLI: 756 tests, 48 files)
- `tsc` clean, `npm run lint` exit 0, `npm run arch:check` 12/12, `cfs check-language` clean
- Live runs of the built binary against fresh projects outside this repository, driving the CLI rather than the filesystem: `register → apply → validate → upgrade → restore → delete`, plus the refusal paths

## Known, and not introduced here

- `validate:templates` fails on `template-mfe`'s nine project-relative `overrides` — red on `develop` too, untouched by this branch, and proven to break the npm graph in a real project.
- `seed` does not work outside this repository: the ecosystem's templates are not published anywhere yet, so the CLI's built-in official defaults are local `path:` origins resolvable only against this checkout, and `defaults` is empty elsewhere. The templates are to be published to a repository of their own as separate future work; the CLI DESIGN now records this as an architectural fact rather than an open question.
- Deterministic validation is green; the semantic half `cfs validate` asks for at the end has not been run for this work.

## Relationship to #564

That PR's documentation is contained here file-for-file, and everything in its ADR set except `0042-cli-machine-readable-output.md` is already on `develop`. Once this merges, #564 has nothing left to add.

Draft: opened for review of the merge with `develop` and the scope above before it is marked ready.


gs-layer:
### Code review

Found 9 issues:

1. `seed` writes `.frontx/project.json` and registers the earlier batch entries before the whole batch is validated, then reports "whole batch aborted, nothing written"; a retry is refused with `INVALID_INPUT` ("already carries a project state document"), so the directory is locked out of `seed` with no CLI recovery (bug: cli-scaffolding FEATURE's seed error scenario says the state file is left uncreated; reproduced against the built CLI)

https://github.com/constructorfabric/gears-frontx/blob/6699d0fec251cad4c7b66dff18b087e285a86ec6/packages/cli/src/commands/seed-repository.ts#L80-L82 and https://github.com/constructorfabric/gears-frontx/blob/6699d0fec251cad4c7b66dff18b087e285a86ec6/packages/cli/src/commands/seed-repository.ts#L93-L100

2. `seed <dir>` on a directory that does not exist yet exits 2 with a bare stderr line and an empty stdout under `--json`: no ADR-0042 envelope, no `error.code` (bug: `createFsCanonicalizeTargetFn` throws before the command runs)

https://github.com/constructorfabric/gears-frontx/blob/6699d0fec251cad4c7b66dff18b087e285a86ec6/packages/cli/src/cli.ts#L1009-L1011 and https://github.com/constructorfabric/gears-frontx/blob/6699d0fec251cad4c7b66dff18b087e285a86ec6/packages/cli/src/adapters/fs-project-io.ts#L335-L339

3. The kit's AI-upgrade adapter still speaks the protocol this PR removed from `frontx upgrade --json`: it waits for a `{changeSet}` stdout line, writes the decision to stdin and expects a final `{ok, status}`; the CLI now answers `CONFIRMATION_REQUIRED` immediately when `--yes` is absent and never emits `status`, so every real invocation ends in "exited without a parseable JSON result" and the review gate never fires (bug: only the argument shape was migrated; tests use a fake spawn so nothing catches it)

https://github.com/constructorfabric/gears-frontx/blob/6699d0fec251cad4c7b66dff18b087e285a86ec6/packages/cyber-pilot-kit-frontx/src/upgrade-orchestration/invoke-upgrade-command.ts#L59-L69 and https://github.com/constructorfabric/gears-frontx/blob/6699d0fec251cad4c7b66dff18b087e285a86ec6/packages/cyber-pilot-kit-frontx/src/upgrade-orchestration/invoke-upgrade-command.ts#L91-L94

4. Every `render*Outcome` maps any `ok:false` to exit 1, so a command's own `INTERNAL` code (produced in `upgrade/commit.ts` and `commands/apply.ts`) exits as a user error and `EXIT_INTERNAL_ERROR` is reachable only from thrown exceptions; cli-invocation FEATURE says "a distinct process exit code for each outcome class — success, user error, and internal error — applied consistently across every dispatched command" (regression of 620fc12b)

https://github.com/constructorfabric/gears-frontx/blob/6699d0fec251cad4c7b66dff18b087e285a86ec6/packages/cli/src/cli.ts#L709-L714 and https://github.com/constructorfabric/gears-frontx/blob/6699d0fec251cad4c7b66dff18b087e285a86ec6/packages/cli/architecture/features/cli-invocation/FEATURE.md#L196-L198

5. `apply` writes through a pre-existing symlink to a path outside the target: `readExistingContentFn` skips symlinks, so a link standing where the payload writes reads as free ground and `writeFileSync` follows it (regression of ebbafe46's fail-closed probe, which is now wired only to `ownership add`; reproduced without `--adopt-existing`: `README.md -> ../outside/escaped.txt` received the template content and `apply` returned `ok:true`)

https://github.com/constructorfabric/gears-frontx/blob/6699d0fec251cad4c7b66dff18b087e285a86ec6/packages/cli/src/adapters/fs-existing-content.ts#L62-L70 and https://github.com/constructorfabric/gears-frontx/blob/6699d0fec251cad4c7b66dff18b087e285a86ec6/packages/cli/src/adapters/fs-project-io.ts#L37-L41

6. An unrecognized command in `--json` mode prints the human usage text, not the `{"ok":false,"error":{"code":"INVALID_INPUT"}}` envelope the FEATURE specifies, although the acceptance criterion for it is ticked `[x]` (spec/code mismatch inside this PR's own FEATURE)

https://github.com/constructorfabric/gears-frontx/blob/6699d0fec251cad4c7b66dff18b087e285a86ec6/packages/cli/src/cli.ts#L1566-L1569 and https://github.com/constructorfabric/gears-frontx/blob/6699d0fec251cad4c7b66dff18b087e285a86ec6/packages/cli/architecture/features/cli-invocation/FEATURE.md#L230-L232

7. The AI entry point this PR wires (`.claude/skills/cf-cyber-pilot-kit-frontx-frontx-project-scaffolding/SKILL.md` loads the kit's `project-scaffolding/SKILL.md`) drives a command surface that no longer exists: it parses `list --json` as `{ok, templates:[{ref}]}` while the CLI emits `{ok, data:{defaults,registered,installed}}` with `origin`/`version`, runs `frontx seed <identity> <dir>` and `frontx add`, and reads `.frontx/provenance.json`, which ADR 0019 says the CLI must refuse (AI-agent docs made wrong by this PR; the same stale forms sit in the kit's top-level `SKILL.md`)

https://github.com/constructorfabric/gears-frontx/blob/6699d0fec251cad4c7b66dff18b087e285a86ec6/packages/cyber-pilot-kit-frontx/skills/project-scaffolding/SKILL.md#L40-L44 and https://github.com/constructorfabric/gears-frontx/blob/6699d0fec251cad4c7b66dff18b087e285a86ec6/packages/cyber-pilot-kit-frontx/skills/project-scaffolding/SKILL.md#L177-L185

8. `README.md` and `QUICK_START.md` (which README names as the full command walkthrough) still document `seed <ref> <dir>`, `add` and `upgrade <root> <version>`; `add` is "Unrecognized command" at HEAD, and none of `register`/`apply`/`assemble`/`ownership`/`delete`/`validate --project` appear in either file (human docs made wrong by this PR's surface change; the three templates' READMEs carry the same retired commands and `ownershipBoundaries`)

https://github.com/constructorfabric/gears-frontx/blob/6699d0fec251cad4c7b66dff18b087e285a86ec6/README.md#L142-L152 and https://github.com/constructorfabric/gears-frontx/blob/6699d0fec251cad4c7b66dff18b087e285a86ec6/QUICK_START.md#L58-L62

9. Two module comments describe the pre-cutover state and contradict the code: `envelope.ts` says "no command is wired to this yet" while every dispatch case renders through `ok`/`err`; `upgrade/validate.ts` says the resolver still returns `INVALID_MANIFEST` for a legacy manifest, which `payload.ts`'s own header says was removed as incoherent (same pattern in `index.ts#L58-L63` "NOT YET wired" and `cli.ts#L526-L533` "ad hoc {ok,status} line")

https://github.com/constructorfabric/gears-frontx/blob/6699d0fec251cad4c7b66dff18b087e285a86ec6/packages/cli/src/envelope.ts#L7-L11 and https://github.com/constructorfabric/gears-frontx/blob/6699d0fec251cad4c7b66dff18b087e285a86ec6/packages/cli/src/upgrade/validate.ts#L193-L197

### Assessment against the review brief

**Three templates, CLI only (live run on fresh folders outside the repo, origins `github:gs-layer/gears-frontx//<template>@feat/cli-registration-impl`).** template-shell: the full cycle is green — register (branch ref pinned to a commit), assemble, apply (378 files, no `frontx-template.json`, no `node_modules`), `validate --project` PASS, `npm install` + `npm run build` + type-check, upgrade to a modified copy (ADD/REPLACE/REMOVE/KEEP_LOCAL verified on disk, doubly-changed file refused whole), `--restore` three times, `delete --dry-run`/`--yes`, unregister; all refusal paths return an envelope with a code and exit 1. template-mfe: `.` conflicts with the shell, and the only accepted target `src-app/mfe_packages` yields `src-app/mfe_packages/src-app/mfe_packages/<mfe>/...` plus the monorepo dev-harness `package.json`; apply, `validate --project` and `npm run build` all exit 0 while `dist/generated-mfe-manifests.json` is `[]` — zero microfrontends. template-design-guardrails: every candidate target is `TARGET_CONFLICT` against the shell owning `.`, and `ownership add` cannot carve out a path that does not exist yet; applied alone it has no root `package.json`. The PR body's "nine project-relative overrides break the npm graph" does not describe what happens: npm succeeds, the payloads are rooted at the project root while the new model roots them at the target, and the four-field cutover dropped the old `exclusiveSubtrees` without re-rooting (https://github.com/constructorfabric/gears-frontx/blob/6699d0fec251cad4c7b66dff18b087e285a86ec6/template-mfe/frontx-template.json#L5, https://github.com/constructorfabric/gears-frontx/blob/6699d0fec251cad4c7b66dff18b087e285a86ec6/template-design-guardrails/frontx-template.json#L5, https://github.com/constructorfabric/gears-frontx/blob/6699d0fec251cad4c7b66dff18b087e285a86ec6/template-shell/frontx-template.json#L5-L7). So one template of three composes through the CLI today.

**Lifecycle coverage.** Every command in the PR body exists in `help` and dispatches. `update-local` is in `help` and the PRD but absent from DESIGN §3.2 and from the cli-invocation dispatch table whose ticked AC asserts a match (https://github.com/constructorfabric/gears-frontx/blob/6699d0fec251cad4c7b66dff18b087e285a86ec6/packages/cli/architecture/DESIGN.md#L260). `seed` is unusable outside the monorepo, stated only in a DESIGN paragraph and an aside of the live-verification README, not in `help` or the PRD FR. ADR 0019's refusal of a legacy `provenance.json` next to a missing `project.json` is not implemented; `loadProjectStateDocument` treats absence as a fresh project (https://github.com/constructorfabric/gears-frontx/blob/6699d0fec251cad4c7b66dff18b087e285a86ec6/packages/cli/src/project-state/io.ts#L109-L124).

**Human, AI, and both.** For a human the only complete reference is `help`; flags, exit codes, the 16 error codes and the `project.json` shape live in ADRs/FEATUREs written for traceability, and the two user-facing docs are stale (issue 8). For an agent the wired skill is stale (issue 7) and its Step 7 tells the agent to read `.frontx/ai` bundles from disk and "not wait for activation", which is the path the new trust gate exists to block (https://github.com/constructorfabric/gears-frontx/blob/6699d0fec251cad4c7b66dff18b087e285a86ec6/packages/cyber-pilot-kit-frontx/skills/project-scaffolding/SKILL.md#L217-L227). The kit DESIGN's division of labour (agent composes one batch, CLI applies it) is right; the shipped instructions do the opposite. `assemble`, `unregister`, `update-local` and `ownership` are run by no document or skill.

**Upgrade engine, below the comment threshold.** The baseline side is no longer identity-checked, only version-checked (2287716f compared both; https://github.com/constructorfabric/gears-frontx/blob/6699d0fec251cad4c7b66dff18b087e285a86ec6/packages/cli/src/upgrade/validate.ts#L150-L176). The commit point writes `set-template` as an unconditional upsert without re-reading the record the plan was computed from (82a95432 aborted on that; https://github.com/constructorfabric/gears-frontx/blob/6699d0fec251cad4c7b66dff18b087e285a86ec6/packages/cli/src/upgrade/commit.ts#L391-L402). `delete` removes a project-local origin folder recorded only in `previous`, leaving `--restore` pointing at a payload that is gone (https://github.com/constructorfabric/gears-frontx/blob/6699d0fec251cad4c7b66dff18b087e285a86ec6/packages/cli/src/scaffold/delete-plan.ts#L116-L121). Latent: `overrides` joined `DEPENDENCY_FIELDS` but the key-order-insensitive comparison sorts one level only.

**Design.** What holds: the module graph is acyclic, adapters import only types from the domain, one resolver serves every acquisition, the error-code vocabulary is one union with no stray literals, `computeExclusionRoots`/`isWithinEffectiveOwnership` are shared by assembler, apply, delete-plan, existing-content, classify and commit, and `cfs validate` is 0/0 at 192/192. What does not: `joinUnderTarget` is declared "the one place" and exists as eight private copies (https://github.com/constructorfabric/gears-frontx/blob/6699d0fec251cad4c7b66dff18b087e285a86ec6/packages/cli/src/scaffold/effective-ownership.ts#L33-L41); the local-origin-folder derivation is repeated about eight times with `LOCAL_ORIGIN_PREFIX` declared three times; `conflict-check.ts#buildReservedGround` re-lists the reserved terms instead of decorating `computeExclusionRoots` (https://github.com/constructorfabric/gears-frontx/blob/6699d0fec251cad4c7b66dff18b087e285a86ec6/packages/cli/src/scaffold/conflict-check.ts#L358-L375); `runCommand` is about 790 lines and the `upgrade` case carries the inventory-promotion and AI-bundle-refresh logic inline; eleven renderers repeat one failure branch; `manifest/legacy-ownership.ts` (160 lines plus a 176-line test) and `upgrade/state.ts` have no production caller yet sit on the public barrel; core modules are 55–63% comment, part of it checkpoint narration that is already stale (issue 9).

Checked and clean: no Claude/Anthropic attribution in commits or diff; no new `@cpt-` markers in template territory; `cfs validate` and `cfs check-language` exit 0; prior review feedback on #564, #548, #545, #541, #533, #498 is addressed, not reintroduced.

PR #615, opened 2026-09-04
feat: move templates to constructorfabric/gears-frontx-templates

tscbmstubp:
## Summary

The FrontX templates (`template-shell/`, `template-mfe/`, `template-design-guardrails/`) move out of this monorepo into their own repository, https://github.com/constructorfabric/gears-frontx-templates. Consumers address them as `github:constructorfabric/gears-frontx-templates//template-shell@<ref>`. This is the physical realization of ADR-0016 (externalized templates) and ADR-0017 (one repository per template family, `//subtree` addressing). The CLI product code needed no change: it already ships zero template content.

Companion PR (receiving side): constructorfabric/gears-frontx-templates#1.

## What changes here

- Removed the three template directories and the template-only guards that read them: `validate:templates`, `policy:template-pin-drift`, `policy:template-lockfile-selflink`, `policy:token-format`, `policy:guideline-index`, `dev:template:link`, `pin-template-ecosystem-to-local.mjs`, `scripts/manual-auth-tests/` (all now live in the templates repo).
- Kept the `packages/*` half of pin drift as `policy:ecosystem-pin-drift` (`scripts/ecosystem-pin-drift-check.mjs`, `scripts/ecosystem-packages.mjs`); `policy:version-bump-on-change` now governs `packages/*` only.
- CI: `main.yml` loses the `template-validate` job and template steps; `template-drift.yml` deleted (re-homed in the templates repo, which checks this repo out as a sibling via `FRONTX_ECOSYSTEM_DIR`); `publish-packages.yml` PART 2 (temporary shell-template publish) removed as its own comment asked.
- `packages/api/vitest.config.ts` no longer imports the shell template's vitest config.
- CLI end-to-end tests run against synthetic fixture templates under `packages/cli/src/__tests__/fixtures/` instead of the real trees; install/seed/add/assembly, ownership-conflict, union-fidelity and declaration-coverage assertions are preserved.
- Dead `packages/cli/templates` / `template-sources` exclusions removed from `.gitignore`, `knip.json`, `.sonarcloud.properties`, `.dependency-cruiser.cjs`, `eslint.config.js`, `.cf-studio/config/artifacts.toml`.
- Architecture: new ADR-0034 `cpt-frontx-adr-template-repository-separation`; ADR-0033 and ADR-0032 reconciled; `ai-project-scaffolding` FEATURE reworded to what the kit can verify.
- Docs: CONTRIBUTING, README, QUICK_START, AGENTS.md, CLAUDE.md point at the templates repo.
- Version bump required by the bump gate: `@gears-frontx/cli` 0.3.0-alpha.6 (develop had already taken alpha.5). `packages/api` gained a `vitest.setup.ts` (real-timers restore, timeouts) replacing what the shell template's shared config used to provide; nothing under its `src/` changed, so no api bump.

## Verification

`build:packages`, `type-check:all`, `lint`, `arch:check`, `arch:edges`, `arch:guards`, all workspace unit suites, `policy:*`, `GITHUB_BASE_REF=develop policy:version-bump-on-change`, `cfs validate` (0 errors), `cfs check-language`: all green in a tree with no template directory.

## Release sequencing

Ecosystem packages publish first; the templates repo re-pins and publishes after. The templates repo's drift workflow checks out this repo's `develop`.


gs-layer:
## Review notes

Reviewed alongside the companion PR (gears-frontx-templates#1). Two things worth fixing before merge, plus one smaller item.

### Worth fixing before merge

**`packages/api/vitest.config.ts`**
The old config (via the now-removed `template-shell/vitest.shared.ts`) set `coverage.thresholds` (70% lines/functions/statements, 60% branches), `coverage.exclude`, a top-level `exclude`, and 4 `include` globs. The rewritten self-contained config keeps only `setupFiles`/`testTimeout`/`hookTimeout`/`passWithNoTests` — the coverage thresholds and most of the include/exclude globs are gone, and the comment at the top doesn't mention this was dropped. Nothing else in the repo enforces coverage for this package now, so coverage can regress silently. Also, `.spec.ts` files and any colocated test outside `__tests__/` won't run anymore under the new `include`.

**`.github/CODEOWNERS`**
This PR removes the `/template-shell/` and `/template-mfe/` entries (makes sense, those dirs are gone), but the new `gears-frontx-templates` repo doesn't have a `CODEOWNERS` file to replace them — worth adding one there before this merges, so the review requirement isn't just dropped.

### Smaller item

**`scripts/mfes-import-boundary-check.mjs`**
This guard's whole reason for existing (per its own docblock) was catching template-side code that instantiates concrete mfes classes instead of going through the facade — and it used to cover `template-shell` since that directory lived in this repo. Now that `template-shell` moved out, this script can't see it anymore, and the new repo's `eslint.config.js`/`.dependency-cruiser.cjs` don't have an equivalent check. `architecture/ADR/0003-mfe-runtime-public-surface.md` (untouched by this PR) still says a CI check "confirms both halves of the boundary" — that's no longer accurate. Worth a follow-up in the templates repo, or at least a note in the ADR.

One more thing, lower confidence: the new `packages/cli/src/__tests__/fixtures/fixture-shell/` and `fixture-overlay/` each add a `frontx-template.json` file, which is literally what your own `CLAUDE.md`/`AGENTS.md` (line 11) defines as "template territory" that's not supposed to remain in this repo. `.cf-studio/config/AGENTS.md` has a narrower "top-level directory" qualifier that would exempt these, and the FEATURE.md update in this same PR suggests this is intentional — just flagging the wording inconsistency between the two docs in case you want to tighten it.

Nice catch on switching the CLI e2e tests to synthetic fixtures instead of the real template trees — checked the old vs. new test assertions and the coverage (ownership-conflict, union-fidelity, declaration-coverage) is preserved.

tscbmstubp:
Thanks for the review. All four points addressed in the current head:

- **`packages/api/vitest.config.ts`**: restored the coverage thresholds (70/70/70 lines/functions/statements, 60 branches), `coverage.exclude`, the top-level `exclude`, and the four include globs, so `.spec.ts` and colocated tests run again. Verified with a coverage run (well over the floor) and a negative control with an unreachable threshold that fails as expected. The header comment now says what the config provides.
- **CODEOWNERS**: the templates repository gains a `CODEOWNERS` with the same owners for the template directories in the companion PR, so the requirement moves rather than disappears.
- **`scripts/mfes-import-boundary-check.mjs`**: the docblock now describes what it still checks here (the barrel half, plus the consumer half over `packages/*` and `internal/*`) and points at the templates repository for the template-side consumer half, which is ported there as its own CI step in the companion PR. ADR-0003's Confirmation now says each half is confirmed over the repository that holds it, and ADR-0034 records the consequence.
- **Fixture manifests vs. "template territory"**: CLAUDE.md and AGENTS.md now define template territory as a top-level directory carrying `frontx-template.json`, matching `.cf-studio/config/AGENTS.md`, and name the CLI fixtures under `packages/*/src/__tests__/` as ecosystem test data inside the artifact chain.

architecture/ADR/0034-template-repository-separation.md
```diff
@@ -0,0 +1,146 @@
...
+* Bad, because a guard whose consumer half walked one tree now has to exist in two, so tightening the rule on one side does not tighten it on the other, and the two copies can drift.
+
+### Confirmation
+
+Compliance is confirmed by the repository state and by continuous integration. No top-level directory in this repository carries a `frontx-template.json` manifest, and template discovery over this repository returns an empty set. No template-only guard, job or publish step remains in this repository's continuous integration. The CLI's end-to-end tests resolve only synthetic fixture templates and no test reads a path that would have been a template directory. The CLI boundary check for `cpt-frontx-constraint-cli-template-independence` continues to pass unchanged, since the CLI bundled no template before this change either. On the other side, the templates repository runs the drift check against an ecosystem checkout and fails when a template's pins no longer match the packages it compares to. Review confirms that a change spanning both repositories publishes the ecosystem packages before the templates repository re-pins to them.
```
itechmeat: [Not a blocker] 🟠 Ordering note rather than a code issue: today the templates repository's main is LICENSE + NOTICE, it has no tags and zero workflow runs (gears-frontx-templates#1 is a fork PR, so its new workflows cannot run before merge). This Confirmation, ADR-0003:63, CONTRIBUTING.md:156, packages/api/CLAUDE.md:9 and the `frontx install ...//template-shell@<ref>` examples in README, QUICK_START and CONTRIBUTING all describe that repository in the present tense, and none of the install commands can be run until a tag exists. Could #1 land and a tag be cut first, so every sentence here is true the moment this merges? Then `<ref>` in the examples can become a real tag or be named as a placeholder explicitly.

architecture/ADR/0034-template-repository-separation.md
```diff
@@ -0,0 +1,146 @@
...
+* **One repository per template** — `template-shell`, `template-mfe` and `template-design-guardrails` each get a repository of their own.
+
+## Decision Outcome
+
+Chosen option: **A dedicated templates repository holding the whole family**, because it is the only option that makes the layout agree with the layer partition while keeping the three templates in one place to review and release together.
+
+The templates leave this repository and live in **`constructorfabric/gears-frontx-templates`**, one top-level directory per template. A consumer addresses one by the subtree form the source-spec already defines, for example `github:constructorfabric/gears-frontx-templates//template-shell@<ref>`. Nothing in the resolution mechanism changes: this decision is the physical realization of `cpt-frontx-adr-template-acquisition-and-location` and of the subtree addressing in `cpt-frontx-adr-source-spec-syntax`, not an amendment to either. The CLI's zero-template-content constraint (`cpt-frontx-constraint-cli-template-independence`) is unchanged and still enforced here.
```
itechmeat: [Not a blocker] 🟠 Question on placement rather than content: this paragraph says the decision amends neither 0016 nor 0017 and changes nothing in the mechanism - so what does a new ADR hold that 0016 "Template Acquisition and Location" (its problem statement is literally "where should installed templates live") cannot? The house convention rewrites a changed decision in place (see 569029ad, which folded the old 0033 back into 0017 for exactly this), and 0017 already weighs "one repository per template". The option comparison and the three consequences here are genuinely new and worth keeping; would they sit in 0016's outcome and consequences instead, with 0033 taking the ownership sentence? Up to the owner, but cheaper to decide before four documents cite this file.

architecture/ADR/0033-template-territory-traceability.md
```diff
@@ -107,7 +107,7 @@ Author a separate nested artifact tree owned by the template and re-point the pa
...
+**Deferred deliberately, and since resolved in part**: whether template territory gets an artifact tree of its own was **not decided here**. It was deferred until template ownership had settled. `cpt-frontx-adr-template-repository-separation` settled the ownership half: template territory now lives in its own repository, `constructorfabric/gears-frontx-templates`. Whether that repository authors an artifact tree for its templates is its own decision to make under its own governance, and is no longer a question this repository holds open. This decision continues to state what it always stated — that the ecosystem artifact tree does not specify template internals, and that `@cpt-` markers found in template payload bind nothing.
```
itechmeat: [Not a blocker] 🟠 This paragraph says the deferred question is now settled, while :55 still says "that is deliberately deferred (see More Information)", :38 still cites "ownership is still moving (the template split)" as a driver, and :62 still leaves the artifact tree "available as a later, deliberate step". :53 and the Confirmation at :68 also keep review rules about touching template files that cannot fire here any more. Could 0033 be rewritten as one current-state decision rather than three patched passages?

architecture/ADR/0003-mfe-runtime-public-surface.md
```diff
@@ -58,7 +58,9 @@ Chosen option: **abstract registry facade with factory injection**, because it i
...
+
+The two halves are no longer confirmed in one place. The **barrel half** is whole here and answers for every consumer anywhere, because a name the barrel does not export cannot be imported from any repository. The **consumer half** reaches only as far as the tree it walks. Template payload used to sit in that tree and no longer does (`cpt-frontx-adr-template-repository-separation`), so the consumer half over template code is confirmed by the same rule ported into the templates repository's own continuous integration, running over its tree. Neither check sees the other's consumers, and a consumer in a repository that runs neither is unchecked — which is why the barrel half, the structural one, is the half that must stay complete.
```
itechmeat: [Not a blocker] 🟡 "no longer confirmed", "used to sit", "and no longer does" - artifacts here describe the current state only, git keeps the past. Same shape in 0033:30/:68/:110, 0034:30 and CONTRIBUTING.md:156. ADR-0032:68 in this same PR shows the clean form ("Synthetic fixture templates, authored for the test..."); could the Confirmation and More Information passages be written the same way?

.github/workflows/main.yml
```diff
@@ -106,59 +103,31 @@ jobs:
...
-    # AND across the ecosystem's own manifests, since `packages/gts-plugin`
-    # pins `mfes` exactly. Catches a version bump to `packages/*` that missed
-    # a pinned site elsewhere. No build-order dependency: reads `package.json`
-    # files directly, not the built CLI.
-    - name: Check template ecosystem pin-drift
-      run: npm run policy:template-pin-drift
+    # still match that package's actual on-disk version - across this repo's
+    # own `packages/*` manifests, since `packages/gts-plugin` pins `mfes`
+    # exactly. Templates that pin these packages now live in their own repo
+    # with their own pin-drift coverage; this is the ecosystem-internal half
```
itechmeat, line 111: [Not a blocker] 🟡 Worth saying out loud somewhere in this repo: the deleted template-drift.yml went red on the gears-frontx PR that broke a template; the re-homed one runs on the templates repo's main plus a daily cron against develop, so the same breakage is now found up to a day later, in another repo, after merge (its own header says so). Fine as a trade-off, but who watches that cron, and should ADR-0034's consequences name it? A repository_dispatch from this CI would close the gap later if it matters.

packages/cli/src/__tests__/template-split.e2e.test.ts
```diff
@@ -379,16 +374,16 @@ describe('Fixture 5 — the conflict check arbitrates nesting and spelling again
...
-      { ground: 'src/ overlaps src', contestants: ['frontx-template-shell', 'slash-mismatch-fixture'] },
+      { ground: 'src/ overlaps src', contestants: ['fixture-shell', 'slash-mismatch-fixture'] },
     ]);
   });
 });
 
-describe('Fixture 6 — declaration coverage: every real file in template-shell/ and template-mfe/ is declared or explicitly allow-listed', () => {
-  it('template-shell/: zero files outside the declared boundary and the {frontx-template.json} allow-list', () => {
-    const manifest = readManifest(TEMPLATE_SHELL_DIR);
+describe('Fixture 6 — declaration coverage: every file in fixture-shell/ and fixture-overlay/ is declared or explicitly allow-listed', () => {
```
itechmeat, line 382: [Not a blocker] 🟡 Fixture 6 used to walk the real template-shell/ and template-mfe/ trees and fail on any file the manifest did not claim. The CLI half is fully preserved here (no assertion dropped), but the product-side guard is gone and not ported: validate:templates checks only what the manifest does claim, never what it does not, and no step in gears-frontx-templates#1 asks that question. Worth a small follow-up in the templates repo (list files, subtract exclusiveSubtrees + allow-list)?

README.md
```diff
@@ -152,6 +152,15 @@ frontx add @acme/web-app ./existing-repo
 frontx upgrade ./my-app 1.1.0
 ```
 
+`acme` above is a placeholder: any repository can publish templates. The
```
itechmeat: [Not a blocker] 🟡 README:155-162 and QUICK_START:52-56 now carry the same sentence, link and command, and CONTRIBUTING:160-161 restates the two consequences that ADR-0034:58/:62 already hold one line before linking to it. Since README:165 already points at QUICK_START for the walkthrough, could the README keep one sentence plus the link and let QUICK_START and the ADR hold the rest?

packages/cyber-pilot-kit-frontx/architecture/features/ai-project-scaffolding/FEATURE.md
```diff
@@ -372,12 +372,13 @@ The system **MUST** leave the ecosystem in a state where the capability is reach
...
 - Entities: `Template`
-- Resource shape: each published template's manifest description (schema owned by `cpt-frontx-feature-template-manifest`)
+- Resource shape: the manifest description a template declares, in a fixture manifest the kit's tests author and in the manifests published from the templates repository (schema owned by `cpt-frontx-feature-template-manifest`)
 - Developer documentation: the repository's quick-start description of which operations a kit capability drives
 - No API surface; no persistent database
 
 **Verifiable clauses**:
-- [ ] Every template manifest the repository publishes declares a non-empty description of what that template establishes and contributes
+- [ ] The kit's own tests demonstrate description-based selection over fixture template manifests they author, so the demonstration depends on no template published from anywhere
+- [ ] Every template manifest published from `constructorfabric/gears-frontx-templates` declares a non-empty description of what that template establishes and contributes - held by review in that repository, since this one publishes no template
```
itechmeat: [Not a blocker] 🟡 A verifiable clause that says in its own text it cannot be verified here reads as satisfied by default. :380 already makes the demonstration independent of any published template; could :381 move out of "Verifiable clauses" into prose, or take DESIGN's "review-held" wording (DESIGN.md:224) so the shape is the one the repo already uses?

scripts/ecosystem-packages.mjs
```diff
@@ -0,0 +1,368 @@
...
+ * Every manifest is read through `readPackageManifest`, so an unreadable or
+ * malformed one aborts naming the file instead of counting as "zero pins here"
+ * (fail-open was the one thing this guard could not afford, since silence is
+ * also what success looks like).
+ *
+ * @param {string} treeDir absolute path
+ * @param {(name: string) => boolean} isEcosystemScopeName
+ * @returns {{ sites: PinSite[]; definedPackageNames: Set<string> }}
+ */
+export function scanTreePins(treeDir, isEcosystemScopeName) {
```
itechmeat: [Not a blocker] 🔵 `scanTreePins` and `findPackageJsonFiles` have no caller left except their own test (the per-template loop that used them is deleted), and the module's `Consumers:` line names two scripts that do not call them. Remove them, or keep them and say the docblock is test-only / kept in step with the templates repo copy?

scripts/mfes-import-boundary-check.mjs
```diff
@@ -17,31 +17,46 @@
...
+ *    every other `packages/*` package, `internal/*`, and the root itself -
+ *    may import a concrete-implementation name from `@gears-frontx/mfes`,
+ *    and no file may deep-import a subpath (`@gears-frontx/mfes/...`) at
+ *    all - a deep import reaches past the barrel into exactly the internals
+ *    the barrel exists to hide. Aliasing does not launder a name: the
+ *    ORIGINAL binding is checked, not the local alias, and `import type`
+ *    counts - a type-level dependency on a concrete class couples the
+ *    consumer to its shape just the same.
+ *
+ * Template territory (`template-shell`, `template-mfe`) is a consumer of
```
itechmeat: [Not a blocker] 🔵 The docblock now says the walk cannot reach template territory, but every runCli case in mfes-import-boundary-check.test.mjs still places the offending consumer under template-shell/ and none under packages/* or internal/*; its :180 docblock also cites the deleted template-shell/vitest.shared.ts. One packages/gts-plugin case plus a prefix swap would make the tests describe the guard they test.

architecture/DESIGN.md
```diff
@@ -104,6 +104,7 @@ CLI (projects orchestration):
 
 * `cpt-frontx-adr-template-acquisition-and-location` — Externalizes templates and resolves them by source-spec at runtime.
 * `cpt-frontx-adr-source-spec-syntax` — Defines the versioned source-spec syntax for template acquisition, including the optional subtree segment that lets one repository publish several addressable templates.
+* `cpt-frontx-adr-template-repository-separation` — Places the templates in a repository of their own, `constructorfabric/gears-frontx-templates`, one addressable subtree per template, realizing the externalization physically.
```
itechmeat: [Not a blocker] 🔵 0033 sits under Foundational at :85, 0034 decides no CLI behaviour - should it sit there too? And :85 still summarises 0033's pre-rewrite rule ("removed as files are touched"); the ADR now says the rule holds wherever the code lives.

CONTRIBUTING.md
```diff
@@ -143,44 +145,22 @@ npm run build:packages:studio
 npm run build:packages:cli
 ```
 
-## Template Development Loop
-
-[`template-shell/`](template-shell) is not a root workspace: it is a standalone npm project that pins `@gears-frontx/api`, `@gears-frontx/mfes` and `@gears-frontx/gts-plugin` to exact registry versions, so that a seeded project installs outside the monorepo. Those pins mean a plain `npm install` inside the template resolves the **published** alpha, not the sources you are editing.
+## Templates
```
itechmeat, line 146: [Not a blocker] 🔵 While this section is being rewritten: :141-144 still list build:packages:framework/react/studio, which no longer exist in package.json, and :168/:171 still describe type-check as "host app + nested MFEs" / "host app only" - there is no host app here any more. Pre-existing, but four lines from the edit.

.github/workflows/publish-packages.yml
```diff
@@ -9,16 +9,15 @@ permissions:
...
   # changed in this push. Ecosystem packages publish INDEPENDENTLY (F1
   # policy) — no layer/dependency ordering.
   #
-  # PART 1 IS A HARD PREREQUISITE OF THE SHELL TEMPLATE (#485): the template
-  # pins @gears-frontx/api, @gears-frontx/mfes and @gears-frontx/gts-plugin to
-  # exact registry versions, so a seeded project cannot install until this job
-  # has published them. Removing the temporary PART 2 (#469 item 5) must leave
-  # PART 1 in place.
+  # The templates that consume these packages (@gears-frontx/api,
+  # @gears-frontx/mfes, @gears-frontx/gts-plugin, ...) now live in their own
```
itechmeat, line 18: [Not a blocker] 🔵 Two open PRs collide with this one and someone should choose the order consciously: #616 extends the pin-substitution mechanism inside the template-validate job this PR deletes (that half of #616 loses its host job if this lands first), and #620 edits 16 files under template-shell/** and template-mfe/** that this PR removes (those edits would need to re-land in the templates repo). Also gears-frontx-templates#1 pins cli 0.3.0-alpha.4 while this PR publishes alpha.6.

.dependency-cruiser.cjs
```diff
@@ -205,10 +206,9 @@ module.exports = {
 
     // @cpt-begin:cpt-frontx-constraint-cli-template-independence:p17:inst-dep-cruiser-rule
     // Scoped to SHIPPED source only (packages/cli/src, excluding the
-    // auto-generated version registry). packages/cli/templates/ and
-    // packages/cli/template-sources/ are fixture/scratch dirs (NOT shipped —
-    // package.json "files": ["dist"]) that legitimately contain template
-    // names/content; they must never trip the CLI-1 boundary check.
+    // auto-generated version registry). The CLI ships zero bundled template
+    // content today (ADR-0016/0017: templates are resolved at runtime by
+    // source-spec), so this rule has nothing to carve an exception for.
```
itechmeat, line 208: [Not a blocker] ⚪ The dropped carve-out was for scratch dirs outside this rule's `from`; the new fixtures under packages/cli/src/__tests__/fixtures/ sit inside it and pass only because they import nothing. Still true tomorrow if a fixture needs an import to mirror a real template shape? A note is enough.

.cf-studio/config/artifacts.toml
```diff
@@ -20,17 +16,13 @@ patterns = ["*/tsup.config.*", "*/vite.config.*", "*/vitest.config.*", "*/postcs
...
... diffs a PR against its merge base and fails any governed package whose substantive changes came without a version bump; scripts/mfes-import-boundary-check.mjs (#540) reads the mfes barrel and its consumers to fail concrete-implementation exports and deep imports across the ADR-0003 boundary; packages/mfes/scripts/verify-dist-import-pragmas.mjs (#504) greps the two built dist formats to fail a build that dropped the webpackIgnore / @vite-ignore comments off the blob-module dynamic import. All three read the tree or the build output and exit non-zero. None of them implements ecosystem behaviour a numbered instruction specifies, so no marker has an instruction to point at. verify-dist-import-pragmas is listed as its own pattern rather than under scripts/ because it lives beside the package it guards and reaches the scan through the root system's packages/ + .mjs catch-all, not through the scripts/ codebase entry. NOTE: scripts/ecosystem-version-policy.mjs (ecosystem-distribution), scripts/package-edge-tests.ts and scripts/verify-guard-configs.ts (ecosystem-governance), and scripts/test-architecture.ts carry @cpt markers and MUST stay scanned — they are deliberately NOT ignored here."
```
itechmeat, line 24: [Not a blocker] ⚪ The removed "STILL UNREGISTERED, deliberately left" paragraph named scripts that left with the templates - and scripts/fs-types.mjs, which is still here, still unmarked and still unregistered. Worth keeping the one sentence that records why?

CLAUDE.md
```diff
@@ -8,6 +8,6 @@ ALWAYS resolve and enforce prerequisites of skills/workflows/commands BEFORE app
...
+Treat `architecture/` as the authority on what the system is and why: PRD for intent, DESIGN for structure, ADRs for decisions, and each FEATURE for the behaviour its numbered instructions specify. Ecosystem code carries `@cpt-` markers back to those instructions; `cfs validate` checks that the chain holds. Template territory is a top-level directory carrying a `frontx-template.json` manifest, and none exists here: the templates live in their own repository, `constructorfabric/gears-frontx-templates`, addressed by source-spec (`github:constructorfabric/gears-frontx-templates//template-shell@<ref>`). `@cpt-` markers found in template territory bind nothing, wherever it lives (`cpt-frontx-adr-template-repository-separation`, `cpt-frontx-adr-template-territory-traceability`). A `frontx-template.json` under `packages/*/src/__tests__/` is a synthetic test fixture, not template territory: it is ecosystem test data and stays inside the artifact chain.
```
itechmeat: [Not a blocker] ⚪ The fixture sentence is already covered by "top-level directory": nothing under packages/*/src/__tests__/ is top-level, and the narrower wording could read as exhaustive (a fixture under internal/ would be exempt by the rule but not named). Feels like a reply to a review comment rather than a rule - I would drop it.

README.md
```diff
@@ -152,6 +152,15 @@ frontx add @acme/web-app ./existing-repo
...
+official FrontX templates are published from
+[`constructorfabric/gears-frontx-templates`](https://github.com/constructorfabric/gears-frontx-templates),
```
itechmeat: [Not a blocker] ⚪ While adding a third repository link here: :242 still says `git clone https://github.com/cyberfabric/FrontX.git`, which is not this repository. Pre-existing, one-line fix.

itechmeat (approved):
Approving - the mechanical half of this move is done carefully: no dangling reference to a removed path anywhere in the tree, every deleted script has no caller left, the surviving guards keep their packages/* half byte-for-byte, the CLI e2e tests lose no assertion on the way to synthetic fixtures, and all four points from the earlier review comment are genuinely addressed at head. What I would settle before pressing merge is ordering: the receiving repository is two files long today and its CI has never run, while several artifacts here already describe it in the present tense - so land gears-frontx-templates#1 and a tag first, then this. The rest inline is artifact hygiene (a new ADR where the house rewrites in place, a patched ADR-0033 that now contradicts itself, history wording) and two follow-ups worth filing on the templates side.

architecture/ADR/0003-mfe-runtime-public-surface.md
GeraBart: Templates as a concept are unrelated to mfes isolation. Mfes isolation is technological layer, templates is productization level. This ADR should be moved to the mfes package, it belongs there, not root SDLC artifacts.

architecture/ADR/0034-template-repository-separation.md
```diff
@@ -0,0 +1,146 @@
...
+- [More Information](#more-information)
+- [Traceability](#traceability)
+
+<!-- /toc -->
+
+**ID**: `cpt-frontx-adr-template-repository-separation`
+
+## Context and Problem Statement
+
+A template is a top-level directory carrying a `frontx-template.json` manifest (`cpt-frontx-adr-template-manifest-contract`). Three of them lived in this repository: `template-shell`, `template-mfe` and `template-design-guardrails`. The architecture already says templates are external. `cpt-frontx-adr-template-acquisition-and-location` decided the CLI bundles no template and resolves each one from an external source at runtime, and `cpt-frontx-adr-source-spec-syntax` decided a reference may name a subtree inside a source repository, so one repository can hold a family of templates. Until now that was true on paper only: the CLI shipped no template, but the templates it resolved sat in the same repository as the packages they consume, and every guard, test and release step written for them ran in the ecosystem's continuous integration.
```
GeraBart: gears-frontx should not be aware of gears-frontx-templates content

GeraBart:
please ensure that this repo has zero mentionings of any specific template. it should have SDLC artifacts describing templates as a concept, but be completely unaware of what is in gears-frontx-templates repo.

tscbmstubp:
Thanks both. Head now follows @GeraBart's direction: this repository names no specific template and knows nothing about the templates repository's content. It keeps the concept only (a template, its manifest, template territory, the repository that publishes templates).

**Architecture**
- ADR-0034 is gone. Its option comparison and consequences are folded into ADR-0016 "Template Acquisition and Location", which now decides both how a template is obtained and which repository publishes them (house convention: rewrite in place).
- ADR-0033 is rewritten as one current-state decision and takes the ownership statement: template territory lives in a repository of its own under its own governance. Template territory is defined as a top-level directory carrying a `frontx-template.json` manifest, consistently with CLAUDE.md/AGENTS.md and the Studio config.
- ADR-0003 no longer carries a template paragraph; the reach of the mfes boundary check is described in `packages/mfes/architecture/DESIGN.md` §3.3. Relocating the whole ADR into the mfes package needs a registry entry in `.cf-studio/config/artifacts.toml` and LAYER-3 retargeting of the root DESIGN trace tables, so that is a separate change.
- ADR-0022 and the scaffolding FEATURE lost their template-specific clauses; the unverifiable clause moved into prose.
- DESIGN.md index updated; drift-detection latency is named as a consequence in ADR-0016.

**Code and tests**
- `validate-manifest.ts` no longer holds a list of template names; it recognises the shape `(frontx-)template-<flavor>` structurally, with tests on fictional names.
- `scanTreePins`/`findPackageJsonFiles` removed; import-boundary tests moved under `packages/*` with a gts-plugin case; `test-architecture.ts` check generalised to the naming convention; depcruise note and `fs-types.mjs` ignore note added.
- Zero hits repo-wide for any template name or for ADR-0034.

**Docs**
- README keeps one sentence and one link; QUICK_START holds the walkthrough with a placeholder source-spec `github:<owner>/<templates-repo>//<template>@<ref>`; CONTRIBUTING links ADR-0016 instead of restating it. Stale build/type-check/publish lines and the wrong clone URL fixed.

All gates green locally: lint, type-check, 2281 unit tests, arch and policy guards, `cfs validate` 0/0.

tscbmstubp:
Rebased onto develop `760b6764` (single signed-off commit, `d4bc80c1`).

- Conflicts were all modify/delete under `template-shell/` and `template-mfe/`; the deletion wins. develop's edits to those trees since the `1531fba3` snapshot (framework types, react MFE hooks incl. the new `useMfeRegistry`, the shared-dep cycle guard test, `mf-gts.ts`, mfe manifests) are not lost: they will be re-snapshotted into gears-frontx-templates after this merges, so the cutover order is: merge this, then re-snapshot and re-pin cli in gears-frontx-templates#1, then merge that. Open PRs still writing into the template directories should re-target the templates repository.
- The `version-bump-on-change` gate now requires `@gears-frontx/api` 0.3.0-alpha.3 (its `src/__tests__/fixtures/mockPlugins.ts` lost a template name) and `@gears-frontx/cyber-pilot-kit-frontx` 0.3.0-alpha.5 (`validate-manifest.ts` changed). Nothing pins either exactly.
- One template path that develop re-introduced in ADR-0012 was generalised.
- Audit is green here: develop's smol-toml 1.7.2 pin came in with the rebase.

GeraBart (approved):

PR #616, opened 2026-09-07
feat(ui-kit): component contracts as the agent knowledge layer, with the package artifact chain

itechmeat:
## The problem

Agents consume UiKit knowledge as prose. READMEs, doc comments and demo screens describe what a component means, when to reach for it and what not to do with it, and none of that is checked against the component. Prose drifts the moment the code moves: a variant gets renamed, a default flips, a prop is removed, and the description keeps reading as though nothing happened. There is no artifact an agent can load that is both semantic enough to guide a choice and validated enough to trust.

The demo in #611 showed one possible shape on one component, deliberately as a discussion piece with a kit-wide red gate. This PR is the mergeable follow-up: the same idea reworked after the review of that demo, on three components of increasing difficulty, with the checks that make it safe to keep.

## What changed after the demo

- **Code is the source of truth.** The hand-written overlay carries only meaning (intent, typical uses, when not to use and what instead, invariants, anti-patterns, examples). Every extractable fact (variant axes, defaults, own props, inherited props) comes from the TypeScript checker, and the compiler rejects an overlay that restates one. One fact, one owner.
- **Semantics shrank to what an agent needs at choice time.** `use_when` became `typical_uses` (at most three), `dont_use_when` is required and its `instead` target is a typed component id, and the fields a validator reads (`dont_use_when`, `composition`, `deprecations`, `coverage`, `family`, `extension_points`) moved into `x-gts-traits` with a generated traits schema on the base component type. Documentation-only fields stay in `x-uikit`.
- **No kit-wide red gate.** Coverage is a report. The blocking checks are scoped to what a pull request touches.

## What this PR materializes

**Contract harness** under `packages/ui-kit/scripts/contracts/`. The extractor unwraps `memo`/`forwardRef`, resolves `cva` through aliases and sibling files, classifies heritage shapes (`Omit`, `Pick`, `VariantProps`, `ComponentProps`, `BaseUIComponentProps`) by symbol rather than by text, splits own props from inherited props by declaration file, and writes anything it cannot classify into `cannot_extract` instead of guessing. Passthrough schemas are generated per origin of the inherited props (`base_ui_<component>_<part>`, `dom_<tag>`) and shared across components, with a collision check on `generated_from`. The compiler validates the overlay against a closed schema, validates the instance against the metamodel and the traits against the generated traits schema, emits `required`, and takes the contract major and metamodel version as parameters. Absolute paths never reach the artifacts.

[...]

PR #620, opened 2026-09-09
Fix: parallel deps fetching

### LINKED ISSUE #618: MFE loader serializes independent dependency fetches instead of running them concurrently
`MfeHandlerMF` loads a microfrontend's shared dependencies and its own exposed-asset chunk graph through loops that `await` each dependency before starting the next fetch, even when nothing requires that ordering.

**Where:**
- `fetchSharedDepSources` — iterates `manifest.shared` with a `for...of` loop, awaiting each dependency's fetch before moving to the next. The full list of shared dependencies is known upfront; nothing depends on fetch order at this stage. Dependency order is only needed afterward, when blob URLs are built from already-resolved source text — a separate step that runs after every fetch has already completed.
- `createBlobUrlChainInternal` — after parsing a chunk's static imports, awaits each sibling import one at a time. Fetching a chunk before its children are known is unavoidable, but siblings discovered at the same level don't depend on each other and could be requested together.

**Why this is a real bug and not just a missed optimization:** `LoadBlobState.inFlight`'s own doc comment explicitly anticipates concurrent sibling requests for the same filename ("concurrent calls for the same filename are deduplicated... sibling ESM modules... trigger overlapping calls for the same dependency") — the dedup guard this would exercise already exists. The call sites simply never issue those calls concurrently, so the guard sits unused.

**Impact:** every microfrontend load pays the *sum* of its dependency fetch times instead of being bounded by the slowest one. This is worse than a normal inefficiency on any origin still limited to HTTP/1.1's 6-connections-per-origin cap (a real, current constraint for at least one consumer, with no near-term fix available at the infrastructure layer) — the loader leaves most of its allowed connections idle by construction, regardless of what the network/infra otherwise allows.

**Suggested fix:** in `fetchSharedDepSources`, fire all shared-dependency fetches concurrently (e.g. `Promise.all`) instead of the sequential loop, keeping each dependency's cache lookup-then-store synchronous relative to its own fetch so the existing cross-MFE `name@version` deduplication is unaffected. In `createBlobUrlChainInternal`, fire sibling static imports concurrently at each discovered level, relying on the existing `inFlight` map for cross-sibling dedup.

### LINKED ISSUE #619: [FEATURE] Allow apps to declare app.mfeRegistry as always-present via type augmentation
## Description

Give shell/app authors a way to declare **once** — via TypeScript type augmentation — that `app.mfeRegistry` is guaranteed present, so consuming code stops needing per-call-site `if (app.mfeRegistry)` / `app.mfeRegistry!` guards.

Reported against a **`@gears-frontx/react` + `@gears-frontx/framework`** stack.

## Problem Statement

`mfeRegistry` is declared as an optional property directly on `FrontXApp`:

```ts
// template-shell/packages/framework/src/types.ts:417
export interface FrontXApp extends FrontXAppRuntimeExtensions {
  /** MFE-enabled MfeRegistry (optional, provided by microfrontends plugin) */
  mfeRegistry?: import('@gears-frontx/mfes').MfeRegistry;
  // ...
}
```

The optionality is correct at the framework level — `microfrontends()` is a plugin, and a `minimal()` / `headless()` build genuinely has no registry. But an application that *does* compose `microfrontends()` (including everything built from `createFrontXApp()` / the `full()` preset) knows statically that the registry is always there, and today has no way to tell the type system that. The result is guard noise at every call site:

```ts
if (!app.mfeRegistry) return;              // dead branch in a full() app
app.mfeRegistry!.registerDomain(...)       // non-null assertions everywhere
const r = app.mfeRegistry; if (!r) throw new Error('expected mfeRegistry');
```

This pattern is already visible across the framework's own tests (`__tests__/plugins/microfrontends/*.test.ts` uses `app.mfeRegistry!` and `getMfeRegistry(app)` throw-helpers), and it repeats in every downstream shell.

Declaration merging cannot fix this from the consumer side: `mfeRegistry` lives directly on `FrontXApp`, and TypeScript does not allow a merged interface declaration to redeclare an existing property with a different optionality — the augmentation conflicts instead of narrowing.

Notably, the framework already has the right mechanism for exactly this — `FrontXAppRuntimeExtensions`, the declaration-merging target that `auth.ts` uses:

```ts
// template-shell/packages/framework/src/plugins/auth.ts:702
declare module '../types' {
  interface FrontXAppRuntimeExtensions {
    auth?: AuthRuntime;
  }
}
```

`mfeRegistry` predates that mechanism and is hard-coded as a special case on `FrontXApp`.

### React layer

`useFrontX()` returns `FrontXApp` verbatim (`react/src/FrontXContext.tsx:30`), so every component that reaches for the registry inherits the optionality and repeats the guard:

```tsx
const app = useFrontX();
if (!app.mfeRegistry) return null;        // dead branch in a full() app
app.mfeRegistry!.registerDomain(...);
```

`@gears-frontx/react` already pays this cost internally — each of its three registry-reading hooks re-does the same narrowing by hand:

- `react/src/mfe/hooks/useDomainExtensions.ts:50-59`
- `react/src/mfe/hooks/useMountedExtensions.ts:60`
- `react/src/mfe/hooks/useRegisteredPackages.ts:61`

each with a variant of:

```ts
const registry = app.mfeRegistry;
if (!registry) {
  throw new Error('useDomainExtensions requires the microfrontends plugin. ...');
}
```

There is **no `useMfeRegistry()` hook exported** — the registry-reading hooks keep the narrowed handle private, so app code has no non-nullable way in.

This constrains the solution: whatever augmentation mechanism lands must flow through `useFrontX()`, i.e. it has to be a property-level change on the shared `FrontXApp` type, not a return-type override on `createFrontXApp()` — `FrontXProvider` / `useFrontX` are typed against `FrontXApp`.

## Proposed Solution

Make `mfeRegistry` narrowable by the application, in the spirit of the existing plugin-extension mechanism. Sketch:

1. Move `mfeRegistry` off `FrontXApp` and onto the augmentation surface, declared by the `microfrontends()` plugin the same way `auth` is:

   ```ts
   // src/plugins/microfrontends.ts
   declare module '../types' {
     interface FrontXAppRuntimeExtensions {
       mfeRegistry?: MfeRegistry;
     }
   }
   ```

2. Expose an opt-in "presence" augmentation an app declares once in its own `frontx.d.ts`, e.g.

   ```ts
   import '@gears-frontx/framework';

   declare module '@gears-frontx/framework' {
     interface FrontXAppGuarantees {
       mfeRegistry: true;   // this app always composes microfrontends()
     }
   }
   ```

   with `FrontXApp` resolving `mfeRegistry` to required vs optional based on that flag (conditional-type lookup on an empty-by-default guarantees interface). Default stays optional — no behavior change for apps that don't opt in.

Exact shape is open — the requirement is: **one declaration per app, zero guards at call sites, the optional default preserved for minimal/headless builds, and the narrowing visible through `useFrontX()`.**

### Companion: export `useMfeRegistry()`

Independently of the augmentation work, a small **`useMfeRegistry(): MfeRegistry`** hook in `@gears-frontx/react` — the existing throw-guard, extracted and exported — would remove most of the day-to-day guard noise for React consumers at near-zero cost, and would let the three internal hooks share one implementation instead of three copies. Worth doing regardless of how the type-augmentation question is resolved; could be split into its own issue.

## Use Case / Example

- A shell built from `createFrontXApp()` (full preset) registers domains and extensions in dozens of places and currently guards `app.mfeRegistry` at each one.
- React components that call `useFrontX()` and then need the registry — every one repeats the guard.
- Plugins/effects in the same app that call `app.mfeRegistry.updateSharedProperty(...)` / `setTheme(...)`.
- Framework and app tests, which today resort to `app.mfeRegistry!` or throw-helpers to get a non-nullable handle.

## Affected Package(s)

- [x] @gears-frontx/framework
- [x] @gears-frontx/react
- [x] @gears-frontx/mfes
- [ ] @gears-frontx/cli
- [ ] @gears-frontx/i18n
- [ ] @gears-frontx/state
- [ ] @gears-frontx/studio

## Alternative Solutions

- **Local helper** — `function requireMfeRegistry(app): MfeRegistry`, throwing when absent. Works today, but pushes a runtime check into every path and still has to be written per app.
- **Per-preset app types** — a distinct `FrontXFullApp` type (with a required `mfeRegistry`) returned by `createFrontXApp()`. **Ruled out as a complete fix:** `FrontXProvider` / `useFrontX` are typed against `FrontXApp`, so the narrowing would never reach React consumers — and it would not help custom `.use(microfrontends())` compositions either.
- **Builder-level type inference** — have `.use(microfrontends())` narrow the built app's type. Most precise, but the largest change to the builder's typing, and it still has to be threaded through `FrontXProvider` / `useFrontX` to help React code.

## Additional Context

Reported by a FrontX consumer:

> Ты не мог бы дать возможность через type augmentation один раз указать явно что mfeRegistry доступен? Очень много if-ов с проверкой на то, что app.mfeRegistry не undefined.
>
> Мы используем react и framework.

References:
- `template-shell/packages/framework/src/types.ts:376` — `FrontXAppRuntimeExtensions` (existing merge target)
- `template-shell/packages/framework/src/types.ts:400-424` — `FrontXApp`, `mfeRegistry?`
- `template-shell/packages/framework/src/plugins/auth.ts:702-706` — precedent for the plugin-augmentation pattern
- `template-shell/packages/react/src/FrontXContext.tsx:30` — `useFrontX(): FrontXApp`
- `template-shell/packages/react/src/mfe/hooks/{useDomainExtensions,useMountedExtensions,useRegisteredPackages}.ts` — three hand-rolled copies of the same guard

## Priority

- [ ] Low - Nice to have
- [x] Medium - Improves workflow
- [ ] High - Blocking or critical for use case

GeraBart:
## Summary by CodeRabbit

- **New Features**
  - Added the `useMfeRegistry` hook for accessing the module registry.
  - Added framework typing support for applications that provide an MFE registry.
  - Added shared-dependency cycle detection during builds.

- **Bug Fixes**
  - Improved concurrent module loading and isolation between independent loads.
  - Added configurable timeout handling for stalled load attempts.
  - Improved diagnostics for circular or unavailable dependencies, with deterministic load failures.

- **Documentation**
  - Updated architecture and feature documentation to clarify resource ordering, discovery, loading, timeout, and failure behavior.


packages/mfes/src/handler/MfeHandlerMF.ts
```diff
@@ -827,52 +1317,239 @@ class MfeHandlerMF extends MfeHandler<MfeEntryMF, ChildMfeBridge> {
...
+      }
+      // @cpt-begin:cpt-frontx-algo-mfe-isolation-blob-url-chain:p1:inst-fetch-source
+      source = await this.fetchSourceText(chunkUrl);
+      // @cpt-end:cpt-frontx-algo-mfe-isolation-blob-url-chain:p1:inst-fetch-source
+    } finally {
+      build.fetchBudget.release();
+    }
+    // @cpt-end:cpt-frontx-algo-mfe-isolation-blob-url-chain:p1:inst-fanout-bounded
+
+    if (build.failed) {
```
itechmeat, line 1819: [Not a blocker] 🟠 The failure flag is per build now, but the join point it guards - `LoadBlobState.inFlight` - is still shared across the whole load. If build A abandons a chunk on this line without minting a blob URL, build B that already joined the same in-flight promise resolves cleanly, finds nothing in `blobUrlMap`, and dies in `rewriteModuleImports` with the "never built" error. I reproduced it with two overlapping `resolveLazyChunk` calls sharing a chunk two levels deep; the existing re-attempt test only covers the sequential case. Still strictly better than develop, where any failure poisoned every later lazy import for good, so fine as a follow-up - but should the joining caller re-check `blobUrlMap` after awaiting a joined promise and re-attempt when it produced nothing? Either way the `ChainBuildState` doc block currently claims cross-build isolation the code does not hold.

packages/mfes/architecture/features/mfe-isolation/FEATURE.md
```diff
@@ -104,52 +104,83 @@ Internal system functions that implement the isolation mechanism.
 
 ### Blob URL Chain Construction
 
-- [x] `p1` - **ID**: `cpt-frontx-algo-mfe-isolation-blob-url-chain`
+- [ ] `p1` - **ID**: `cpt-frontx-algo-mfe-isolation-blob-url-chain`
```
itechmeat: [Not a blocker] 🟠 The spec now says the opposite of the code. Both algorithms went from `[x]` to `[ ]`, and all twenty new instructions (failure scope, ancestor/join cycle, in-flight registration, failure-signal branches, duplicate names) are `[ ]` with no code markers - yet the code for every one of them is in this diff. The acceptance criteria at :286/:288 are `[x]` while the steps behind them are `[ ]`, and mfe-loading has the reverse for the timeout (:249/:251). One marker is also wrong: the `inst-return-inflight` region at MfeHandlerMF.ts:1348 returns a resolved promise, not the in-flight one - that branch is `inst-return-join-cycle`, which is left unmarked. Could you mark the regions, tick the boxes, and fix that id, so the chain describes what shipped?

template-shell/package.json
```diff
@@ -70,8 +70,8 @@
   },
   "dependencies": {
     "@gears-frontx/api": "0.3.0-alpha.2",
-    "@gears-frontx/gts-plugin": "0.3.0-alpha.4",
-    "@gears-frontx/mfes": "0.3.0-alpha.4",
+    "@gears-frontx/gts-plugin": "0.3.0-alpha.5",
```
itechmeat, line 73: [Not a blocker] 🟡 Not about the red check itself - that is the known bump-before-publish gap (#607 merged through the same failure; #613's substitution only covers the composition step and the extension to this one is in #616). Just flagging that Template Validate is also the only job that compiles, lints and tests template-shell/packages/{framework,react}, so nothing green has checked the new conditional type, the new hook or the three refactored hooks. A local run of build + type-check + lint + test:unit inside an isolated template-shell copy would be reassuring.

architecture/ADR/0013-mfe-asset-discovery.md
```diff
@@ -32,7 +32,7 @@ To load a microfrontend the runtime must discover where its chunk assets live, w
...
-* Completeness for an isolated load — the description must carry everything an isolated load needs: the base location for assets, the file backing each exposed entry, the entry's stylesheet assets, and the shared dependencies in resolvable order — so a load needs no out-of-band lookup.
+* Completeness for an isolated load — the description must carry everything an isolated load needs: the base location for assets, the file backing each exposed entry, the entry's stylesheet assets, and the full set of shared-dependency declarations in a stable enumeration order — so a load needs no out-of-band lookup. That enumeration order fixes the sequence in which a consumer visits the declarations; it is not a load order. Resolving the declarations leaves-first remains the consumer's own pass over the sources it fetches, which is where dependency-order correctness and circular-dependency handling are decided (`cpt-frontx-adr-mfe-load-isolation`). Every mention of ordered declarations elsewhere in this decision carries this meaning.
```
itechmeat: [Not a blocker] 🟠 This sentence (and :95) points at ADR 0011 as the place where circular-dependency handling is decided, but 0011 does not mention cycles at all. And this PR does add a real exclusion to 0011's isolation guarantee - the cycle fallback resolves the closing import to its origin URL, so one module evaluates outside the isolated graph (the warn at MfeHandlerMF.ts:1436 says as much). Worth adding that exclusion to 0011's Consequences in place so the pointer becomes true?

packages/mfes/src/handler/MfeHandlerMF.ts
```diff
@@ -278,11 +560,17 @@ class MfeHandlerMF extends MfeHandler<MfeEntryMF, ChildMfeBridge> {
...
     }
     // @cpt-begin:cpt-frontx-flow-mfe-isolation-load:p1:inst-else-new-load
+    // @cpt-begin:cpt-frontx-algo-mfe-loading-attempt-timeout:p1:inst-lto-per-attempt-budget
+    // The timeout wraps the ATTEMPT, inside the retry loop — every retry of
+    // a failed attempt is raced against its own fresh budget rather than
+    // sharing the first attempt's, so a load's worst-case wall clock is the
+    // budget times the attempt count plus `RetryHandler`'s backoff.
     const promise = this.retryHandler.retry(
-      () => this.loadInternal(entry),
+      () => this.withLoadTimeout(this.loadInternal(entry), entry.id),
```
itechmeat: [Not a blocker] 🟡 Does the retry actually get a fresh attempt here? `fetchSourceText` caches the in-flight promise and only evicts it in `.catch`, so a fetch that never settles is never evicted - attempts 2 and 3 reuse the same hung promise and time out again (3 x 30s by default). Bounding the hang is still worth it, but either the timeout should evict the URL from `sourceTextCache` / `sharedDepTextCache` when it wins the race, or the doc block should say "bounded failure" rather than "recovery path".

packages/mfes/src/handler/MfeHandlerMF.ts
```diff
@@ -317,6 +605,87 @@ class MfeHandlerMF extends MfeHandler<MfeEntryMF, ChildMfeBridge> {
...
+  // @cpt-dod:cpt-frontx-dod-mfe-loading-attempt-timeout:p1
+  private withLoadTimeout<T>(
+    attempt: Promise<T>,
+    entryId: string
+  ): Promise<T> {
+    // @cpt-begin:cpt-frontx-algo-mfe-loading-attempt-timeout:p1:inst-lto-read-budget
+    const timeoutMs = this.config.timeout;
+    // @cpt-end:cpt-frontx-algo-mfe-loading-attempt-timeout:p1:inst-lto-read-budget
+    // @cpt-begin:cpt-frontx-algo-mfe-loading-attempt-timeout:p1:inst-lto-if-disabled
+    if (timeoutMs === undefined || timeoutMs <= 0) {
```
itechmeat, line 821: [Not a blocker] 🟡 `timeout` was dead config before this PR and is live now, so every app that passes no config silently gets a 30s-per-attempt budget where a hang used to hang - and neither linked issue asked for it. Fine as a change, but it needs a sentence in the PR description, `MfeLoaderConfig` exported from the package index with doc comments on its fields, and the "non-positive disables the race" rule stated somewhere a consumer reads rather than inside a private method's doc block.

packages/mfes/src/handler/MfeHandlerMF.ts
```diff
@@ -89,6 +273,80 @@ interface LoadBlobState {
...
+ *
+ * This is deliberately NOT part of {@link LoadBlobState}: `LoadBlobState`
+ * lives for the whole load (page lifetime, per the never-revoke invariant)
+ * and is shared by every lazy chunk resolved through it, but "a fetch
+ * failed somewhere in this build" must not survive past the build it
+ * happened in. A `LoadBlobState`-scoped flag would latch permanently after
+ * the first failed lazy import — poisoning every later, otherwise
+ * independent, lazy import on the same already-mounted MFE, with no retry
+ * path short of a full page reload. Scoping the flag to one build instead
+ * means a failed lazy import fails only its own `resolveLazyChunk` call;
```
itechmeat, line 274: [Not a blocker] 🟡 A lot of the new prose argues against designs that only ever existed inside this PR - the `ChainBuildState` block answering "resetting a load-wide flag would NOT fix this", the "versus the sequential loop this replaced" asides at :180/:420/:1080, and the "before the fix" sentences in the tests (`LoadBlobState` never had a `failed` field on develop). The invariant in the first sentence of each block is worth keeping; the "what it replaced" half reads like the PR description and this file is now 64% comment lines. Same for "rather than awaiting each sibling's entire subtree" in FEATURE.md:137 - describe what it does.

packages/mfes/architecture/features/mfe-isolation/FEATURE.md
```diff
@@ -252,5 +283,10 @@ The system **MUST** accept an entry's manifest either as the document itself or
...
+- [x] A dependency absent from the per-load blob URL map for any reason other than the deliberate cycle fallback fails the load with a diagnostic naming the referring chunk and the unbuilt dependency, rather than silently emitting an origin URL.
+- [x] A chain build that fails does not affect any later independent chain build of the same load: a lazy import that follows a failed one re-attempts construction of every chunk the failed build abandoned, including chunks the two builds share.
+- [x] Sibling static-import dependencies are fetched concurrently, and the number of chunk-source fetches in flight for one chain build never exceeds the configured width no matter how deep or how wide the dependency graph is; the failure reported for a group of siblings is still the first in declaration order regardless of completion order.
+- [ ] A manifest declaring the same shared-dependency package name more than once fails the load with a diagnostic naming that package, before any shared-dependency source is fetched.
```
itechmeat: [Not a blocker] 🟡 Three new refusals a caller can observe (duplicate shared name, unbuilt dependency at rewrite, attempt timeout) and neither flow's Error Scenarios list changed. Could they get their entries, and this criterion its tick - the duplicate-name guard is implemented and tested? Small thing on the same block: `assertUniqueSharedDepNames` passes `manifest.id` as `entryTypeId`, so the message renders as "Failed to load MFE '<manifest id>'".

packages/mfes/src/handler/MfeHandlerMF.ts
```diff
@@ -47,6 +47,182 @@ import {
...
+ * manifest).
+ *
+ * The width is chosen to stay well short of the browser's per-origin
+ * connection pool, so that added concurrency still shortens wall-clock time
+ * instead of just queuing behind that pool, while an unbounded fan-out
+ * risks minting page-lifetime blob URLs (see the never-revoke invariant on
+ * {@link createBlobUrlChainInternal}) for work a failed load will never
+ * use.
+ */
+const MAX_CONCURRENT_FETCHES = 6;
```
itechmeat, line 69: [Not a blocker] 🟡 The spec says "the configured width" but nothing configures it - it is a module const, next to `timeout` and `retries` that hosts can set. Either the wording changes or it moves into `MfeLoaderConfig`; also is it 6 or 12 in flight when the shared-dep and chunk fan-outs run at once? And the decision itself - bound concurrency to a fixed width chosen against the browser's per-origin pool - is in no ADR, only in the instruction's why-clauses.

packages/mfes/architecture/features/mfe-isolation/FEATURE.md
```diff
@@ -104,52 +104,83 @@ Internal system functions that implement the isolation mechanism.
...
 
-- [x] `p1` - **ID**: `cpt-frontx-algo-mfe-isolation-build-shared-dep-blob-urls`
+- [ ] `p1` - **ID**: `cpt-frontx-algo-mfe-isolation-build-shared-dep-blob-urls`
 
 **Input**: MFE manifest containing the shared-dependency list (name, version, chunk path) in dependency order (leaves first)
 
 **Output**: Map of shared-dependency package name to blob URL, covering all shared dependencies declared in the manifest
 
+**Note**: "In declaration order" in `inst-for-each-dep` governs enumeration and cache-key precedence only — the order in which declarations are visited and, when two declarations collide on the same `name@version` key, which one claims the cross-MFE cache entry. It imposes no ordering on when fetches are issued or when they complete, so the enumeration may issue its fetches concurrently. What keeps that deduplication race-free however the fetches are scheduled is that the entry `inst-fetch-and-cache` stores is the in-flight fetch promise, recorded under the key before it is awaited. Dependency-order correctness for blob construction comes solely from `inst-resolve-order`, which orders the already-fetched sources leaves-first and falls back to partial rewrites on cycles.
```
itechmeat: [Not a blocker] 🟡 The Input line at :162 still says "in dependency order (leaves first)" and the flow step at :91 says the same, while this Note says manifest order is enumeration only and correctness comes solely from `inst-resolve-order`. Could the two older sentences be rewritten instead of adding a paragraph that reinterprets them? `**Note**` is also not a slot the CDSL algorithm shape defines.

architecture/ADR/0013-mfe-asset-discovery.md
itechmeat: [Not a blocker] 🟡 "Every mention of ordered declarations elsewhere in this decision carries this meaning" is there because :41 and :67 still say "the ordered shared-dependency declarations" unchanged - could those two be fixed instead? The same explanation is now written out at :35, :47, :61, :95 and again in mfe-isolation FEATURE.md:166; one place plus pointers would read better.

packages/mfes/src/handler/MfeHandlerMF.ts
```diff
@@ -47,6 +47,182 @@ import {
...
+ * bound holds for the whole build regardless of depth or sibling-group
+ * count.
+ *
+ * A slot is held ONLY across the source fetch, never across a recursion
+ * into a dependency: a parent holding a slot while waiting for its
+ * children's fetches would deadlock as soon as the graph is deeper than
+ * the width. Waiters are released strictly first-in-first-out, so siblings
+ * admitted from one fan-out enter in declaration order.
+ */
+class FetchBudget {
```
itechmeat, line 89: [Not a blocker] 🟡 `FetchBudget`, `boundedMap` and `firstRejection` know nothing about MFEs - own module? The handler is 2k lines now, and the two primitives would be easy to unit-test on their own the way `LruCache` is.

packages/mfes/src/handler/__tests__/MfeHandlerMF.test.ts
```diff
@@ -103,3 +213,715 @@ describe('MfeHandlerMF — unresolved publicPath placeholder guard', () => {
...
+        ) => string;
+      }
+    ).rewriteModuleImports(source, loadState, chunkFilename, build);
+
+  it("resolves a cycle-short-circuited dependency to its origin chunk URL", () => {
+    const handler = new MfeHandlerMF(ENTRY_BASE_ID, { retries: 0 });
+    const loadState = buildLoadState();
+    const build = {
+      failed: false,
+      fetchBudget: null,
```
itechmeat: [Not a blocker] 🟡 Eleven of the twenty cases reach private methods through `as unknown as` casts, and this one hands in a `ChainBuildState` with `fetchBudget: null` that cannot exist at runtime. The stated reason (no blob `import()` under Node) is real, which is usually the sign of a missing seam - an injectable `importBlobModule` or an observable `blobUrlMap` - rather than of tests needing casts. Fine to leave for a follow-up, but worth naming as one.

template-shell/packages/framework/src/types.ts
```diff
@@ -397,7 +397,38 @@ export interface FrontXAppRuntimeExtensions {
...
+ * (no longer `MfeRegistry | undefined`), so call sites no longer need an
+ * `if (!app.mfeRegistry)` guard or a `app.mfeRegistry!` assertion.
+ */
+export interface FrontXAppGuarantees {}
+
+type MfeRegistrySlot = FrontXAppGuarantees extends { mfeRegistry: true }
+  ? { mfeRegistry: MfeRegistry }
+  : { mfeRegistry?: MfeRegistry };
+
+export interface FrontXApp extends FrontXAppRuntimeExtensions, MfeRegistrySlot {
```
itechmeat, line 443: [Not a blocker] 🟡 The new `FrontXAppGuarantees` block landed between the old "FrontX App Interface" doc and `FrontXApp` itself, so `FrontXApp` now has no JSDoc and the stranded block's `@example` still shows `if (app.mfeRegistry) {` - the guard #619 exists to remove. Same slip in MfeHandlerMF.ts:608, where `withLoadTimeout` sits under `loadInternal`'s doc block. Both are a one-line move.

template-shell/packages/react/src/mfe/hooks/index.ts
```diff
@@ -7,6 +7,7 @@
 export { useMfeBridge } from './useMfeBridge';
 export { useSharedProperty, type UseSharedPropertyOptions } from './useSharedProperty';
 export { useHostAction } from './useHostAction';
+export { useMfeRegistry } from './useMfeRegistry';
```
itechmeat, line 10: [Not a blocker] 🟡 Also docs: react's CLAUDE.md (shipped in `files`) has a section per mfe hook and a summary list, and `useMfeRegistry` is in neither, nor in llms.txt. On the framework side `FrontXAppGuarantees` appears only in types.ts and index.ts - where would an app author learn to write the `declare module ... { mfeRegistry: true }` line that #619 is about?

template-shell/packages/react/src/mfe/hooks/useMfeRegistry.ts
```diff
@@ -0,0 +1,60 @@
...
+ *
+ * @param callerName - Name of the calling hook, used in the error message
+ * @returns The MFE-enabled registry
+ *
+ * @example
+ * ```ts
+ * const registry = useMfeRegistry();
+ * ```
+ */
+export function useMfeRegistry(callerName: string = 'useMfeRegistry'): MfeRegistry {
```
itechmeat: [Not a blocker] 🟡 Nobody passes `callerName` - the three refactored hooks call `resolveMfeRegistry` directly - so the public hook gains a string argument that means nothing to an app author, and it has no test. The comment at :43 justifying the two-function split says a hook "must not read the context twice", but `useFrontX()` is just `useContext` plus a guard; a second call is free. Could this shrink to `useMfeRegistry(): MfeRegistry` plus a test?

template-shell/packages/framework/src/types.ts
```diff
@@ -397,7 +397,38 @@ export interface FrontXAppRuntimeExtensions {
...
+ * }
+ * ```
+ *
+ * With that declaration in scope, `app.mfeRegistry` is typed as `MfeRegistry`
+ * (no longer `MfeRegistry | undefined`), so call sites no longer need an
+ * `if (!app.mfeRegistry)` guard or a `app.mfeRegistry!` assertion.
+ */
+export interface FrontXAppGuarantees {}
+
+type MfeRegistrySlot = FrontXAppGuarantees extends { mfeRegistry: true }
```
itechmeat, line 419: [Not a blocker] 🔵 What stops a `minimal()` app from declaring `mfeRegistry: true`? Nothing checks it against the composed plugins, and `createFrontX.ts:186` casts `aggregated.registries.mfeRegistry as FrontXApp['mfeRegistry']`, so the framework's own composition is the first place the declaration makes unsound. Probably fine as an opt-in assertion, but a sentence in the doc block naming that cost would help.

packages/cyber-pilot-kit-frontx/package.json
```diff
@@ -47,6 +47,6 @@
   ],
   "license": "MIT",
   "dependencies": {
-    "smol-toml": "^1.6.1"
+    "smol-toml": "1.7.2"
```
itechmeat, line 50: [Not a blocker] 🔵 The smol-toml pin is a legitimate audit fix, but it is a third concern in a "parallel deps fetching" PR and the only thing telling a reader why a kit version and two toml files moved would be the PR description - which is currently just CodeRabbit's auto-summary. Could you add an author description (what changed, what it deliberately does not touch, the timeout behaviour change, the spec instructions realised, checks run)? Half the commit subjects also label rather than describe ("fix: findings fixes"), though that matters less if the merge squashes.

packages/mfes/src/handler/__tests__/MfeHandlerMF.test.ts
```diff
@@ -103,3 +213,715 @@ describe('MfeHandlerMF — unresolved publicPath placeholder guard', () => {
...
+      const url = [...dispatchedAt.keys()].find((u) => u.endsWith(suffix));
+      expect(url, `expected a fetch call for ${suffix}`).toBeDefined();
+      return dispatchedAt.get(url as string) as number;
+    });
+
+    const spread = Math.max(...depTimes) - Math.min(...depTimes);
+    // Sequential dispatch would spread these by ~40-80ms (one full response
+    // per iteration before the next fetch is even issued); concurrent
+    // dispatch issues all three within a handful of milliseconds.
+    expect(spread).toBeLessThan(30);
```
itechmeat, line 373: [Not a blocker] 🔵 A 30ms window on real timers - has this flaked in CI yet? The event-log assertion at :331 is deterministic; the same shape would carry this case and the two timeout ones.

packages/mfes/src/handler/MfeHandlerMF.ts
```diff
@@ -1227,24 +1932,56 @@ class MfeHandlerMF extends MfeHandler<MfeEntryMF, ChildMfeBridge> {
...
       const resolved = this.resolveRelativePath(chunkFilename, relPath);
-      const blobUrl = blobUrlMap.get(resolved);
+      const blobUrl = loadState.blobUrlMap.get(resolved);
       if (blobUrl) return blobUrl;
-      // If resolved is already absolute (dep of a cross-origin portable chunk),
-      // use the URL as-is instead of prepending baseUrl.
-      if (resolved.startsWith('http://') || resolved.startsWith('https://')) {
-        return resolved;
+      // @cpt-begin:cpt-frontx-algo-mfe-isolation-blob-url-chain:p1:inst-if-dep-cycle-absent
+      if (build.cycleShortCircuits.has(resolved)) {
```
itechmeat: [Not a blocker] 🔵 `cycleShortCircuits` is keyed by filename only, so a referrer that was never on the cycle also takes the origin fallback if the entry is missing from `blobUrlMap`. Mostly shadowed by the map lookup, but worth a sentence in the comment.

packages/mfes/src/handler/MfeHandlerMF.ts
```diff
@@ -827,52 +1317,239 @@ class MfeHandlerMF extends MfeHandler<MfeEntryMF, ChildMfeBridge> {
...
+   */
+  private recordCycleShortCircuit(
+    build: ChainBuildState,
+    filename: string,
+    lineage: ReadonlySet<string>,
+    detectedVia: string
+  ): void {
+    if (build.cycleShortCircuits.has(filename)) return;
+    build.cycleShortCircuits.add(filename);
+    console.warn(
```
itechmeat: [Not a blocker] 🔵 `console.warn` here - the spec's cycle instructions describe a silent short-circuit, so this is observable behaviour with nothing behind it, and it bypasses whatever error pipeline the host has. Either an instruction step or an injected reporter?

architecture/ADR/0006-mfe-handler-resolution.md
```diff
@@ -89,7 +89,7 @@ A table maps concrete entry types to handlers.
...
+The present concrete instantiation of the handler abstraction is the abstract `MfeHandler` class (`packages/mfes/src/handler/types.ts`), which declares `handledBaseTypeId`, `priority`, the `bridgeFactory` it provides, and `load(entry, extensionId)`; its documented resolution rule is that the registry matches entries using `typeSystem.isTypeOf(entryTypeId, handledBaseTypeId)` rather than the handler choosing itself. The shipped default handler is the Module Federation handler `MfeHandlerMF` (`packages/mfes/src/handler/MfeHandlerMF.ts`), constructed at priority `0`. The handler's live loading path is manifest-driven and is decided separately in `cpt-frontx-adr-mfe-asset-discovery`. The opaque type-substrate port that the injected type system satisfies is decided in `cpt-frontx-adr-runtime-type-system-coupling`, and the runtime surface that exposes the handler-bearing registry is decided in `cpt-frontx-adr-mfe-runtime-public-surface`.
```
itechmeat, line 92: [Not a blocker] 🔵 The screensets -> mfes path fix landed in 0006/0012/0013, but 0003, 0004, 0005, 0009, 0010 and 0011 still carry `packages/screensets`, and 0011's Confirmation points at a file that does not exist while this PR leans on that ADR. One sweep, here or in a follow-up?

packages/mfes/src/handler/MfeHandlerMF.ts
```diff
@@ -47,6 +47,182 @@ import {
...
+ * Result of scanning a batch of settled results for the first rejection.
+ *
+ * A plain `reason | undefined` return value cannot distinguish "no
+ * rejection found" from "the first rejection found was itself the value
+ * `undefined`" (a task can reject with no reason at all). Both call sites
+ * need that distinction to decide whether to propagate a failure, so this
+ * returns an explicit `found` flag alongside the reason instead of relying
+ * on `reason`'s own value to carry that signal.
+ */
+interface FirstRejectionResult {
```
itechmeat: [Not a blocker] ⚪ Why not `settled.find((r): r is PromiseRejectedResult => r.status === 'rejected')`? The settled result object already carries found-vs-not, so the interface and the `found` flag seem to restate it.

packages/mfes/src/handler/__tests__/MfeHandlerMF.test.ts
```diff
@@ -103,3 +213,715 @@ describe('MfeHandlerMF — unresolved publicPath placeholder guard', () => {
...
+/**
+ * Concurrent-fetch behaviour of the two dependency-fanout loops
+ * (`fetchSharedDepSources`'s shared-dep loop and
+ * `createBlobUrlChainInternal`'s sibling loop) and the cycle-detection
+ * fallback that keeps them from deadlocking on circular dependencies. Prior
+ * coverage of these loops only mocked `fetch` to reject immediately, so the
+ * concurrent-dispatch path, the cross-caller `inFlight`/`blobUrlMap` dedup
+ * path, and the cycle-detection fallback were never actually exercised.
+ */
+describe('MfeHandlerMF — concurrent dependency fetch', () => {
```
itechmeat, line 326: [Not a blocker] ⚪ The file header still says "publicPath resolution guard tests" while four new describe blocks cover fetch concurrency, lazy-import isolation, cycle fallback and LRU eviction. Also docs.

template-shell/packages/react/src/mfe/hooks/useDomainExtensions.ts
```diff
@@ -43,24 +40,10 @@ import type { Extension } from '@gears-frontx/framework';
  * }
  * ```
  */
-// @cpt-begin:cpt-frontx-flow-react-bindings-use-domain-extensions:p1:inst-call-domain-extensions
-// @cpt-begin:cpt-frontx-dod-react-bindings-observation-hooks:p1:inst-call-domain-extensions
 export function useDomainExtensions(domainId: string): Extension[] {
   const app = useFrontX();
-  const registry = app.mfeRegistry;
+  const registry = resolveMfeRegistry(app, 'useDomainExtensions');
```
itechmeat, line 45: [Not a blocker] ⚪ Markers for the context-guard algo were removed from three hooks and types.ts but stay in FrontXContext.tsx:7, :29 and :47 - worth finishing the retirement for that one concept while you are here?

itechmeat (approved):
Approving - the fix is real and well-tested: the serial awaits from #618 are gone, the fan-out is bounded, cycle detection covers the cross-branch case, the #619 type augmentation behaves exactly as documented, and I could not make the new tests pass against a reverted implementation. Nothing inline blocks, but two items are worth a look before or right after merge: a narrow case where one chain build still inherits another's failure through the shared in-flight registry (strictly better than develop, where any failure poisoned every later lazy import, but the new comment claims full isolation), and the FEATURE artifacts recording implemented behaviour as not implemented, so the traceability chain no longer describes the code. The rest is comment volume, docs and a few questions.

frontgeeks (approved):

frontgeeks (approved):

PR #625, opened 2026-09-11
Fix: shared dep dedup

GeraBart:
## Summary by CodeRabbit

- **New Features**
  - Manifests can publish optional emitted-chunk content hashes for shared dependencies.
  - Shared dependency reuse uses producer content hashes when available, with a resolved URL fallback.
  - Added diagnostics for unresolved declared module references and warnings for undeclared package imports.

- **Bug Fixes**
  - Prevents incorrect shared dependency reuse across independently built microfrontends with matching package versions.
  - Shared dependency adoption notices are deduplicated with bounded tracking.

- **Documentation**
  - Added architecture guidance covering shared dependency identity, compatibility, and adoption.


template-shell/package.json
```diff
@@ -70,8 +70,8 @@
   },
   "dependencies": {
     "@gears-frontx/api": "0.3.0-alpha.2",
-    "@gears-frontx/gts-plugin": "0.3.0-alpha.5",
-    "@gears-frontx/mfes": "0.3.0-alpha.5",
+    "@gears-frontx/gts-plugin": "0.3.0-alpha.6",
+    "@gears-frontx/mfes": "0.3.0-alpha.6",
```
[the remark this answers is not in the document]
GeraBart: Not applying this one — the stated mechanism does not hold.

`scripts/ecosystem-version-policy.mjs`'s exact-pin rule (`isExactPin`, used by the edge-compatibility check) applies to dependency **edges between ecosystem packages**, not to template manifests. The only edge it evaluates here is `@gears-frontx/mfes` → `@gears-frontx/gts-plugin`, which declares `^0.3.0-alpha.0` and passes.

`npm run policy:version-check` is green on this branch with these exact pins in place:

> Edge-compatibility check passed: @gears-frontx/mfes declares "^0.3.0-alpha.0" on @gears-frontx/gts-plugin (satisfied by: 0.3.0-alpha.6). Independent publication holds for: @gears-frontx/mfes@0.3.0-alpha.6, @gears-frontx/gts-plugin@0.3.0-alpha.6, ...

Exact pins in templates are also the deliberate convention rather than a violation: `scripts/template-pin-drift-check.mjs` exists to verify that every template pinning an ecosystem-scope package at an exact registry version matches the version actually on disk. Replacing these with ranges would remove them from that gate's coverage.

architecture/ADR/0034-shared-dep-dedup-key.md
```diff
@@ -0,0 +1,127 @@
...
+
+**Open question, recorded rather than resolved here**: the published manifest's concrete field-level schema has no FEATURE that owns it, unlike other contracts whose ownership `cpt-frontx-adr-contract-schema-ownership` assigns by name. This decision does not assign that ownership; closing that gap is a separate decision.
```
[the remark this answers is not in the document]
GeraBart: Agreed that the gap is real, but it predates this PR and the ADR records it deliberately as an open question rather than deciding it — assigning a schema owner is its own decision under the ADR 0027 convention, not something to settle inside a bug fix.

Filed as #626, covering both halves: who owns `MfManifest`'s concrete field-level schema, and what keeps the JSON Schema and the TypeScript types in agreement (generation, a conformance check, or an accepted duplication with a stated reason).

One correction on the detail: `EnrichedSharedEntry` in `template-shell/src/build/mf-gts.ts` was consolidated in this PR to import `MfManifestShared` from the published package rather than redeclare it, so the producer's build-time type is no longer a separate declaration. `OutMfManifestShared` in `generate-mfe-manifests.ts` still is, and that is captured in #626.

packages/mfes/src/handler/__tests__/MfeHandlerMF.test.ts
```diff
@@ -1395,3 +1405,790 @@ describe('LruCache — capacity eviction and MRU re-insertion', () => {
...
+
+    expect(thrown).toBeInstanceOf(MfeLoadError);
+    const message = (thrown as Error).message;
+    expect(message).toContain('dep-b');
+    expect(message).toContain('dep-a');
+    expect(message).toContain('ext-declared-bare-specifier-survives');
+
+    blobModuleStub.current = undefined;
+    rewriteSpy.mockRestore();
+    fetchSpy.mockRestore();
```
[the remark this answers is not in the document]
GeraBart: Leaving this as-is for now, with a reason rather than a dismissal.

Checked concretely: the file has 16 `vi.spyOn` calls against 52 `mockRestore`/`restoreAllMocks` calls, and the suite is green — spies are restored, generally in `finally`. The residual exposure is narrower than "shared test state is not cleaned up": it is a spy leaking if a test throws *before* reaching its restore.

The clean fix for that is `restoreMocks: true` in the Vitest config, which applies repo-wide and would change behaviour for every test file in the package — not something this PR should carry in alongside a runtime fix. Better as its own change, where any test that silently depended on a leaked spy shows up in isolation rather than mixed into this diff.

packages/mfes/src/handler/MfeHandlerMF.ts
```diff
@@ -635,6 +639,15 @@ class MfeHandlerMF extends MfeHandler<MfeEntryMF, ChildMfeBridge> {
...
 
+  /**
+   * Tracks which `name@version` + manifest id pairs have already received
+   * the one-time adoption notice emitted when a shared-dep entry declares
+   * no `contentHash` (see `inst-emit-adoption-notice`). Scoped to the
+   * handler instance, same lifetime as `sharedDepTextCache`, so the notice
+   * fires once per pair across every load this handler serves — not once
+   * per load.
+   */
+  private readonly sharedDepAdoptionNoticesEmitted = new Set<string>();
```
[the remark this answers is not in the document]
GeraBart: Valid finding — fixed.

`sharedDepAdoptionNoticesEmitted` is now an `LruCache<string, true>` bounded by a new `SHARED_DEP_ADOPTION_NOTICE_CACHE_CAPACITY = 64`, sitting beside `SOURCE_TEXT_CACHE_CAPACITY` and `SHARED_DEP_TEXT_CACHE_CAPACITY` and reusing the same `LruCache` rather than introducing a second eviction mechanism. 64 covers one entry per distinct (shared dep, manifest) pair lacking a `contentHash`; a single manifest contributes only a handful, so it sits below the 128 the shared-dep text cache uses.

Worth noting this also changed a stated guarantee, so the FEATURE was amended first: `inst-emit-adoption-notice` said the notice was **one-time**, which a bounded ledger cannot promise — after eviction the same pair may warn again. It now specifies deduplication per pair *while the ledger retains it*, plus the bound itself, with the capacity left to the code. Existing once-per-pair coverage still passes and a new test asserts the ledger does not grow past its capacity (159/159; `cfs validate` 0 errors).

packages/mfes/src/handler/MfeHandlerMF.ts
```diff
@@ -1310,19 +1352,46 @@ class MfeHandlerMF extends MfeHandler<MfeEntryMF, ChildMfeBridge> {
...
+        // @cpt-begin:cpt-frontx-algo-mfe-isolation-build-shared-dep-blob-urls:p1:inst-compute-key-hash
+        cacheKey = `${dep.name}@${dep.version}@${dep.contentHash}`;
+        // @cpt-end:cpt-frontx-algo-mfe-isolation-build-shared-dep-blob-urls:p1:inst-compute-key-hash
+      } else {
+        // @cpt-begin:cpt-frontx-algo-mfe-isolation-build-shared-dep-blob-urls:p1:inst-else-no-hash
+        // @cpt-begin:cpt-frontx-algo-mfe-isolation-build-shared-dep-blob-urls:p1:inst-compute-key-fallback
+        cacheKey = `${dep.name}@${dep.version}@${absoluteUrl}`;
+        // @cpt-end:cpt-frontx-algo-mfe-isolation-build-shared-dep-blob-urls:p1:inst-compute-key-fallback
+        // @cpt-begin:cpt-frontx-algo-mfe-isolation-build-shared-dep-blob-urls:p1:inst-emit-adoption-notice
+        const noticeKey = `${dep.name}@${dep.version} ${manifest.id}`;
```
itechmeat: [Not a blocker] 🟠 There is a raw U+0000 byte committed between `${dep.version}` and `${manifest.id}` here - the only NUL in any .ts file in the repo. grep and ripgrep now classify this 2500-line file as binary and print nothing for it (`rg` says "binary file matches"), which is very likely why CodeRabbit could not confirm the thread-3 fix. NUL is a sound separator, but could it be the `\0` escape, or the plain `@` the key one line above uses?

packages/mfes/src/handler/MfeHandlerMF.ts
```diff
@@ -635,6 +651,22 @@ class MfeHandlerMF extends MfeHandler<MfeEntryMF, ChildMfeBridge> {
     SHARED_DEP_TEXT_CACHE_CAPACITY,
   );
 
+  /**
+   * Tracks which `name@version` + manifest id pairs have already received
```
itechmeat, line 682: [Not a blocker] 🟠 The new ledger's doc is right, but the field twenty lines up (:639-643) still says "Keyed by `name@version` ... cache hit regardless of which server hosts the MFE" - that sentence is the bug this PR fixes, written as the design. Same at :466, :497 and :1322-1324, and :469 has a "now bounded" that reads as a diff note. Also :1347-1349 cites "the FEATURE Note on inst-for-each-dep" about name@version precedence, and this PR deleted that Note. Worth one pass over the comments in this file?

template-shell/src/build/mf-gts.ts
```diff
@@ -764,6 +803,54 @@ class StandaloneEsmBuilder {
...
+    const source = fs.readFileSync(outfile, 'utf-8');
+    const normalized = source
+      // pnpm's store-path segment (with its enclosing node_modules/ on
+      // both sides), with or without leading '../' hops.
+      .replace(
+        /(?:\.\.\/)*(?:node_modules\/)?\.pnpm\/[^/"'\s]+\/node_modules\//g,
+        'node_modules/'
+      )
+      // Remaining depth-only '../' hops immediately preceding node_modules.
+      .replace(/(?:\.\.\/)+node_modules\//g, 'node_modules/');
```
itechmeat: [Not a blocker] 🟠 Everything `contentHash` is worth depends on these two regexes making two builds byte-identical, and nothing tests them - while the cycle guard #620 added to this same file is an exported pure function with its own test in `__tests__/build/`. Could `normalizeEmbeddedModulePaths` and `computeContentHash` get the same treatment, ideally with a two-builds-same-hash case? The ADR's own closing line says the byte-level agreement is "confirmed by measurement".

template-shell/src/build/mf-gts.ts
```diff
@@ -764,6 +803,54 @@ class StandaloneEsmBuilder {
   static normalizeDepName(name: string): string {
     return name.replace(/^@/, '').replace(/\//g, '-');
   }
+
+  /**
+   * Canonicalizes the embedded `node_modules` source paths esbuild writes
+   * into non-minified output, so the same dependency/version/externals
+   * combination produces byte-identical output regardless of package
```
itechmeat: [Not a blocker] 🟡 "Regardless of package manager layout or depth" is wider than what the two regexes do. Measured with the plugin's own esbuild options: a workspace-linked `@gears-frontx/mfes` emits `../packages/mfes/dist/index.js` (no `node_modules/` segment at all) and stays untouched, so a linked build and a registry build of the same version hash differently and never share. Safe direction, but could the comment say exactly which shapes are normalized?

packages/mfes/architecture/features/mfe-isolation/FEATURE.md
```diff
@@ -159,29 +159,44 @@ Internal system functions that implement the isolation mechanism.
...
    1. [x] - `p1` - Rewrite bare shared-dep specifiers in the source to the already-resolved blob URLs - `inst-rewrite-specifiers`
-   2. [x] - `p1` - Wrap the rewritten source in a blob, create a fresh blob URL, and add it to the shared-dep blob URL map - `inst-create-dep-blob`
+   2. [x] - `p1` - Assert that no shared-dependency name the manifest declares survives the rewrite as a bare specifier in the rewritten source, checked in the same two import forms the rewrite handles — `from "x"` and `import "x"` — because that check is the exact inverse of a rewrite performed per declared name and so cannot mistake ordinary code for an import; a dynamically imported specifier (`import("x")`) is outside this surface and is not asserted, because the rewrite does not handle it either - `inst-assert-shared-dep-no-bare-specifier`
+   3. [x] - `p1` - **IF** a declared shared-dependency name survives that assertion - `inst-if-shared-dep-bare-specifier`
+      1. [x] - `p1` - **RETURN** error — fail the load with a diagnostic naming the chunk, the surviving specifier, and the microfrontend - `inst-raise-shared-dep-bare-specifier`
```
itechmeat, line 189: [Not a blocker] 🟡 Two new observable outcomes - the load-failing refusal here and the undeclared-specifier warning at :190 - and the flow's Error Scenarios list (:75-82) that already carries the sibling cycle refusal gained nothing, nor did the acceptance criteria. Could both get their entries, plus one criterion for the adoption notice?

packages/mfes/src/handler/MfeHandlerMF.ts
```diff
@@ -1441,13 +1521,50 @@ class MfeHandlerMF extends MfeHandler<MfeEntryMF, ChildMfeBridge> {
...
     const rewritten = this.rewriteBareSpecifiers(source, blobUrls);
     // @cpt-end:cpt-frontx-algo-mfe-isolation-build-shared-dep-blob-urls:p1:inst-rewrite-specifiers
     // @cpt-end:cpt-frontx-algo-mfe-loading-manifest-discovery:p1:inst-md-rewrite-specifiers
+    // @cpt-begin:cpt-frontx-algo-mfe-isolation-build-shared-dep-blob-urls:p1:inst-assert-shared-dep-no-bare-specifier
+    const survivor = findSurvivingDeclaredSharedDepSpecifier(rewritten, sharedNames);
+    // @cpt-end:cpt-frontx-algo-mfe-isolation-build-shared-dep-blob-urls:p1:inst-assert-shared-dep-no-bare-specifier
+    // @cpt-begin:cpt-frontx-algo-mfe-isolation-build-shared-dep-blob-urls:p1:inst-if-shared-dep-bare-specifier
+    if (survivor !== undefined) {
+      // @cpt-begin:cpt-frontx-algo-mfe-isolation-build-shared-dep-blob-urls:p1:inst-raise-shared-dep-bare-specifier
+      throw new MfeLoadError(
```
itechmeat: [Not a blocker] 🟡 The cycle refusal at :1486 goes through `markDeterministicLoadFailure`; this one does not, so with the default `retries: 2` it costs three attempts and ~3s of backoff before surfacing (measured). Same deterministic shape - should it be marked the same way?

packages/mfes/src/handler/MfeHandlerMF.ts
```diff
@@ -1441,13 +1521,50 @@ class MfeHandlerMF extends MfeHandler<MfeEntryMF, ChildMfeBridge> {
...
+        `shared-dep chunk '${depName}' still imports the bare specifier ` +
+          `'${survivor}' after rewriting its declared shared ` +
+          `dependencies, for microfrontend '${extensionId}'. Every ` +
+          'declared shared-dependency name that survives rewriting must ' +
+          'resolve to a blob URL; this indicates a rewrite defect rather ' +
```
itechmeat: [Not a blocker] 🟡 This branch is reachable without any rewrite defect: `isDepReadyToResolve` skips `other === name`, so a shared-dep chunk that imports its own package name gets rewritten against a map that cannot contain itself yet and lands here (reproduced at head with no stubs). The message then points at a handler bug when the cause is the producer's build. Worth naming that case in the message, or excluding self-imports upstream?

packages/mfes/src/handler/mf-dynamic-module-ops.ts
```diff
@@ -48,6 +48,39 @@ export function sourceImports(source: string, packageName: string): boolean {
...
+ * Build the regex that matches the exactly two import forms this trust
+ * kernel handles, for one exact package name: `from "x"` (static imports/
+ * re-exports) and `import "x"` (bare side-effect imports). Quote style is
+ * captured (group 2) so callers can preserve it, and the specifier text is
+ * captured (group 3). Deliberately does not match `import("x")` (dynamic
+ * import) — neither this function's callers rewrite that form, so nothing
+ * may assert against it either.
+ *
+ * `rewriteBareSpecifier` and `findSurvivingDeclaredSharedDepSpecifier` both
+ * build on this single per-name pattern so the two can never drift apart:
```
itechmeat: [Not a blocker] 🟡 "Both build on this single per-name pattern so the two can never drift apart" - `rewriteBareSpecifier` does, but `findSurvivingDeclaredSharedDepSpecifier` goes through `sourceImports` (:145), which has its own regex and its own escape at :47-48. Equivalent today, but the guarantee the comment (and FEATURE.md:186) rests on is not there. One line makes it true, or the sentence can stop claiming it. Also `bareSpecifierPattern` is exported with no external consumer.

template-shell/src/build/mf-gts.ts
```diff
@@ -764,6 +803,54 @@ class StandaloneEsmBuilder {
...
+      // both sides), with or without leading '../' hops.
+      .replace(
+        /(?:\.\.\/)*(?:node_modules\/)?\.pnpm\/[^/"'\s]+\/node_modules\//g,
+        'node_modules/'
+      )
+      // Remaining depth-only '../' hops immediately preceding node_modules.
+      .replace(/(?:\.\.\/)+node_modules\//g, 'node_modules/');
+
+    if (normalized !== source) {
+      fs.writeFileSync(outfile, normalized, 'utf-8');
```
itechmeat, line 881: [Not a blocker] 🟡 The normalization rewrites the shipped bundle, and the second regex is unanchored over the whole file - so `../node_modules/` inside a dependency's own string or regex literal gets rewritten too. "Only two path-bearing forms occur" holds for the current fixtures, not for whatever a vendor template declares. Since the runtime never compares the hash to the bytes, would hashing a normalized copy and serving the untouched chunk give the same stability with none of the exposure?

template-shell/src/build/mf-gts.ts
```diff
@@ -228,14 +230,42 @@ class MfeJsonEnricher {
...
+   * same dependency at the same version hash differently. Returns
+   * `undefined` when the chunk is missing rather than emitting a hash for
+   * bytes that were never actually published — an absent `contentHash` is a
+   * valid, well-handled manifest state; a wrong one is not.
+   */
+  private computeContentHash(chunkFilePath: string): string | undefined {
+    try {
+      const bytes = fs.readFileSync(chunkFilePath);
+      return createHash('sha256').update(bytes).digest('hex');
+    } catch {
```
itechmeat: [Not a blocker] 🟡 When can the chunk be missing here? `StandaloneEsmBuilder.build()` just wrote it at exactly this path, and an esbuild failure rejects out of `closeBundle`, so this catch only fires on a real I/O fault - and turns it into a manifest the runtime reads as "producer has not adopted yet", with nothing in the build log. A loud warning naming the path, at least?

packages/gts-plugin/src/frontx.mfes/schemas/mfe/mf_manifest.v1.json
```diff
@@ -64,6 +64,11 @@
               { "type": "string", "minLength": 1 },
               { "type": "null" }
             ]
+          },
+          "contentHash": {
+            "type": "string",
+            "minLength": 1,
```
itechmeat: [Not a blocker] 🟡 The contract says a contentHash is any non-empty string, and the runtime keys reuse on it verbatim without ever checking it against the bytes - so a foreign producer emitting a 4-char hash is conforming and brings the cross-MFE mis-reuse straight back. A `pattern` of `^[0-9a-f]{64}$` here (and a non-empty guard on the runtime branch at :1365) would make the ADR's safety property a contract rather than a convention. Fine to defer to #626 if that is where the schema question lands.

template-shell/src/build/mf-gts.ts
```diff
@@ -228,14 +230,42 @@ class MfeJsonEnricher {
...
-    }));
+    return declaredDeps.map((name) => {
+      const fileName = `${StandaloneEsmBuilder.normalizeDepName(name)}.js`;
+      const contentHash = this.computeContentHash(
+        path.join(sharedOutputDir, fileName)
+      );
+      return {
+        name,
+        version: this.resolvePackageVersion(name),
+        chunkPath: `shared/${fileName}`,
```
itechmeat, line 244: [Not a blocker] 🟡 Question rather than a change: with the hash beside a content-independent `shared/react.js`, a manifest that goes stale relative to its chunk (partial rebuild, CDN TTL skew) advertises H_old over new bytes, and another MFE that genuinely built H_old is handed foreign code - the exact defect the ADR closes, reachable through deployment. Was `shared/react.<hash>.js` considered? It is not among the ADR's four options; even a sentence on why not would help.

template-shell/scripts/generate-mfe-manifests.ts
```diff
@@ -536,6 +538,7 @@ export class ManifestGenerator {
         version: s.version,
         chunkPath: s.chunkPath,
         unwrapKey: s.unwrapKey,
+        ...(s.contentHash !== undefined ? { contentHash: s.contentHash } : {}),
```
itechmeat, line 541: [Not a blocker] 🟡 Also docs: this is the one place that knows per MFE whether a hash was published, and it stays silent - the only signal is a browser `console.warn` deduplicated and evicted at 64 entries. The generator already prints a per-MFE summary; one line naming the MFEs without a hash would give operators and app authors a way to see who is on the fallback. And `contentHash` appears in no README, QUICK_START or llms.txt yet.

architecture/ADR/0034-shared-dep-dedup-key.md
```diff
@@ -0,0 +1,127 @@
+---
+status: proposed
+date: 2026-09-10
+---
+
+# What Identity Should the Cross-MFE Shared-Dependency Source-Text Cache Key On?
```
itechmeat, line 6: [Not a blocker] 🟡 Open PR #560 also adds an ADR numbered 0034 (`0034-region-union-marker-grammar-and-host-document.md`), and #564 announces 0035-0042 in its title (0042 exists as a file today), so whichever lands second inherits a duplicate number. Worth agreeing who renumbers now rather than at merge time - 0043 looks like the first free one.

packages/mfes/src/handler/__tests__/MfeHandlerMF.test.ts
```diff
@@ -1395,3 +1405,846 @@ describe('LruCache — capacity eviction and MRU re-insertion', () => {
...
+ * `name@version`. A shared-dep chunk is built per consuming microfrontend,
+ * so two manifests can legitimately declare the same name and version while
+ * shipping structurally different chunk bytes; whichever manifest's load
+ * reached the cache first served its chunk's text to every later manifest
+ * declaring the same name and version, leaving the later manifest with an
+ * unrewritten bare specifier baked into a `blob:` module and a silent mount
+ * failure. The fix (not implemented by these tests) keys reuse on a
+ * declared content hash when the manifest provides one, falls back to the
+ * resolved absolute chunk URL (which is unique per microfrontend) when it
+ * does not — emitting a one-time adoption notice per `name@version` and
```
itechmeat: [Not a blocker] 🟡 This docblock still promises a "one-time" notice while the test at :1661 is titled "the earliest pair is renotified" and the FEATURE was amended off "one-time" in thread 3. Two blocks nearby (:1409-1424, :1815-1831) narrate the old contract and "the REJECTED prior design" - that is ADR 0034 material, not what these cases assert. Keep the rule (the scan must never fail a load), drop the history?

packages/mfes/src/handler/__tests__/MfeHandlerMF.test.ts
```diff
@@ -1395,3 +1405,846 @@ describe('LruCache — capacity eviction and MRU re-insertion', () => {
...
+    }
+
+    expect(thrown).toBeInstanceOf(MfeLoadError);
+    const message = (thrown as Error).message;
+    expect(message).toContain('dep-b');
+    expect(message).toContain('dep-a');
+    expect(message).toContain('ext-declared-bare-specifier-survives');
+
+    blobModuleStub.current = undefined;
+    rewriteSpy.mockRestore();
```
itechmeat: [Not a blocker] 🟡 On CodeRabbit's thread: the repo-wide `restoreMocks` argument is right, but this particular spy is a prototype stub on the class under test replaced with an identity function, and the three asserts above it sit outside any `try` - one failed expect leaves the rest of this 2250-line file running a handler that does not rewrite specifiers. The four tests right above use `} finally {`; two lines here would match them without touching the config question.

packages/mfes/architecture/features/mfe-isolation/FEATURE.md
```diff
@@ -159,29 +159,44 @@ Internal system functions that implement the isolation mechanism.
...
+   4. [x] - `p1` - **IF** the rewritten source still carries, in those same two import forms, a specifier the manifest never declared which is well-formed as a package module specifier — an optionally scoped package name with an optional subpath, carrying no whitespace, quote, parenthesis, colon or line break - `inst-if-undeclared-specifier`
+      1. [x] - `p1` - Emit a diagnostic naming the chunk, that specifier, and the microfrontend, and proceed with the load — because such a specifier cannot resolve inside an isolated module and the browser's own resolution failure names neither the chunk nor the microfrontend, while this detection reads the chunk's text without parsing it and therefore must never be able to fail a load that would otherwise succeed - `inst-warn-undeclared-specifier`
+   5. [x] - `p1` - Wrap the rewritten source in a blob, create a fresh blob URL, and add it to the shared-dep blob URL map - `inst-create-dep-blob`
 7. [x] - `p1` - **RETURN** the complete shared-dep blob URL map - `inst-return-map`
 
+**Notes** (deferred, not decided here):
```
itechmeat, line 195: [Not a blocker] 🔵 Open questions parked inside a FEATURE algorithm - two of the three belong in ADR 0034, and the hash-verification one is already there as the Review trigger (:113). "This amendment" also dates the text to the change. Keep the expose-chain one, move the rest?

packages/mfes/architecture/features/mfe-isolation/FEATURE.md
```diff
@@ -159,29 +159,44 @@ Internal system functions that implement the isolation mechanism.
 
 - [x] `p1` - **ID**: `cpt-frontx-algo-mfe-isolation-build-shared-dep-blob-urls`
 
-**Input**: MFE manifest containing the shared-dependency list (name, version, chunk path) in the manifest's published enumeration order, which fixes only the sequence in which the declarations are visited (`cpt-frontx-adr-mfe-asset-discovery`)
+**Input**: MFE manifest containing the shared-dependency list (name, version, chunk path, and an optional content hash of the entry's emitted chunk) in the manifest's published enumeration order, which fixes only the sequence in which the declarations are visited (`cpt-frontx-adr-mfe-asset-discovery`)
```
itechmeat, line 163: [Not a blocker] 🔵 Five of the new instructions exist because of ADR 0034, but §1.4 References (:55) still lists only `cpt-frontx-adr-mfe-load-isolation`. Worth adding `cpt-frontx-adr-shared-dep-dedup-key` so the reference runs both ways?

packages/mfes/src/handler/mf-dynamic-module-ops.ts
```diff
@@ -48,6 +48,39 @@ export function sourceImports(source: string, packageName: string): boolean {
...
+ * whatever the rewriter can rewrite for a given declared name is exactly
+ * what the assertion checks for that name's survival. A generic (no package
+ * name) form of this matcher intentionally does not live here — see
+ * `findUndeclaredWellFormedSpecifiers` in `mf-shared-dep-specifier-scan.ts`,
+ * which needs no interpolation and so carries no trust-kernel residency
+ * requirement.
+ *
+ * @safety-reviewed 2026-04-20
```
itechmeat: [Not a blocker] 🔵 New function, April date? The file header says adding an export here needs a security review, and the lint only checks the tag is present - so a copied date is exactly what nothing catches. Same for `findSurvivingDeclaredSharedDepSpecifier`.

packages/mfes/src/handler/MfeHandlerMF.ts
```diff
@@ -1441,13 +1521,50 @@ class MfeHandlerMF extends MfeHandler<MfeEntryMF, ChildMfeBridge> {
...
+        entryId
+      );
+      // @cpt-end:cpt-frontx-algo-mfe-isolation-build-shared-dep-blob-urls:p1:inst-raise-shared-dep-bare-specifier
+    }
+    // @cpt-end:cpt-frontx-algo-mfe-isolation-build-shared-dep-blob-urls:p1:inst-if-shared-dep-bare-specifier
+    // @cpt-begin:cpt-frontx-algo-mfe-isolation-build-shared-dep-blob-urls:p1:inst-if-undeclared-specifier
+    const undeclared = findUndeclaredWellFormedSpecifiers(rewritten, sharedNames);
+    for (const specifier of undeclared) {
+      // @cpt-begin:cpt-frontx-algo-mfe-isolation-build-shared-dep-blob-urls:p1:inst-warn-undeclared-specifier
+      console.warn(
```
itechmeat, line 1598: [Not a blocker] 🔵 The adoption notice twenty lines up got a bounded dedup ledger; this warning fires on every load of the same MFE (four loads, four identical lines, measured). Intentional asymmetry? And since only the producer can act on it, is the host's console the right reader?

template-shell/package.json
itechmeat, line 74: [Not a blocker] 🔵 Not about the red check itself - just that `template-shell/package-lock.json` still pins both packages at alpha.4 (identical to base) while `template-mfe`'s lock was refreshed in this PR, so publishing alpha.6 alone will not turn Template Validate green until this lock is regenerated too. Neither pin gate reads lockfiles, so nothing flags it.

packages/mfes/src/handler/MfeHandlerMF.ts
```diff
@@ -1441,13 +1521,50 @@ class MfeHandlerMF extends MfeHandler<MfeEntryMF, ChildMfeBridge> {
...
+          'declared shared-dependency name that survives rewriting must ' +
+          'resolve to a blob URL; this indicates a rewrite defect rather ' +
+          'than a missing declaration.',
+        entryId
+      );
+      // @cpt-end:cpt-frontx-algo-mfe-isolation-build-shared-dep-blob-urls:p1:inst-raise-shared-dep-bare-specifier
+    }
+    // @cpt-end:cpt-frontx-algo-mfe-isolation-build-shared-dep-blob-urls:p1:inst-if-shared-dep-bare-specifier
+    // @cpt-begin:cpt-frontx-algo-mfe-isolation-build-shared-dep-blob-urls:p1:inst-if-undeclared-specifier
+    const undeclared = findUndeclaredWellFormedSpecifiers(rewritten, sharedNames);
```
itechmeat, line 1595: [Not a blocker] ⚪ `depName`, `entryId`, `extensionId` are three strings in a row on `createRewrittenBlobUrl`, and `extensionId` is drilled through five signatures for two log lines. A small diagnostics context object would carry it without the transposition risk - not for this PR necessarily.

packages/mfes/src/handler/__tests__/MfeHandlerMF.test.ts
```diff
@@ -1395,3 +1405,846 @@ describe('LruCache — capacity eviction and MRU re-insertion', () => {
...
+ * declaring the same name and version, leaving the later manifest with an
+ * unrewritten bare specifier baked into a `blob:` module and a silent mount
+ * failure. The fix (not implemented by these tests) keys reuse on a
+ * declared content hash when the manifest provides one, falls back to the
+ * resolved absolute chunk URL (which is unique per microfrontend) when it
+ * does not — emitting a one-time adoption notice per `name@version` and
+ * manifest id in that case — and asserts no bare specifier survives a
+ * shared-dep chunk's rewrite before minting its blob.
+ */
+describe('MfeHandlerMF — shared-dep cross-MFE cache key (issue #621)', () => {
```
itechmeat, line 1422: [Not a blocker] ⚪ The header's "Covers, in file order" list (:4-16) stops at LruCache, so the two new suites are outside the file's own index. Also docs.

architecture/ADR/0034-shared-dep-dedup-key.md
```diff
@@ -0,0 +1,127 @@
+---
+status: proposed
```
itechmeat, line 2: [Not a blocker] ⚪ `proposed` while the runtime already keys on it - deliberate pending the measurement the Neutral consequence asks for, or should it be `accepted` like everything since July?

architecture/ADR/0034-shared-dep-dedup-key.md
```diff
@@ -0,0 +1,127 @@
...
+* Good, because it is unconditionally safe: an absolute chunk URL is unique per microfrontend, so reuse never crosses a build boundary it should not.
+* Good, because it needs no change to the published manifest contract at all.
+* Bad, because a microfrontend's own asset base location is unique to it, so this option forfeits cross-microfrontend reuse entirely rather than preserving it where it is sound — every microfrontend refetches every shared dependency regardless of whether an identical build already exists in cache.
+
+### Keep the version-based key and require every producing build to bundle a shared dependency identically
+
+Retain the package-name-and-version key, and instead constrain every producing build to bundle a given shared dependency the same way, so that any two chunks sharing a key are guaranteed to be byte-identical by construction.
+
+* Good, because it would recover safe reuse without adding any field to the published manifest.
+* Bad, because the runtime cannot verify that a third-party producing build actually followed the convention — the manifest gives it no way to check, so a violation reproduces exactly today's defect.
```
itechmeat, line 94: [Not a blocker] ⚪ "today's defect" here and "the unsafe prior behaviour" at :51 date the text to the change; "the defect this decision closes" / "unsafe cross-microfrontend reuse" would read the same in a year.

itechmeat (approved):
Approving - the fix is real and provable: the cache key is derived in one place, all five hash/URL combinations behave as the ADR says (reuse only on a matching hash or the same per-MFE URL), fail-safe holds everywhere, and reverting the key to name@version fails exactly the three tests that should fail. The new ADR is legitimate too - nothing existing owned the question. Nothing inline blocks. The items worth a look before merge: a raw NUL byte committed into MfeHandlerMF.ts that makes grep and ripgrep treat the file as binary (very likely why CodeRabbit could not verify the thread-3 fix), four doc blocks in the same file still describing the old name@version key as the design, and the path normalization the hash's stability rests on having no test while its comment claims more than the two regexes do. The rest is docs, symmetry and a few questions.

template-shell/src/build/mf-gts.ts
gs-layer: [HIGH] Current producer should not silently omit `contentHash` when hashing its own just-built shared chunk fails.

`computeContentHash()` catches every read/hash failure and returns `undefined`, so `buildSharedEntries()` publishes the shared dependency without `contentHash`. Runtime fallback is useful for old or external manifests, but this producer controls the emitted `shared/${fileName}` output; if that file is missing or unreadable here, the manifest is internally inconsistent and should fail before publication instead of deferring the breakage to runtime.

Suggested fix: make this build-side path throw with the chunk path when hashing fails, and add a test that simulates a missing shared output file and expects manifest enrichment/build to fail.

packages/mfes/src/handler/MfeHandlerMF.ts
gs-layer: [MEDIUM] Please replace the raw NUL byte in this source string.

There is an actual U+0000 byte between `${dep.version}` and `${manifest.id}` in the `noticeKey` template literal. TypeScript can parse it, but common source tools now treat this file as binary (`rg` reports a binary match), which makes review/search/static analysis around this file brittle. An escaped separator such as `\u0000` / `\u001f`, or a structured key like `JSON.stringify([dep.name, dep.version, manifest.id])`, would preserve uniqueness without committing a control byte into the source file.

packages/mfes/src/handler/MfeHandlerMF.ts
gs-layer, line 682: [LOW] The shared dependency cache comments need to be brought in sync with the new keying model.

This area still has comments describing `sharedDepTextCache` as keyed by `name@version` / cache-hit-regardless-of-server, while the implementation now keys by absolute URL and content hash to avoid cross-host stale reuse. There are a few related stale references nearby as well (for example the compatibility/load comments later in this file). Please update the comments so future changes do not reintroduce the old unsafe `name@version` assumption.

GeraBart:
Thanks — these were sharp, and several were measured rather than guessed. Worked through all 16. Every claim was verified against the code before acting; three turned out to explain things I had misread myself.

## Fixed in this PR

**The NUL byte** (`MfeHandlerMF.ts:1375`) — confirmed, a literal U+0000 in the notice key, the only one in any `.ts` under `packages/`. Replaced with `@`, matching the key a line above. This also explains a symptom I had written off as a tooling quirk: `grep`/`rg` were silently returning nothing on that file. Verified 0 NUL bytes and grep working again.

**Deterministic failure retried** — confirmed: the bare-specifier refusal threw a bare `MfeLoadError` while the sibling cycle refusal wrapped its throw in `markDeterministicLoadFailure`. Now marked, so it surfaces immediately instead of after three attempts and ~3s of backoff.

**"Can never drift apart" was false** — confirmed: the assertion reached `sourceImports`, which built its own regex with its own escape. Made the claim true rather than deleting it: `sourceImports` now delegates to `bareSpecifierPattern`, so there is exactly one definition of the import surface. (Care taken with the `/g` regex's `lastIndex` — a fresh pattern per call.)

**Copied `@safety-reviewed` dates** — confirmed on two September additions. Corrected, with per-function `@why` reasoning rather than an inherited neighbour's.

**Stale comments** — the `name@version` descriptions at `:466`, `:497`, `:639-643`, `:1322-1324`, the "now bounded" diff-note, and the citation of a FEATURE Note this PR deleted: all rewritten to describe the current two-tier key.

**Refusal message blamed the wrong party** — the self-import case you reproduced now reads accurately for both causes.

**Test docblocks narrating the rejected design** — dropped the history and the stale "one-time" promise (the FEATURE was amended off it); the rules those blocks state are kept.

**Non-failure-safe spy** — now `try/finally`, matching the four tests above it. You were right that this is narrower and more actionable than the repo-wide `restoreMocks` argument I had replied with earlier.

**Silent `computeContentHash` catch** — now warns with the chunk path and the underlying error before falling back, so a real I/O fault is distinguishable from an unadopted producer.

**Comment overclaimed "regardless of layout or depth"** — your workspace-linked counter-example is right (`../packages/mfes/dist/index.js` has no `node_modules/` segment and is untouched). The docblock now names the two shapes actually normalized and notes the rest pass through.

**No tests for the normalization regexes** — the strongest point, and taken. `normalizeEmbeddedModulePathsInSource` and `computeContentHash` are now exported pure functions with 12 tests under `template-shell/__tests__/build/`, following the precedent of #620's cycle guard in that same file — including the two-builds-same-hash case and a differing-code case.

**Schema accepted any non-empty string as a hash** — real hole, closed: `"pattern": "^[0-9a-f]{64}$"` (still optional), plus a runtime guard so a malformed value falls back to URL keying instead of being trusted.

**Artifacts** — `cpt-frontx-adr-shared-dep-dedup-key` added to §1.4 References; the LRU-sizing note moved to ADR 0034 Consequences; the hash-verification note dropped as duplicate (it is already the ADR's Review trigger); "this amendment" phrasing removed; Error Scenarios and acceptance criteria gained entries for the refusal, the warn-and-proceed diagnostic, and the notice's dedup scope — each `[x]` tied to a named test.

### One your comment caused me to find

Tightening the hash guard broke a test — correctly: the fixtures used toy values (`'same-hash'`, `'hash-a'`) that the guard now rejects. Fixing them exposed that the *differing*-hash test had been passing for the wrong reason: both values were rejected as malformed, so it would have passed even with hash comparison entirely broken. Fixtures are now real sha256 digests, and I mutation-tested it — breaking the comparison now fails the test, as it should.

## Deferred, with reasons

- **Schema ownership** → filed as #626. Real, predates this PR, and assigning an owner is its own decision under the ADR 0027 convention.
- **Hash a normalized copy, ship the untouched chunk** — agreed in principle: the second regex is unanchored over the whole file, so a vendor's string literal could be rewritten. It is a design change to the producer rather than a fix, and I would rather it land on its own with its own gate.
- **Content-addressed chunk filenames vs stale-manifest skew** — a genuine deployment-level hazard and not among the ADR's four options; ADR-level, worth its own decision.
- **Generator reporting which MFEs lack a hash, plus docs** — agreed, and the generator is indeed the only place that knows. Follow-up.
- **ADR number collision with #560** — not renumbering: this repo's ADR numbering is consecutive, so `0034` is correct as the successor to `0033`. Whichever PR lands second takes the next free number.

All gates green: 160/160 tests, `cfs validate` 0 errors, `arch:check` 12/12, and the browser done-gate passed (both MFEs mount, cross-MFE reuse demonstrated — `fixture_b` served all 12 deps from `fixture_a`'s cache).

template-shell/package.json
```diff
@@ -70,8 +70,8 @@
   },
   "dependencies": {
     "@gears-frontx/api": "0.3.0-alpha.2",
-    "@gears-frontx/gts-plugin": "0.3.0-alpha.5",
-    "@gears-frontx/mfes": "0.3.0-alpha.5",
+    "@gears-frontx/gts-plugin": "0.3.0-alpha.6",
```
gs-layer, line 73: [HIGH] Please regenerate the template lockfiles after these version bumps.

`template-shell/package.json` now asks for `@gears-frontx/gts-plugin@0.3.0-alpha.6`, `@gears-frontx/mfes@0.3.0-alpha.6`, and `@gears-frontx/ui-kit@0.4.0-alpha.3`, but the checked-in `template-shell/package-lock.json` still records older versions. A clean install from the PR tree fails with `npm ci` because the manifest and lockfile are out of sync; `template-mfe` has the same problem for `@gears-frontx/ui-kit`. Please regenerate the affected template lockfiles and rerun the isolated template validation.

packages/mfes/src/handler/MfeHandlerMF.ts
```diff
@@ -349,6 +351,21 @@ function isDeterministicLoadFailure(error: unknown): boolean {
   );
 }
 
+/**
+ * Whether `hash` is a well-formed lowercase-hex SHA-256 digest — the shape
+ * `contentHash` is expected to carry. The JSON Schema for the manifest
+ * currently accepts any non-empty string in that field, so a malformed
```
gs-layer, line 357: [LOW] This comment is now stale relative to the schema update.

The manifest schema now constrains `contentHash` with `^[0-9a-f]{64}$`, so it no longer accepts any non-empty string. The runtime guard is still useful for stale or untrusted manifests, but the comment should describe that overlap accurately so future readers do not think the schema remains permissive.

PR #628, opened 2026-09-14
feat(ui-kit): align the kit with the designer catalogue - tokens in both themes, radius base 10, motion, size axes, table resize

itechmeat:
## Description

The kit is brought onto the designer's live catalogue, the Storybook at `studio.constructor.rocks/foundations` with its handoff package. The owner made it the single source for component anatomy, variants, sizes, states and token values; the Figma file is frozen and now only explains inherited values. Two rules govern the diff: token and prop names follow shadcn/ui, and values are taken literally from the catalogue in both themes, never corrected kit-side. Where the catalogue draws nothing, the kit keeps its value. The owner-facing analysis lives outside the repository (`.ai-notes`, gitignored) because it names the vendor's design files.

## What changed

**Colour tokens, light and dark.** Every role the catalogue defines takes its value: the flat dark page `#0f172a`, the ring that no longer lifts in dark, `--destructive` and `--danger` at `#fb7185` in dark, `--secondary`, `--input`, the sidebar roles, `--muted` on the page value, and three new status text tones (`--success-foreground`, `--warning-foreground`, `--danger-foreground`). `--accent` becomes the brand blue with white text as drawn; the selected states of Toggle, segmented ToggleGroup and Tabs move to `--secondary` with the drawn 1px hairline, which is how the catalogue paints them. The guard tests that pinned the earlier corrected values now assert the drawn ones.

**Scales.** The radius scale is rebuilt on the drawn base 10: `sm` 6, `md` 8, `lg` 10, `xl` 12, new `2xl` 16; every module reads the step whose value it drew, so fields sit on 10 (Input, Textarea, Select, InputGroup, Combobox, NativeSelect), popovers on 10, cards on 12. A motion layer is published as `--duration-fast/default/slow/...` and `--ease-standard` with the catalogue's values, replacing the raw durations that matched them. A chart palette `--chart-1..5` plus four categorical hues feeds the Chart component. Two scrim tokens replace the single inverting overlay: modal `#0f172a` at 48% in both themes, side panel black at 10% with blur.

**Geometry on the drawn values.** InputGroup keeps its three-step axis with the drawn 32 as `default`; Badge is the drawn 20px box plus the undrawn `xs` 18 count step; Checkbox 16 with radius 4 and a 14px glyph; Dialog `100% - 32` capped at 384 (`lg` kept); Sheet 75% capped at 384 without a radius; Popover radius 10; Select 32 and 28 with the drawn 10/8 inset (`filter` 36 kept, undrawn); the segmented ToggleGroup container drops its border, inset and height. Tabs move to the drawn model: one 3px indicator sliding over 240ms on `--ease-standard`, trigger radius 8, padding 2/6, a filled 32px track; the `size` axis stays with the drawn step as `default`.

**New surface.** Button variants `navigation`, `avatar`, `utility` and the tinted `destructive`, on the current 32/36/40 scale. Card `typography="panel"`. DropdownMenu `indicatorSide`, Avatar group pieces, NavigationMenu variants, Toggle `steel`, Tooltip delay 0, ScrollArea track 10. Table gains column resizing in `TableHead` (pointer and keyboard, 12px steps, minimum widths, live `aria-label`) and the dense collection view (sticky header, 64px rows, focus ring, pending row). Column-width presets stay with templates.

**Component contract.** Button is one of the three components #616 enrolled in the contract harness. Its committed `button.contract.json` is recompiled so the variant enum carries `navigation`, `avatar` and `utility`; the compat check against develop reports the change as backward compatible, and the freshness test passes.

**Version.** `@gears-frontx/ui-kit` `0.4.0-alpha.4` (develop after #616) to `0.4.0-alpha.5` with the lockfile workspace entry.

## Visible and breaking for consumers

Radii shift by 2px on cards, popovers and sheets. The omitted InputGroup height is 32. Badge sizes `sm`/`default` (24/28) are gone; the box is 20. `--accent` reads as a brand-blue fill, so anything painting `--accent` as a pale surface changes look. The dark page, card and popover share one colour as drawn. Tabs indicator animates. The control scale change the catalogue also draws (buttons 24/28/32/36, fields 32) is deliberately not here: it ships as a separate major-bump release.

## Verification

- Repo root after the rebase onto develop with #616: `npm ci`, `build:packages`, `type-check:all`, `lint`, `test:unit --coverage`, `test:consumer`, `arch:check`, `arch:deps:core`, `arch:edges`, `arch:guards`, `policy:ecosystem-pin-drift`, `policy:version-bump-on-change`, `policy:mfes-import-boundary`, `policy:contracts` (guard and compat against develop), `cfs validate`, `validate-kits` green.
- Runtime, Chrome against the kit demo, light and dark: every token computed and every size, radius and width measured with `getBoundingClientRect`; the Tabs indicator sampled across frames to prove it slides; column resize dragged and keyed with the width sum held constant. An independent QA pass repeated the measurements in both themes.
- Added lines contain no vendor names or design-file references.

## Out of scope, deliberately

The control scale and field height change (separate major release), the catalogue's product families (studio, artifacts, layout, table presets, ai, findings, features), `--studio-*` tokens, and anything the catalogue does not draw.


packages/ui-kit/src/components/table/table.module.css
```diff
@@ -263,6 +317,74 @@
...
+  /* The base table, unchanged. */
+}
+
+.variantCollection {
+  table-layout: fixed;
+  min-width: 960px;
+}
+
+.variantCollection .tableHead {
+  position: sticky;
```
SysoevAndrey, line 345: This header won't stick when the page scrolls. `.tableContainer` has `overflow-x: auto`, which makes it the sticky container (overflow-y becomes auto too), and it has no height, so there's nothing to stick inside. Consumers can't give it a height either, since `className` lands on `<table>`, not the wrapper.

packages/ui-kit/src/components/table/table.tsx
```diff
@@ -1,8 +1,257 @@
...
+  minSelf: number,
+  minNeighbour: number,
+): number {
+  const self = cells[index];
+  if (!self) {
+    return selfStart;
+  }
+  const neighbour = neighbourStart === null ? null : cells[index + 1];
+  const ceiling =
+    neighbourStart === null || !neighbour
```
SysoevAndrey, line 135: For the last column the table width is supposed to absorb the change, but `.table` is `width: 100%` (and `min-width: 960px` in collection), so it can't get narrower. With `table-layout: fixed` the browser hands the freed width back to the other columns, so shrinking the last column doesn't hold, and the `aria-label` ends up announcing a width that isn't on screen.

packages/ui-kit/src/components/table/table.tsx
```diff
@@ -1,8 +1,257 @@
...
+    lockPage(false);
+    const handle = handleRef.current;
+    if (drag && handle) {
+      capturePointer(handle, drag.pointerId, false);
+    }
+  }, []);
+
+  // Releases the lock if the header unmounts mid-drag: the page would
+  // otherwise keep the resize cursor and the suppressed selection forever.
+  useEffect(() => endDrag, [endDrag]);
```
SysoevAndrey, line 229: This runs `lockPage(false)` on every unmount, not just mid-drag, so it resets `body.style.cursor` / `userSelect` even when this resizer never set them (column set change, route change, StrictMode double mount). Worth only unlocking when `dragRef.current` is set.

packages/ui-kit/src/components/table/table.module.css
itechmeat: Fixed in f8d761f3: the wrapper takes `containerClassName` / `containerStyle`, so it scrolls itself and the header sticks inside it. Docs and demo updated.

packages/ui-kit/src/components/table/table.tsx
itechmeat: Fixed in 08787cce: the first drag pins the table width to the column sum, so the last column narrows (down to `min-width`) and the label matches the rendered width.

packages/ui-kit/src/components/table/table.tsx
itechmeat: Fixed in d9fbfccb: `endDrag` returns early when no drag is active.

SysoevAndrey (approved):

GeraBart (approved):

